PR #20, opened 2021-04-07
feat: add Test API

fubuloubu:
requires: #8, #33

TODO:
- [ ] Run some tests using the `ethereum:development:tester` Provider


fubuloubu:
Still in work, will re-open

Issue #36, opened 2021-04-23
Add ContractAPI to ProjectManager and EcosystemAPI


fubuloubu:
- [ ] Create `ContractAPI` using `ProviderAPI` (maybe using #33)

Issue #35, opened 2021-04-23
Add TransactionAPI and TransactionReceiptAPI to EcosystemAPI


fubuloubu:
- Add `.transfer(AddressAPI, amt)`
- Add `.deploy(ContractType, *args)` functions to `AccountAPI` using `ProviderAPI`

PR #38, opened 2021-05-03
feat: add transactions

fubuloubu:
Adds transactional support for both Ether transfers and contract deployments

fixes: #35
fixes: #36

~~requires: #49 #70~~

TODO: 
- [ ] testing


src/ape/api/accounts.py
```diff
@@ -1,81 +1,96 @@
...
 
-    @abstractmethod
-    def sign_transaction(self, txn: dict) -> Optional[SignedTransaction]:
-        ...
+        txn_cost = txn.gas_limit * txn.gas_price
+        if "value" in kwargs:
+            txn_cost += kwargs["value"]
+
+        if txn_cost > self.balance:
+            raise  # Transfer value meets or exceeds account balance
```
unparalleled-js, line 80: Is there an exception in the stack at this point?
I see this all over the code and in other situations, I have arrived there with no exceptions in the stack, causing the error to be odd.

It would be better to momentarily use `raise Exception("stuff")` and in the future, replace it with custom or more thought out errors. Unless there is something I don't understand here.

fubuloubu: You're totally right, was just using these as bookmarks, but probably way better to use `Exception` and customize them later

Was thinking it might be time to customize anyways, just haven't found the time

pyproject.toml
```diff
@@ -22,7 +22,7 @@ markers = "fuzzing: Run Hypothesis fuzz test suite"
 line_length = 100
 force_grid_wrap = 0
 include_trailing_comma = true
-known_third_party = ["IPython", "click", "dataclassy", "eth_account", "eth_utils", "github", "hypothesis", "hypothesis_jsonschema", "importlib_metadata", "pluggy", "pytest", "requests", "setuptools", "web3", "yaml"]
+known_third_party = ["IPython", "click", "dataclassy", "eth_abi", "eth_account", "eth_utils", "github", "hypothesis", "hypothesis_jsonschema", "importlib_metadata", "pluggy", "pytest", "requests", "setuptools", "web3", "yaml"]
```
unparalleled-js, line 25: why both pyproject.toml and setup.py?

src/ape/api/address.py
```diff
@@ -0,0 +1,64 @@
...
+    def nonce(self) -> int:
+        return self.provider.get_nonce(self.address)
+
+    @property
+    def balance(self) -> int:
+        return self.provider.get_balance(self.address)
+
+    @property
+    def code(self) -> bytes:
+        # TODO: Explore caching this (based on `self.provider.network` and examining code)
```
unparalleled-js, line 47: Could you use `@lru_cache`?
https://www.geeksforgeeks.org/python-functools-lru_cache/

fubuloubu: Caching code is a more complex thing, it basically depends on which network you're own and how it was deployed. I can tell you more about how it works, but a simple lru cache wouldn't really work here.

src/ape/api/base.py
```diff
@@ -9,7 +9,8 @@ class AbstractDataClassMeta(DataClassMeta, ABCMeta):
     pass
 
 
-abstractdataclass = partial(dataclass, meta=AbstractDataClassMeta)
+abstractdataclass = partial(dataclass, kwargs=True, meta=AbstractDataClassMeta)
```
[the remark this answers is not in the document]
fubuloubu: Haha yes, this is a flag provided to the dataclassy `dataclass` decorator that allows handling kwargs in the constructor

src/ape_ethereum/ecosystem.py
```diff
@@ -10,8 +21,92 @@
...
-# TODO: Define VM-specific stuff here
+    def decode_event(self, abi: ABI, receipt: "ReceiptAPI") -> "ContractLog":
+        filter_id = keccak(to_bytes(text=abi.selector))
+        raw_event = next(log for log in receipt.logs if log["filter_id"] == filter_id)
+        input_names = [i.name for i in abi.inputs]
+        input_type = ",".join(i.canonical_type for i in abi.inputs)
+        event_data = abi_decode(input_type, raw_event)
+        return ContractLog(  # type: ignore
+            name=abi.name,
+            inputs={k: v for k, v in zip(input_names, event_data)},
```
lost-theory, line 111: Using `dict(zip(...))` also works and is shorter:

```
>>> x, y = ['a', 'b', 'c'], [1, 2, 3]
>>> {k: v for (k, v) in zip(x, y)}
{'a': 1, 'b': 2, 'c': 3}
>>> dict(zip(x, y))
{'a': 1, 'b': 2, 'c': 3}
```

src/ape/api/contracts.py
```diff
@@ -0,0 +1,259 @@
...
+        def get_name(abi: ABI):
+            return abi.name
+
+        try:
+            if attr_name in map(get_name, self._contract_type.events):
+                selected_abis = [
+                    abi for abi in self._contract_type.events if get_name(abi) == attr_name
+                ]
+                assert len(selected_abis) == 1
+                return self._event_class(  # type: ignore
```
lost-theory, line 149: The `self._event_class` property was part of the old implementation.

lost-theory: Also these three blocks are very similar, so you could generalize that inner function to handle three different cases, something like this:

```
def __getattr__(self, attr_name: str) -> Any:
    handlers = {
        'events': EventHandler,
        'calls': ContractCallHandler,
        'transactions': ContractTransactionHandler,
    }
    def get_abi(abi_type -> str, abi_handler -> Any):
        selected = [abi for abi in getattr(self._contract_type, abi_type) if abi.name == attr_name]
        if not selected:
            return
        kwargs = {'provider': self.provider, 'abis': selected[0] if abi_type == "events" else selected}
        assert len(selected) == 1 if abi_type == "events" else True
        return handlers[abi_type](**kwargs)
    for (abi_type, abi_handler) in handlers.items():
        handler = get_abi(abi_type, abi_handler)
        if handler:
            return handler
```

It might only be worth it to do that though if there are more than 3 cases.

lost-theory:
I like the new architecture that relies less on adding new `abstractdataclass`es! Moving the encode/decode logic into one fatter Ecosystem class seems easier to understand vs. having it spread out over many smaller classes. Same goes for moving e.g. `selector` up into `ABI`.

Left a comment on one potential bug (`_event_class` not defined, I think mypy will catch this if `type: ignore` is removed on that line) and two minor style issues. Code LGTM to me otherwise.

src/ape_ethereum/ecosystem.py
fubuloubu: oh yeah duh

fubuloubu:
> I think mypy will catch this if `type: ignore` is removed on that line) and two minor style issues.

I would love to figure out what about `dataclassy` makes it not possible to recognize kwargs even if I have `kwargs=True` set

src/ape/api/contracts.py
```diff
@@ -0,0 +1,303 @@
...
+    abis: List[ABI]
+
+    def __call__(self, *args, **kwargs) -> Any:
+        selected_abi = None
+        for abi in self.abis:
+            if len(args) == len(abi.inputs):
+                selected_abi = abi
+
+        if not selected_abi:
+            raise Exception("Number of args does not match")
```
lost-theory, line 95: Feel free to ignore, but here's a way to do this in fewer lines:
```
selected_abi = [abi for abi in self.abis if len(args) == len(abi.inputs)]
assert len(selected_abi) == 1, "Number of args does not match"
return ContractCall(abi=selected_abi[0], ...
```

src/ape/api/contracts.py
```diff
@@ -0,0 +1,303 @@
...
+    @property
+    def address(self) -> str:
+        return self._address
+
+    def __getattr__(self, attr_name: str) -> Any:
+        def get_name(abi: ABI):
+            return abi.name
+
+        try:
+            if attr_name in map(get_name, self._contract_type.events):
```
lost-theory, line 183: How about something like this to DRY up the three repeated blocks:

```
def match_name(abi: ABI):
    return abi if abi.name == attr_name else None
selections = [
    (ContractEvent, filter(match_name, self._contract_type.events)),
    (ContractCallHandler, filter(match_name, self._contract_type.calls)),
    (ContractTransactionHandler, filter(match_name, self._contract_type.transactions)),
]
for (klass, selected) in selections:
    selected = list(selected)
    if selected:
        return klass(provider=self.provider, address=self.address, abis=selected)
```

src/ape/api/contracts.py
```diff
@@ -0,0 +1,303 @@
...
+    def __call__(self, *args, **kwargs) -> TransactionAPI:
+        constructor = ContractConstructor(  # type: ignore
+            abi=self._contract_type.constructor,
+            provider=self._provider,
+            deployment_bytecode=self._deployment_bytecode,
+        )
+        return constructor.encode(*args, **kwargs)
+
+
+def Contract(
```
lost-theory, line 262: I think it's a little confusing to have three different ways to import `Contract`:

```
In [1]: from ape import Contract as C1
In [2]: from ape.api import Contract as C2
In [3]: from ape.api.contracts import Contract as C3
In [4]: C1 == C2, C2 == C3
Out[4]: (False, True)
```

If you want "C1" above to be the real class used by end users, maybe C2 and C3 should be named `_Contract` in order to hide that implementation detail.

lost-theory (approved):
LGTM, just left a few minor comments.

src/ape/api/contracts.py
fubuloubu: Fair. This is what we want.

src/ape/api/accounts.py
fubuloubu: Resolved all outstanding `raises` without `Exception`

src/ape/api/contracts.py
fubuloubu: refactored in 8afd46

fubuloubu:
@lost-theory any further comments? or good to merge?

lost-theory (approved):
LGTM.

src/ape/types/abstract.py
```diff
@@ -50,7 +50,7 @@ def to_dict(v: Any) -> Optional[Union[list, dict, str, int, bool]]:
         return v
 
     else:
-        raise  # Unhandled type
+        raise Exception(f"Unhandled type '{type(v)}'")
```
sabotagebeats, line 53: supercedes https://github.com/ApeWorX/ape/pull/80 and fixes https://github.com/ApeWorX/ape/issues/71

PR #41, opened 2021-05-05
refactor: cache manifest

lost-theory:
Finishing off @fubuloubu's earlier refactor/cache-manifest work.

Test plan: 

* Made an ape project with one dep and one vyper contract. Ran `ape console`.
* First `project.contracts` lookup caused compilation.
* Subsequent lookups used the cache.
* Changing contract contents triggered a re-compile of only the changed contract.
* Adding a new contract triggered a compilation of just the one new file.
* Deleting a file caused the contract to disappear from `project.contracts`.
* Repeated all the above with both `.vy` and `.json` contract files.


src/ape/api/compiler.py
```diff
@@ -11,6 +11,13 @@ class CompilerAPI(ABC):
     def name(self) -> str:
         ...
 
+    @property
+    @abstractmethod
+    def versions(self) -> List[str]:
```
fubuloubu, line 16: More I think about it, more I think this is necessary
```suggestion
    def versions(self, all_paths: List[Path]) -> List[str]:
```

To explain, this would be given all paths (not just ones to recompile) and return the set of versions necessary to compile them. This is important for driving the `compilers` key in the manifest

src/ape/managers/config.py
```diff
@@ -28,8 +31,8 @@ def __init__(self):
 
         for plugin_name, config_class in self.plugin_manager.config_class:
             if plugin_name in user_config:
-                user_override = user_config[plugin_name]
-                del user_config[plugin_name]  # For checking if all config was processed
+                user_override = user_config.pop(plugin_name)
+                # NOTE: `dict.pop()` is used for checking if all config was processed
             else:
                 user_override = {}
```
fubuloubu, line 37: This can be reduced to:
```python
user_override = user_config.pop(plugin_name, {})
```

src/ape/managers/config.py
```diff
@@ -41,6 +44,11 @@ def __init__(self):
 
             self._plugin_configs[plugin_name] = config
 
+        # Top level config items
+        self.name = user_config.pop("name", "")
+        self.version = user_config.pop("version", "")
+        self.dependencies = user_config.pop("dependencies", [])
+
```
fubuloubu, line 51: Maybe this should move to above the plugin loop to prevent a malicious plugin from front-running these settings

src/ape/managers/project.py
```diff
@@ -1,34 +1,78 @@
...
 
 
+def compute_checksum(source: str, algorithm: str = "md5") -> str:
+    if algorithm == "md5":
+        hasher = md5()
+        hasher.update(source.encode("utf-8"))
+        return hasher.hexdigest()
+
+    else:
+        raise  # Unknown algorithm
```
fubuloubu, line 22: Should move to `utils`

src/ape/managers/project.py
```diff
@@ -1,34 +1,78 @@
...
     compilers: CompilerManager
 
-    depedendencies: Dict[str, "ProjectManager"] = dict()
+    dependencies: Dict[str, "ProjectManager"] = dict()
 
     def __init__(self):
-        pass  # Look for depedencies from config
+        if isinstance(self.path, str):
+            # TODO: Handle URIs for external package references
+            self.path = Path(self.path)
```
fubuloubu, line 35: I did this upstream I think

src/ape/managers/project.py
```diff
@@ -1,34 +1,78 @@
...
+
+        self.dependencies = {
+            manifest.name: manifest
+            for manifest in map(self._extract_manifest, self.config.dependencies)
+        }
+
+    def _extract_manifest(self, manifest_uri: str) -> PackageManifest:
```
fubuloubu, line 42: Technically, dependencies is a dict of `str: ProjectManager`, that might need to get updated

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
 
         return files
 
+
+    def _load_contracts(self) -> Dict[str, ContractType]:
+        # Load a cached or clean manifest (to use for caching)
+        manifest = self.cached_manifest or PackageManifest()
```
fubuloubu, line 92:
```suggestion
    def load_contracts(self, use_cache: bool=True) -> Dict[str, ContractType]:
        # Load a cached or clean manifest (to use for caching)
        manifest = use_cache and self.cached_manifest or PackageManifest()
```

NOTE: Can make use of this in the compile CLI

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
     def _tests_folder(self) -> Path:
         return self.path / "tests"
 
     # TODO: Make this work for generating and caching the manifest file
+    @property
+    def compiler_data(self) -> List[Compiler]:
+        compilers = []
+
+        for compiler in self.compilers.registered_compilers.values():
+            for version in compiler.versions:
```
fubuloubu, line 176: To match up with change above (turn it into not a `property`)
```suggestion
        for extension, compiler in self.compilers.registered_compilers.items():
            for version in compiler.versions(file for file in self.sources if file.suffix == extension):
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
+        )
+
+        # NOTE: This property is optional here, but mandatory for publishing
+        if self.config.name:
+            manifest.name = self.config.name
+
+        # NOTE: This property is optional here, but mandatory for publishing
+        if self.config.version:
+            manifest.version = self.config.version
+
```
fubuloubu, line 196:
```suggestion

        # NOTE: This property is optional here, but mandatory for publishing
        if self.config.meta:
            manifest.meta = self.config.meta
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
-
     # def publish_manifest(self):
     #     manifest = self.manifest.to_dict()  # noqa: F841
-    #     # TODO: Clean up manifest
+    #     if not manifest.name:
+    #         raise
+    #     if not manifest.version:
+    #         raise
+    #     if self.meta:
+    #         manifest.meta = self.meta
```
fubuloubu, line 210:
```suggestion
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
-    #         ),
-    #     )
-
     # def publish_manifest(self):
     #     manifest = self.manifest.to_dict()  # noqa: F841
-    #     # TODO: Clean up manifest
+    #     if not manifest.name:
+    #         raise
+    #     if not manifest.version:
+    #         raise
```
fubuloubu, line 208:
```suggestion
    #         raise  # Need version to release manifest
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
-    #             Compiler(c.name, c.version)  # type: ignore
-    #             for c in self.compilers.registered_compilers.values()
-    #         ),
-    #     )
-
     # def publish_manifest(self):
     #     manifest = self.manifest.to_dict()  # noqa: F841
-    #     # TODO: Clean up manifest
+    #     if not manifest.name:
+    #         raise
```
fubuloubu, line 206:
```suggestion
    #         raise  # Need name to release manifest
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
-    #         compilers=list(
-    #             Compiler(c.name, c.version)  # type: ignore
-    #             for c in self.compilers.registered_compilers.values()
-    #         ),
-    #     )
-
     # def publish_manifest(self):
     #     manifest = self.manifest.to_dict()  # noqa: F841
-    #     # TODO: Clean up manifest
+    #     if not manifest.name:
```
fubuloubu, line 205:
```suggestion
    #     if not manifest["name"]:
```

src/ape/managers/project.py
```diff
@@ -42,41 +86,128 @@ def sources(self) -> List[Path]:
...
+    #         raise
+    #     if not manifest.version:
```
fubuloubu, line 207:
```suggestion
    #     if not manifest["version"]:
```

fubuloubu (approved):

src/ape/api/compiler.py
lost-theory: Alright, I've updated the ABC and ethpm `PackageCompiler`. I'll make a PR for ape-vyper.

src/ape/managers/project.py
lost-theory: I removed the TODO if that's what you mean. Anything else needed?

src/ape/managers/project.py
lost-theory: Updated the type to be `Dict[str, PackageManifest]`.

src/ape/managers/project.py
lost-theory: Ok, it works:

```
$ ape console
In [1]: project.compiler_data
Out[1]: [Compiler(name='vyper', version='0.2.11', settings=None, contractTypes=None)]
```

src/ape/managers/project.py
lost-theory: I deleted this property because it wasn't being used and because the code was buggy (`sources=self.sources` was a `List[Path]` when it should have been a `Dict[str, Source]`). Want me to add it back?

I'm also confused by having both a `manifest` and `cached_manifest` properties. Which one do you use when? Does that `self.confg...` logic also need to apply to the `cached_manifest`? Right now `load_contracts` doesn't use `self.manifest` AFAICS.

lost-theory:
Ready for another review @fubuloubu. Take a look at the `sourcePath` code I added in order to make it possible to detect when ethpm JSON files are deleted (the `sourceId` we were using had no connection to the JSON file on disk).

```
$ ( rm -rf .build/ ; ape console )

In [1]: project.contracts.keys()
INFO: Compiling 'contracts/escrow.json'
INFO: Compiling 'contracts/contract.vy'
Out[1]: dict_keys(['Escrow', 'SafeSendLib', 'contract'])

In [2]: project.contracts['Escrow'].sourceId
Out[2]: 'Escrow.sol'  # <-- this can't be used to detect filesystem changes

In [3]: project.contracts['Escrow'].sourcePath
Out[3]: PosixPath('/home/cuboid/src/ape/example-ignore/contracts/escrow.json')  # <-- this new attribute can
```

src/ape/managers/project.py
fubuloubu: Nope, looks good. Extra data sanitation is always good haha

src/ape/types/contract.py
```diff
@@ -103,6 +108,26 @@ def load_content(self):
...
+
+        if not self.content:
+            raise ValueError("Content not loaded yet. Can't compute checksum.")
+
+        if algorithm == "md5":
+            hasher = md5
+        else:
+            raise  # Unknown algorithm
+
+        self.checksum = Checksum(  # type: ignore
```
fubuloubu, line 127: Use the function from utils?

src/ape/utils.py
```diff
@@ -108,6 +108,15 @@ def load_config(path: Path, expand_envars=True, must_exist=False) -> Dict:
         return {}
 
 
+def compute_checksum(source: Path, algorithm: str = "md5") -> str:
+    if algorithm == "md5":
+        hasher = md5()
+        hasher.update(source.read_bytes())
+        return hasher.hexdigest()
```
fubuloubu, line 115: I like this return statement outside of the if branch, like the other one

src/ape/types/manifest.py
```diff
@@ -47,13 +47,28 @@ class PackageManifest(FileMixin, SerializableType):
...
                 # NOTE: This was inserted by us, remove it
                 del data["contractTypes"][name]["contractName"]
+                # convert Path to str, or else we can't serialize this as JSON
+                data["contractTypes"][name]["sourceId"] = str(
+                    data["contractTypes"][name]["sourceId"]
+                )
+                if "sourcePath" in data["contractTypes"][name]:
+                    data["contractTypes"][name]["sourcePath"] = str(
+                        data["contractTypes"][name]["sourcePath"]
+                    )
```
fubuloubu, line 71: Maybe this file path normalization logic should live in `src/ape/abstract.py` as a separate handling case for `Path` objects in `SerializableType`'s `to_dict` and `from_dict` methods

fubuloubu (approved):

src/ape/types/manifest.py
lost-theory: Gave that a try and couldn't find an easy way to roundtrip `Path` objects somewhere in the nested structure. If you recursively convert all `Path`s to `str`s, then in the `from_dict` you don't have enough info to determine if an incoming string is a `Path` or not.

It might be possible to get that to work using type annotations, but I'd rather do it in a separate PR and land this.

lost-theory:
OK to land @fubuloubu? Do you want me to squash or rewrite the history?

I will land https://github.com/ApeWorX/ape-vyper/pull/5 as well (the remaining CI failures there are happening in master).

src/ape/types/manifest.py
fubuloubu: What if they were URIs, and you could check if a URI is in the string object.

Either that, or sublclass `from_dict` manually for those items

fubuloubu:
> OK to land @fubuloubu? Do you want me to squash or rewrite the history?

Yes. Squash please.

> I will land https://github.com/ApeWorX/ape-vyper/pull/5 as well (the remaining CI failures there are happening in master).

Awesome!

PR #44, opened 2021-05-19
refactor: update how compile CLI works to select contracts

fubuloubu:
This PR changes `ape compile` CLI endpoint behavior, allowing to filter selections from the current project when compiling (default is all paths)

### DIFF
--- a/src/ape_compile/_cli.py
+++ b/src/ape_compile/_cli.py
@@ -1,13 +1,28 @@
+from itertools import chain
+from pathlib import Path
+
 import click
 
 from ape import project
 from ape.utils import notify
 
+flatten = chain.from_iterable
 
-@click.command(short_help="Compile the contract source files")
-@click.argument("contracts", nargs=-1, type=click.Path())
+
+@click.command(short_help="Compile select contract source files")
+@click.argument(
+    "filepaths",
+    nargs=-1,
+    type=click.Path(exists=True, path_type=Path),
+)
 @click.option(
-    "-a", "--all", "compile_all", default=False, is_flag=True, help="Recompile all contracts"
+    "-f",
+    "--force",
+    "use_cache",
+    flag_value=False,
+    default=True,
+    is_flag=True,
+    help="Force recompiling selected contracts",
 )
 @click.option(
     "-s",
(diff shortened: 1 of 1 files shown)


src/ape_compile/_cli.py
```diff
@@ -26,28 +33,53 @@ def cli(contracts, compile_all, display_size):
...
+            if path.is_dir():
+                contract_filepaths.extend(list(path.iterdir()))
+
+            else:
+                contract_filepaths.append(path)
+
+        # Filter by what's in our project's source tree
+        contract_filepaths = [c for c in contract_filepaths if c in project.sources]
+
+    if len(contract_filepaths) == 0:
```
lost-theory, line 58: I think it's slightly more idiomatic python to say `if not contract_filepaths` (same as what you did on line 37).

lost-theory (approved):

src/ape_compile/_cli.py
```diff
@@ -62,3 +94,8 @@ def cli(contracts, compile_all, display_size):
             click.echo(f"  {name:<{indent}}  -  {size:>6,}B  ({pct:.2%})")
 
         click.echo()
+
+    # TODO: Remove this once compiling process works correctly
+    else:
+        # HACK: This force compiles all contracts
+        project.contracts
```
lost-theory, line 101: What does "correctly" mean? Maybe we want to add a `project.compile()` method so that we can get rid of the `HACK` comment and not rely on the side effect of looking up the property.

src/ape_compile/_cli.py
fubuloubu:
```suggestion
    if not contract_filepaths:
```

src/ape_compile/_cli.py
fubuloubu: Definitely agreed. I didn't get to updating this for your caching PR yet

src/ape_compile/_cli.py
```diff
@@ -26,33 +41,55 @@ def cli(contracts, compile_all, display_size):
...
+        else:
+            selected_paths = str(Path.cwd() / "contracts")
+
+        notify("WARNING", f"No project files detected in '{selected_paths}'")
         return
 
-    # NOTE: This compiles all types in `contracts/`
-    contract_types = list(project.contracts.values())
+    # TODO: only compile selected contracts
+    contract_types = project.load_contracts(use_cache)
```
fubuloubu, line 73: @lost-theory still need a better API for only compiling the selected files, but this is good enough for now I think

PR #45, opened 2021-05-19
chore: add `release-drafter`

fubuloubu:
NOTE: Won't work until installed

### DIFF
--- a/.github/workflows/draft.yaml
+++ b/.github/workflows/draft.yaml
@@ -1,0 +1,16 @@
+name: Release Drafter
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/draft.yaml
```diff
@@ -0,0 +1,16 @@
+name: Release Drafter
+
+on:
+  push:
+    branches:
+      - master
```
fubuloubu, line 6:
```suggestion
      - main
```

lost-theory (approved):
Rebase to see if test failures go away.

PR #47, opened 2021-05-19
fix: only compile interface JSON files

fubuloubu:


src/ape_pm/compiler.py
```diff
@@ -21,19 +21,16 @@ def compile(self, filepaths: List[Path]) -> List[ContractType]:
...
+            if not isinstance(data, list):
+                raise  # Not a valid ABI interface JSON file
 
             else:
                 contract_types.append(
                     ContractType(  # type: ignore
                         contractName=path.stem,
                         abi=data,
+                        sourceId=str(path),
                         sourcePath=path,
```
lost-theory, line 34: You removed the `sourcePath` attribute from the dataclass, so I think this can be removed.

lost-theory (approved):
Left one comment, LGTM otherwise.

src/ape_pm/compiler.py
fubuloubu:
```suggestion
```

PR #48, opened 2021-05-19
refactor: plugin cli

fubuloubu:


fubuloubu:
This one is good to go

src/ape/api/config.py
```diff
@@ -30,3 +30,8 @@ def serialize(self) -> Dict:
 
     def validate_config(self):
         pass
+
+
+class ConfigDict(ConfigItem):
+    def __post__init__(self):
```
lost-theory, line 36: Should be `__post_init__` (one underscore in the middle)?

fubuloubu:
```suggestion
    def __post_init__(self):
```

lost-theory (approved):

PR #61, opened 2021-05-27
test: add test for console items that they don't have issues

fubuloubu:


lost-theory:
It's not directly related to ipython, the segfault in ipython just covers up the max recursion depth traceback which makes it harder to debug. Here's the chain of nested dataclass instances which causes the infinite recursion:

```
__repr__ <class 'ape.managers.accounts.AccountManager'>
__repr__ <class 'ape.managers.config.ConfigManager'>
__repr__ <class 'ape.managers.networks.NetworkManager'>
__repr__ <class 'ape.managers.config.ConfigManager'>
__repr__ <class 'ape_test.providers.LocalNetwork'>
__repr__ <class 'ape.api.networks.create_network_type.<locals>.network_def'>
__repr__ <class 'ape_ethereum.ecosystem.Ethereum'>
__repr__ <class 'ape.managers.networks.NetworkManager'>
__repr__ <class 'ape.managers.config.ConfigManager'>
...the last 5 repeat in a cycle after that...
```

This seems to be a bug in dataclassy, so I created an issue for them: https://github.com/biqqles/dataclassy/issues/48

fubuloubu:
yeah seems like a custom `__repr__` would be an easy solution for this, maybe to the `manager` objects

lost-theory (approved):
Some CI failures but code and dataclassy version bump LGTM.

lost-theory:
In case we want to add back the test for `--verbose`, here's a workaround:

https://github.com/ipython/ipykernel/issues/91#issuecomment-168670274

Change `faulthandler.enable()` to `faulthandler.enable(file=sys.__stderr__)`.

PR #67, opened 2021-05-31
fix: small updates

unparalleled-js:
* Use error text color for Abort error
* Return empty dict when config is None on load
* Fix f-str, missing `f`
* Fix spellings

Additional questions:

* Do docs work / exist? Seems like there is a plan for readthedocs and also the link on the site for docs does not really go anywhere. I just want to make sure I have all the material.
* Plans for more thorough unit tests?
* Loading the config at module load time seems bad because you miss out on available click features, such as `click.Exception`.

I will continue on tomorrow, just wanted to open some dialogue and contribute, even if small. The big change here is the text color for `Abort` exceptions. Another big change is that is fixing a weirdness when you try to load an empty `yml` file as a config, the parse actually returns `None` which caused an expected code-flow / exception

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# IDE
+.idea/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# IDE
-.idea/
-
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -53,7 +53,7 @@ class ConfigManager:
             self._plugin_configs[plugin_name] = config
 
         if len(user_config.keys()) > 0:
-            raise  # Unprocessed config items
+            raise Exception("Unprocessed config items")
 
     def get_config(self, plugin_name: str) -> ConfigItem:
         if plugin_name not in self._plugin_configs:
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -26,7 +26,7 @@ class ConfigManager:
         config_file = self.PROJECT_FOLDER / CONFIG_FILE_NAME
 
         if config_file.exists():
-            user_config = load_config(config_file) or {}
+            user_config = load_config(config_file)
         else:
             user_config = {}
 
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -26,7 +26,7 @@ class ConfigManager:
         config_file = self.PROJECT_FOLDER / CONFIG_FILE_NAME
 
         if config_file.exists():
-            user_config = load_config(config_file)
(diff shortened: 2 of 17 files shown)


fubuloubu:
> * Do docs work / exist? Seems like there is a plan for readthedocs and also the link on the site for docs does not really go anywhere. I just want to make sure I have all the material.

No docs yet, definitely need to get on this

> * Plans for more thorough unit tests?

Yes, definitely needed soon. I got started on re-architecting the test suite a bit for a series of example projects that would be executed via parametrizing the `project` fixture

> * Loading the config at module load time seems bad because you miss out on available click features, such as `click.Exception`.

Can you explain this one more?

.gitignore
```diff
@@ -1,3 +1,6 @@
+# IDE
+.idea/
```
fubuloubu, line 2: Usally I ask people put their own IDE stuff in their global gitignore, saves on updating this file constantly

src/ape/managers/config.py
```diff
@@ -53,7 +53,7 @@ def __post_init__(self):
             self._plugin_configs[plugin_name] = config
 
         if len(user_config.keys()) > 0:
-            raise  # Unprocessed config items
+            raise Exception("Unprocessed config items")
```
fubuloubu, line 56: I like this, can you do this across the board really quick?

Should do better exceptions once #38 is finished

fubuloubu (approved):
Very nice! Thanks so much for doing this!

.gitignore
unparalleled-js: Ah ok, I don't remove this.
However, some projects include some shared IDE settings from this directory, so I can't put it in my global

unparalleled-js:
@fubuloubu Regarding loading the config file:

Right now, if you have any config-file related errors, you can't raise clean-looking CLI errors (via `Abort`). I am guessing this is because the config is parsed as soon as the python module is loaded since it happens globally. If we instead let `click` load this config file at a slightly later time, we can raise prettier errors if there are issues with it.

One way to do this might involve letting click's context object handle calling a method that parses and loads the config file... I could try to find an example..

fubuloubu (approved):

fubuloubu:
> Right now, if you have any config-file related errors, you can't raise clean-looking CLI errors (via `Abort`). I am guessing this is because the config is parsed as soon as the python module is loaded since it happens globally. If we instead let `click` load this config file at a slightly later time, we can raise prettier errors if there are issues with it.
> 
> One way to do this might involve letting click's context object handle calling a method that parses and loads the config file... I could try to find an example..

So, part of the problem here is that `ape` needs to be used both as a library (for writing tests and scripts e.g. `from ape import accounts`) and as a CLI application. Some of the plugins that need to be used are configurable via user config files, and needs to load prior to them being available. I had a heckuva time trying to rectify the loading sequence, and landed at the CLI being loaded via entrypoints, and all of the internals being loaded with the core package itself, including their configurations.

I'm definitely welcome to revisiting this, but think it's a good spot for another PR if you'd like to prototype how it might work differently.

PR #68, opened 2021-06-04
feat: Replace entry_points with add_command + fuzzy cmd suggestions

unparalleled-js:
* Removes entry_point list of 1st level plugins
* Imports them directly and calls add_command
* Refactor CLI namespace
* Includes a custom Group object that will output fuzzy suggestions if possible. e.g. type `ape acccount` and it will say `Error: No such command 'acccount'. Did you mean accounts?`


src/ape/_cli/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+@click.group(
+    cls=ApeCLI,
+    context_settings=CONTEXT_SETTINGS,
+    invoke_without_command=True,
+    no_args_is_help=True,
+)
+@shared_options()
```
fubuloubu, line 23: This should be the only place that these are used (`ape --version` at least)

I'm open to adding `ape_config` for first displaying the config and then potentially working with it too

src/ape/_cli/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+def cli():
+    ...
```
fubuloubu, line 25: `pass`?

src/ape_accounts/_cli.py
```diff
@@ -5,13 +5,14 @@
...
 from ape import accounts
+from ape._cli import ApeCLI
 from ape.utils import notify
 
 # NOTE: Must used the instantiated version of `AccountsContainer` in `accounts`
 container = accounts.containers["accounts"]
 
 
-@click.group(short_help="Manage local accounts")
+@click.group(cls=ApeCLI, name="accounts", short_help="Manage local accounts")
```
fubuloubu, line 15: Is `cls=ApeCLI` supposed to be there?

src/ape/_cli.py
```diff
@@ -1,66 +0,0 @@
...
-        if not self._commands:
-            entry_points = metadata.entry_points()
-
-            if "ape_cli_subcommands" not in entry_points:
-                raise Exception("Missing registered cli subcommands")
-
-            self._commands = {
-                clean_plugin_name(entry_point.name): entry_point.load
-                for entry_point in entry_points["ape_cli_subcommands"]
-            }
```
fubuloubu, line 41: So how would this change work with external plugins?

This code is how it's being handled now, via `entry_points`.

setup.py
```diff
@@ -74,13 +74,6 @@
     ],
     entry_points={
         "console_scripts": ["ape=ape._cli:cli"],
-        "ape_cli_subcommands": [
-            "ape_accounts=ape_accounts._cli:cli",
-            "ape_compile=ape_compile._cli:cli",
-            "ape_console=ape_console._cli:cli",
-            "ape_plugins=ape_plugins._cli:cli",
-            "ape_networks=ape_networks._cli:cli",
-        ],
```
fubuloubu, line 83: NOTE: The use of `entry_points` for these "1st party plugins" is also prototyping how 2nd/3rd party plugins would integrate into the CLI.

src/ape/_cli/__init__.py
unparalleled-js: What I would like to do is create a custom context object that is attached the CLI via a stateful object that gets set up automatically by reading the file.

Then, in order to use that config object, you'd have to re-use this `@shared_options` and it will give you another param. It won't matter that it's also here.

I have done this same pattern in a few other click projects to handle automatically creating an HTTP client for a particular service. I have also done a JSON config type situation.

src/ape/_cli/__init__.py
unparalleled-js: Yeah, I can use `pass`

unparalleled-js: The reason I used `...` instead is because I saw it somewhere else in the code base I am pretty sure, and it's the same thing, right?

src/ape_accounts/_cli.py
unparalleled-js: Yess

unparalleled-js: I made `ApeCLI` a different type than it was before

setup.py
unparalleled-js: If we go with `click.extensions`, we can have 2nd/3rd party plugins be installed via `pip` instead

src/ape/_cli.py
unparalleled-js: For external plugins, they would install a separate `setup.py` which will it attach it to the `ape` CLI under the hood! So we wouldn't have to support this custom code for doing it

src/ape/_cli/__init__.py
fubuloubu: ah gotcha, okay I'll keep my eye out for that

Could we at least split them into the `version_option` and the `config_option`?

src/ape/_cli/__init__.py
fubuloubu: I used it with abstract base classes, I think `abstractmethod` uses `...` to determine abstract methods somehow

src/ape_accounts/_cli.py
fubuloubu: Should it be with all the rest of them then?

src/ape/_cli.py
fubuloubu: I'm definitely looking forwards to this, this code isn't in the PR yet though right? I couldn't see it

setup.py
fubuloubu: Would we be able to have the `ape_*` packages under `src` install in this way? would be nice to use the same mechanism for all of them.

src/ape_accounts/_cli.py
unparalleled-js: Oh oops, you are right, this shouldn't be here

src/ape/_cli.py
unparalleled-js: I can either do multiple small PRs in a staging branch or just keep pushing to this branch

setup.py
unparalleled-js: We can add `setup.py` to those as well!

fubuloubu: I'm open to restructuring the repo to be a series of projects, I just wanted it to be all installed via 1 `setup.py`

src/ape/_cli.py
fubuloubu: Seems like would be easier to work with another PR

src/ape/_cli/__init__.py
fubuloubu:
```suggestion
    pass
```

setup.py
unparalleled-js: Ok, I can look into this. I am pretty sure it's possible to install local `setup.py`s from a `setup.py` but that might not be a good solution in the end.

Does this mean users will not be able to uninstall them, or do we still want that?

I am curious as to the benefit of this. I understand that a consistent plugin installation is nice, but these might as well get installed the easy way -- by being imported / packaged together.

fubuloubu: We don't have to do this change in this PR, I think this is okay for now.

> Does this mean users will not be able to uninstall them, or do we still want that?

Yeah, the "1st class plugins" should never be uninstallable, they're sort of the core set of functionality required for ape to do anything useful at all. But still, I wanted them to be "1st class" instead of directly into the Core package (e.g. `ape`) to 1) help prototype the plugin interfaces, and 2) create a distinction between the core APIs and management functions and the "core" CLI/plugins provided by installing the package

fubuloubu: Small point: I like having the CLI endpoints set by the plugin's name (which is set via package name) to prevent malicious plugins from overriding core plugin's functionality. As these plugins have access to sensitive data stored on disk or provide privileged access to signing keys, I think it's important to take a reasonable amount of care to prevent impact as much as we can.

unparalleled-js: We should need to stress this.. Users can always use `pip` on their own without `ape plugins add`, which may cause these issues?

unparalleled-js:
closing for now but will keep in mind

setup.py
fubuloubu: Yeah, the idea wasn't necessarily to block people from using `pip`, but give a default endpoint that avoided some gotchas. Can always reconfigure manually with `pip`

PR #70, opened 2021-06-09
refactor: change how tests are organized and run

fubuloubu:
Tests before were not organized well and sparse, making it difficult to ascertain how it is supposed to work


tests/integration/cli/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+import ape
+from ape import Project
+
+TEST_PROJECTS_FOLDER = Path(__file__).parent / "data" / "projects"
+
+
+@pytest.fixture(params=[p for p in TEST_PROJECTS_FOLDER.iterdir() if p.is_dir()])
+def project_folder(request, config):
+    project_source_dir = request.param
+    project_folder = config.PROJECT_FOLDER / project_source_dir.name
```
lost-theory, line 16: Doesn't seem to cause problems but using `project_folder` as the function name and a variable name made me do a double take (to make sure it wasn't recursive).

lost-theory (approved):
LGTM. Maybe add a module level docstring to the `__init__.py`s explaining what the difference is between functional and integration.

PR #78, opened 2021-06-28
feat: add script runner

fubuloubu:
### What I did

https://github.com/orgs/ApeWorX/projects/1#card-54989179

TODO:
- [ ] Pass context variables to `cli.Group`/`cli.Command` invocation


src/ape_run/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+from ape.utils import Abort
+
+
+def _run_script(script_file, interactive=False):
+    # Load the python module and find our hook functions
+    ns: dict = {}
+    try:
+        py_module = compile(script_file.read_text(), str(script_file), "exec")
+        eval(py_module, ns, ns)
```
lost-theory, line 13: Not sure what the tradeoffs would be but I think you can also do this step using `importlib`:

https://docs.python.org/3/library/importlib.html

lost-theory (approved):
LGTM. Left one minor observation.

src/ape_run/__init__.py
fubuloubu: `importlib` seems clearer!

lost-theory (approved):
LGTM.

src/ape_run/_cli.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _run_script(script_file, interactive=False):
+    # Load the python module and find our hook functions
+    import_str = ".".join(script_file.parts[:-1] + (script_file.stem,))
+    try:
+        py_module = import_module(import_str)
+
+    except Exception as e:
+        raise Abort(e.with_traceback()) from e
```
lost-theory, line 20: I think you want:

```
raise Abort().with_traceback(e.__traceback__)
```

Or:

```
raise Abort() from e
```

As-is this causes an error:

```
TypeError('with_traceback() takes exactly one argument (0 given)')
```

lost-theory (commented):
Actually I do see one issue from CI. Left a comment.

PR #87, opened 2021-07-13
refactor: warning on duplicate package

benjyz:


fubuloubu (approved):
I think I might want it to raise instead, because it sort of hurts the next step of picking the first

But I could be convinced to leave it at WARNING or ERROR (without failing out)

Issue #3, opened 2021-02-16
Argument Conversion API


fubuloubu:
## API Objects
(Implemented by plugin)

```python
class ArgumentConversionAPI(Generic[ABIType]):
    def is_convertible(self, value: str) -> bool:
        """
        Returns `True` if string value provided by `value` is convertible using `self.convert(value)`
        """

    def convert(self, value: str) -> ABIType:
        """
        Implements any conversion logic on `value` to produce `ABIType`.

        Must throw if not convertible.
        """
```

## API Registration Functions
(Defined by plugin)

```python
def ape_register_argument_conversion() -> Iterable[Tuple[arg_type: ABIType, converter: ArgumentConversionAPI]]: ...
```

PR #89, opened 2021-07-13
feat: add conversion api

fubuloubu:
### What I did
Added support for argument conversions via API

fixes: #3 
~~depends: #92~~

### How I did it
Added `ConversionManager` and `ConverterAPI`

### How to verify it
(In `ape console`):

```python
>>> convert("uint256", "1 ETH") == int(1e18)
True
```

### Checklist

- [x] Integrate deeper into relevant sections of the codebase
- [x] Test with https://github.com/ApeWorX/ape-tokens
- [x] Test with https://github.com/ApeWorX/ape-ens


fubuloubu:
Python 3.6 bug, see: https://github.com/biqqles/dataclassy/issues/52

sabotagebeats:
I was trying to verify it but I got the following error: 

```
In [2]: convert("uint256", "1 ETH") == int(1e18)
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-2-51251206c2a7> in <module>
----> 1 convert("uint256", "1 ETH") == int(1e18)

~/work/apeworx/ape/src/ape/managers/converters.py in convert(self, value, type)
     83         if type not in self._converters:
     84             converter_types = ", ".join(map(lambda t: t.__name__, self._converters))
---> 85             raise Exception(f"ABI Type '{type}' must be one of [{converter_types}]")
     86
     87         if self.is_type(value, type):

Exception: ABI Type '1 ETH' must be one of [ChecksumAddress, bytes, int]
```

sabotagebeats:
I put a breakpoint at 83: 
```
(Pdb) type
'1 ETH'
(Pdb) type in (self._converters)
False
```

sabotagebeats:
```
In [1]: convert("1 Eth",int) == int(1e18)
Out[1]: True

In [2]: convert("1 Eth",int)
Out[2]: 1000000000000000000
```

fubuloubu:
> I was trying to verify it but I got the following error:

I swapped the order of the two, does it work now?

src/ape/managers/converters.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def is_type(self, value: Any, type: Type) -> bool:
+        if type is AddressType:
+            return is_checksum_address(value)
+
+        else:
+            return isinstance(value, type)
+
+    def convert(self, value: Any, type: Type) -> Any:
+        if type not in self._converters:
```
sabotagebeats, line 83: `convert("1 Eth",int)` works but `convert("1 Eth","uint256")` or `convert("uint256","1 Eth")` throw because `uint256` is not a type included in `self._converters`

sabotagebeats (changes requested):
@fubuloubu no i think it has to be `int` not `"uint256"`
> I swapped the order of the two, does it work now?

```
In [1]: convert("1 Eth","int") == int(1e18)
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-1-0c519f26f9f1> in <module>
----> 1 convert("1 Eth","int") == int(1e18)

~/work/apeworx/ape/src/ape/managers/converters.py in convert(self, value, type)
     83         if type not in self._converters:
     84             converter_types = ", ".join(map(lambda t: t.__name__, self._converters))
---> 85             raise Exception(f"ABI Type '{type}' must be one of [{converter_types}]")
     86
     87         if self.is_type(value, type):

Exception: ABI Type 'int' must be one of [ChecksumAddress, bytes, int]
```

src/ape/managers/converters.py
fubuloubu: Yes, this is as expected. These converters convert to "raw" python types, not ABI types. The ABI conversions happens in another spot in `api/contracts.py`

tests/functional/conversion/test_ether.py
```diff
@@ -0,0 +1,31 @@
...
+    value=st.decimals(
+        min_value=-(2 ** 255),
+        max_value=2 ** 256 - 1,
+        allow_infinity=False,
+        allow_nan=False,
+    ),
+    unit=st.sampled_from(list(ETHER_UNITS.keys())),
+)
+def test_ether_conversions(value, unit):
+    assert convert(value=f"{value} {unit}", type=int) == int(value * ETHER_UNITS[unit])
```
sabotagebeats, line 21: this test looks right, and passes

sabotagebeats (approved):

sabotagebeats:
@fubuloubu ok looks good to me as long as those caveats I listed above - 

to use syntax needs to be like `convert("1 Eth",int)`

Issue #85, opened 2021-07-13
compile filepath


benjyz:
### Environment information

* `ape` and plugin versions:

0.1.0a14.dev2+gf51ff39

* Python Version: x.x.x
* OS: osx/linux/win

### What went wrong?

I call compile and have a greeter.sol contract in contracts folder, but not found.
```
ape compile
WARNING: No project files detected in '/Users/ben/dev/apetest/contracts'
ls contracts 
Greeter.sol
```

### How can it be fixed?

investigating

benjyz:
more info

```
INFO: no filepath specified
INFO: project.path '/Users/ben/dev/apetest'
INFO: project.sources []
WARNING: No project files detected in '/Users/ben/dev/apetest/contracts'. 
contract_filepaths: '[]'
```

fubuloubu:
So this bug happens when you have no compiler plugins installed. The error message should be better (maybe summarizing the extensions and saying "no compiler installed to handle extension `*/sol`")

PR #99, opened 2021-07-20
fix: error warning not accurate for uninstalled compiler

sabotagebeats:
### What I did

changed error messages to be more reflective of actual error regarding uninstalled compiler. 

Related issue: # fix: https://github.com/ApeWorX/ape/issues/85

### How I did it

modified the compiler _cli file to look for .vy and .sol files in os.listdir contract folder to base error messages on

### How to verify it

try to compile vyper files with no vyper plugin, it should tell you no vyper plugin is installed. same with solidity plugin. 
if there are no .vy or .sol in contracts folder it should say so, and if there is no contracts folder it should say so. 

### Checklist

### DIFF
--- a/src/ape_compile/_cli.py
+++ b/src/ape_compile/_cli.py
@@ -1,5 +1,4 @@
 from itertools import chain
-from os import listdir
 from pathlib import Path
 
 import click
--- a/src/ape_compile/_cli.py
+++ b/src/ape_compile/_cli.py
@@ -1,5 +1,6 @@
 from itertools import chain
 from pathlib import Path
+from os import listdir
 
 import click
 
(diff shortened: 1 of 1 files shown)


src/ape_compile/_cli.py
```diff
@@ -67,7 +68,13 @@ def cli(filepaths, use_cache, display_size):
...
             selected_paths = str(project.path / "contracts")
 
-        notify("WARNING", f"No project files detected in '{selected_paths}'")
+        if any(".sol" or ".vy" in s for s in listdir(selected_paths)):
+            if any(".sol" in s for s in listdir(selected_paths)):
+                notify("WARNING", "The required compiler plugin Ape-Solidity is not installed")
+            if any(".vy" in s for s in listdir(selected_paths)):
+                notify("WARNING", "The required compiler plugin Ape-Vyper is not installed")
+        else:
+            notify("WARNING", f"No contract files detected in '{selected_paths}'")
```
fubuloubu, line 77: So, for ape core, we don't take any particular view on which compilers to install, but it is based on extension so we can make that more apparent for the error message
```suggestion
        extensions = ", ".join(set(f.suffix for f in selected_paths))
        notify("WARNING", f"No compilers detected for the following extensions: {extensions}")
```

sabotagebeats: looks good to me :)

fubuloubu (approved):

PR #103, opened 2021-07-22
fix: add support for eth-account v0.5.5

fubuloubu:
### What I did
`eth-account` v0.5.5 was just released today, causing an incompatibility (because we are using an internal-only set of functions to encode transaction types)

### How I did it
duplicated the name change


sabotagebeats:
Ape broke for me when I pulled this. v0.1.0a16 still works for me though. I'll file an issue.

sabotagebeats:
https://github.com/ApeWorX/ape/issues/106

Issue #110, opened 2021-07-30
Have HTTP Provider raise if the RPC value of Chain ID is not as expected


fubuloubu:

fubuloubu:
basically if `self._web3.eth.chain_id != self.network.chain_id` then raise

PR #113, opened 2021-07-30
fix: raise if not connected to correct chain via HTTP provider

fubuloubu:
fixes: #110
depends: #114


fubuloubu:
Checkout integrating https://hub.docker.com/r/ethereum/client-go to resolve the CI issue

PR #131, opened 2021-09-01
chore: upgrade python syntax

unparalleled-js:
### What I did

I ran `pyupgrade` to upgrade the code base to use syntax for python 3.7 or greater.

During the last meeting, I showed some of the changes. I have concluded that:

1.) The team does not always want to use the latest syntax during cases they find it confusing
2.) The team is not immediately attracted to the idea of another pre-commit hook modifying code.

And also, some contrary things about `pyupgrade`:

1.) `pyupgrade` works best a pre-commit hook. This is because it takes a list of python files as arguments.
2.) `pyupgrade` does not have an opt-out feature natively, but there is a fork that does.

Thus, it does not seem like a good time to automate the usage of `pyupgrade` because of these conflicts.

Related issue: # n/a

### How I did it

I made the precommit hook, ran it, and then removed the precommit. I could have instead ran `pyupgrade` on all the python files.  (also used the `--py37-plus` flag).

### How to verify it

Ensure the changes align with practices we believe are good, as far as using upgraded syntax goes.

### Checklist

### DIFF
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -144,7 +144,7 @@ def load_config(path: Path, expand_envars=True, must_exist=False) -> Dict:
         return config or {}
 
     elif must_exist:
-        raise IOError(f"{path} does not exist!")
+        raise OSError(f"{path} does not exist!")
 
     else:
         return {}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ cd ape-accounts
 python3 -m venv venv
 source venv/bin/activate
 
-# install brownie into the virtual environment
+# install ape into the virtual environment
 python setup.py install
 
 # install the developer dependencies (-e is interactive mode)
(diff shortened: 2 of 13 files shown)


src/ape/utils.py
```diff
@@ -144,7 +144,7 @@ def load_config(path: Path, expand_envars=True, must_exist=False) -> Dict:
         return config or {}
 
     elif must_exist:
-        raise IOError(f"{path} does not exist!")
+        raise OSError(f"{path} does not exist!")
```
unparalleled-js, line 147: IOError was removed from the exception hierarchy since python 3.3: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

it is an alias for OSError.

Here is a GitHub action where this is is discussed: https://github.com/asottile/pyupgrade/issues/141

I kind of like `IOError` sometimes, but I am not sure here is one of those times...

fubuloubu (approved):

PR #134, opened 2021-09-02
chore: refactor and test networks list command

unparalleled-js:
### What I did

Whilst trying to figure out how to add a network (similarly to `brownie networks add`), I noticed that the `network list` command could be refactored. I also noticed there was not a test for it, so I added that as well.

### How I did it

Pulled out shared logic, broke up the large command-method's implementation to use sub-methods, added a simple test.

### How to verify it

Run `ape networks list`, the output should be the same as it was before.

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -1,15 +1,15 @@
 from typing import Dict
 
-try:
-    from importlib import metadata  # type: ignore
-except ImportError:
-    import importlib_metadata as metadata  # type: ignore
-
 import click
 import yaml
 
 from ape.plugins import clean_plugin_name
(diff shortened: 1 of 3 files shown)


src/ape/_cli.py
```diff
@@ -1,16 +1,16 @@
...
-except ImportError:
-    import importlib_metadata as metadata  # type: ignore
-
 import click
 import yaml
 
 from ape.plugins import clean_plugin_name
 from ape.utils import notify
 
+try:
```
unparalleled-js, line 9: `isort` did this

fubuloubu (approved):

Issue #21, opened 2021-04-07
Add Docs


fubuloubu:
Options:
- https://www.mkdocs.org/
- https://pdoc3.github.io/pdoc/
- https://www.sphinx-doc.org/en/master/index.html

Tools:
- https://docs.python.org/3/library/doctest.html
- https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html
- https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
- https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html
- https://github.com/PyCQA/pydocstyle
- https://github.com/astropy/pytest-doctestplus

fubuloubu:
- https://github.com/squidfunk/mkdocs-material
- https://myst-parser.readthedocs.io/en/latest/using/intro.html
- https://github.com/mkdocstrings/mkdocstrings/
- https://www.it3.be/2020/08/14/using-mkdocs-with-gh-pages/

fubuloubu:
https://github.com/click-contrib/sphinx-click

lost-theory:
@sabotagebeats and I talked about splitting up the docs work on Monday. Here are some next steps:

* eip712 `feature/sphinx-docs` branch:
  * [ ] add the doctest and autosummary extensions
  * [ ] figure out how to build and host the docs on Github Pages
  * [ ] merge the branch and port the changes over to `project-template`
* [ ] copy the `project-template` Sphinx configs to each repo (`ape` at least, maybe the plugin repos too?)
* [ ] increase documentation coverage everywhere Sphinx is set up (documenting each module, class, and function)

fubuloubu:
## End User Docs
- "`ape` as an application" usage guide
  - Project layout
  - Configuration file
  - "core plugin CLI" endpoint usage guide (Click generated docs?)
    - `ape accounts`
    - `ape compile`
    - `ape test`
    - `ape console`
  - Plugin installation guide
    - `ape plugins`
- "`ape` as a package" usage guide
  - `project` object (and `Project` class)
  - `accounts` object
  - `networks` object
  - other objects... (`compilers`, etc.)
- Using the Console
  - Inspecting and Debugging Transactions
- Writing and Using Scripts
- Writing and Running Tests
  - Debugging Tests
  - "Fork Mode" testing
  - Measuring and Viewing Coverage
  - Property Tests
  - Stateful Testing
- Deploying Packages

## Plugin Writer Docs
- Overall architecture flow
[...]

PR #135, opened 2021-09-02
docs: work in progress

sabotagebeats:
### What I did
adding docs, work in progress

Related issue: # https://github.com/ApeWorX/ape/issues/21

### How I did it
Adding docs in docs folder

### How to verify it
build and review

### Checklist

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -1,0 +1,66 @@
+Writing Plugins
+###############
+
+Overall architecture flow
+*************************
+
+Plugin Registration process flow
+================================
+    API Object registration
+    CLI registration
+
+
+Compilation process flow
+========================
+    contracts/ folder
+    Project manager contains all the items in the projects folder, 
+    including the contracts folder. 
+    The contracts folder is where the compiler will be looking for your contracts in order to compile them.
+    The file extension is used to determine which compiler extension should be used to compile the contract code.
+    
+    The contracts are then grouped by compiler type and fed into the compiler to compile them.
+    
+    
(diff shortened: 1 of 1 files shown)


src/ape/api/compiler.py
```diff
@@ -13,6 +13,9 @@ class CompilerAPI:
     def name(self) -> str:
         ...
 
+    """
+    Retrieve avaiailable compiler versions
+    """
     @abstractmethod
     def get_versions(self, all_paths: List[Path]) -> Set[str]:
         ...
```
fubuloubu, line 21: You put the docstrings inside the method like so (we don't need the `...` when we add a docstring)

```suggestion
    @abstractmethod
    def get_versions(self, all_paths: List[Path]) -> Set[str]:
        """
        Retrieve available compiler versions
        """
```

unparalleled-js:
could you expand on the instructions in `How to verify it`?


Something like ?
```
python build_docs.py
python -m http.server --directory "_build/html" 1337
```

and then 

```
open http://127.0.0.1:1337
```

(I am not exactly sure)

fubuloubu:
> could you expand on the instructions in `How to verify it`?
> 
> Something like ?
> 
> ```
> python build_docs.py
> python -m http.server --directory "_build/html" 1337
> ```
> 
> and then
> 
> ```
> open http://127.0.0.1:1337
> ```
> 
> (I am not exactly sure)

Yes, exactly. Sometimes it helps to display larger sections that have new rendering to see if it looks/reads right

Issue #132, opened 2021-09-01
Custom Exceptions


unparalleled-js:
### Overview

Implement custom exceptions. Replace the usage of `raise Exception("message")` (where apt) with well-defined custom exceptions that can be imported by users.  This allows users to catch the exceptions in their scripts and/or plugins. Also, this allows the CLI to catch custom exceptions and handle them by outputting them to the user in a CLI-like way.

### Specification

Allow exceptions to be importable from a module in "ape core".
Replace the usage of `raise Exception("message") with raising the custom exception
Handle errors from Ape Core in the CLI

### Dependencies

Coordination with the object logging feature (#57), but not dependent on it.

unparalleled-js:
Work completed

PR #136, opened 2021-09-03
fix: accounts failure exit codes

unparalleled-js:
### What I did

Fixes issue where `ape accounts import <alias>` and `ape accounts generate <alias>` would return 0 exit codes when they failed to import or generate accounts.

Additionally, changed `notify()` to detect when it's `ERROR` and print to `stderr` instead. Thus, `Abort()` can just use `notify()` now in its implementation.

I added tests - I am trying to get a sense of how to write tests for this project.

Also, I added `pytest-mock`, I hope that is okay? It is magical!

This is all part of the efforts of custom exceptions + musing about exception handling. It by no means resolves that issue (I have another branch for that), but it is helpful to have while thinking about exceptions as a whole.

Related issue: #132

### How I did it

I tried to write a unit test and it failed so I realized it was not doing what I thought was expected behavior.l

### How to verify it

Do:

```
ape accounts import <alias-that-already-exists>
echo $?
>>> 1
```

Notice that it's `1` now (was `0` before!)

### Checklist

### DIFF
--- a/src/ape_accounts/_cli.py
+++ b/src/ape_accounts/_cli.py
@@ -6,7 +6,7 @@ from eth_account import Account as EthAccount  # type: ignore
 from eth_utils import to_bytes
 
 from ape import accounts
-from ape.utils import notify
+from ape.utils import Abort, notify
 
 # NOTE: Must used the instantiated version of `AccountsContainer` in `accounts`
 container = accounts.containers["accounts"]
@@ -57,8 +57,7 @@ def _list():
 @click.argument("alias")
 def generate(alias):
     if alias in accounts.aliases:
-        notify("ERROR", f"Account with alias '{alias}' already exists")
-        return
+        raise Abort(f"Account with alias '{alias}' is already in use")
 
     path = container.data_folder.joinpath(f"{alias}.json")
     extra_entropy = click.prompt(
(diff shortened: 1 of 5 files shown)


src/ape_accounts/_cli.py
```diff
@@ -57,8 +57,7 @@ def _list():
 @click.argument("alias")
 def generate(alias):
     if alias in accounts.aliases:
-        notify("ERROR", f"Account with alias '{alias}' already exists")
-        return
+        raise Abort(f"Account with alias '{alias}' is already in use")
```
unparalleled-js, line 60: This is the bugfix, raising an Exception. It has the same effect, as far as CLI output goes, but the exist code is 1 instead of 0.

This is in `_import()` as well

tests/integration/cli/test_accounts.py
```diff
@@ -61,25 +65,52 @@ def test_account():
...
+    result = invoke_import()
+    assert result.exit_code == 0
+    result = invoke_import()
+    assert_failure(result, f"Account with alias '{ALIAS}' is already in use")
+
+
+def test_import_account_instantiation_failure(
+    mocker, ape_cli, runner, test_account, test_keyfile_path
+):
+    eth_account_from_key_patch = mocker.patch("ape_accounts._cli.EthAccount.from_key")
```
unparalleled-js, line 89: Using `.patch` here because the error can many different things from `EthAccount.from_key` (our dependency in this command).

fubuloubu (approved):
Great catch!

PR #138, opened 2021-09-08
feat: custom exceptions

unparalleled-js:
### What I did

Created and used custom Ape exceptions

Related issue: #132

### How I did it

Created a python module `ape.exceptions` that contains custom exceptions that end users can easily import if needed (for either raising in their custom plugin implementations _or_ handling when `ape` raises them naturally).

Then, I used these custom exceptions 

### How to verify it

* Make sure ape still works as it did before
* Make sure exceptions make sense and named in a way that is agreeable and include data that is helpful for users who are debugging.
* Make sure there are not more places where custom exceptions seem useful that should be added as part of this work.

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -1,6 +1,6 @@
 import difflib
 import re
-from typing import Any, Dict
+from typing import Dict
 
 import click
 import yaml
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -1,10 +1,11 @@
-from typing import Dict
+from typing import Any, Dict
 
 import click
 import yaml
 
+from ape.exceptions import ApeException
 from ape.plugins import clean_plugin_name
-from ape.utils import notify
+from ape.utils import Abort, notify
 
 try:
     from importlib import metadata  # type: ignore
@@ -28,13 +29,19 @@ def display_config(ctx, param, value):
 class ApeCLI(click.MultiCommand):
     _commands = None
 
+    def invoke(self, ctx: click.Context) -> Any:
+        try:
+            return super().invoke(ctx)
+        except ApeException as err:
+            raise Abort(str(err)) from err
+
     @property
     def commands(self) -> Dict:
         if not self._commands:
             entry_points = metadata.entry_points()
 
             if "ape_cli_subcommands" not in entry_points:
-                raise Exception("Missing registered cli subcommands")
+                raise Abort("Missing registered cli subcommands")
 
             self._commands = {
                 clean_plugin_name(entry_point.name): entry_point.load
@@ -50,8 +57,11 @@ class ApeCLI(click.MultiCommand):
         if name in self.commands:
             try:
                 return self.commands[name]()
-            except Exception:
-                notify("WARNING", f"Error loading cli endpoint for plugin 'ape_{name}'")
+            except Exception as err:
+                notify(
+                    "WARNING",
+                    f"Unable to load CLI endpoint for plugin 'ape_{name}'.\n\t{err}",
+                )
 
         # NOTE: don't return anything so Click displays proper error
 
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -1,6 +1,6 @@
 import difflib
 import re
-from typing import Dict
+from typing import Any, Dict
 
 import click
 import yaml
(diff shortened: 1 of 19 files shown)


src/ape/exceptions.py
```diff
@@ -0,0 +1,309 @@
+from typing import List, Set, Type
+
+
+class ApeError(Exception):
```
fubuloubu, line 4: Prefer `ApeException` for the base class case

fubuloubu: Also wonder if we should subclass `Abort` too from `click` :thinking:, so all unhandled exceptions cause a CLI Abort by default

unparalleled-js: Yes, we could do that way.
It does have that behavior currently because of the overridden `invoke()` method that catches it and raises `Abort`.
, but perhaps using that base class is easier.

I was trying to not worry about click or CLI related stuff for the exceptions module though, but maybe unnecessarily

fubuloubu: Yeah, I mean maybe not super over-complicate it, but if there's a way it makes sense to simplify our handling of the command line, this would be the way to do it

src/ape/_cli.py
```diff
@@ -28,13 +29,19 @@ def display_config(ctx, param, value):
 class ApeCLI(click.MultiCommand):
     _commands = None
 
+    def invoke(self, ctx: click.Context) -> Any:
+        try:
+            super().invoke(ctx)
```
fubuloubu, line 34: because of the return `Any`, should this statement be:
```suggestion
            return super().invoke(ctx)
```

src/ape/utils.py
```diff
@@ -154,7 +158,7 @@ def compute_checksum(source: bytes, algorithm: str = "md5") -> str:
     if algorithm == "md5":
         hasher = md5
     else:
-        raise Exception("Unknown algorithm")
+        raise ApeException(f"Unknown algorithm `{algorithm}`")
```
fubuloubu, line 161: Should avoid using the Base exception directly when raising (catching is fine as a catchall)

src/ape/exceptions.py
unparalleled-js: We could subclass `Abort` (or just move `Abort` to `exceptions` and rename it to `ApeException`) but we can not use the shared util method `notify` anymore.... Is that better or worse than the way it is now, where we catch `ApeException` and raise `Abort`? I could go either way.

src/ape/utils.py
unparalleled-js: Should be resolved now, everywhere. Thanks

src/ape/exceptions.py
fubuloubu: for sure, I guess we can add a blanket `except` for `ApeException` and `raise Abort(...) from e` for that

src/ape/_cli.py
```diff
@@ -28,13 +29,19 @@ def display_config(ctx, param, value):
 class ApeCLI(click.MultiCommand):
     _commands = None
 
+    def invoke(self, ctx: click.Context) -> Any:
+        try:
+            return super().invoke(ctx)
+        except ApeException as err:
+            raise Abort(str(err))
```
unparalleled-js, line 36: @fubuloubu

fubuloubu: To preserve the stack trace
```suggestion
            raise Abort(str(err)) from err
```

unparalleled-js: TIL

unparalleled-js:
How do we feel about custom warnings?

We could something like 

```
def _format_warning(message, *args, **kwargs):
    """ Return a warning message, formatted for Visual Studio """
    return f"{style_message('WARNING', message)}\n"


warnings.formatwarning = _format_warning
```

where `style_message` is the styling part from the `notify()` method.

This would allow us to use `warnings.warn("warn warn warn", ApeWarning)` and it would have the same behavior

fubuloubu (approved):

fubuloubu:
> How do we feel about custom warnings?
> 
> We could something like
> 
> ```
> def _format_warning(message, *args, **kwargs):
>     """ Return a warning message, formatted for Visual Studio """
>     return f"{style_message('WARNING', message)}\n"
> 
> 
> warnings.formatwarning = _format_warning
> ```
> 
> where `style_message` is the styling part from the `notify()` method.
> 
> This would allow us to use `warnings.warn("warn warn warn", ApeWarning)` and it would have the same behavior

ummmm, not really sure I have an opinion, I haven't used warnings like this very much

PR #140, opened 2021-09-09
fix: help text fixes

unparalleled-js:
### What I did

* Corrects help text in yes options for plugins (said “remove” in places where it should have said “add”)
* Adds missing help text for the verbose option in the `ape console` command
* Adds shorthand `-v` to the verbose option in the `ape console` command
* Moves shared CLI options to an `options.py` module, including the `network_option` plus its choice class

### How I did it

Noticed that `NetworkChoice` was being imported from one 1st class plugin to another, and that had a bad smell to it.
Sought to move it to a share module for plugins to use, but then noticed that the `--yes` option in the `plugins` command had mistakenly said `remove` when it should have said `add`, so fixed that, since I was analyzing shared-options anyway, it felt like it fit. I then noticed that the verbose option could also be shared but more importantly, it was missing a `help` text in the `ape console` command.

### How to verify it

1.) Do

```
ape console --help
```

Notice there is now a help text for the `--verbose` options.

2.) Do

```
ape plugins add --help
ape plugins remove --help
ape plugins install --help
```

Notice each of the `--yes` option makes sense per the command (before, they all just spoke like it was for the `remove` command).

### Checklist


fubuloubu (approved):
I was thinking about positioning this under a `cli_utils` submodule or something, so that plugins can also import and make use of these features. It might also make sense to move `Abort` and some of the things from `utils` to there as well.

unparalleled-js:
> I was thinking about positioning this under a `cli_utils` submodule or something, so that plugins can also import and make use of these features. It might also make sense to move `Abort` and some of the things from `utils` to there as well.

I like that, we'd have 2 utils, one that imports `click` and one that does not.

If you want, I can sneak those changes in this PR?

Changes:

* rename `options.py` to `cli_utils.py`
* Move `Abort` and `notify` to `cli_utils.py`

fubuloubu:
> > I was thinking about positioning this under a `cli_utils` submodule or something, so that plugins can also import and make use of these features. It might also make sense to move `Abort` and some of the things from `utils` to there as well.
> 
> I like that, we'd have 2 utils, one that imports `click` and one that does not.
> 
> If you want, I can sneak those changes in this PR?
> 
> Changes:
> 
> * rename `options.py` to `cli_utils.py`
> * Move `Abort` and `notify` to `cli_utils.py`
> 
> Edit: Else, if we think it is important to have a changelog entry for the new `cli_utils`, I could do that in a separate PR with that as the title, so they both end up showing up there.

Thinking about this more, it might make sense to have a public `cli` submodule in `ape`, used for plugin writers building their own CLIs. Then, this kind of stuff can be under there e.g.

```
src/
  ape/
    ...
    cli/
      options.py
      utils.py
    ...
    _cli.py  # still keep this, it's not "public"
```

PR #142, opened 2021-09-10
refactor: create cli module

unparalleled-js:
### What I did

Create a share CLI module

Related to: #145

### How I did it

Discussed here: #140  on how to refactor the CLI to re-use shared options and utils, both for us and for 3rd party plugins.

### How to verify it

Make sure you can import the shared CLI definitions

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -5,9 +5,9 @@ from typing import Any, Dict
 import click
 import yaml
 
+from ape.cli.utils import Abort, notify
 from ape.exceptions import ApeException
 from ape.plugins import clean_plugin_name
-from ape.utils import Abort, notify
 
 try:
     from importlib import metadata  # type: ignore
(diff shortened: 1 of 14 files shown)


src/ape/_cli.py
```diff
@@ -5,9 +5,9 @@
 import click
 import yaml
 
+from ape.cli.utils import Abort, notify
```
fubuloubu, line 8: We can probably just export these at the top level
```suggestion
from ape.cli import Abort, notify
```

src/ape_run/_cli.py
```diff
@@ -5,8 +5,9 @@
 import click
 
 from ape import config, networks
-from ape.options import network_option, verbose_option
-from ape.utils import Abort, get_relative_path
+from ape.cli.options import network_option, verbose_option
```
fubuloubu, line 8: Same here
```suggestion
from ape.cli import network_option, verbose_option
```

src/ape/_cli.py
unparalleled-js: This creates a bit of stir, regarding circular imports and stuff..
I am going to try and dive and deeper into the structure and see what might be best.

unparalleled-js:
closing until logging work is done

PR #143, opened 2021-09-10
fix: remove production asserts

unparalleled-js:
### What I did

Remove a few `assert` statements from production code.
Reason:
When compiling optimized python coded (`.pyo` files), `assert` statements get removed. They should be avoided in non-debugging or testing scenarios.

fixes: #147 

More info:
https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html#b101-test-for-use-of-assert

Related issue: #132

### How I did it

Noticed the issue from a `bandit` scan.
Realized these should probably be errors instead of `assert` statements.
Changed them to raise `ValueError()` instead.

### How to verify it

I wrote unit tests for this one.

_however_ I am still not exactly sure where to put these types of functional unit tests. Should it be in more of an abstract collection such as `test_paths`?

I _really_ prefer tests to match more of the projects structure. This way, if I want to see some examples (As a contributor) to how a method or class works, I know exactly where to find it. It helps audit the tests, outside of just coverage. That is why I put it in `test_utils`. However, that does not seem to be the way it is done on this project. Help!

Writing the tests for `get_relative_path()` makes me completely understand how the method behaves (so I know I am not breaking it, in this case).

### Checklist

### DIFF
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -45,14 +45,15 @@ def get_relative_path(target: Path, anchor: Path) -> Path:
     which may or may not share a common ancestor.
     NOTE: Both paths must be absolute
     """
-    assert anchor.is_absolute()
-    assert target.is_absolute()
+    if not target.is_absolute():
+        raise ValueError("'target' must be an absolute path")
+    if not anchor.is_absolute():
+        raise ValueError("'anchor' must be an absolute path")
 
     anchor_copy = Path(str(anchor))
     levels_deep = 0
     while not is_relative_to(anchor_copy, target):
         levels_deep += 1
-        assert anchor_copy != anchor_copy.parent
         anchor_copy = anchor_copy.parent
 
     return Path("/".join(".." for _ in range(levels_deep))).joinpath(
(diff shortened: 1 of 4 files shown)


src/ape/utils.py
```diff
@@ -45,14 +45,15 @@ def get_relative_path(target: Path, anchor: Path) -> Path:
...
+    if not target.is_absolute():
+        raise ValueError("'target' must be an absolute path")
+    if not anchor.is_absolute():
+        raise ValueError("'anchor' must be an absolute path")
 
     anchor_copy = Path(str(anchor))
     levels_deep = 0
     while not is_relative_to(anchor_copy, target):
         levels_deep += 1
-        assert anchor_copy != anchor_copy.parent
```
unparalleled-js, line 55: Does this ever actually happen?  I could not produce this error so I removed it.

If this actually happens, may I have the scenario so I can handle it and write a test for it?

fubuloubu: I usually use asserts in scenarios where I'm _pretty sure_ the condition won't happen, but still want to catch it if it does

unparalleled-js: It's just that this `python` command option: https://docs.python.org/3/using/cmdline.html#cmdoption-o
Causes `assert` statements to get removed, which I don't think we want.

unparalleled-js:
- [ ] Make path in tests more generic (and did variable name too) - (feedback from meeting)

fubuloubu (approved):

Issue #57, opened 2021-05-27
Create a logger object and add `click-logging`


fubuloubu:
https://pypi.org/project/click-logging/

lost-theory:
+1 on a logging setup in general. I could use it with the `ape-hardhat` work (logging network calls, subprocess calls, retries, etc.).

lost-theory:
Also for dumping captured stdout/stderr from the hardhat process, like error messages.

fubuloubu:
@lost-theory let's talk about this tomorrow

lost-theory:
I started on this last week.

PR #146, opened 2021-09-13
feat: logging

unparalleled-js:
### What I did

Opened early because working with @lost-theory on this

fixes: #57 

### How I did it

1.) ^ Creates a logger that is injected into `click` and is able to have its verbosity set via `--verbosity` flag (`click-logging` bulitin), thanks @lost-theory 
2.) Replaces uses of `notify` with `click-logging`.  Meaning: if you did `notify("WARNING, msg)` before, now you do `logger.warning(msg)` and it is virtually the same.
3.) Create a `PluginHelper` class that has the logger as a dependency.  It also has an  `abort()` method. This is injectible into click commands (plugins).

### How to verify it

* Run commands and notice the same output as before
* Try using the `plugin_helper` in a plugin

### Checklist


src/ape/_cli.py
```diff
@@ -94,6 +95,7 @@ def get_command(self, ctx, name):
 
 
 @click.command(cls=ApeCLI, context_settings=dict(help_option_names=["-h", "--help"]))
+@plugin_helper()
```
unparalleled-js, line 98: @lost-theory so instead of having to import logger and the `click_logging` decorator, you can just import this `plugin_helper()` and you automatically get a class that has all the logging methods available.

It also has a helpful `.abort()` method.

src/ape/options.py
```diff
@@ -1,6 +1,37 @@
...
+        return self._logger.info(f"SUCCESS: {msg}")
+
+    @staticmethod
+    def abort(msg, base_error=None):
+        if base_error:
+            raise Abort(msg) from base_error
+        raise Abort(msg)
+
+
+def plugin_helper():
```
unparalleled-js, line 28: This decorator both sets the logger in `click_logging` as well as makes it a pass decorator to enable having access to the class where it is needed, all in one go

src/ape/utils.py
```diff
@@ -89,27 +91,12 @@ def get_package_version(obj: Any) -> str:
...
-    )
-
-
 class Abort(click.ClickException):
     """Wrapper around a CLI exception"""
 
     def show(self, file=None):
         """Override default ``show`` to print CLI errors in red text."""
-        notify("ERROR", self.format_message(), file=file)
+        logger.error(self.format_message())
```
unparalleled-js, line 99: `Abort()` now logs using the global `ape` logger.
`notify()` is not used anymore! Everything uses the logger.

The `CLI` gets a decorator helper that can log for users as well as `abort()`.

src/ape/options.py
```diff
@@ -1,6 +1,44 @@
 import click
+import click_logging
 
 from ape import networks
+from ape.logging import logger as logger
+from ape.utils import Abort
+
+
+class PluginHelper:
```
unparalleled-js, line 9: I was even wondering about making the managers that you normally access via `from ape import <manager<` accessible here, but maybe that is not that useful.

src/ape/_cli.py
```diff
@@ -86,15 +88,13 @@ def get_command(self, ctx, name):
             try:
                 return self.commands[name]()
             except Exception as err:
-                notify(
-                    "WARNING",
-                    f"Unable to load CLI endpoint for plugin 'ape_{name}'.\n\t{err}",
-                )
+                logger.warning(f"Unable to load CLI endpoint for plugin 'ape_{name}'.\n\t{err}")
```
unparalleled-js, line 91: Unable to use `ctx.obj.log_warning()` here because this happens before the context object exists

src/ape/logging.py
```diff
@@ -0,0 +1,80 @@
+import logging
```
unparalleled-js, line 1: This is main new work added

src/ape/logging.py
```diff
@@ -0,0 +1,80 @@
...
+CLICK_STYLE_KWARGS = {
+    "EXCEPTION": dict(fg="bright_red"),
+    "CRITICAL": dict(fg="bright_red"),
+    "ERROR": dict(fg="bright_red"),
+    "WARNING": dict(fg="bright_red"),
+    "INFO": dict(fg="blue"),
+    "DEBUG": dict(fg="blue"),
+    "SUCCESS": dict(fg="bright_green"),
+}
+CLICK_ECHO_KWARGS = {
```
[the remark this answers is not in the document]
fubuloubu: Only keep the ones we need yeah

src/ape/options.py
fubuloubu: See about, thinking of more descriptive names for this :thinking:

src/ape/options.py
```diff
@@ -5,6 +5,67 @@
...
+        self._logger = logger
+
+    def log_info(self, msg: str):
+        self._logger.info(msg)
+
+    def log_warning(self, msg: str):
+        self._logger.warning(msg)
+
+    def log_success(self, msg: str):
+        self._logger.success(msg)  # type: ignore
```
fubuloubu, line 27: Why not just let people use `self.logger` directly?

src/ape_compile/_cli.py
```diff
@@ -4,8 +4,8 @@
 
 import click
 
+from ape.options import plugin_helper
```
fubuloubu, line 7: Still think an `ape.cli` package containing types and helpers for building CLI extensions for ape should be exposed (different than `_cli.py` which is the actual cli endpoint for ape)

src/ape_accounts/_cli.py
```diff
@@ -56,41 +57,44 @@ def generate(alias):
...
+@plugin_helper()
+def _import(helper, alias):
     path = container.data_folder.joinpath(f"{alias}.json")
     key = click.prompt("Enter Private Key", hide_input=True)
     try:
         account = EthAccount.from_key(to_bytes(hexstr=key))
     except Exception as error:
-        raise Abort(f"Key can't be imported: {error}")
+        helper.abort(f"Key can't be imported: {error}")
+        return
```
fubuloubu, line 74: Isn't this unnecessary?

fubuloubu (approved):
Some smaller issues and discussion on semantic naming

src/ape/_cli.py
```diff
@@ -104,5 +104,5 @@ def get_command(self, ctx, name):
     callback=display_config,
     help="Show configuration options (using `ape-config.yaml`)",
 )
-def cli():
-    pass
+def cli(helper):
```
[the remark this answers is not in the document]
fubuloubu: Or maybe just `ape_cli`?

idk not great with names today

src/ape/options.py
```diff
@@ -5,6 +5,67 @@
...
+
+    @staticmethod
+    def abort(msg: str, base_error: Exception = None):
+        if base_error:
```
fubuloubu, line 31: Do we also want to log aborts?

src/ape_compile/_cli.py
unparalleled-js: Yes, I agree and have already implemented this but I kind of needed this logging stuff taken care first.
The `ape.cli` package will be a fast follow!

src/ape_accounts/_cli.py
unparalleled-js: The `return` is yes. I just put it there to make my IDE happy

src/ape/logging.py
```diff
@@ -1,51 +1,45 @@
...
 
 import click
 
 
-class Levels:
-    ERROR = "ERROR"
-    WARNING = "WARNING"
-    INFO = "INFO"
-    SUCCESS = "SUCCESS"
+class Level(Enum):
```
fubuloubu, line 10: Still think `LogLevel` to be a little clearer

src/ape_compile/_cli.py
fubuloubu: Sounds good!

src/ape_accounts/_cli.py
fubuloubu: Haha, okay

src/ape_compile/_cli.py
unparalleled-js: Well, maybe I can sort of initialize that here

src/ape_accounts/_cli.py
unparalleled-js: I will remove the return and suppress the warning

src/ape/_cli.py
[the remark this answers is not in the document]
unparalleled-js: Here are some ideas:

* `plugin_helper`
* `cli_helper`
* `cli_tool`
* `utilities`
* `ape`
* `plugin_tools`
* `plugin`
* `sdk`
* `ape_sdk`
* `ape_cli_sdk`

 So far I like `utilities`, `helper`, and `ape_cli_sdk` the best.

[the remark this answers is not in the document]
unparalleled-js: We often just call this `context` in click apps.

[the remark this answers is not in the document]
unparalleled-js: Maybe `ape_cli_tools` is the most descriptive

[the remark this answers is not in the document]
fubuloubu: `ape` would be totally confusing lmao

`ape_cli` would not be too bad (`ape_cli.logger`, `ape_cli.abort(...)`, etc.) It would be unlikely to be re-used in that way (e.g. `from ape import cli as ape_cli` wouldn't be as nice as `from ape.cli import verbose_option`)

Don't want to be too verbose here, as there will likely be many other arguments to such CLI functions. Food for thought.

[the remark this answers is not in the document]
unparalleled-js: `ape_cli_helper` is more like what it is though (or `ape_cli_context`). `ape_cli` makes me think it is an instance of a `click.Group` or something.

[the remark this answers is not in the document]
fubuloubu: Definitely better

fubuloubu (approved):

Issue #156, opened 2021-09-21
3rd party account APIs show up in `accounts list` even though can't necessarily do other `accounts` commands on them


unparalleled-js:
### Environment information

* ape from the main branch
* ape-ledger (my implementation branch)

```
$ ape --version
# 0.1.0a23 (local from source)

$ ape plugins list
# ...copy and paste result of above command here...
```
  ape_solidity (0.1.0a2)
  ape_trezor (0.1.0a2.dev0+g2bb7300.d20210916)
  ape_ledger (0.1.dev2+ge8f73a3.d20210921)
* Python Version: 3.9.7
* OS: osx

### What went wrong?

First, implement the `AccountAPI` in new plugin.

Second, add an account however you do that in your plugin

Third, try to remove the account using the native ape command.

```bash
ape accounts delete <alias>
```

`ape` is able to find and load the account and then it attempts to call a method called `delete()`. However, this method is not part of the `AccountAPI` and thus will only exist by chance

Error output: 

```
  File "/Users/nifl/Projects/ape/src/ape_accounts/_cli.py", line 114, in delete
    account.delete()
AttributeError: 'LedgerAccount' object has no attribute 'delete'
```

### How can it be fixed?

We should add an abstract method `delete` to the `AccountAPI` so that developers know that they may need to implement it.

unparalleled-js:
As it is encouraged for account plugins to create their own `delete` command, this still is an unexpected behavior.

I think it should attempt to call a Delete API, which should only fail with some sort of `NotImplementedError`()

fubuloubu:
Note that only the KeyfileAccount implements this method https://github.com/ApeWorX/ape/blob/main/src/ape_accounts/accounts.py#L105

This is probably confusing, since the `ape accounts` only really works with keyfile accounts, but it's the default account type we use here

unparalleled-js:
Should `ape accounts list` not include accounts from `ledger` and `trezor` (external account APIs) then? That would make users less likely to try other `ape accounts` commands on those and would resolve this issue.

unparalleled-js:
I am not sure if it's a big deal, but the `ape accounts delete <alias>` will still work for external accounts if their custom Account class happens to have a `delete()` method.

We _could_ check the type of the account when loading it in the `delete` command and raise a custom error...

unparalleled-js:
It seems like in general, if we want this, that we need a more comfortable way to loop through accounts by container type, re: the `__iter__` method of the `AccountsManager`.
[...]

PR #157, opened 2021-09-22
fix: limit ape_accounts plugin to KeyfileAccounts

unparalleled-js:
### What I did

fixes: #156 

### How I did it

Loop through the `accounts` container rather than all the accounts containers accounts.

### How to verify it

1.) Install another account API plugin
2.) Add an account
3.) Run `ape accounts list` and notice that account does not show up in the list

### Checklist

### DIFF
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -20,6 +20,10 @@ class AddressAPI:
 
         return self._provider
 
+    @provider.setter
+    def provider(self, value: ProviderAPI):
+        self._provider = value
+
     @property
     def _receipt_class(self) -> Type[ReceiptAPI]:
         return self.provider.network.ecosystem.receipt_class
--- a/src/ape/options.py
+++ b/src/ape/options.py
@@ -1,5 +1,6 @@
-import click
 from typing import List, Type
+
+import click
 
 from ape import accounts, networks
 from ape.api.accounts import AccountAPI
@@ -50,11 +51,15 @@ class Alias(click.Choice):
     @property
     def choices(self) -> List[str]:  # type: ignore
         # NOTE: This is a hack to lazy-load the aliases so CLI invocation works properly
-        return accounts.get_typed_aliases(self._account_type)
+        return [
+            a.alias
+            for a in accounts.get_accounts_by_type(self._account_type)
+            if a.alias is not None
+        ]
 
 
-def _require_non_existing_alias(arg, type_: Type[AccountAPI] = KeyfileAccount):
-    if arg in accounts.get_typed_aliases(type_):
+def _require_non_existing_alias(arg):
+    if arg in accounts.aliases:
         raise AliasAlreadyInUseError(arg)
     return arg
 
--- a/src/ape/options.py
+++ b/src/ape/options.py
@@ -1,6 +1,8 @@
 import click
+from typing import List
 
-from ape import networks
+from ape import accounts, networks
+from ape.exceptions import AliasAlreadyInUseError
 
 
 class NetworkChoice(click.Choice):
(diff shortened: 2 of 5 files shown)


unparalleled-js:
- [ ] Figure out best way to handle `delete()` when it happens to be an account from another plugin

src/ape/options.py
```diff
@@ -31,3 +35,31 @@ def verbose_option(help=""):
...
         help=help,
     )
+
+
+class Alias(click.Choice):
+    """Wraps ``click.Choice`` to load account aliases for the active project at runtime."""
+
+    name = "alias"
+
+    def __init__(self, account_type: Type[AccountAPI] = KeyfileAccount):
```
unparalleled-js, line 45: Now, different account plugins can import this and initialize it with their own custom type.
This will help make it so you only select the accounts for your type.

A couple PRs from now, I will update the ledger (and trezor) plugins!

src/ape/options.py
```diff
@@ -31,3 +36,31 @@ def verbose_option(help=""):
...
+        # NOTE: we purposely skip the constructor of `Choice`
+        self.case_sensitive = False
+        self._account_type = account_type
+
+    @property
+    def choices(self) -> List[str]:  # type: ignore
+        # NOTE: This is a hack to lazy-load the aliases so CLI invocation works properly
+        return [a.alias for a in accounts.get_accounts_by_type(self._account_type)]
```
unparalleled-js, line 54: Now, the choice aliases and confound to a specific type.
Thus, in the `ape_accounts` plugin, we limit them to `KeyfileAccounts`.

This affords the `delete` command to not be able to delete accounts from other plugins (if they happen to have a `delete()` method implemented).

src/ape/options.py
```diff
@@ -31,3 +36,31 @@ def verbose_option(help=""):
...
+
+
+def _require_non_existing_alias(arg):
```
unparalleled-js, line 57: This callback, however, still checks against all the aliases...

I am not sure how important that it, it _could_ probably still work to have aliases with the same name in separate containers. That seems out of scope though, and not that important...

src/ape/options.py
```diff
@@ -31,3 +36,35 @@ def verbose_option(help=""):
...
+    name = "alias"
+
+    def __init__(self, account_type: Type[AccountAPI] = KeyfileAccount):
+        # NOTE: we purposely skip the constructor of `Choice`
+        self.case_sensitive = False
+        self._account_type = account_type
+
+    @property
+    def choices(self) -> List[str]:  # type: ignore
+        # NOTE: This is a hack to lazy-load the aliases so CLI invocation works properly
```
unparalleled-js, line 53: I am not sure if this comment is still needed...
CLI invocation seems to work fine and this does not seem all that "lazy" as it did before

src/ape/options.py
fubuloubu: Curious if we might want an `--all` flag to display all the accounts w/ their plugin

src/ape/api/address.py
```diff
@@ -20,6 +20,10 @@ def provider(self) -> ProviderAPI:
 
         return self._provider
 
+    @provider.setter
+    def provider(self, value: ProviderAPI):
+        self._provider = value
```
fubuloubu, line 25: Hmm... Wonder if we could just remove the `_provider` field, since we can inject it manually via `a = AddressAPI(); a.provider = ...`

fubuloubu: Or even more generally, make "injectable properties" which are properties that are injected after initialization :thinking:

fubuloubu (approved):

src/ape/api/address.py
unparalleled-js: Interesting!

src/ape/options.py
unparalleled-js: I could get behind that.

src/ape/api/address.py
fubuloubu: Yeah, have to think of this one more. This property in particular is extremely awkward, and I couldn't figure out how to make it work properly. Same with the managers as properties (which is the source of the issue w/ pydantic refactor)

src/ape/options.py
fubuloubu: no, the aliases are a global construct, so we want to make sure they don't collide

src/ape/options.py
fubuloubu: Can the default be `None` which means "select all"?

unparalleled-js: Ah, that makes sense! I suppose some non-account API related plugins may want to easily allow selection from all accounts.

fubuloubu: Exactly!

src/ape/api/address.py
fubuloubu: this should be a new PR, not a blocker on this

Issue #158, opened 2021-09-23
Handle keyboard interrupts more gracefully.


unparalleled-js:
### Overview

Currently, when using keyboard interrupts, you get massive stack traces. It would be nice to interrupt without stack traces.
Note that a lot logic happens during the import statements from `ape.__init__.py`.

### Specification

From the file `ape.__init__.py`, before importing anything else, add this statement to the top of the file:

```python
import signal

signal.signal(signal.SIGINT, lambda s, f: _sys.exit(1))
```

Then, continue to import everything else.

This way, the signal handler is registered extremely early before ape is fully set up, so you can exit right away and avoid nasty looking unhelpful stack-traces in your terminal.

PR #159, opened 2021-09-23
Feat/handle interupts

unparalleled-js:
### What I did

Related issue: #158 

### How I did it

Add this code to the top of `ape.__init__.py` (earliest point in execution).
```python
import signal

signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
```

Note that it is best to put it here because it is a lengthy process to import the managers from ape, so you can easily interrupr the process at that point and it is not graceful.

Additionally, this violates `flake8` E402, which is definitely a good rule but I feel at this time that this is an edge case so I did a `flake8` per-file ignore for this rule in the `flake8` config.

### Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,4 +5,5 @@ exclude =
 	docs
 	build
 per-file-ignores =
-    src/ape/__init__.py: E402  # Need signal handler before imports
+    # Need signal handler before imports
+    src/ape/__init__.py: E402
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,5 @@ exclude =
 	venv*
 	docs
 	build
+per-file-ignores =
+    src/ape/__init__.py: E402  # Need signal handler before imports
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -1,3 +1,7 @@
+import signal
+
+signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
+
 import sys as _sys
 from functools import partial as _partial
 from pathlib import Path as _Path


setup.cfg
```diff
@@ -4,3 +4,6 @@ exclude =
 	venv*
 	docs
 	build
+per-file-ignores =
+    # Need signal handler before imports
+    src/ape/__init__.py: E402
```
fubuloubu, line 9: why not `# noqa: E402` in the file?

src/ape/__init__.py
```diff
@@ -1,3 +1,7 @@
+import signal
+
+signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
```
fubuloubu, line 3: Have to try this locally...

setup.cfg
unparalleled-js: I could put `# noqa: E402` after every import in the file, but that is effectively that same as this.

I couldn't figure out a way to ignore the entire file using a comment within the file, unless you know?

src/ape/__init__.py
unparalleled-js: I am wondering if this a better place to register this.
I generally don't put them here but more like in the `_cli.py` but this file actually takes a long time to load.

setup.cfg
fubuloubu: ah, I misunderstood. this is fine lol

I think there's a way to do it at the top of the file, but I don't remember how.

src/ape/__init__.py
fubuloubu: `_cli.py` might make more sense. that's the actual entry point for running this code I believe

setup.cfg
fubuloubu: Wonder if you could just get `isort` to ignore the ordering of the import :thinking:

fubuloubu (approved):
Tried it locally, looks great!

Did the following:
1. Run `ape`
2. Hit `ctrl+C` prior to loading
3. Observe lack of stack trace

src/ape/__init__.py
unparalleled-js: For some reason, lol, `__init__` happens first.
I put print statements at the top of both modules and run `ape`.  The `__init__.py` one ran first.  Is this unexpected? Kind of interesting.

setup.cfg
unparalleled-js: I think `isort` is actually okay with this. `flake8` i the one with the problem.

src/ape/__init__.py
fubuloubu: hmm, entry points must load the module first. weird.

PR #165, opened 2021-09-27
Feature: Add dockerfile for automated builds to dockerhub

sabotagebeats:
### What I did
added dockerfile and updated readme to include information about dockerhub

fixes: #

### How I did it
added dockerfile
updated readme

### How to verify it
we will verify it in the integration 

### Checklist

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu:latest
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install python3-pip python3 python-is-python3 -y
+RUN pip install --upgrade eth-ape
+RUN ape plugins add ens -y
+RUN ape plugins add etherscan -y
+RUN ape plugins add infura -y
+RUN ape plugins add solidity -y
+RUN ape plugins add vyper -y
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ cd ape
 python3 setup.py install
 ```
 
+### via `docker`
+
+Please visit our [Dockerhub](https://hub.docker.com/repository/docker/apeworx/ape) for more details on using Ape with Docker.
+
 ## Quick Usage
 
 Ape is primarily meant to be used as a command line tool. Here are some things you can use ape to do:


dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:latest
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install python3-pip python3 python-is-python3 -y
```
fubuloubu, line 3: Pin to Python 3.9

fubuloubu (approved):

dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN pip install --upgrade eth-ape
+RUN ape plugins add ens -y
```
unparalleled-js, line 5: It might be nice to actually make this a docker compose project...

We could have all the base plugins automatically added, but then we could have a bunch more commented out the docker compose `.yml` file and then users can plug and play by uncommenting out the plugins they want.

This would also help users make their own images with custom plugins.

dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:latest
```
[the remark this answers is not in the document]
sabotagebeats: no I haven't built this on alpine yet

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:latest
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install python3-pip python3 python-is-python3 -y
+RUN pip install --upgrade eth-ape
+RUN ape plugins add ens -y
```
unparalleled-js, line 5: For scripts, it is generally more readable to use the longer option names, such as `--yes` in this case instead of `-y`. Shorthand options are great for adhoc, long-hand options are great for scripts.

sabotagebeats: I'll make the requested changes for this. I also I got an email with your note about docker compose, not sure where that comment went, but would love to chat with you about this, maybe at Wednesday meeting.

dockerfile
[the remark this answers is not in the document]
sabotagebeats: do you know if I can do this one after another in the dockerfile? do you know how I can make sure that each build is automatically tagged properly as ubuntu or apline in the dockerhub repo?

dockerfile
fubuloubu: Pinning 3.8 is great. Kind of wild to have to select a timezone, must be a way around that 🤔

sabotagebeats:
Should I add docker build to testing for this?

dockerfile
sabotagebeats: it has to do with tzdata which i think is a requirement in apt-get in python3.9

Dockerfile
unparalleled-js: Ah yeah, the docker-compose thing might be more complicated than I though. It might make sense for ape plugin dependencies though, like running the hardhat network.

Another usability for `ape` would be to copy other package managers and allow multiple args for the plugins when adding. That way, you can just do:

```
ape plugins add ens etherscan infura solidity 
```

and it will install all 4.

dockerfile
sabotagebeats:
```ARG DEBIAN_FRONTEND=noninteractive
ENV TZ=Europe/Moscow
RUN apt-get install -y tzdata```

 if we want to pin to 3.9

https://serverfault.com/questions/949991/how-to-install-tzdata-on-a-ubuntu-docker-image

File: Dockerfile (hunk 4 above)
Comment (unparalleled-js) [reply Dockerfile]: One thing we can do is create a `build_ape.sh` script that takes parameters such as which plugins to use. Or maybe it figures it out other ways. Either way, this script can be copied in the container and ran with parameters set from environment variables e.g. `--build-arg ARG_NAME=ARG_VALUE`

File: Dockerfile [hunk 5]
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:20.04
+RUN apt-get update && apt-get upgrade --yes
+RUN apt-get install python3-pip python3.8 --yes
+RUN pip3 install --upgrade eth-ape
+RUN ape plugins add ens --yes
+RUN ape plugins add etherscan --yes
+RUN ape plugins add infura --yes
+RUN ape plugins add solidity --yes
+RUN ape plugins add vyper --yes
```
unparalleled-js, line 9: A lot of CLI Dockerfiles I see use ENTRYPOINT, so it is like running ape when you run the container, which means if you don't give it commands, it just prints the help and exits... I think you mentioned you looked into this pattern, but just thought I'd say that it is common from what I have seen. How much have we considered this approach?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:20.04
+RUN apt-get update && apt-get upgrade --yes
+RUN apt-get install python3-pip python3.8 --yes
+RUN pip3 install --upgrade eth-ape
```
unparalleled-js, line 4: Hmm.... I'd actually expect this to build from source. That way, you download the repo and can build the container for dev purposes

Dockerfile
unparalleled-js: Here is GETH example: https://github.com/islishude/geth-docker/blob/master/Dockerfile

unparalleled-js: Notice it builds uses alpine, uses `geth` command as the entrypoint, and it builds from source!

Dockerfile
sabotagebeats: Yes I believe the reason we went with dockerfile and not compose is that we didn't want to have it spinning up an app, we wanted it to be a CLI command.

dockerfile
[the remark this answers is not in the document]
sabotagebeats: This is the build that isn't working for me: 
```
FROM python:alpine3.13
RUN pip3 install --upgrade eth-ape
RUN ape plugins add ens --yes
RUN ape plugins add etherscan --yes
RUN ape plugins add infura --yes
RUN ape plugins add solidity --yes
RUN ape plugins add vyper --yes
```

Dockerfile
sabotagebeats: so like this? 
```
RUN git clone https://github.com/ApeWorX/ape
RUN pip install -e ./ape
```

unparalleled-js: I was thinking something like `python setup.py sdist` and then `pip install dist/eth-ape*` since this Dockerfile is in the same repo.

From CI/CD perspective, if we want to run a job to deploy the image to Dockerhub on release, there might be a race condition if relying on the build being available on pypi first

unparalleled-js: Oh youd have to copy the source files to the container as well as use a `.dockerignore`

unparalleled-js: Lol maybe `git` is better.

Dockerfile
unparalleled-js: It would be what ever command the user types in!

They would run the container with the same args that they run `ape` with. So if they run the container with `console` as the argument, it would launch a console, etc.

unparalleled-js: Like this `docker run apeworx/ape console

unparalleled-js (commented):
I might have approved too early, I think there is some more to figure out.
The main thing, I think, is to not build from pypi but to use the source files. Additionally, I think it is best to have the container function the exact same as the CLI tool itself, so running the container should exit immediately and it should use the simple `ape` as its ENTRYPOINT. It might take some figuring out. Finally, we should try to use alpine to cut the cruft of the image. This should be able to build fast!

Dockerfile
sabotagebeats: > It would be what ever command the user types in!
> They would run the container with the same args that they run `ape` with. So if they run the container with `console` as the argument, it would launch a console, etc.

yes then the problem here then becomes that they can't do other things if they need to with the cli within that image, right?

Dockerfile
sabotagebeats: I was testing this with git and i'm having a bunch of failures. I'll keep trying and post the failures if i can't *git* it to work

sabotagebeats:
> I might have approved too early, I think there is some more to figure out. The main thing, I think, is to not build from pypi but to use the source files. Additionally, I think it is best to have the container function the exact same as the CLI tool itself, so running the container should exit immediately and it should use the simple `ape` as its ENTRYPOINT. It might take some figuring out. Finally, we should try to use alpine to cut the cruft of the image. This should be able to build fast!

ape as an entry point then the user can't do other things if they need to with the cli within that image, right? Not sure if this is a problem or not. 

I wasn't able to successfully get alpine to build unfortunately. There are many dependency errors and issues with pip install.

sabotagebeats:
This article is why I originally decided to go with Ubuntu as a base image. 

https://megamorf.gitlab.io/2020/05/06/why-it-s-better-not-to-use-alpine-linux-for-python-projects/

sabotagebeats:
This article details why I'm running into errors trying to build with alpine: https://pythonspeed.com/articles/alpine-docker-python/

sabotagebeats:
> This article details why I'm running into errors trying to build with alpine: https://pythonspeed.com/articles/alpine-docker-python/

as per this article I tried to reduce the build size by using python:3.8-slim however then I came into a bunch of errors with `unable to execute 'gcc': No such file or directory`: 

```
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for bitarray
```
  
  
```
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for lru-dict
```

```
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for pyethash
```

```
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for pysha3
```

unparalleled-js:
> This article is why I originally decided to go with Ubuntu as a base image.
> 
> https://megamorf.gitlab.io/2020/05/06/why-it-s-better-not-to-use-alpine-linux-for-python-projects/

Ok, that makes sense!  Ubuntu works.

unparalleled-js:
> > I might have approved too early, I think there is some more to figure out. The main thing, I think, is to not build from pypi but to use the source files. Additionally, I think it is best to have the container function the exact same as the CLI tool itself, so running the container should exit immediately and it should use the simple `ape` as its ENTRYPOINT. It might take some figuring out. Finally, we should try to use alpine to cut the cruft of the image. This should be able to build fast!
> 
> ape as an entry point then the user can't do other things if they need to with the cli within that image, right? Not sure if this is a problem or not.
> 
> I wasn't able to successfully get alpine to build unfortunately. There are many dependency errors and issues with pip install.

Correct, they wouldn't be able to do anything that is not an `ape` command.
Maybe we don't want that, maybe that is too limiting.. It is just what I see most CLI tools do for their Dockerfiles, so we should at least consider it.

dockerfile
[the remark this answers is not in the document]
fubuloubu: Skipping Alpine for now

Dockerfile
sabotagebeats: as per conversation with @fubuloubu:
copy the files from the local directory into the docker image and then use pip install 
benefit: not relying on deployed version, it will be whatever is on the master branch and be easier to debug and avoid race condition

sabotagebeats:
todo: 
entrypoints and source code copy 

skip for now: 
alpine and 3.8 slim

Dockerfile
VoR0220: I would agree with copying files from local directory as that is also easiest to build for the purposes of development and testing new features on a local instance via a built image.

sabotagebeats:
source code copying is working except that installing vyper step is failing. 
``$3aa926e91d69 ape plugins list
WARNING: Error loading plugin package 'ape_vyper'
WARNING: $GITHUB_ACCESS_TOKEN not set, skipping 2nd class plugins
Installed plugins:
  ape_ens (0.1.0a1)
  ape_etherscan (0.1.0a1)
  ape_infura (0.1.0a2)
  ape_solidity (0.1.0a2)
```
then when I go to compile docker 
```$3aa926e91d69 ape compile
WARNING: Error loading plugin package 'ape_vyper'
WARNING: No compilers detected for the following extensions: .vy```

**Comment by fubuloubu** [comment]
@sabotagebeats that warning pops up when the plugin isn't currently functioning properly. Try debugging it locally, and then fixing the `ape-vyper` plugin

**Comment by sabotagebeats** [comment]
this is a preexisting issue: https://github.com/ApeWorX/ape-vyper/issues/22

**Comment by fubuloubu** [comment]
@sabotagebeats yes, so the solution to that issue will solve this one and allow the plugin to function in the image. **However** we should take this oppurtunity to think through how the image is built here, and make sure that all plugins are functioning for when we distribute the image.

Perhaps this needs to be "tested" in the image by making sure that `ape plugins` has no warnings. WIll have to think more on this one.

**Comment by sabotagebeats** [comment]
> Perhaps this needs to be "tested" in the image by making sure that `ape plugins` has no warnings. WIll have to think more on this one.

yes maybe we can do build testing for compilation functionality of the docker image?

**Comment by unparalleled-js** [comment]
It should be pretty straight forward to upgrade the vyper plugin. Let me know if not!
I was going to upgrade all the plugins probably after the next alpha release as the CLI SDK work is wrapping up.

**Comment by sabotagebeats** [comment]
```
$ docker build .
Sending build context to Docker daemon  16.12MB
Step 1/10 : FROM ubuntu:latest
 ---> 597ce1600cf4
Step 2/10 : RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip python3-dev --yes
 ---> Using cache
 ---> 676ae34f4e32
Step 3/10 : COPY ./src ./setup.py ./README.md ./pyproject.toml ./setup.cfg ./
 ---> 82edcc2114b7
Step 4/10 : RUN python3 ./setup.py install
 ---> Running in 79fdd6aa6e4a
  File "./setup.py", line 54, in <module>
    setup(
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 144, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.8/distutils/core.py", line 108, in setup
    _setup_distribution = dist = klass(attrs)
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 447, in __init__
    _Distribution.__init__(self, {
  File "/usr/lib/python3.8/distutils/dist.py", line 292, in __init__
    self.finalize_options()
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 740, in finalize_options
    ep.load()(self)
  File "/.eggs/setuptools_scm-6.3.2-py3.8.egg/setuptools_scm/integration.py", line 94, in infer_version
    dist.metadata.version = _get_version(config)
  File "/.eggs/setuptools_scm-6.3.2-py3.8.egg/setuptools_scm/__init__.py", line 177, in _get_version
    parsed_version = _do_parse(config)
  File "/.eggs/setuptools_scm-6.3.2-py3.8.egg/setuptools_scm/__init__.py", line 135, in _do_parse
    raise LookupError(
LookupError: setuptools-scm was unable to detect version for '/'.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
The command '/bin/sh -c python3 ./setup.py install' returned a non-zero code: 1

```

fubuloubu:
> filed issue #178

See response (and closure) e.g. just change to `VyperToken.vy` and it should compile

unparalleled-js:
> > > @sabotagebeats build it locally first, and the move the `version.py` file over to the image as well, this is generated by `setuptools-scm` from your git repo during installation.
> > 
> > 
> > So I did try to build this locally but I don't have the `version.py` file. Maybe I'm doing something wrong? It works if I copy the entire directory.
> 
> the `.git` folder is what's needed, which if you don't copy all the files it will install as "dirty" because some of the files are gone
> 
> let's stick with copying everything over.can always fix it

You should be able to bind it during the install command, e.g. `RUN --mount=source=.git,target=.git,type=bind python3 ./setup.py install`

fubuloubu:
> looks like merging main fixed the types issue :)

Is this good to merge and see how docker hub takes it?

sabotagebeats:
> > looks like merging main fixed the types issue :)
> 
> Is this good to merge and see how docker hub takes it?

yes I believe we can merge this and then I can review the integration to make sure it was successful.

sabotagebeats:
this is building properly on dockerhub @fubuloubu @unparalleled-js 
It is building immediately and not on version upgrade. Should we build every time a commit is merged, or should we build every time the version is updated?

Dockerfile
sabotagebeats: https://github.com/ApeWorX/ape/issues/383

fubuloubu:
`latest` should be building every time
`stable` and the tagged versions should only build on new version tags

I think this happened today because we did 0.1.1 by accident, and Docker also caught it

Issue #166, opened 2021-09-27
Unable to display unsupported extensions in `ape compile`


unparalleled-js:
### Environment information

* `ape` and plugin versions:

$ ape --version
0.1.0a23

$ ape plugins list
  ape_infura (0.1.0a2)
  ape_solidity (0.1.0a2)
  ape_vyper (0.1.0a6).   # NOTE: I removed this one to further reproduce the issue of not having compilers
  ape_ledger (0.1.dev2+ge8f73a3.d20210921)
  ape_hardhat (0.1.dev12+gfd19de4)

* Python Version: 3.9.7
* OS: macOS 11.6

Also, note that I am working with both `vyper` and `solidity` contracts.

### What went wrong?

There are two small bugs that occur when running `ape compile` when not having the correct compilers **or** not having an source files to compile (empty `contracts/` directory)

1. You get an error: `AttributeError: 'str' object has no attribute 'suffix'`

This is because part of the code assumes the type to be `Path` instead of a `str`.

After fixing that issue, the error message is still strange:

```
WARNING: No compilers detected for the following extensions:
```

notice that is not able to determine the actual extensions.
This is because they are previously excluded from the list that it checks from.

### How can it be fixed?

* Use Paths instead of str when calculating the extensions
* Get the unsupported extensions earlier in the process
* Handle when the contracts directory is empty separately
* Refactor where helpful

fubuloubu:
One other thing to note is that I don't think the "package compiler" (`.json` extension) handles non-compliant json files well

In the event that it does not contain the right structure, it should be ignored instead of throwing. (and potentially raise a warning)

PR #167, opened 2021-09-27
fix: handle missing compiler extensions

unparalleled-js:
### What I did

fixes: #166

### How I did it

* Make sure to work with paths on path methods
* Correctly calculate the extensions that are missing from registered compilers
* Handle the case where the contracts/ directory is empty (As far as error message goes).

### How to verify it

Start off without any compiler plugins

1. Create a `contracts/` directory with a Solidity file and a Vyper file
2. Run `ape compile`

You should get the error message: `WARNING: No compilers detected for the following extensions: .vy, .sol`

3. Add a Vyper plugin via `ape plugins add vyper`

It should compile the solidity file and warn about the vyper file:

`WARNING: No compilers detected for the following extensions: .sol`

4. Add the Solidity plugin

It should compile both files and not warn.

5. Move both files out of the directory and run `ape compile`

It should warn with 

`WARNING: No compilers detected for the following extensions: .sol`

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -73,6 +73,9 @@ class ProjectManager:
 
     @property
     def sources(self) -> List[Path]:
+        """All the source files in the project.
+        Excludes files with extensions that don't have a registered compiler.
+        """
         files: List[Path] = []
         for extension in self.compilers.registered_compilers:
             files.extend(self._contracts_folder.rglob("*" + extension))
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
In general, I wish that some of this logic (where appropiate) migrated to the `Package` class, so that we could handle compiling dependencies (from Github, IPFS, etc.)  just as easily

This CLI should be as minimal as possible

unparalleled-js:
> In general, I wish that some of this logic (where appropriate) migrated to the `Package` class, so that we could handle compiling dependencies (from Github, IPFS, etc.) just as easily.
> 
> This CLI should be as minimal as possible

Yeah that makes sense.

Issue #153, opened 2021-09-17
Replace `raise Exception` in 1st class plugins with raising custom ape errors


unparalleled-js:
### Overview

This must have gotten missed during the custom exceptions work.
There are still places where we are raising the base exception.

PR #169, opened 2021-09-30
feat: custom errors in 1st class plugins

unparalleled-js:
### What I did

fixes: #153

### How I did it

* Replaced `raise Exception()` in 1st class plugins with custom exceptions that, for the most part, already existed from previous work.

### How to verify it

* Make sure the custom exceptions make sense.

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -12,6 +12,7 @@ from eth_utils import keccak, to_bytes, to_int
 from hexbytes import HexBytes
 
 from ape.api import ContractLog, EcosystemAPI, ReceiptAPI, TransactionAPI, TransactionStatusEnum
+from ape.exceptions import ContractCallError, TransactionError
 from ape.types import ABI, AddressType
 
 NETWORKS = {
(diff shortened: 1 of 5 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -95,8 +96,8 @@ def decode_calldata(self, abi: ABI, raw_data: bytes) -> Any:
         try:
             return abi_decode(output_types, raw_data)
 
-        except InsufficientDataBytes as e:
-            raise Exception("Output corrupted") from e
+        except InsufficientDataBytes as err:
+            raise ContractCallError("Output corrupted") from err
```
unparalleled-js, line 100: Does `decode_calldata()` only have to do with contract calls?
I am trying to make sense of the all the use-cases of this method so I know if this is the best custom exception type to use.

fubuloubu: Contract transactions can also return data that needs to be decoded (not sure if this is implemented yet)

I would say that `DecodingError` might be a better name

fubuloubu (approved):

src/ape_ethereum/ecosystem.py
unparalleled-js: Ok, so I:

1. made a base exception `ContractError`.
2. made `TransactionError` and `ContractCallError` subclasses of `ContractError`
3. made new error `DecodingError` that is also a subclass of `ContractError`

I think that models the relationships a little better here.

fubuloubu (approved):

Issue #151, opened 2021-09-17
Refactor plugins key for config file


fubuloubu:
```yaml
plugins:                                                                                        
    ape-solidity: 0.1.0a2
```
Would rather it look more like this
```yaml
plugins:                                                                                        
    - solidity:          
        version: 0.1.0a2  # optional
```

PR #170, opened 2021-09-30
fix: improve plugins install config schema

unparalleled-js:
### What I did

fixes: #151 

### How I did it

* Created a helper method that extracts the module name and the package name from the config, following the new schema defined in the corresponding issue.

### How to verify it

* The following yaml should install a specific version of the solidity plugin and the latest version of the ledger plugin:

```yaml
plugins:
  - solidity:
      version: 0.1.0a2
  - ledger:

```

* A missing config, empty config, or empty `plugins:` in a config should not install anything nor should it error, it just should do nothing.

e.g. 

```yaml
plugins:

```

* A malformed plugin config entry item should raise an error and display the expected schema for installing plugins via the config.

### Checklist


fubuloubu:
> * The following yaml should install a specific version of the solidity plugin and the latest version of the ledger plugin:
> 
> ```yaml
> plugins:
>   - solidity:
>       version: 0.1.0a2
>   - ledger:
> ```

I was thinking of reformatting it to look more like this:

```yaml
plugins:
  - name: solidity
    version: 0.1.0a2
  - name: ledger
    # No need to specify `version`
```

unparalleled-js:
Yeah that schema definitely is more yaml-y, I like it.

I am thinking some of this github stuff and the methods I added should just move to a `utils.py` file within `ape_plugins` and then it would be more testable.

fubuloubu:
> Yeah that schema definitely is more yaml-y, I like it.
> 
> I am thinking some of this github stuff and the methods I added should just move to a `utils.py` file within `ape_plugins` and then it would be more testable.

agreed

src/ape_plugins/_cli.py
```diff
@@ -116,26 +117,49 @@ def add(cli_ctx, plugin, version, skip_confirmation):
         subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", plugin])
 
 
+def _get_config_error() -> ConfigError:
+    expected_format = "plugins:\n  - name: <plugin-name>\n    version: <plugin-version>  # optional"
+    return ConfigError(f"Config item mis-configured. Expected format:\n\n{expected_format}\n")
+
+
+def _get_plugin_names(item: Dict) -> Tuple[str, str]:
```
fubuloubu, line 125: Naming here doesn't match what it returns. suggest `_extract_module_and_package_names`

src/ape_plugins/_cli.py
```diff
@@ -116,26 +117,49 @@ def add(cli_ctx, plugin, version, skip_confirmation):
...
+    specified in the config.
+    """
+    try:
+        name = item["name"]
+        module_name = f"ape_{name.replace('-', '_')}"
+        package_name = f"ape-{name}"
+        version = item.get("version")
+
+        if version:
+            package_name = f"{package_name}=={version}"
```
fubuloubu, line 138: `package_name` doesn't quite match what the purpose is, maybe something more like `package_install_name`

fubuloubu (approved):

PR #171, opened 2021-10-02
Feat/network cmd

unparalleled-js:
### What I did

Made the network context manager happen automatically by either using a `network_bound_command` decorator
You can also use the custom `ape_group` decorator on a `click.group` and then from there, do `cli.network_bound_command()`.

### How I did it

First, I noticed that this didn't happen automatically, and I wanted it to, for ease of development.
Next, I created the custom `NetworkBoundCommand`
Then I made it accessible as a decorator, including off a custom group so it can be used either in command or group-command.

### How to verify it

Add
`print(networks.active_provider.network.name)` to the run or console commands and set `--network` and see that it gets sets.

Also, try this:

```
from ape.cli import ape_group, network_option

@ape_group()
def cli():
    """My custom Ethereum tools"""
   
@cli.network_bound_command()
@network_option
def cmd(network):
  print(networks.active_provider.network.name)
```

### Checklist


src/ape/cli/commands.py
```diff
@@ -0,0 +1,93 @@
...
+    return click.command(cls=NetworkBoundCommand, *args, **kwargs)
+
+
+class NetworkBoundCommand(click.Command):
+    """A command that uses the network option.
+    It will automatically set the network for the duration of the command execution.
+    """
+
+    def invoke(self, ctx: Context) -> Any:
+        with networks.parse_network_choice(ctx.params["network"]):
```
unparalleled-js, line 32: Here is where the context manager gets used

src/ape_console/_cli.py
```diff
@@ -1,23 +1,23 @@
...
+    short_help="Load the console",
+    context_settings=dict(ignore_unknown_options=True),
+)
 @verbose_option(help="Display more information in the console")
 @network_option
 def cli(verbose, network):
     """
     Opens a console for the local project."""
 
-    with networks.parse_network_choice(network):
```
unparalleled-js, line 19: Now, plugin writers can just use the network_bound_command + the network_option, and they can write their command logic.

src/ape/cli/commands.py
```diff
@@ -0,0 +1,93 @@
...
+            super().invoke(ctx)
+
+
+class ApeCLI(click.MultiCommand):
```
unparalleled-js, line 36: I only moved this here because I wanted to get all the custom click stuff in the `cli` folder.
However, it is not really needed and I could put it back.

src/ape_console/_cli.py
fubuloubu: Two comments:
1. Would `network_option` work without `network_bound_command`?
2. Vice versa (would `network_bound_command` work without `network_option`?)

If the answer to both is "no", would it make sense to combine them into just `network_bound_command`?

src/ape/cli/commands.py
```diff
@@ -0,0 +1,20 @@
+from typing import Any
+
+import click
+from click import Context
+
+from ape import networks
+
+
+def network_bound_command(*args, **kwargs):
```
fubuloubu, line 9: Past tense, would `bind_network_command` be a better name?

src/ape/cli/commands.py
fubuloubu: Love this!

src/ape/cli/groups.py
```diff
@@ -0,0 +1,12 @@
+import click
+
+from ape.cli.commands import NetworkBoundCommand
+
+
+def ape_group(*args, **kwargs):
+    return click.group(cls=ApeGroup, *args, **kwargs)
+
+
+class ApeGroup(click.Group):
```
fubuloubu, line 10: Note: some CLI options don't need a network connection e.g. `ape compile`

src/ape_console/_cli.py
```diff
@@ -1,23 +1,23 @@
...
-from ape.cli import network_option, verbose_option
+from ape.cli import network_bound_command, network_option, verbose_option
 from ape.version import version as ape_version  # type: ignore
 
 
-@click.command(short_help="Load the console", context_settings=dict(ignore_unknown_options=True))
+@network_bound_command(
+    short_help="Load the console",
+    context_settings=dict(ignore_unknown_options=True),
+)
```
fubuloubu, line 13: Ahhh, nvm my previous comment about `bind_network_command`. If I read this correctly, this overrides the use of `click.command`, correct?

Would there be another way to integrate this decorator that works with the context of the naked `click.command`?

I would say if not, we should look into bundling `click.command` into our CLI SDK so that plugin writers don't have to switch between both packages (and also allows us to refactor more effectively)

fubuloubu (commented):
Excellent refactor!

Some architectural discussion here before moving forwards

src/ape_console/_cli.py
unparalleled-js: Yes, so hence the past tense, re: overriding use of `click.command`

We could do something like that, like make an `ape_command` decorator. the `ApeGroup` was kind of along those lines, since it is just a normal `click.group` with the ability to create `bound_network_commands` if needed

src/ape/cli/groups.py
unparalleled-js: Well, this is just a normal click group. You can still make non-network commands with this group. You just now have the ability to create network commands if needed

unparalleled-js:
@fubuloubu  Regarding https://github.com/ApeWorX/ape/pull/171#discussion_r720726180,
You can use `network_option` without it needed to be a bound-network command. A use-case for this might be like `ape network show <network>` that does not require a provider but still operates on the user's networks

src/ape/cli/groups.py
unparalleled-js: It is so you can do `cli.network_bound_command()` from within a plugin where `cli` is a `click.Group` (an `ApeGroup` specifically!)

src/ape_console/_cli.py
unparalleled-js: Hmm... I will have to think about this more. It seems like I am trying to _extend_ click here. I am trying to avoid abstracting all of click away because that may be just mostly passthroughs (like `click.echo`, `click.option` etc).
We could also go the reverse and requires everyone to import the class `NetworkBoundCommand` and set that via `cls=` on their `click.commands()`, which is fine. I just pushed a commit that makes it so they get the `network_option` for free though, so it would just require more setting up on the user end to make a network bound command.

Yeah I am going to think about this more. I am building out a custom ape plugin (`ape-jules`) to test that stuff.. It just makes JSON-RPC requests and prints stuff to the terminal, nothing fancy, but it is helpful in QA-ing this stuff.

src/ape_console/_cli.py
unparalleled-js: `network_bound_command` now get the `network_option` for free.
However, you can still use `network_option` without it being a network bound command, if it makes sense.

src/ape/cli/commands.py
fubuloubu: Better: `command_with_network_option` now would make a lot more sense

unparalleled-js: What about something like `command_using_network_option`? That way it still conveys that the network option get used upon invocation

fubuloubu: yes! and then me (a plugin writer) parsing this out:
"oh, okay this makes sense, if I add `network` to my args I get access to the network option that `command_using_network_option` provides!"

src/ape_console/_cli.py
fubuloubu: > Hmm... I will have to think about this more. It seems like I am trying to _extend_ click here. I am trying to avoid abstracting all of click away because that may be just mostly passthroughs (like `click.echo`, `click.option` etc). We could also go the reverse and require everyone to import the class `NetworkBoundCommand` and set that via `cls=` on their `click.commands()`, which is fine. I just pushed a commit that makes it so they get the `network_option` for free though, so it would just require more setting up on the user end to make a network bound command.

Yes, seems like we have a choice here to fully wrap click in everything we're doing, or be more transparent and let click do more of the work. I'm okay with either choice, but we need to commit to one. We've been sort of trending down this path towards wrapping everything click into a "CLI SDK" for plugin writers

---

> Yeah I am going to think about this more. I am building out a custom ape plugin (`ape-jules`) to test that stuff.. It just makes JSON-RPC requests and prints stuff to the terminal, nothing fancy, but it is helpful in QA-ing this stuff.

Consider making this a permanent addition to the test suite, e.g. `ape-dummy`, so we have a way to test integration of a new plugin for all the features that are not tested via the 1st class plugins already in the repo

fubuloubu (approved):

Issue #174, opened 2021-10-05
Debug logging


unparalleled-js:
### Overview

Add support DEBUG verbosity when logging. That way, users can turn on debug logging to better diagnose issues.

### Specification

`--verbosity DEBUG` or -`v DEBUG` (already existing flag).
Log more exception info, such as full stack traces.

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #175, opened 2021-10-05
feat: debug logging

antazoey:
### What I did

fixes: #174 

### How I did it

* Add DEBUG option for logging verbosity 
* Log entire stack trace in a couple critical spots where debug is enabled.

### How to verify it

* Have a non-working plugin, do `ape -h -v DEBUG`. You should now see a large stack trace output.
* Raise some exception in a command that triggers `Abort()`, you should now see a large stack trace in the output.

For example, when DEBUG is on, you will get a trace from this:

```python
from ape.exceptions import ApeException
raise ApeException("test")
```

Otherwise, you only get `ERROR: test`

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -87,7 +87,7 @@ class ApeCLI(click.MultiCommand):
             try:
                 return self.commands[name]()
             except Exception as err:
-                logger.log_verbose_error(
+                logger.verbosely_warn_from_error(
                     err, f"Unable to load CLI endpoint for plugin 'ape_{name}'"
                 )
 
--- a/src/ape/logging.py
+++ b/src/ape/logging.py
@@ -112,7 +112,12 @@ class CliLogger:
         if message:
             self._logger.error(message)
 
-    def log_verbose_error(self, err: Exception, message: str):
+    def verbosely_warn_from_error(self, err: Exception, message: str):
+        """
+        Warn the user with the given message,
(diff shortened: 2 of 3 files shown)


fubuloubu (approved):

lost-theory (approved):
LGTM. Though as a developer I like it when full tracebacks are shown by default.

unparalleled-js:
> LGTM. Though as a developer I like it when full tracebacks are shown by default.

Maybe we could have a config option that lets you always run with with a debug-level verbosity?  Tracebacks seems like they should log at the DEBUG level to me, I'd only care to see them if I am trying to figure out what is going on.  Else, the output is really verbose. This is a pattern I have seen elsewhere in python apps too.

One thing that might be nice though is a suggestion in the warning message:  `Use -v DEBUG for full tracebacks`

fubuloubu:
> One thing that might be nice though is a suggestion in the warning message: `Use -v DEBUG for full tracebacks`

The first time one appears would be great (and only the first time?)

> LGTM. Though as a developer I like it when full tracebacks are shown by default.

Could add a config file option :thinking:

fubuloubu:
> > What command option?
> 
> Like `ape --verbosity DEBUG` versus the log-level configured in the `ape-config.yaml` file, which one would take precedence?

command option for sure

we'll see if the config option becomes necessary, was just a suggestion

Issue #172, opened 2021-10-04
Custom click ape_commands and command_using_network_option


unparalleled-js:
### Overview

* Be able to use networks option automatically in command without having to wrap whole method in context manager `with` block, as discussed in #171 .
* This calls for more general click abstraction to make plugin development dependent on single framework.

### Specification

Follow the WIP Pr #171 

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #176, opened 2021-10-05
fix: warn for missing compilers recursively

unparalleled-js:
### What I did

* Fixes issue where if you had nested files in a directory with an extension that is not a registered compiler key, it would fail to warn for the missing compiler.

### How I did it

While working on the CLI SDK refactor, I realized we had an opportunity for a `AllFilesPath` `click.ParamType` which essentially just converts a directory structure to a list of file paths (if just given a single file path, it returns a list with that file path).

This gets used in the source file args, but they call now have a callback method that handles flattening the `AllFilesPaths` lists, de-duplicating, and calling `resolve()`.

Additionally, I moved some of this logic to the `ProjectManager`, which now can tell me which extensions do not have compilers and if sources are just generally missing. This makes the `ape_compile._cli.cli()` method much more focused on `cli`.

I moved `arguments` to their own module `arguments` as well as `Alias` to `choices` in the process of this work because they are bugging me.

All of the solution to fixing this bug helps keep my PR smaller in the CLI refactor I am working on as part of #172 

### How to verify it

* Have `ape-solidity` installed
* Move a `.vy` file to a directory inside `contracts/`
* Run `ape compile`. It should work for missing compilers.  Notice that on the main branch, it currently does not do that correctly.
* Try giving it a path, it should only warn for extensions given (if it needs to warn)
* ^ also there is an integration test for these

https://github.com/ApeWorX/ape-vyper/pull/23/checks may help fully validate this

### Checklist

### DIFF
--- a/src/ape_compile/_cli.py
+++ b/src/ape_compile/_cli.py
@@ -1,51 +1,22 @@
 from itertools import chain
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict
 
 import click
 
 from ape.cli import ape_cli_context
+from ape.cli.paramtype import AllFilePaths
 from ape.types import ContractType
 
-flatten = chain.from_iterable
-
-
-class _ContractsSource:
-    """
-    A helper class that is able to figure out which source files to
-    compile.
-    """
-
-    def __init__(self, use_cache: bool):
-        # NOTE: Lazy load so that testing works properly
-        from ape import project
-
-        self._project = project
-        self._use_cache = use_cache
-
-    @property
-    def root(self) -> Path:
-        return self._project.path / "contracts"
-
-    def select_paths(self, file_paths: Optional[List[Path]]):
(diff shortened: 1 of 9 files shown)


src/ape_compile/_cli.py
```diff
@@ -65,52 +36,40 @@ def compile(self) -> Dict[str, ContractType]:
...
     back to the manifest.
 
     Note that ape automatically recompiles any changed contracts each time
     a project is loaded. You do not have to manually trigger a recompile.
     """
-    source = _ContractsSource(use_cache)
-    missing_source = not source.root.exists() or not source.root.iterdir()
-    if not filepaths and missing_source:
+    # NOTE: Lazy load so that testing works properly
+    from ape import project
```
fubuloubu, line 48: This doesn't work at the top?

fubuloubu (approved):

PR #197, opened 2021-10-19
fix: extract revert msg and handle ganache-cli

antazoey:
### What I did

* Handle when error args[0] is of type `str` (such as what happens on `ganache-cli`).
* Extract the revert reason using a known prefix.

### How I did it

* Tried reverts again using ganache, realized it was not working.
* Made it work. Realized for both hardhat and ganache, there was an annoying prefix to the revert message.
* I stripped out the prefix for both respectively

### How to verify it

I tested all the gas related errors again as well as a txn I knew would revert and made sure I only got the expected message for both hardhat node and ganache-cli.

**NOTE: ape-http is not really the way to go in the future. It is great for now for getting up and running, but I think once every provider has its own plugin, they should handle this web3 errors uniquely and separately.  This means ape-hardhat should probably no longer inherit from the ape-http provider. I don't want to try and make ape-http work for every provider possible. I think it is ok for now since we are focusing on development networks and it is nice for demo-ing and working with providers that we don't yet support natively.**

### Checklist

### DIFF
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -150,12 +150,22 @@ def get_tx_error_from_web3_value_error(web3_value_error: ValueError) -> Transact
         # Not known from provider
         return TransactionError(base_err=web3_value_error)
 
-    err_data = web3_value_error.args[0]
-    if not isinstance(err_data, dict):
-        return TransactionError(base_err=web3_value_error)
+    def _extract_revert_message(msg: str, _prefix: str) -> str:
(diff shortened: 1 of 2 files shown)


unparalleled-js:
I think this work is helpful to have but I think we https://github.com/ApeWorX/ape-hardhat/issues/4 to happen and if we ever made an `ape-ganache`, we do a similar thing there.

Issue #204, opened 2021-10-25
Geth Out-of-the-Box support


unparalleled-js:
### Overview

* When no HTTP provider set, start up a dev geth process with funded accounts.

### Specification

TBD

### Dependencies

geth, maybe py-geth

PR #210, opened 2021-10-29
feat: start ephemeral geth when no http provider active

unparalleled-js:
### What I did

fixes: #204

### How I did it

Using `py-geth`, stand up an epheremal geth when using the http provider and they do not have a provider running at that URI.

### How to verify it

Deploy a script to the ephemeral geth.

First, you need to create an account.

```bash
ape account import geth_dev_0
```
And supply `0xdd23ca549a97cb330b011aebb674730df8b14acaee42d211ab45692699ab8ba5`
your address should be '0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C'

Next, deploy a simple contract:

Contract:

```solidity
pragma solidity ^0.6.0;

contract MyNumber {
    int public number = 42;
}
```

Script (get_my_number.py)

```python
from ape import accounts
from ape import project


def main():
    account = accounts.load("get_dev_0")

    contract_type = project.MyNumber
    contract = test_account.deploy(contract_type)

    my_num = contract.number(sender=test_account)
    print(my_num)

```

When you run the script:

```bash
ape run get_my_number
```

Follow the prompt to unlick your account.
It should print the number `42`.

Before that, it will have compiled the contracts if it needed as well as start up an ephemeral geth.

I can open the Testing PR after this that includes filtering out test accounts etc.

### Checklist


src/ape/utils.py
```diff
@@ -142,42 +146,43 @@ def compute_checksum(source: bytes, algorithm: str = "md5") -> str:
     return hasher(source).hexdigest()
 
 
-def get_tx_error_from_web3_value_error(web3_value_error: ValueError) -> TransactionError:
+def generate_dev_accounts(
```
unparalleled-js, line 149: Some providers need to also generate the dev account while some can simply supply a mnemonic.
Both eth-tester and geth will use this.

src/ape_http/providers.py
```diff
@@ -1,15 +1,103 @@
...
+    TransactionError,
+)
+from ape.logging import logger
+from ape.utils import generate_dev_accounts, get_gas_estimation_revert_error_message
 
-DEFAULT_SETTINGS = {"uri": "http://localhost:8545"}
+DEFAULT_SETTINGS = {"uri": "http://localhost:8545", "chain_name": "devchain"}
+
+
+class EphemeralGeth(LoggingMixin, DevGethProcess):
```
unparalleled-js, line 32: Note: Unitl https://github.com/ethereum/py-geth/pull/77 is merged and released, the `LoggingMixing` causes errors.

src/ape_http/providers.py
```diff
@@ -134,12 +270,70 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         try:
             txn_hash = self._web3.eth.send_raw_transaction(txn.encode())
         except ValueError as err:
-            raise get_tx_error_from_web3_value_error(err) from err
+            raise _get_tx_error(err) from err
+
+        receipt = self.get_transaction(txn_hash.hex())
+
+        if receipt.status == TransactionStatusEnum.FAILING and receipt.gas_used == txn.gas_limit:
+            raise OutOfGasError()
```
unparalleled-js, line 278: For geth, this is the only OutOfGasErrors occur. Other providers raise them normally. I have never seen any other errors come this far for any other reason in geth.. Reverts / other VM errors get caught and handled above.

src/ape/api/providers.py
```diff
@@ -166,3 +166,17 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
     @abstractmethod
     def get_events(self, **filter_params) -> Iterator[dict]:
         ...
+
+
+class TestProviderAPI(ProviderAPI):
```
unparalleled-js, line 171: As part of the test work, we filter out accounts that implement this API and `ape test` _only_ gets accounts that implement this API.

src/ape/api/providers.py
```diff
@@ -166,3 +166,17 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
+    @abstractmethod
+    def snapshot(self) -> Dict:
+        ...
+
+    @abstractmethod
+    def revert(self, snapshot: Dict):
```
unparalleled-js, line 177: Using `Dict` because `EthTester` require `dicts`.
And for `hardhat` or `ganache`, provider plugins can just return `{"snapshot_id": 123}` or something.

src/ape_http/providers.py
```diff
@@ -174,20 +174,19 @@ def connect(self):
 
             if not self._web3.isConnected():
                 self._geth.disconnect()
-                raise ConnectionError("Unable to connect web3 to geth.")
+                raise ConnectionError("Unable to connect to locally running geth.")
 
         self._web3.eth.set_gas_price_strategy(rpc_gas_price_strategy)
 
+        if self.network.name not in ("mainnet", "ropsten"):
```
fubuloubu, line 181: I would flip this, because I don't believe local networks need this either, and neither would some other testnets for PoW chains
```suggestion
        if self.network.name in ("rinkeby", "goerli", "kovan"):
```

Also might want to do this via chain_id.... :thinking:

unparalleled-js: Done and Done..

Wait, would local PoA networks need this though? I just added some code that checks the consensus via `admin.node_info()` and will set this middleware if `clique` was configured

src/ape_http/providers.py
```diff
@@ -1,15 +1,111 @@
...
+                "istanbulBlock": 0,
+                "berlinBlock": 0,
+                "londonBlock": 0,
+                "ethash": {},
+            },
+            "alloc": {a.address: {"balance": str(initial_balance)} for a in accounts},
+        }
+        geth_cmd_args = {
+            "rpc_addr": hostname,
+            "rpc_port": str(port),
```
unparalleled-js, line 75: As soon as the next py-geth is released, we won't have to call `str()` on the ints, we won't need to specify and `allow_insecure_unlock`

unparalleled-js:
will hold off on merging this until py-geth gets release

unparalleled-js:
- [ ] Switch to PoA or rule it out

PR #215, opened 2021-11-01
feat: ape test command

unparalleled-js:
### What I did

Make `ape test` work normally.

### How I did it

* Borrowed stuff from @fubuloubu 's started work
* Overrode some stuff in `EthTester` (kind of hackily) so we can support our own accounts - can be better if my open PR there merges and that gets released
* There is quite a bit of shared code between this PR and #210 -- Either can merge before the other and the left over one will be smaller
* Made `EthTester` provider handle errors expectedly.
* Create a `RevertsContextManager` and detect reverts from contracts that way.

### How to verify it

Have a demo project like this https://github.com/unparalleled-js/ape-demo-project

Run :

```bash
ape test
```

^ it should use the test provider and run all the tests nicely

_____

Next, make sure you have added one of the test accounts as a key file account, such as 

```bash
ape account import dev_0
```
and provide `0xdd23ca549a97cb330b011aebb674730df8b14acaee42d211ab45692699ab8ba5` as the pkey.

Then run the deploy script in the dem project like this:

```bash
ape run deploy --network ::test
```

Note you will have to specify to use the test network on this one because it is not the default for scripts / console.

### Checklist


src/ape_test/providers.py
```diff
@@ -1,8 +1,94 @@
...
+from hexbytes import HexBytes
 from web3 import EthereumTesterProvider, Web3
 
-from ape.api import ProviderAPI, ReceiptAPI, TransactionAPI
+from ape.api import ProviderAPI, ReceiptAPI, TransactionAPI, TransactionStatusEnum
+from ape.exceptions import ContractLogicError, OutOfGasError, VirtualMachineError
+from ape.utils import generate_dev_accounts
+
+
+def _setup_tester_chain(
```
unparalleled-js, line 21: This part can be removed and this module can be a lot simpler once https://github.com/ethereum/eth-tester/pull/209 merges and gets released

unparalleled-js:
Probably should hold off on this until the eth-tester PR goes through and gets released.

unparalleled-js:
closing for now, will reopen when its time

PR #216, opened 2021-11-03
feat: test config that includes mnemonic and number of accounts

unparalleled-js:
### What I did

Creating a config for the test plugin for setting a mnemonic and the number of accounts. This work is also present on the testing wip PR (#215), so it will help focus that PR.

Using this, I can create a testing account and provider setup, which I can open a PR for after this (also work present on the testing wip PR). And I have tested that setup with both geth and ethtester.

I know @lost-theory was working on this, let me know if you want to merge.

### How I did it

Added a config api implementation to the test plugin with the required fields

### How to verify it

Verify using the #215 if you want

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Callable, Iterator, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Callable, Iterator, List, Optional, Type, Union
 
 from ape.types import (
     AddressType,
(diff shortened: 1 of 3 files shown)


src/ape/api/accounts.py
```diff
@@ -134,6 +137,7 @@ def deploy(self, contract_type: ContractType, *args, **kwargs) -> ContractInstan
 class AccountContainerAPI:
     data_folder: Path
     account_type: Type[AccountAPI]
+    config_manager: "ConfigManager"
```
unparalleled-js, line 140: Any container that is for testing needs to know the configuration of the test config

fubuloubu (approved):

PR #219, opened 2021-11-03
fix: only handle errors from geth in ape-http provider plugin

unparalleled-js:
### What I did

Makes error handling geth focused for ape-http.
This work is really needed to make the other testing PRs smaller and it is useful by itself.

### How I did it

Checkout this repo https://github.com/unparalleled-js/hdpath-geth
Run `make`

Make sure you have a test account in ape form that mnemonic uses in that project 
e.g. private key `0xdd23ca549a97cb330b011aebb674730df8b14acaee42d211ab45692699ab8ba5` (first account)

Mess around with causing reverts of different kinds

*This work will greatly benefit the geth work and also help with the testing work in general*

### Checklist


src/ape/utils.py
```diff
@@ -1,17 +1,15 @@
 import collections
```
unparalleled-js, line 1: The changes to this file comprise of two things:

1. Delete a method related provider error handling. We have since learned providers should handle their errors on their own since they are always unique.
2. Add a couple new methods that get used elsewhere.

src/ape_http/providers.py
```diff
@@ -1,15 +1,23 @@
-from typing import Iterator
```
unparalleled-js, line 1: Main changes of PR are in this file.

We just stop using the util method and use a more custom error handling approach to geth, defined in the util method at the bottom of the module.

src/ape/utils.py
```diff
@@ -170,42 +168,39 @@ def generate_dev_accounts(
     return accounts
 
 
-def get_tx_error_from_web3_value_error(web3_value_error: ValueError) -> TransactionError:
+def get_gas_estimation_revert_error_message(tx_error: Exception) -> str:
```
fubuloubu, line 171: this name seems super long

tests/functional/http/test_providers.py
```diff
@@ -21,24 +39,55 @@ def test_send_when_web3_error_raises_transaction_error(
...
+            "message": "Test Error Message",
         }
         mock_web3.eth.send_raw_transaction.side_effect = ValueError(web3_error_data)
-        with pytest.raises(VirtualMachineError) as err:
+        with pytest.raises(TransactionError) as err:
             provider.send_transaction(mock_transaction)
 
         assert web3_error_data["message"] in str(err.value)
+
+    def test_send_transactions_failing_and_used_all_gas_raises_out_of_gas_error(
```
fubuloubu, line 53: lol this method name

fubuloubu (approved):

tests/functional/http/test_providers.py
unparalleled-js: It is a test!

`test_<condition>_<expectation>()`

unparalleled-js: I trimmed it down some

unparalleled-js:
Thank you!

PR #224, opened 2021-11-07
feat: start ephemeral development geth when no HTTP provider running

unparalleled-js:
### What I did

fixes: #204 

I can open a concise PR for the first pass of `ape test` as soon as this gets merged in.
I have tested these changes with my `ape test` work. The tests run super duper fast!

### How I did it

Geth PoA chain using genesis file to spawn accounts from the test configured mnemonic
Deletes data dir on shutdown... Acts ephemerally.

### How to verify it

Run a simple script.
Notice that it first spawns up a local PoA blockchain and then runs the script.
Make sure to use an account that is based off the test-mnemonic!

### Checklist


src/ape_http/providers.py
```diff
@@ -35,24 +129,64 @@ class NetworkConfig(ConfigItem):
...
+        super().__init__(
+            "geth is not installed and there is no local provider running.\n"
+            "Things you can do:\n"
+            "\t1. Install geth and try again\n"
+            "\t2. Run geth separately and try again\n"
+            "\t3. Use a different ape provider plugin"
+        )
+
+
 class EthereumProvider(ProviderAPI):
```
fubuloubu, line 143: Might make sense to rename to `GethProvider`

fubuloubu (approved):

src/ape_http/providers.py
unparalleled-js: Should we also rename the plugin to `ape-geth` ?

PR #225, opened 2021-11-08
feat: test command

unparalleled-js:
### What I did

`ape test` command (synchronous)

### How I did it

Copied stuff from fubu's branch.
Copied stuff from brownie.

### How to verify it

Checkout https://github.com/unparalleled-js/ape-demo-project
Run `ape test`.
The tests should run and pass.

It should compile before testing if needbe.

Note: make sure you have geth installed.

Note: unable to easily add tests using geth as the provider.
Once we are supporting London fork changes, we can add all sorts of tests a lot easier using `EthTester` as the provider, which is great for CI/CD.

Note: xdist is going to be a follow up

### Checklist


pyproject.toml
```diff
@@ -25,7 +27,7 @@ markers = "fuzzing: Run Hypothesis fuzz test suite"
 line_length = 100
 force_grid_wrap = 0
 include_trailing_comma = true
-known_third_party = ["IPython", "click", "dataclassy", "eth_abi", "eth_account", "eth_typing", "eth_utils", "geth", "github", "hexbytes", "hypothesis", "hypothesis_jsonschema", "importlib_metadata", "pluggy", "pytest", "requests", "setuptools", "web3", "yaml"]
+known_third_party = ["IPython", "_pytest", "click", "dataclassy", "eth_abi", "eth_account", "eth_typing", "eth_utils", "geth", "github", "hexbytes", "hypothesis", "hypothesis_jsonschema", "importlib_metadata", "pluggy", "pytest", "requests", "setuptools", "web3", "yaml"]
```
fubuloubu, line 30: why underscore?

src/ape_test/_cli.py
```diff
@@ -0,0 +1,23 @@
...
+@click.command(
+    add_help_option=False,  # NOTE: This allows pass-through to pytest's help
+    short_help="Launches pytest and runs the tests for a project",
+    context_settings=dict(ignore_unknown_options=True),
+)
+@network_option()
+@click.argument("pytest_args", nargs=-1, type=click.UNPROCESSED)
+def cli(network, pytest_args):
+    if "--network" not in pytest_args:
+        pytest_args = [*pytest_args, "--network", network]
```
fubuloubu, line 18: Is `pytest_args` overriding the network option?

src/ape_test/contextmanagers.py
```diff
@@ -0,0 +1,29 @@
...
+            ) from exc_value
+
+        actual = exc_value.revert_message
+
+        if self.expected_message is not None and self.expected_message != actual:
+            raise AssertionError(
+                f"'{self.expected_message}' not found in revert message '{actual}'."
+            )
+
+        return True  # Assertion passes
```
fubuloubu, line 29: wait, why is this returning True? is this just how it integrates with pytest?

src/ape_test/plugin.py
```diff
@@ -0,0 +1,70 @@
...
+    # Enable verbose output if stdout capture is disabled
+    config.option.verbose = config.getoption("capture") == "no"
+
+    # Inject the runner plugin (must happen before fixtures registration)
+    session = PytestApeRunner(config, project, networks)
+    config.pluginmanager.register(session, "ape-test")
+
+    # Only inject fixtures if we're not configuring the x-dist master runner
+    has_xdist = "numprocesses" in config.option
+    if not has_xdist or not config.getoption("numprocesses"):
```
fubuloubu, line 52: Seems like this code be abstracted out with the lines above in `_get_runner_class`

pyproject.toml
unparalleled-js: I am not sure, this got added automatically

src/ape_test/_cli.py
unparalleled-js: If you do `ape test --network <network>`, it gets added to pytest_args.
If you do `ape --network <network> test`, it gets added to the network option.
If you do the latter, we just inject it to the pytest args. This way, we don't start up any providers unless tests are collected

src/ape_test/contextmanagers.py
unparalleled-js: I copied it from brownie, but you are right, it does not seem to be needed

src/ape_test/plugin.py
unparalleled-js: Just realized was not using `_get_runner_class()` (probably will do something like that for xdist).

It is probably best to just remove checking for xdist for now!

pyproject.toml
unparalleled-js: It was because we were importing the `Config` class from `_pytest` but that was related to the xdist check (which I just removed).

src/ape_test/contextmanagers.py
unparalleled-js: I just found out why it is needed: https://docs.python.org/3/library/contextlib.html

^ From line:  `By returning true values, these callbacks can suppress exceptions the same way context manager __exit__() methods can.`

That implies that returning `True` from `__exit__` causes the exceptions to not get raised, which is what we want here

src/ape_test/_cli.py
unparalleled-js: I guess `ape --network <network> test` does not actually work so this is not really needed, I just removed!

src/ape/managers/networks.py
```diff
@@ -86,7 +86,9 @@ def get_provider_from_choice(
         provider_settings: Optional[Dict] = None,
     ) -> ProviderAPI:
         if network_choice is None:
-            return self.default["development"].get_provider(provider_settings=provider_settings)
+            return self.default_ecosystem["development"].get_provider(
```
unparalleled-js, line 89: There is no `self.default`, I wonder why mypy didn't complain
So this is like a bug fix to an unreleased piece of work I am sneaking in...

src/ape_test/contextmanagers.py
fubuloubu: lmao tf

unparalleled-js: Problem something to keep in mind when working with context managers.

Also note that `pytest.raises()` does this, as well as `brownie.reverts()`

PR #228, opened 2021-11-09
feat: upgrade black un pre-commit hook

unparalleled-js:


unparalleled-js:
Ah maybe, we should include this in the release first

fubuloubu (approved):

Issue #229, opened 2021-11-09
Second class plugins are included with first class plugins with installed from pypi


unparalleled-js:
### Environment information

python 3.9.7
macOS
ape 0.1.0a26

### What went wrong?

I am seeing a strange issue from installing ape from production along with a second class plugin (solidity in this case, also from production). They get included as first class plugins and can't be uninstalled:

```python
import ape
from ape_plugins import utils
print(utils.FIRST_CLASS_PLUGINS)
```

Outputs `{'ape_plugins', 'ape_networks', 'ape_solidity-0.1.0a3.dist-info', 'ape_solidity', 'ape_console', 'ape_compile', 'ape_pm', 'ape_test', 'ape_ethereum', 'ape_run', 'ape_accounts', 'ape_geth'}`

^ Notice that `ape_solidity` is in there twice as well

### How can it be fixed?
Finding another way to distinguish the first class plugins

unparalleled-js:
Ah, the reason this is happening is because all the plugins, whether first or second or third or anything, get installed in `site-packages`, so every ape plugin is considered a 1st class plugin here.

PR #230, opened 2021-11-09
fix: issue where all plugins got counted as 1st class plugins

unparalleled-js:
### What I did

fixes: #229 

### How I did it

Set `py_modules` from magic list in ape
Rely on same list in plugins plugin

### How to verify it

Install ape without using -e
Install plugin
Uninstall plugin

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
-from os import path
+from pathlib import Path
 from typing import Dict
 
 from setuptools import find_packages, setup  # type: ignore
 
-here = path.abspath(path.dirname(__file__))
+here = Path(__file__).parent.absolute()
 packages_data: Dict = {}
-with open(path.join(here, "src", "ape", "__modules__.py"), encoding="utf8") as modules_file:
+with open(here / "src" / "ape" / "__modules__.py", encoding="utf8") as modules_file:
     exec(modules_file.read(), packages_data)
 
 extras_require = {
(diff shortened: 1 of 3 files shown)


unparalleled-js:
I can help resolve merge conflicts if this interferes with @Ninjagod1251 's work (or we can wait on this.. However, this is kind of a critical bug!)

src/ape_plugins/utils.py
```diff
@@ -1,18 +1,15 @@
...
 from ape.logging import logger
 
-# Plugins included with ape core
-FIRST_CLASS_PLUGINS: Set[str] = {
-    d.name for d in Path(__file__).parent.parent.iterdir() if d.is_dir and d.name.startswith("ape_")
-}
-
 # Plugins maintained OSS by ApeWorX (and trusted)
+
+FIRST_CLASS_PLUGINS = set(__modules__)
```
unparalleled-js, line 12: I tried finding a way to use `pkg_resources` or `distutils` to find the 1st class packages but could not find a way to do it. 

I also tried `set([p.__name__ for p in plugin_manager.get_plugins()])` but that also contains alllll the plugins, not just the first class

setup.py
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env python
+from os import path
```
fubuloubu, line 2: use `pathlib`

fubuloubu (approved):

PR #231, opened 2021-11-09
feat: web3 provider base class

unparalleled-js:
### What I did

Creates a base class for providers that use web3 rpc.
I have a branch where hardhat will also use this base class (to help upgrade it and make it fit for ape-test)

### How I did it

Copy shared code to base class and put in Provider api module.

### How to verify it

- [x] Geth provider still works
- [ ] Check out my branch feat/use-new-baseclass in ape-hardhat and see that you can run tests and scripts and stuff using a version of the provider that uses this base class.

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -4,6 +4,7 @@ from typing import Iterator, List, Optional
 
 from dataclassy import as_dict
 from hexbytes import HexBytes
+from web3 import Web3
 
 from ape.logging import logger
 from ape.types import TransactionSignature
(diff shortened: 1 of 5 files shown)


unparalleled-js:
Here is an example using it in hardhat: https://github.com/ApeWorX/ape-hardhat/pull/10/files

unparalleled-js:
Also, tried it out with Infura! https://github.com/ApeWorX/ape-infura/pull/10/files

src/ape/api/providers.py
```diff
@@ -187,3 +189,95 @@ def snapshot(self) -> str:
...
+
+class Web3Provider(ProviderAPI):
+    """
+    A base provider that is web3 / RPC based.
+    """
+
+    _web3: Web3 = None  # type: ignore
+
+    def connect(self):
+        self._web3 = Web3(HTTPProvider(self.uri))
```
fubuloubu, line 202: what is `self.uri`? perhaps add it as an abstractmethod property

unparalleled-js: Talked elsewhere, plan is to keep `connect` and `disconnect` abstract for this class because not all web3 providers will need a URI

fubuloubu (approved):

PR #234, opened 2021-11-10
fix: keep type info for all ape packages up-to-date

unparalleled-js:
### What I did

Adds missing packages to packages_data.. Fixes issue where not all packages had their typing including.
Makes it so it goes off the universal source of truth, so it should not happen in the future!

### How I did it

Noticed package_data. Noticed it was missing some packages. Noticed we could use the same source as truth as `py_modules`.

### How to verify it

Install ape, both editable mode and not, make sure it installs expectedly.
Shouldn't need type ignore anymore if was importing from these missing modules? (not exactly sure)

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,16 +108,7 @@ setup(
     keywords="ethereum",
     packages=find_packages("src"),
     package_dir={"": "src"},
-    package_data={
-        "ape": ["py.typed"],
-        "ape_accounts": ["py.typed"],
-        "ape_compile": ["py.typed"],
-        "ape_ethereum": ["py.typed"],
-        "ape_geth": ["py.typed"],
-        "ape_run": ["py.typed"],
-        "ape_test": ["py.typed"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -108,16 +108,7 @@
...
-        "ape": ["py.typed"],
-        "ape_accounts": ["py.typed"],
-        "ape_compile": ["py.typed"],
-        "ape_ethereum": ["py.typed"],
-        "ape_geth": ["py.typed"],
-        "ape_run": ["py.typed"],
-        "ape_test": ["py.typed"],
-        "ape_pm": ["py.typed"],
-    },
+    package_data={p: ["py.typed"] for p in packages_data["__modules__"]},
```
unparalleled-js, line 111: Just realized that `__modules__` should maybe be `__packages__`.

fubuloubu (approved):

PR #236, opened 2021-11-11
feat: allow iteration/existance checking contract types in project

fubuloubu:
### What I did
quick little PR to allow checking if a contract name is in a project e.g. `"MyContract" in project`

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, Iterator, List, Optional
 
 import requests
 from dataclassy import dataclass
@@ -168,6 +168,9 @@ class ProjectManager:
     @property
     def contracts(self) -> Dict[str, ContractType]:
         return self.load_contracts()
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self.contracts)
 
     def __getattr__(self, attr_name: str):
         contracts = self.load_contracts()


unparalleled-js (approved):
LGTM

I tested it out and will use this in my demo project.

unparalleled-js:
Hmm... Actually, when I do 

```
    account = get_account()
    for contract in project:
        return account.deploy(project[contract], *args, **kwargs)
```

I get error `TypeError: 'ProjectManager' object is not subscriptable`

unparalleled-js:
I _am_ able to check if it `"Fund" in project` though

unparalleled-js:
Maybe we should also either make the ProjectManager subscriptable or make a `get_contract()` method so we can loop through the contract and deploy them all. Otherwise, I have to use `getattr()` to access each contract type

fubuloubu:
> I _am_ able to check if it `"Fund" in project` though
> 
> Edit: Realize I am also able to iterate but wondering about the UX of accessing the contracts while looping

beginning to think that since we already have `project.contracts`, we should just use that instead e.g. `if "MyContract" in project.contracts: return project.MyContract`

unparalleled-js:
> > I _am_ able to check if it `"Fund" in project` though
> > Edit: Realize I am also able to iterate but wondering about the UX of accessing the contracts while looping
> 
> beginning to think that since we already have `project.contracts`, we should just use that instead e.g. `if "MyContract" in project.contracts: return project.MyContract`

Explicit is good.. Looping through a project might be confusing, like I am looping through the project's test and scripts or contracts, it might not be clear as day

fubuloubu:
closing as not needed

use `project.contracts` instead

Issue #192, opened 2021-10-14
Add support for TypedTransactions


fubuloubu:
### What went wrong?
The London hardfork will add typed transaction envelopes ([EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)). `eth-account` version 0.5.5 [added support](https://eth-account.readthedocs.io/en/latest/release_notes.html#eth-account-v0-5-5-2021-07-21) for these upgrades.

### How can it be fixed?
Add support to the Ethereum ecosystem plugin for typed transaction support. It might take some better abstractions to get it right.

Issue #104, opened 2021-07-22
Add support for EIP-1559


fubuloubu:
depends: #192

### What went wrong?
The London hardfork will change the format for gas costs via [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559). `eth-account` version 0.5.5 [added support](https://eth-account.readthedocs.io/en/latest/release_notes.html#eth-account-v0-5-5-2021-07-21) for these upgrades.

### How can it be fixed?
Add support to Transaction type for eip-1559 style gas abstractions. It might take some better abstractions to get it right.

fubuloubu:
Idea: abstract the `gas_price` kwarg into a `GasStrategyAPI` plugin system, allowing users to select a gas strategy via kwarg e.g. `contract.my_transaction(..., gas_strategy=GasStrategyPlugin(), ...)` (with some handling for EIP1559 behaviors):

```python
contract.my_transaction(..., max_fee="10 gwei", ...)  # Use post-EIP1559 behavior (built-in Constant Conversion to `gas_strategy`)

contract.my_transaction(..., gas_strategy=GasNow("fast"), ...)  # Use gasnow API "fast" gas (no resends)

contract.my_transaction(..., gas_strategy=LinearGasStrategy(start_gas="10 gwei", end_gas="50 gwei"), ...)
```

unparalleled-js:
This should probably be two separate issues, one for EIP-1559 alone and one for typed transactions.

PR #237, opened 2021-11-11
feat: support dynamic fee transactions

unparalleled-js:
### What I did

fixes: #192
fixes: #104

Note that these changes will try to use the dynamic model as the default.
Running scripts, tests, etc still all works the same as before when you are not specifying any parameters like `gas_price` etc!

### How I did it

* Create `BaseTransaction` and subclasses `StaticFeeTransaction` and `DynamicFeeTransaction` from the original `Transaction` in `ape_ethereum.ecosystems`.
* Moved some encoding specific logic from `TransactionAPI` to `ape_ethereum` transaction classes. Created API method `set_defaults(provider)` that will set these now. Each transaction subclass has their own implementation but both call the base class's as well.

I have tested this out with `EthTester` and it works!
I have also tested with our geth provider.

### How to verify it

Send transactions using `gas_price=x`.
Send transactions using `max_priority_fee=x`


src/ape/api/accounts.py
```diff
@@ -60,18 +60,9 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
-        if txn.gas_price is None:
-            txn.gas_price = self.provider.gas_price
-        # else: assume user specified a correct price, or will take too much time to confirm
-
-        # NOTE: Allow overriding gas limit
-        if txn.gas_limit is None:
-            txn.gas_limit = self.provider.estimate_gas_cost(txn)
-        # else: assume user specified the correct amount or txn will fail and waste gas
-
+        txn.set_defaults(self.provider)
```
unparalleled-js, line 63: The above logic now happens in implementation's call of the `set_defaults(provider)` method.

src/ape/api/address.py
```diff
@@ -24,14 +24,6 @@ def provider(self) -> ProviderAPI:
     def provider(self, value: ProviderAPI):
         self._provider = value
 
-    @property
```
unparalleled-js, line 27: These were not used, so I deleted them?

src/ape_test/providers.py
```diff
@@ -32,7 +32,17 @@ def chain_id(self) -> int:
 
     @property
     def gas_price(self) -> int:
-        # NOTE: Test chain doesn't care about gas prices
+        """Returns 0 because test chains do not care about gas prices."""
```
unparalleled-js, line 35: Fun fact: test chains actually do care about gas now....

But, I have a branch (that relies on this one) that makes eth-tester work all over.

setup.py
```diff
@@ -85,7 +85,7 @@
         "singledispatchmethod ; python_version<'3.8'",
         "IPython>=7.25",
         "pytest>=6.0,<7.0",
-        "web3[tester]>=5.18.0,<6.0.0",
+        "web3[tester]>=5.24.0,<6.0.0",
```
unparalleled-js, line 88: - [ ] Wait for 5.25.0 and require that....

src/ape/api/accounts.py
```diff
@@ -60,18 +60,9 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
         if send_everything:
-            txn.value = self.balance - txn.gas_limit * txn.gas_price
+            txn.value = self.balance - txn.max_fee
```
unparalleled-js, line 65: `max_fee` is calculated differently per transaction type

src/ape/api/networks.py
```diff
@@ -33,7 +33,7 @@ class EcosystemAPI:
     data_folder: Path
     request_header: str
 
-    transaction_class: Type["TransactionAPI"]
+    transaction_class_map: Mapping[str, Type["TransactionAPI"]]
```
unparalleled-js, line 36: Now, ecosystems can support N number of transaction types

src/ape_ethereum/ecosystem.py
```diff
@@ -90,7 +182,10 @@ def decode(cls, data: dict) -> ReceiptAPI:
 
 
 class Ethereum(EcosystemAPI):
-    transaction_class = Transaction
+    transaction_class_map = {
+        "0x0": StaticFeeTransaction,
```
unparalleled-js, line 186: Does anyone use type `0x1` ?
I think we may need an intermediary tx class to support it but I am not sure why it would be needed

src/ape_test/providers.py
fubuloubu: rekt by EIP-1559

I think I was talking with steve or you at one point about setting the base fee in genesis to 0, but yeah I guess that would still affect this target gas calculation.... :thinking: 

if target gas were high enough (maybe 50m gas), and base fee were zero, then it should be impossible to have the algorithm trigger an upward adjustment, which only happens if the gas used in a block exceeds the target.

src/ape_ethereum/ecosystem.py
fubuloubu: Yes, static fee transactions are still used by people on mainnet, and other chains that haven't migrated yet. also, combing through history

src/ape/api/providers.py
```diff
@@ -6,10 +6,10 @@
 from hexbytes import HexBytes
 from web3 import Web3
 
-from ape.logging import logger
 from ape.types import TransactionSignature
 
 from ..exceptions import ProviderError
+from ..logging import logger
```
fubuloubu, line 12: just a little nit, but I usually don't like using double dot like this as it makes the code dependent on structure a little more

src/ape/api/providers.py
```diff
@@ -23,24 +23,40 @@ class TransactionAPI:
     nonce: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
     value: int = 0
     gas_limit: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
-    gas_price: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
     data: bytes = b""
+    type: str = ""
```
fubuloubu, line 27: Can type be an enum?

src/ape/api/providers.py
```diff
@@ -23,24 +23,40 @@ class TransactionAPI:
...
         """
-        # TODO Support EIP-1559
-        return (self.gas_limit or 0) * (self.gas_price or 0) + self.value
+        return self.value + self.max_fee
+
+    def set_defaults(self, provider: "ProviderAPI"):
+        """
+        Set default values in a transaction with the help of the given provider,
+        such as setting the `gas_limit` from an RPC call to ``eth_gasLimit``.
+        """
```
fubuloubu, line 59: Wondering if the dependency injection might be made a little more explicit

src/ape/api/providers.py
```diff
@@ -159,6 +181,16 @@ def estimate_gas_cost(self, txn: TransactionAPI) -> int:
...
+    @property
+    @abstractmethod
+    def priority_fee(self) -> int:
+        ...
+
+    @property
+    @abstractmethod
+    def base_fee(self) -> int:
+        ...
+
```
fubuloubu, line 193: Should these raise `NotImplementedError`s by default? In case a network doesn't have EIP-1559

src/ape_ethereum/ecosystem.py
```diff
@@ -25,8 +33,18 @@
...
 
-# TODO: Fix this to add support for TypedTransaction
-class Transaction(TransactionAPI):
+class BaseTransaction(TransactionAPI):
+    type: str = None  # type: ignore
+
+    def set_defaults(self, provider: ProviderAPI):
+        """
+        Sets ``gas_limit`` if it is None.
+        Sub-classes should likely should call this method.
```
fubuloubu, line 42:
```suggestion
        Sub-classes would likely call this method.
```

src/ape_ethereum/ecosystem.py
```diff
@@ -75,7 +92,83 @@ def encode(self) -> bytes:
...
+    gas_price: int = None  # type: ignore
+    type: str = "0x0"
+
+    @property
+    def max_fee(self) -> int:
+        return (self.gas_limit or 0) * (self.gas_price or 0)
+
+    def set_defaults(self, provider: ProviderAPI):
+        if self.gas_price is None:
+            self.gas_price = provider.gas_price
```
fubuloubu, line 109: This would query the value of gas price only at this point in time. Perhaps it should be more lazily querying.

src/ape_test/providers.py
```diff
@@ -32,7 +32,17 @@ def chain_id(self) -> int:
...
+        return self.base_fee + 1
```
fubuloubu, line 36:
```suggestion
        return self.base_fee  # no miner tip
```

src/ape_test/providers.py
unparalleled-js: In a test env, `base_fee + 1` should suffice, right? because everything just gets mined.

src/ape/api/providers.py
unparalleled-js: Yeah, and I was actually wondering if you'd be interested in just going 100% absolute paths everywhere, I think `isort` created a whole new import section for relative imports rather than including them with the 1st party ones.

src/ape/api/providers.py
unparalleled-js: I think maybe the API property should be a native type, like `str`, so it is most extensible? Otherwise, there is no good way to have like a "base" enum. `ape-ethereum` will use the `0x0` and `0x2` but not every ecosystem may use those same transaction versions (some may exclude `0x2` altogether for example).

unparalleled-js: I can use enum in `ape-ethereum` at the very least though, to validate

unparalleled-js: `TransactionStatusEnum` is similar, except less of a big deal because int-enums can be argued as just being arbitrarily ordered.

src/ape/api/providers.py
unparalleled-js: This is not really "injecting" the provider here, it is setting defaults on the transaction by making calls to the provider.

I don't really like this approach that much though... I just don't know how else to make it so that different transaction types can request different defaults to be set from the provider based on their type / needs, if that makes sense... Something to keep thinking about! Please help.

src/ape/api/providers.py
unparalleled-js: I was imagining each of those networks to raise the error themselves, but yeah this makes it so they don't have to.

Done!

src/ape_ethereum/ecosystem.py
unparalleled-js: This only gets called right before sending a transaction in `AccountAPI.call()`.
Would a warning suffice for this method? Like, `Only set the defaults on the transaction before sending it otherwise they may become outdated."` ?

Otherwise, the transaction almost need a provider as its dependency (I guess it kind of does in this sense).

src/ape/api/providers.py
unparalleled-js: I moved the `set_defaults` to be on the provider. Now the provider just checks the type of the transaction and sets the defaults accordingly... It is an extra `if` statement but I think the design makes more sense, having the provider take the txn as the arg rather than the other way around, I think this at least partially remedies some of your feedback.

unparalleled-js: ^ Scratch that lol... I just deleted the `set_defaults()` method for now and do type checking in the accounts api next to all all the stuff before sending a txn.

src/ape/api/networks.py
```diff
@@ -33,7 +33,7 @@ class EcosystemAPI:
     data_folder: Path
     request_header: str
 
-    transaction_class: Type["TransactionAPI"]
+    transaction_class_map: Mapping["TransactionType", Type["TransactionAPI"]]
```
fubuloubu, line 36:
```suggestion
    transaction_types: Mapping["TransactionType", Type["TransactionAPI"]]
```

src/ape/api/providers.py
```diff
@@ -1,20 +1,25 @@
...
 
-from ..exceptions import ProviderError
 from . import networks
 from .base import abstractdataclass, abstractmethod
 from .config import ConfigItem
 
 
+class TransactionType(Enum):
+    STATIC = "0x0"
+    DYNAMIC = "0x2"
```
fubuloubu, line 20:
```suggestion
    DYNAMIC = "0x2"  # EIP-1559
```

src/ape/api/networks.py
```diff
@@ -33,7 +33,7 @@ class EcosystemAPI:
     data_folder: Path
     request_header: str
 
-    transaction_class: Type["TransactionAPI"]
+    transaction_types: Mapping["TransactionType", Type["TransactionAPI"]]
```
fubuloubu, line 36: Should this just be a `Dict` type? To be more explicit

fubuloubu (dismissed):

src/ape/api/networks.py
unparalleled-js: `mypy` was getting made about something I think, but let me try it again

unparalleled-js: Done!

fubuloubu (approved):

PR #239, opened 2021-11-13
test: allow CLI test projects to opt-out of tests

unparalleled-js:
### What I did

Make it so that the projects in `tests.integration.cli` can have a `tests-config.yaml` file that includes a list of `skip-tests`. Any test in the skip-tests data will not run for that project!

Additionally, I refactored some of the integration tests to use this new approach instead of having early returns and stuff.

This _should_ help us all add integration tests... and make our lives easier.

### How I did it

The hook `pytest_collection_modifyitem()` and filtering.

Also, reading the `tests-config.yaml` file for each project and all the tests in `tests.integration.cli`

It still uses the parametrized fixture approach. See the note below for more historical details!

___

I started off working with that branch that @fubuloubu and I started on together during a peer-programming session.
On that branch, we were fiddling with creating custom test items from a parent node that is the project's `tests-config.yaml` file. I got something almost working, but it required copying a lot of internal code from `pytest` and it was not working out that great. I then learned of `pytest_collection_modifyitems()` and how it can be used to filter out tests. This approach has turned out to give us exactly what we need without tweaking the internals of `pytest` too much.. I structured the code in this PR to be re-usable for if we do end up needing another approach.

Also, with what I have learned about creating custom `pytest` nodes, I think that approach could still make sense for `vyper` or elsewhere.

### How to verify it

The tests should run!  If you copy only the test changes to the main branch, they fail because the `if` checks are missing and tests are configured to run per project elsewhere (in the `tests-config.yaml` file).
Also try messing with the `tests-config.yaml`!

### Checklist


tests/integration/cli/test_run.py
```diff
@@ -1,19 +1,24 @@
 BAD_COMMAND = "not-a-name"
 
 
+def test_run_no_scripts_dir(ape_cli, runner, project):
```
unparalleled-js, line 4: One thing I don't like about this is that it is not obvious what projects are skipped per test when looking at the test alone...
I am pretty sure using a very similar approach, I can make it work from a custom decorator, something like:

```python
@skip_projects(["interface", "script"])
def test_run_no_scripts_dir(ape_cli, runner, project):
  ...
```

if that seems interesting

fubuloubu: I think this would definitely help. I also think this decorator should throw if it doesn't match a valid project name to prevent ensuing madness

tests/integration/cli/test_compile.py
```diff
@@ -1,26 +1,37 @@
...
+    result = runner.invoke(ape_cli, ["compile"])
+    assert result.exit_code == 0
+    assert "WARNING" in result.output
+    assert "No 'contracts/' directory detected" in result.output
+    return  # Nothing else to test for this project
+
+
+def test_missing_extensions(ape_cli, runner, project):
+    """
+    Run for project 'unregistered-contracts`.
```
fubuloubu, line 14: yeah i definitely like the decorator idea, these are weirdly duplicitous

fubuloubu (commented):
Do the decorator set up instead

tests/integration/cli/test_run.py
unparalleled-js: Check out the decs!

I made two:

`skip_projects()`
and 
`skip_projects_except()`.

Because sometimes you literally want to test a weird part of just one project and that test would make zero sense anywhere else.  Now, it is quite intuitive to do this! I like this decorator approach a lot.

tests/integration/cli/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    def do_skip(self, project: str, module: str, test: str) -> bool:
+        """
+        Returns ``True`` if a test has been marked to be
+        skipped for the given project using the ``skip_project`` or
+        ``skip_project_except`` decorators.
+        """
+        result = test in self.projects[project].get(module, [])
+        return result
+
+    def _raise_if_not_exists(self, project_name: str, node_id: str):
```
unparalleled-js, line 56: @fubuloubu Any project specified that does not exist will raise an error, that should be helpful when writing tests if you butcher the name of a project or something. 

This addresses that part of your feedback from another comment! :)

tests/integration/cli/test_compile.py
```diff
@@ -33,9 +38,9 @@ def test_compile(ape_cli, runner, project):
     for file in project.path.glob("contracts/**/*"):
         assert file.stem not in result.output
 
-    if not any(c.deploymentBytecode for c in project.contracts.values()):
-        return  # Only interfaces
 
+@skip_projects_except([])
```
fubuloubu, line 42: is this necessary?

fubuloubu (approved):

PR #240, opened 2021-11-15
feat: contract CLI options

unparalleled-js:
### What I did

Creates a new custom `ape.cli.option` for contract.
* The contract type _must_ be a contract within the current project
* The command fails in the arg collection stage when not in a project with the given contract (which is before starting up a network!).

Use case:

So you want to make a custom command that queries logs for a given contract. You have a provider like Infura and you have the compiled ABI etc. This makes it easy to pass in a CLI option for specifying the contract type.

### How I did it

Custom click option in the cli module using `from ape import projects`.

### How to verify it

```python
import click
import json
from ape.cli import contract_option

@click.group()
def cli():
    pass


@cli.command()
@contract_option(help="The name of the contract to dump the ABI from.", required=True)
def abi(contract):
    abi_dict = {"abi": [abi.to_dict() for abi in contract.abi]}
    output = json.dumps(abi_dict, indent=4)
    click.echo(output)


@cli.command()
@contract_option(help="The name of the contracts to list the extensions for.", multiple=True, required=True)
def list_ext(contract):
    exts = []
    for con in contract:
        ext = str(con.sourceId).split(".")[-1]
        exts.append(ext)

    click.echo(f"The given contracts have the following extensions: {exts}")
```

run the first one like `ape jules abi --contract Vault`. It should dump the vault source abi.

run the second one like `abi jules list-ext --contract Vault --contract Fund`. It should list the extensions `[.vy, .sol]`

### Checklist

### DIFF
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -123,7 +123,7 @@ def _load_contracts(ctx, param, value) -> Optional[Union[ContractType, List[Cont
     return [create_contract(c) for c in value] if is_multiple else create_contract(value)
 
 
-def contract_type_option(help=None, required=False, multiple=False):
+def contract_option(help=None, required=False, multiple=False):
     """
     Contract(s) from the current project.
     If you pass ``multiple=True``, you will get a list of contract types from the callback.
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -1,9 +1,13 @@
+from typing import List, Optional, Union
+
 import click
 
-from ape import networks
+from ape import networks, project
 from ape.cli.choices import AccountAliasPromptChoice, NetworkChoice
 from ape.cli.utils import Abort
+from ape.exceptions import ContractError
 from ape.logging import LogLevel, logger
+from ape.types import ContractType
 
 
 class ApeCliContextObject:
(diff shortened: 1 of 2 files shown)


fubuloubu (dismissed):
Interesting feature, seems appropiate for use w/ deploy scripts

src/ape/cli/options.py
```diff
@@ -99,3 +103,32 @@ def account_option_that_prompts_when_not_given():
...
+
+    is_multiple = isinstance(value, (tuple, list))
+
+    def create_contract(contract_name: str) -> ContractType:
+        if contract_name not in project.contracts:
+            raise ContractError(f"No contract named '{value}'")
+
+        return project.contracts[contract_name]
+
+    return [create_contract(c) for c in value] if is_multiple else create_contract(value)
```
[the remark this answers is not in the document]
unparalleled-js: I think if the user passes in `multiuple=True`, they are expecting a list. Otherwise, they are not.

Like, if the user did not know they _could_ possible `multiple=True` for some reason, it would be a little surprising to get a list... Like in the `How To Verify It` first example.

unparalleled-js:
> Interesting feature, seems appropiate for use w/ deploy scripts

Yeah! That is a good use case. Another use case is getting logs for contracts

src/ape/cli/options.py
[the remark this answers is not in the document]
fubuloubu: gotcha, okay this is fine then

src/ape/cli/options.py
```diff
@@ -99,3 +103,32 @@ def account_option_that_prompts_when_not_given():
...
+
+
+def _load_contracts(ctx, param, value) -> Optional[Union[ContractType, List[ContractType]]]:
+    if not value:
+        return None
+
+    if len(project.contracts) == 0:
+        raise ContractError("Project has no contracts.")
+
+    is_multiple = isinstance(value, (tuple, list))
```
fubuloubu, line 115: Wait, you don't opt-in to this behavior? Could make it harder to use

unparalleled-js: You do opt in.
See the examples in the `How to verify it` section!

unparalleled-js: When you pass in `multiple=True`, it works the exact same was when doing it for `click.option()`, you get a list as the `value`.

The default value is `False`, which makes `value` not a list but a single item.

Here, we detect if the user is doing multiple mode or not and return the same type of value (list or single item)

unparalleled-js: I will add a comment

fubuloubu: It seems like it would be cleaner to access that kwarg somehow, but I could see how this is not available

fubuloubu (approved):

fubuloubu:
For whatever reason, this PR started causing docker build failures
https://hub.docker.com/repository/docker/apeworx/ape/builds

PR #241, opened 2021-11-15
docs: customize the CSS and graphics

ahoymrag:
### What I did

fixes: #

### How I did it
Updated the CSS file a bit - for the docs 
- Changed the background colors
- Changed the Font Family
- Basic Padding and Leading
### Checklist

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,1 +1,124 @@
-#version-picker-label { color: white; display: inline; }
+@import url('https://fonts.googleapis.com/css?family=Space+Mono&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Rubik&display=swap');
+
+#version-picker-label { color: #ffffff; display: inline; }
+
+body {
+    font-weight: 400;
+    font-family: 'Rubik';
+    background-color: #f5f5f5;
+    min-height: 100%;
+    background: #d7d7d7;
+
+}
+
+p {
+	font-weight: 400;
+    font-family: 'Rubik';
+}
+
+h1 {
+    font-size: 65px;
+    line-height: 65px;
+    font-family: 'Space Mono';
+    text-transform: uppercase;
+    letter-spacing: -0.05em; 
+
+
+    
+}
+
+h2 {
+    font-size: 42px;
+    line-height: 42px;
+    font-family: 'Space Mono';
+    letter-spacing: -0.05em; 
+
+}
+
+a {
+    color: #FF7E7E;
+    text-decoration: none;
+    cursor: pointer;
+}
+
+a:visited {
+  color: #FF7E7E;
+}
+
+a:hover {
+  color: #000000;
+}
+
+.wy-side-scroll {
+    width: 320px;
+    background-color: #ffffff;
+    position: relative;
+    overflow-x: hidden;
+    overflow-y: scroll;
+    height: 100%;
+}
+
+
+.wy-nav-content-wrap {
+    margin-left: 300px;
+    background: #d7d7d7;
+    min-height: 100%;
+}
+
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -77,7 +82,7 @@ def fixpath(p):
     get away with having only one copy of the CSS and web fonts.
     """
     new = f"/{project}/latest/_static/" + p.split("_static")[1]
-    return new
+    return p
```
unparalleled-js, line 85: Is the variable `new` still needed then? What is the nature of this change?

docs/_static/custom.css
```diff
@@ -1 +1,124 @@
...
+    min-height: 100%;
+    color: #9b9b9b;
+    background: #343131;
+    z-index: 200;
+}
+
+footer {
+    color: grey;
+    display: none;
+}
```
unparalleled-js, line 124: Can you add a newline at the end of this file?

docs/conf.py
```diff
@@ -58,6 +58,11 @@ def setup(sphinx):
     sphinx.connect("autoapi-skip-member", autoapi_skip_members)
 
 
+# -- AG CUSTOM CODE BABY
+def setup(app):
+    app.add_css_file('custom.css')
```
unparalleled-js, line 63: nit pick, all other strings seems to use double quotes in this file

docs/conf.py
```diff
@@ -58,6 +58,11 @@ def setup(sphinx):
     sphinx.connect("autoapi-skip-member", autoapi_skip_members)
 
 
+# -- AG CUSTOM CODE BABY
```
unparalleled-js, line 61: What is AG?

docs/_static/custom.css
```diff
@@ -1 +1,124 @@
-#version-picker-label { color: white; display: inline; }
+@import url('https://fonts.googleapis.com/css?family=Space+Mono&display=swap');
```
unparalleled-js, line 1: Can you format this file?
There are lots of unneeded empty lines and some of the indentation is off. There is probably a tool like `prettier` that can help!

unparalleled-js:
Wow this looks awesome!
I am not 100% it is loading correctly for me, can I see a screenshot so I know? Otherwise the rest of my feedback may not make complete sense:

* the font under `User API Reference` seems small. It seems smaller than other fonts like it.
* the font for `Table of Contents` seems wrong and unusual, lime times new roman or something
* the color of the links might be a tad bright for being on a white background, that orange-ish color. I love the color but it is just a tad bright, maybe?
* The user api reference guides are very white, I like maybe the main headers of the guides should be black like they are for the quickstart guides

docs/conf.py
fubuloubu: This section needs work. This was a hack to get it working for, but we should figure out exactly what's going on here and why it doesn't work before merging

docs/conf.py
fubuloubu: @ahoymrag's initials lol

docs/conf.py
```diff
@@ -58,6 +58,11 @@ def setup(sphinx):
...
+def setup(app):
```
fubuloubu, line 62: Type checker is yelling about this function name shadowing an imported function. Should rename this

fubuloubu (commented):
Run `black docs/conf.py` and amend previous commit to fix linting bug

docs/conf.py
unparalleled-js: immortalized!

unparalleled-js:
Closing as this these changes made it from #302 ! Thank you so much @ahoymrag 
It should be much easier to contribute to the docs now.

PR #246, opened 2021-11-17
refactor: add project property to CLI context object

unparalleled-js:
### What I did

To limit amount of lazy importing required, I added `project` to the `cli_ctx` object. Now, the project will get lazily imported in the background, keeping the nasty out of the CLI implementations.

### How I did it

* Add a property on `ApeCliContextOjbect` called `project` that will lazily import the project and cache that and use it on subsequent calls.
* Replace usages of importing the project within a `_cli.py` command implementation. I didn't change the `ape console`, as `ape` needs to be imported anyway so there was no reason to.

### How to verify it

`ape compile` still works

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -23,8 +23,10 @@ def display_config(ctx, param, value):
     if not value or ctx.resilient_parsing:
         return
 
+    from ape import project
+
     click.echo("# Current configuration")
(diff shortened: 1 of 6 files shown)


unparalleled-js:
I should also add that this will help some work I am doing on making `ape compile` only compile its given contracts

PR #247, opened 2021-11-18
feat: only compile given contracts

unparalleled-js:
### What I did

Allows only compiling the given contracts.

I am trying to get on the path of making ape smart enough to only compile the contracts it needs for the given commands.. Rather than just always compiling everything. Lazy compiling essentially. This is the first step!

I added CLI integration tests!

### How I did it

* make `load_contracts` take `file_paths` and honor them.
* make `eth-ath[test]` require `ape-solidity` so it can compile the contracts in the tests
* make `file_paths` arg actually honored in `ape compile` command.
* move cli contract file paths logic to apecli
* add tests! (this required adding a new project, which I called `hello-world` and I plan to use this project for some other tests down the line, once eth-tester works
* had to fetch tags in github actions that required `eth-ape[test]` so it can correctly install `ape-solidity`
* added `project` to cli ctx obj and accessed that way in the compile related hooks / functions

**
I like to be able to more lazily compile contracts, the work in this PR will help get us there.
It'd be nice to have a big contract like Vault in my test project but to only compile if it I am actually using it in the tests.
Right now, it just always gets compiled. I believe this work will help us make a more lazily compilation process, like only compile the give contract (and its dependencies?) once it is requested... This is down-the-line though
**

### How to verify it

Example usages (these are from the tests):

`ape compile HelloWorld`
`ape compile HelloWorld.sol`
`ape compile contracts/HellowWorld.sol`
`ape compile foo bar`
^ Examples are in the tests!

### Checklist


src/ape_compile/_cli.py
```diff
@@ -10,12 +10,34 @@
 _flatten = chain.from_iterable
 
 
+def _create_contracts_paths(ctx, param, value):
+    from ape import project
```
unparalleled-js, line 14: We can just use the `ctx` param here if `https://github.com/ApeWorX/ape/pull/246` merges

tests/integration/cli/projects/hello-world/contracts/HelloWorld.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
```
unparalleled-js, line 1: I'd like to add a deploy script as well as some tests as well as more features to the contract for this `HelloWorld` project, so it will be like a "real' project. This will come down the line as needed!

.github/workflows/test.yaml
```diff
@@ -32,6 +32,11 @@ jobs:
         steps:
         - uses: actions/checkout@v2
 
+        - name: Get history and tags for SCM versioning to work
+          run: |
+            git fetch --prune --unshallow
+            git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+
```
fubuloubu, line 39: What is this addition doing?

setup.py
```diff
@@ -16,6 +16,7 @@
         "pytest-mock",  # For creating mocks
         "hypothesis>=6.2.0,<7.0",  # Strategy-based fuzzer
         "hypothesis-jsonschema==0.19.0",  # JSON Schema fuzzer extension
+        "ape-solidity",  # Required for compiling test contracts
```
fubuloubu, line 19: Is there a way to test this functionality with the built-in package compiler instead?

.github/workflows/test.yaml
unparalleled-js: This is needed if we include `ape-solidity` in `setup.py`. Otherwise, `ape` has a weird version in the GH Action. Anything dependent on `ape`, like plugins, are unable to install when `ape` has a weird, non-alpha version that.

setup.py
unparalleled-js: - [ ] Hmm.. Maybe we don't! I can try that.

However, I think having hello-world contracts in the tests is worth considering, for future testing purposes.

unparalleled-js: Easy Peasy! I removed all the stuff related to solidity in the tests and am just using the json compiler.

fubuloubu (approved):

PR #249, opened 2021-11-19
fix: pin typing extensions dependency to fix docker build

unparalleled-js:
### What I did

fix issue where docker image was not building

### How I did it

https://pypi.org/project/typing-extensions/ 

^ this happened on the same day, is the culprit
Some loose dependency out there broke us..  I am not sure.

### How to verify it

Docker should now build locally. and hopefully in CI.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,6 +85,7 @@ setup(
         "singledispatchmethod ; python_version<'3.8'",
         "IPython>=7.25",
         "pytest>=6.0,<7.0",
+        "typing-extensions>=3.10,<4.0.0",
         "web3[tester]>=5.24.0,<6.0.0",
     ],
     entry_points={


unparalleled-js:
looking at this closer, lol, really weird to have to put typing extensions in the install requires..

It does seem to fix the problem but what the heck.

unparalleled-js:
If this is not an immediate need, we should wait and see if this can merge: https://github.com/ethereum/py-trie/pull/122
As then we can pin `py-trie` and that will make slightly more sense than pinning `typing-extensions` (lol).

unparalleled-js:
I only made this a draft because I don't think it is the ideal solution

unparalleled-js:
OK, so the real issue appears to be in `eth-typing` and it is currently in the process of being resolved by ethereum devs,

PR #253, opened 2021-11-22
fix: in geth plugin, verify the provider is geth when connecting to already running node

unparalleled-js:
### What I did

Before, we just took it for granted that the node running was actually geth
Now, we verify first to make sure it is. If not, we raise a `ProviderError`.

Reason:  When messing around with `hardhat`, if I were to accidentally use `geth`, I would be confused at the reasons things would fail. This earlier failure will save us from that confusion and tell us what we are doing wrong!

### How I did it

Use an `else` statement after the `if` that starts the connection when it is not already running.
In that `else` logic, check the `clientVersion` on `web3`.

### How to verify it

Start geth.
Run a script.
Notice it says it is connecting to already running geth

Start hardhat
Run a script but specify `geth`
Notice it fails saying it tried to connect the geth provider to hardhat.

### Checklist


src/ape_geth/providers.py
```diff
@@ -172,6 +172,17 @@ def connect(self):
...
                 raise ConnectionError("Unable to connect to locally running geth.")
+        else:
+            # Verify is actually a geth provider
+            client_version = self._web3.clientVersion
+
+            if "geth" not in client_version.lower():
+                network_name = client_version.split("/")[0]
+                raise ProviderError(
+                    f"Attempted to connect geth plugin to non-geth network ('{network_name}')."
+                )
```
fubuloubu, line 183: Can this be a warning instead? I feel like it shouldn't outright fail in case of using an alternative node implementation like Erigon or OpenEthereum

fubuloubu (approved):

PR #255, opened 2021-11-22
feat: support mainnet-fork

unparalleled-js:
### What I did

This is some work that I did in order to make `mainnet-fork` work for downstream providers.

I am opening this PR to get design feedback too. See comments in PR for more discussion points.

### How I did it

* Create `UpstreamProvider` with property `connection_str`. HTTP providers should use their `uri` for this.
* `yield "ethereum", "mainnet-fork", NetworkAPI` from `ape_ethereum` to include the mainnet-fork as a network option.
* Had to make networks okay with not having any providers (it would raise an error before)

### How to verify it

My example (from my hardhat branch):

```python
class HardhatMainnetForkProvider(HardhatProvider):
    """
    A Hardhat provider that uses ``--fork``, like:
    ``npx hardhat node --fork <upstream-provider-url>``.

    Set the ``upstream_provider`` in the ``hardhat`` config section
    of your ``ape-config.yaml` file to specify which provider
    to use as your archive node.
    """

    def _create_process(self) -> HardhatProcess:
        upstream_provider_name = self.config.upstream_provider
        mainnet = self.network.ecosystem.mainnet
        upstream_provider = mainnet.get_provider(upstream_provider_name)

        if not isinstance(upstream_provider, UpstreamProvider):
            raise HardhatProviderError(
                f"Provider '{upstream_provider_name}' is not an upstream provider."
            )

        fork_url = upstream_provider.connection_str
        if not fork_url:
            raise HardhatProviderError("Upstream provider does not have a ``connection_str``.")

        return HardhatProcess(
            self._base_path,
            self.port,
            self._mnemonic,
            self._number_of_accounts,
            fork_url=fork_url,
        )

```

^ The code in the example in still in development / being tested, but it at least works for the easy-pass and is a POC

### Checklist


src/ape_ethereum/__init__.py
```diff
@@ -20,5 +20,6 @@ def networks():
     for network_name, network_params in NETWORKS.items():
         yield "ethereum", network_name, create_network_type(*network_params)
 
-    # NOTE: This works for `geth --dev` as it gets chain_id from itself
+    # NOTE: This works for development environments, as they get chain_id from themselves
     yield "ethereum", "development", NetworkAPI
+    yield "ethereum", "mainnet-fork", NetworkAPI
```
unparalleled-js, line 25: I believe `brownie` makes you add this via a CLI command.
Do we want to have that be how `mainnet-fork` gets added, or should we just add it here as a default, since it is rather common?

src/ape/api/providers.py
```diff
@@ -320,3 +320,16 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
         return receipt
+
+
+class UpstreamProvider(ProviderAPI):
+    """
+    A provider that can also be set as another provider's upstream.
+    """
+
+    @property
+    def connection_str(self) -> str:
```
unparalleled-js, line 331: We needed a way to have the URI be part of the API, as it wasn't.
However, I didn't want to assume it was always a URI, hence `connection_str`. Thoughts?

src/ape/api/networks.py
```diff
@@ -78,6 +78,7 @@ def __getitem__(self, network_name: str) -> "NetworkAPI":
         return self._try_get_network(network_name)
 
     def __getattr__(self, network_name: str) -> "NetworkAPI":
+        network_name = network_name.replace("_", "-")
```
unparalleled-js, line 81: This makes it so `mainnet-fork` can be the network name rather than `mainnet_fork`.

src/ape/api/networks.py
```diff
@@ -257,11 +258,7 @@ def providers(self):  # -> Dict[str, Partial[ProviderAPI]]
                     request_header=self.request_header,
                 )
 
-        if len(providers) > 0:
-            return providers
-
-        else:
-            raise NetworkError("No network providers found")
+        return providers
```
unparalleled-js, line 261: `mainnet-fork` network is empty until you add a plugin like `ape-hardhat`

src/ape_ethereum/__init__.py
fubuloubu: You know, I was sort've hoping that the plugin can add extra networks... but I'm not sure about the security implications of that

this works too

src/ape/api/providers.py
fubuloubu: This is because Hardhat has to provide an HTTP provider connection string to connect upstream? Seems kind of hacky, but I don't see a better way :thinking:

src/ape/api/networks.py
fubuloubu: See above comment about adding new networks

src/ape_ethereum/__init__.py
unparalleled-js: I tried to do that first! It was not showing up in the list for some reason, IDK.

Then I thought about ganache and how there could be multiple providers of mainnet-fork.

It seems like we either have to support adding custom networks or have mainnet-fork be a default network

src/ape/api/providers.py
unparalleled-js: Yeah, ultimately, this is the URL value that goes in `npx hardhat node --fork <url>` (or the config file, but it is easier in our case to use the CLI arg for this).

src/ape_ethereum/__init__.py
fubuloubu: Fair. I guess this seems as good of an approach as any here

src/ape/api/providers.py
```diff
@@ -320,3 +320,16 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
         receipt = self.get_transaction(txn_hash.hex())
         return receipt
+
+
+class UpstreamProvider(ProviderAPI):
+    """
+    A provider that can also be set as another provider's upstream.
+    """
+
+    @property
```
fubuloubu, line 330: should this be an abstract property?

fubuloubu (approved):

src/ape_ethereum/__init__.py
unparalleled-js: Hmm..... Hardhat actually supports forking any of the testnets as well, I don't think we should limit it `mainnet`... Thinking!

unparalleled-js: Should we just also add:

```
ethereum -> rinkeby-fork -> hardhat
ethereum -> infura-fork -> hardhat
ethereum -> kovan-fork -> hardhat
ethereum -> goerli-fork -> hardhat
```

?

src/ape_networks/_cli.py
```diff
@@ -37,11 +37,13 @@ def make_sub_tree(data: Dict, create_tree: Callable) -> Tree:
             ecosystem_tree = make_sub_tree(ecosystem, Tree)
             _networks = ecosystem["networks"]
             for network in _networks:
-                network_tree = make_sub_tree(network, ecosystem_tree.add)
                 providers = network["providers"]
-                for provider in providers:
-                    make_sub_tree(provider, network_tree.add)
+                if providers:
```
unparalleled-js, line 41: This makes it so `mainnet-fork` does `mainnet-fork` (or any provider, really) not show up if there are no providers in it.

fubuloubu (approved):

PR #257, opened 2021-11-23
feat: use rich tree when outputting networks list

unparalleled-js:
### What I did

Added a rich Tree view of `ape networks list`.

I realized when adding new networks to `ape` that it is difficult to parse `yaml` with my eyeballs.

This tree view looks pretty and I think makes this command a lot more useful.

### How I did it

Using the `rich` library:
https://rich.readthedocs.io/en/stable/
which claims to suppose the macOS, linux, and Windows, as well as python 3.6+, so we should be good with the dependency, but let me know your thoughts, this PR is also a proposal of the idea.

Also, I allowed the `yaml` version to remain an option via `--format yaml`

### How to verify it

`ape networks list`
`ape networks list --format yaml`

Screen Shot 2021-11-23 at 16 41 33
(image is a picture of the tree output format, which is the new default when not specifying `--format`)

###

- [ ] New test


fubuloubu:
This package looks awesome, and will come in real handy to support transaction trace viewing. I wonder if Brownie is even using this package :thinking:

fubuloubu (approved):

unparalleled-js:
> This package looks awesome, and will come in real handy to support transaction trace viewing. I wonder if Brownie is even using this package 🤔

I looked to see if brownie was because it looks really familiar but didn't see it in their requirements file, so IDK.
It does seem quite useful, I am always a little skeptical with new packages though, but this one seems popular and kind of standard for what we want.

Issue #265, opened 2021-11-24
`transfer` method does not work


unparalleled-js:
### Environment information

Latest `ape`

### What went wrong?

Attempting to transfer funds from one account to another fails.

It looks like `AccountsAPI` has not actually had `self._transaction_class()`,

### How can it be fixed?

`AccountAPI` is using `self._transaction_class` which does not exist.

PR #268, opened 2021-11-24
fix: issue preventing the transfer() method from working on accounts

unparalleled-js:
### What I did

fixes: #265

Makes transfer work again.
This was a regression (my fault)! So I added an integration test (a script that calls transfer) to prevent this bug regression again.

### How I did it

Made a `create_transaction()` public API method on `EcosystemAPI` that did what the `_extract_transaction_type()` thing did before except it also just constructs the txn at that point and returns it.

### How to verify it

I added an integration test to ensure transfer works!

### Checklist

### DIFF
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -24,9 +24,6 @@ class AddressAPI:
     def provider(self, value: ProviderAPI):
         self._provider = value
 
-    def _create_transaction(self, **kwargs):
-        return self.provider.network.ecosystem.create_transaction(**kwargs)
-
     @property
     @abstractmethod
     def address(self) -> AddressType:
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -24,6 +24,9 @@ class AddressAPI:
     def provider(self, value: ProviderAPI):
         self._provider = value
 
+    def _create_transaction(self, **kwargs):
+        return self.provider.network.ecosystem.create_transaction(**kwargs)
+
     @property
     @abstractmethod
     def address(self) -> AddressType:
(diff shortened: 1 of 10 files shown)


tests/integration/test_transfer.py
```diff
@@ -0,0 +1,8 @@
+from ape import accounts, networks
+
+
+def test_transfer():
```
unparalleled-js, line 4: Originally had put this in a script and put it in the projects, but that seemed weird... Because, here, I just want to test that `account.transfer()` works (in an integration-test sort of way), not that `ape run` works.

src/ape/api/address.py
```diff
@@ -24,6 +24,9 @@ def provider(self) -> ProviderAPI:
     def provider(self, value: ProviderAPI):
         self._provider = value
 
+    def _create_transaction(self, **kwargs):
+        return self.provider.network.ecosystem.create_transaction(**kwargs)
+
```
fubuloubu, line 29: This actually seems to make the most sense under `AccountAPI`, since a generic address (or contract account) cannot produce a transaction

tests/integration/test_transfer.py
fubuloubu: Agree

tests/integration/test_transfer.py
```diff
@@ -0,0 +1,8 @@
...
+    with networks.parse_network_choice("::test"):
+        sender = accounts.test_accounts[0]
+        receiver = accounts.test_accounts[1]
+        sender.transfer(receiver, "1 gwei")
```
fubuloubu, line 8: add an `assert` clause showing it works

fubuloubu (approved):

fubuloubu (approved):

Issue #129, opened 2021-08-24
testing: Add `chain` (ChainManager) top-level object


fubuloubu:
### Overview
To be able access chain information (using the `network.active_provider`), there should be a top-level API that provides access to convenience functions, such as transaction history, chain manipulation, and other advanced, non-account-related features.

Initial ticket will be limited in scope and should include wrapping snapshots. Additional tickets will add #344 mining control and #345 querying blockchain information: fee info, transaction info, timestamp manipulation.

### Specification
```python
class ChainManager:
    # Container of all confirmed transactions that have been made
    history: Mapping[AddressType, TxnHistoryContainer]  # get a transaction history for an account (requires heavy processing)
    # Container of all blocks (wrapper around RPC calls)
    blocks: BlockContainer  # Head at `len(chain.blocks)`
    def new_blocks(self) -> Iterator[Block]: ... # Awaits new blocks and allows handling them
    # Timestamp getter/setter
    pending_timestamp: Timestamp  # UNIX timestamp of "pending block" (defaults to computer time)
    pending_timestamp.setter  # Temporarily override timestamp of "pending block" (resets to default after mining)
    # Chain info
    chain_id: int
    def get_transaction(self, txn_hash: TransactionId): ...
    gas_price: int
    base_fee: int
    miner_tip: int
    # Chain Manipulation
    def set_storage(self, address: AddressType, slot: bytes32, value: bytes32): ...
    def set_code(self, address: AddressType, code: Bytecode): ...
    # Mining management
    def mine(self, num_blocks: int=1): ...
    def enable_automining(self): ...
    def disable_automining(self): ...
    # Snapshot management
    def snapshot(self) -> SnapshotId: ...
    def restore(self, sid: SnapshotId=last_snapshot): ...
    def reset(self): ...
    # Control transaction ordering
    def undo(self) -> Transaction: ...
    def redo(self, txn: Transction=last_undone): ...
```

### Dependencies
No dependencies

PR #279, opened 2021-12-01
feat: block API

unparalleled-js:
### What I did

Add a `BlockAPI` abstract class, `Block` class in `ape-ethereum`, and an abstract method `get_block(block_id: str)` on the `ProviderAPI` and implemented in `Web3ProviderAPI` where it returns the block using the `decode()` method from the ecosystem's `.block_class`.

**This is needed in the work related to the Receipt API so that it can use block numbers to track txn confirmations.**

### How I did it

Followed the existing design pattern of `TransacitonAPI` and `ReceiptAPI`.

### How to verify it

See the parametrized test I added. It passes in all the variants of the block ID.

### Checklist

### DIFF
--- a/src/ape/_compat.py
+++ b/src/ape/_compat.py
@@ -1,7 +1,0 @@
-import sys
-
-# We can remove this once we stop supporting python3.7.
-if sys.version_info >= (3, 8):
-    from typing import Literal
-else:
-    from typing_extensions import Literal
--- a/src/ape/_compat.py
+++ b/src/ape/_compat.py
@@ -2,12 +2,6 @@ import sys
 
 # We can remove this once we stop supporting python3.7.
 if sys.version_info >= (3, 8):
-    from functools import cached_property  # type: ignore
-    from functools import singledispatchmethod  # type: ignore
-    from importlib import metadata  # type: ignore
     from typing import Literal
 else:
-    import importlib_metadata as metadata  # type: ignore
-    from backports.cached_property import cached_property  # type: ignore
-    from singledispatchmethod import singledispatchmethod  # type: ignore
     from typing_extensions import Literal
--- a/src/ape/_compat.py
+++ b/src/ape/_compat.py
@@ -1,0 +1,13 @@
+import sys
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from functools import cached_property  # type: ignore
+    from functools import singledispatchmethod  # type: ignore
+    from importlib import metadata  # type: ignore
+    from typing import Literal
+else:
+    import importlib_metadata as metadata  # type: ignore
+    from backports.cached_property import cached_property  # type: ignore
+    from singledispatchmethod import singledispatchmethod  # type: ignore
+    from typing_extensions import Literal
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -150,7 +150,7 @@ class BlockGasAPI:
 
     @classmethod
     @abstractmethod
-    def decode(cls, data: Dict):
+    def decode(cls, data: Dict) -> "BlockGasAPI":
         ...
 
 
@@ -161,7 +161,7 @@ class BlockConsensusAPI:
 
     @classmethod
     @abstractmethod
-    def decode(cls, data: Dict):
+    def decode(cls, data: Dict) -> "BlockConsensusAPI":
         ...
 
 
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -1,3 +1,4 @@
+import sys
 from enum import Enum, IntEnum
 from pathlib import Path
 from typing import Dict, Iterator, List, Optional, Union
@@ -8,7 +9,6 @@ from eth_utils import add_0x_prefix
 from hexbytes import HexBytes
 from web3 import Web3
 
-from ape._compat import Literal
 from ape.exceptions import ProviderError
 from ape.logging import logger
 from ape.types import TransactionSignature
@@ -16,6 +16,12 @@ from ape.types import TransactionSignature
 from . import networks
 from .base import abstractdataclass, abstractmethod
 from .config import ConfigItem
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 
 class TransactionType(Enum):
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -139,9 +139,8 @@ class ReceiptAPI:
 @abstractdataclass
 class BlockGasAPI:
     @property
-    @abstractmethod
-    def base_fee(self) -> Optional[int]:
-        ...
+    def base_fee(self) -> int:
+        raise NotImplementedError("base_fee is not implemented by this block.")
 
     @property
     @abstractmethod
@@ -309,7 +308,7 @@ class Web3Provider(ProviderAPI):
     @property
     def base_fee(self) -> int:
         block = self.get_block("latest")
-        return block.gas_data.base_fee or 0
+        return block.gas_data.base_fee
 
     def get_block(
         self, block_id: Union[str, int, HexBytes, Literal["latest"], Literal["pending"]]
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -168,6 +168,7 @@ class BlockConsensusAPI:
 @abstractdataclass
 class BlockAPI:
     gas_data: BlockGasAPI
+    consensus_data: BlockConsensusAPI
     hash: HexBytes
     number: int
     parent_hash: HexBytes
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -153,7 +153,7 @@ class BlockGasAPI:
 
 @abstractdataclass
 class BlockAPI:
-    gas_fee_data: BlockGasAPI
+    gas_data: BlockGasAPI
     hash: HexBytes
     number: int
     parent_hash: HexBytes
@@ -311,7 +311,7 @@ class Web3Provider(ProviderAPI):
     @property
     def base_fee(self) -> int:
         block = self.get_block("latest")
-        return block.gas_fee_data.base_fee or 0
+        return block.gas_data.base_fee or 0
 
     def get_block(
         self, block_id: Union[str, int, HexBytes, Literal["latest"], Literal["pending"]]
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -149,6 +149,11 @@ class BlockGasAPI:
 
 
 @abstractdataclass
+class BlockConsensusAPI:
+    ...
+
+
+@abstractdataclass
 class BlockAPI:
     gas_data: BlockGasAPI
     hash: HexBytes
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -144,13 +144,12 @@ class ReceiptAPI:
 
 @abstractdataclass
 class BlockGasAPI:
-    @property
-    def base_fee(self) -> int:
-        raise NotImplementedError("base_fee is not implemented by this block.")
+    gas_limit: Optional[int] = None
+    gas_used: Optional[int] = None
+    base_fee: Optional[int] = None
 
-    @property
     @abstractmethod
-    def total_gas_used(self) -> int:
+    def decode(self, data: Dict):
         ...
 
 
@@ -318,8 +317,14 @@ class Web3Provider(ProviderAPI):
 
     @property
     def base_fee(self) -> int:
+        """
+        Returns the current base fee from the latest block.
+
+        NOTE: If your chain does not support base_fees (EIP-1559),
+        this method returns 0.
+        """
         block = self.get_block("latest")
-        return block.gas_data.base_fee
+        return block.gas_data.base_fee or 0
 
     def get_block(
         self, block_id: Union[str, int, HexBytes, Literal["latest"], Literal["pending"]]
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -138,8 +138,6 @@ class ReceiptAPI:
 
 @abstractdataclass
 class BlockGasAPI:
-    raw_data: Dict
-
     @property
     @abstractmethod
     def base_fee(self) -> Optional[int]:
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -137,9 +137,23 @@ class ReceiptAPI:
 
 
 @abstractdataclass
+class BlockGasAPI:
+    raw_data: Dict
+
+    @property
+    @abstractmethod
+    def base_fee(self) -> Optional[int]:
+        ...
+
+    @property
+    @abstractmethod
+    def total_gas_used(self) -> int:
+        ...
+
+
+@abstractdataclass
 class BlockAPI:
-    gas_fee_data: Dict
-    consensus_data: Dict
+    gas_fee_data: BlockGasAPI
     hash: HexBytes
     number: int
     parent_hash: HexBytes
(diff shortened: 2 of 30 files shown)


src/ape/api/providers.py
```diff
@@ -133,6 +134,27 @@ def decode(cls, data: dict) -> "ReceiptAPI":
         ...
 
 
+@abstractdataclass
+class BlockAPI:
+    base_fee_per_gas: Optional[int] = None
```
fubuloubu, line 139: Wonder if gas fee info could be abstracted out into `gas_fee_data: dict` or into a computed property, so that non-EIP1559 blocks can produce the same sorts of data you'd get from EIP1559 blocks

src/ape/api/providers.py
```diff
@@ -133,6 +134,27 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
+    difficulty: int
```
fubuloubu, line 140: Same idea, but with `consensus_data: dict` to abstract out difficulty/PoA stuff, miner/coinbase, etc.

src/ape/api/providers.py
fubuloubu: `gas_limit` and `gas_used` could be here as well :thinking:

src/ape/api/providers.py
```diff
@@ -133,6 +134,27 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
+    gas_limit: int
+    gas_used: int
+    hash: HexBytes
+    miner: str
+    nonce: HexBytes
```
fubuloubu, line 145: Maybe a part of consensus data?

src/ape/api/providers.py
```diff
@@ -133,6 +134,27 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
 
 
+@abstractdataclass
+class BlockAPI:
+    base_fee_per_gas: Optional[int] = None
+    difficulty: int
+    gas_limit: int
+    gas_used: int
+    hash: HexBytes
+    miner: str
```
fubuloubu, line 144: Note to self: need to add and use `AddressType` more in codebase

src/ape/api/providers.py
```diff
@@ -192,6 +214,10 @@ def priority_fee(self) -> int:
     def base_fee(self) -> int:
         raise NotImplementedError("base_fee is not implemented by this provider")
 
+    @abstractmethod
+    def get_block(self, block_id: str) -> BlockAPI:
```
fubuloubu, line 218: Add `BlockId = Union[int, hash, Literal["latest"], Literal["pending"]]` as type?

fubuloubu (commented):
Overall looks good, thinking more abstractly about different block types

src/ape/api/providers.py
```diff
@@ -133,6 +135,22 @@ def decode(cls, data: dict) -> "ReceiptAPI":
         ...
 
 
+@abstractdataclass
+class BlockAPI:
+    gas_fee_data: Dict
+    consensus_data: Dict
```
fubuloubu, line 141: I like these, wonder if we could make them more explicit types, and offer relevant data from them e.g. `average_fee`, `total_gas_used`, `average_gas_used` and other relevant factors.

This is actually getting me thinking more about querying data (e.g. a data model for block that allows computing data over many blocks)

fubuloubu:
```py
@abstractdataclass
class BlockGasDataAPI:
    @property
    @abstractmethod
    def average_fee(self) -> int:
        ...
```

and the like

src/ape/_compat.py
```diff
@@ -0,0 +1,13 @@
...
+if sys.version_info >= (3, 8):
+    from functools import cached_property  # type: ignore
+    from functools import singledispatchmethod  # type: ignore
+    from importlib import metadata  # type: ignore
+    from typing import Literal
+else:
+    import importlib_metadata as metadata  # type: ignore
+    from backports.cached_property import cached_property  # type: ignore
+    from singledispatchmethod import singledispatchmethod  # type: ignore
+    from typing_extensions import Literal
```
unparalleled-js, line 13: So `Literal` s not available until python 3.8 so that caused issues.

I don't think we want end-users importing compat methods from our `ape.utils` because they shouldn't need to. Keeping them here makes it easier to track (for deleting this code!) and makes it more protected rather than public.

src/ape/api/providers.py
unparalleled-js: I like the easier querying this would make

src/ape/_compat.py
fubuloubu: Definitely would like to see this as a separate PR

unparalleled-js: No prob! https://github.com/ApeWorX/ape/pull/282

src/ape/api/providers.py
```diff
@@ -136,10 +136,24 @@ def decode(cls, data: dict) -> "ReceiptAPI":
         ...
 
 
+@abstractdataclass
+class BlockGasAPI:
+    raw_data: Dict
```
fubuloubu, line 141: still really love Pydantic for here, because you can do `__root__: dict` to have it work like a dict with extra methods and validatorss

src/ape/api/providers.py
```diff
@@ -136,10 +136,24 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
+
+    @property
+    @abstractmethod
+    def base_fee(self) -> Optional[int]:
```
fubuloubu, line 145: I think `base_fee` might be lower level, my idea was thinking about other properties that might be a bit higher level and still very relevant

Although specifically, I think knowing `base_fee` would be very useful assuming EIP1559 is active and we want to be able to price gas costs for a new transaction

src/ape/api/providers.py
```diff
@@ -136,10 +136,24 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
+    @property
+    @abstractmethod
+    def total_gas_used(self) -> int:
+        ...
+
+
 @abstractdataclass
 class BlockAPI:
-    gas_fee_data: Dict
-    consensus_data: Dict
```
fubuloubu: deleted `consensus_data`

I definitely don't see using this data a whole lot, except for some rare queries made against the protocol itself

src/ape/api/providers.py
```diff
@@ -136,10 +136,24 @@ def decode(cls, data: dict) -> "ReceiptAPI":
...
+    gas_fee_data: BlockGasAPI
```
fubuloubu, line 156: seems like this should just be `gas_data` come to think of it

or maybe just `gas` (`block.gas.total_gas_used` :thinking:)

src/ape_ethereum/ecosystem.py
```diff
@@ -159,22 +160,24 @@ def decode(cls, data: dict) -> ReceiptAPI:
...
-            "gas_limit": data.get("gasLimit"),
-            "gas_used": data.get("gasUsed"),
-        }
-        consensus_data = {
-            "miner": AddressType(data["miner"]) if "miner" in data else None,
-            "difficulty": data.get("difficulty"),
-            "total_difficulty": data.get("totalDifficulty"),
-        }
+        gas_fee_data = BlockGasFee(  # type: ignore
+            raw_data={k: v for k, v in data.items() if k in ["baseFeePerGas", "gasUsed"]}
```
fubuloubu, line 177: even for EIP1559 chains, I think pre-EIP1559 fork blocks need to handle scenarios where this data might not exist :thinking:

src/ape/api/providers.py
unparalleled-js: Yeah I figured we could re-add it later when we had a better idea of what the API should be? I just am unsure about that one, as far as making higher level goes.

src/ape_ethereum/ecosystem.py
unparalleled-js: What needs to happen when it does not exist?

src/ape/api/providers.py
unparalleled-js: Yeah we are actually calling this somewhere. Should we make subclasses or something?

src/ape/api/providers.py
unparalleled-js: done!

src/ape/api/providers.py
unparalleled-js: This is half the reason I need the `BlockAPI` to exist right now! So to be able to access this easily would be nice but I will settle for using dict-key access if you prefer.

unparalleled-js: Lol I don't know what to do here!

I think we should do something that is easy to iterate on, whatever is, and add stuff on a per-needs basis.

I will think on it.

unparalleled-js: We have `base_fee` on the `ProviderAPI` too, but it just raised `NotImplemented` by default. This is the same idea kind of except it returns `None`, which is a little nicer. 

Maybe raising `NotImplented` would be better though?

src/ape/api/providers.py
```diff
@@ -133,6 +136,33 @@ def decode(cls, data: dict) -> "ReceiptAPI":
         ...
 
 
+@abstractdataclass
+class BlockGasAPI:
+    @property
+    def base_fee(self) -> int:
```
unparalleled-js, line 142: Now, it will always exist but just raise `NotImpementedError` by default, just like the `ProviderAPI`.

This way, the real implementation does not have to have a return type of `Optional`, so that is a win.

src/ape_ethereum/ecosystem.py
```diff
@@ -158,12 +160,45 @@ def decode(cls, data: dict) -> ReceiptAPI:
         )
 
 
+class BlockGasFee(BlockGasAPI):
```
unparalleled-js, line 163: People can always iterate on this class as they see fit, I'd love to see some better block querying properties.

src/ape/api/providers.py
fubuloubu: I think returning `None` is fine, it is a little awkward way to abstract, because old style only has one parameter, and new style has two. Let's just do our best here and see how it feels?

src/ape/api/providers.py
fubuloubu: just `block.consensus["difficulty"]` maybe

src/ape_ethereum/ecosystem.py
fubuloubu: I think they make the blocks pre-EIP1559 compatible in some way, but I'm not sure how to handle it or if the way they are handling it makes sense for our use case. Can probably see this in web3py where they made fixes for EIP1559

src/ape/api/providers.py
fubuloubu: This works too

src/ape_ethereum/ecosystem.py
fubuloubu: Yeah, this needs more thought, you'd probably want to do some sort of query like:

```py
avg_base_fee = sum(chain.query_blocks(start=START_BLOCK, end=END_BLOCK).block.gas.base_fee) / (END_BLOCK - START_BLOCK)
```

fubuloubu: Or probably more pythonically

```py
avg_base_fee = sum(
    blk.gas.base_fee for blk in chain.query_blocks(start=START_BLOCK, end=END_BLOCK)
) / (END_BLOCK - START_BLOCK)
```

fubuloubu: This gets into https://github.com/ApeWorX/ape/issues/129

src/ape/api/providers.py
unparalleled-js: I added the class for it so we can add to it and made sure difficulty was in there from `ape_ethereum` (provided it exists, its optional).

src/ape/api/providers.py
unparalleled-js: 99.9% of the time, people will only be calling this on the latest block while crafting a dynamic transaction

src/ape_ethereum/ecosystem.py
unparalleled-js: In this querying use-case, they need to make sure the `START_BLOCK` post London fork. I think an exception getting raised when trying to query a base fee pre London makes sense. I don't think there is any concept of a base fee pre for block pre-London.

fubuloubu: We can look into handling that when it becomes a problem

src/ape/_compat.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
fubuloubu: I don't think `typing_extensions` is in `setup.py`, although it is a dependency of one of our dependencies. I typically like to handle this in `install_requires` with `; < python_version<'3.8'` modifier

fubuloubu (approved):

src/ape/_compat.py
unparalleled-js: Added! I am scared to pin a version lol... Because of what is happening in all the ethereum packages since 4.0.0 came out.

src/ape/api/providers.py
```diff
@@ -14,6 +17,12 @@
 from .base import abstractdataclass, abstractmethod
 from .config import ConfigItem
 
+# We can remove this once we stop supporting python3.7.
```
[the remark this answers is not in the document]
fubuloubu: I wasn't opposed to the file, just needed the install pinned in there

src/ape/_compat.py
fubuloubu: Yeah, always an issue with their libraries, ugh

src/ape/api/providers.py
unparalleled-js: Yeah maybe returning `None` for everything is the best solution...

1. It is relaxed, like `dict`, in that those keys don't have to exist.
2. Any new keys that end up be needed can always be added and kept optional. Chains that to use them know should be there.

src/ape_ethereum/ecosystem.py
unparalleled-js: As far as I can tell, the `baseFeePerGas` just is not present on pre-London blocks. I think just returning `None` for all gas-related block properties is the safest way. `ProviderAPIs` that need to access this data will know whether it should be there or not, or handle it however if it is not.

Maybe `gas_used` does not need to be optional though. Are there any props we can always count on being part of the `BlockGasAPI` ?

src/ape_ethereum/ecosystem.py
unparalleled-js: Yeah I think returning `None` is better again. The key is not there on the data pre London forks, so it is like the same thing, just safer: no `KeyErrors`.

src/ape_ethereum/ecosystem.py
fubuloubu: `None` seems like the path forward. `gas_used` should always be there
other things like average and total too

src/ape/api/providers.py
```diff
@@ -271,8 +325,39 @@ def priority_fee(self) -> int:
...
-        block = self._web3.eth.get_block("latest")
-        return block.baseFeePerGas  # type: ignore
+        """
+        Returns the current base fee from the latest block.
+
+        NOTE: If your chain does not support base_fees (EIP-1559),
+        this method returns 0.
+        """
+        block = self.get_block("latest")
+        return block.gas_data.base_fee or 0
```
fubuloubu, line 335: should this maybe throw `NotImplementedError` if EIP1559 isn't implemented?

src/ape/api/providers.py
```diff
@@ -271,8 +325,39 @@ def priority_fee(self) -> int:
...
+
+    def get_block(
+        self, block_id: Union[str, int, HexBytes, Literal["latest"], Literal["pending"]]
```
fubuloubu, line 338: was thinking to add this union to `ape.types` as `BlockId = Union[str, int, HexBytes, Literal["latest"], Literal["pending"]`

src/ape/api/providers.py
```diff
@@ -271,8 +316,42 @@ def priority_fee(self) -> int:
...
+        return block.gas_data.base_fee
+
+    def get_block(self, block_id: BlockID) -> BlockAPI:
+        """
+        Returns a block for the given ID.
+
+        Args:
+            block_id: The ID of the block to get. Set as
+              "latest" to get the latest block,
+              "earliest" to get the earliest block,
```
fubuloubu, line 340: you note "earliest" here, but it's not covered in `BlockId`

Also, I can just type 0

fubuloubu (approved):

fubuloubu (approved):

PR #282, opened 2021-12-02
refactor: use _compat.py module

unparalleled-js:
### What I did

Create compat module to prevent people from using our compat methods from `ape.utils`. Also, easier to track and delete the code when necessary. Keeps it protected and isolated basically and makes more sense when importing, such as `Literal` from the `typing` module.

### How I did it

Move all compat methods to new module `ape._compat.py`

### Checklist

### DIFF
--- a/src/ape/_compat.py
+++ b/src/ape/_compat.py
@@ -1,0 +1,13 @@
+import sys
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from functools import cached_property  # type: ignore
+    from functools import singledispatchmethod  # type: ignore
+    from importlib import metadata  # type: ignore
+    from typing import Literal
+else:
+    import importlib_metadata as metadata  # type: ignore
+    from backports.cached_property import cached_property  # type: ignore
+    from singledispatchmethod import singledispatchmethod  # type: ignore
+    from typing_extensions import Literal
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -15,16 +15,6 @@ from importlib_metadata import PackageNotFoundError, packages_distributions, ver
 
 from ape.logging import logger
 
-try:
-    from functools import cached_property  # type: ignore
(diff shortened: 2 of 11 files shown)


src/ape/_compat.py
```diff
@@ -0,0 +1,13 @@
+import sys
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from functools import cached_property  # type: ignore
+    from functools import singledispatchmethod  # type: ignore
```
unparalleled-js, line 6: This does break vyper and solidity but I can make sure to release hot-fixes for those.

I think each plugin should manage their own compatibility imports individually to be safe.. Otherwise, they could import from the protect `_compat.py` module and that is okay I think since they are 2nd class plugins

src/ape/utils.py
```diff
@@ -195,7 +185,6 @@ def extract_nested_value(root: Mapping, *args: str) -> Optional[Dict]:
 
 
 __all__ = [
-    "cached_property",
```
fubuloubu, line 198: So I did sort of intend these to be exported for use by external libraries (2nd/3rd class plugins) because I generally thought them to be useful. Also this way in case we needed to patch any of their functionality to work with how ape's API classes work, I wanted plugins to source these utilities from us to make that patching easier.

Overall unsure of the pros/cons of this approach.

unparalleled-js: Ah, I figured we were only doing that because of python 3.7.

It's like, if python 3.7 goes away, I'd like to just delete this code without worrying about making breaking changes, that is the main benefit. But if there is a chance of making custom implementations for these methods, than it does makes sense to keep them in `ape.utils.py`

fubuloubu: Yeah, so like I wouldn't modify them too much, but let's say there was a bug in the implementation that caused us problems, I'd want it to be easier to fix by modifying here instead of all downstream users that need the same fix

Issue #149, opened 2021-09-17
Specify Dependencies


fubuloubu:
### Overview
Allow users to specify dependencies via `ape-config.yaml`

### Specification
```yaml
dependencies:
    dep1_name: https://github.com/org/dep1_name.git 
    dep2_name: ipfs://bafybeic5cdjintzeuegosx2y3nc27p3t7rfetj4pa5lprcylbfwytdllia
    ... # other types
```

unparalleled-js:
As part of this issue, we can refactor `ape-compile` to pull out the logic that is shareable from the `_cli` module and move it to the `Package` class.

Referencing: https://github.com/ApeWorX/ape/pull/167#pullrequestreview-764874264

fubuloubu:
Note: allow compiler plugin to do remappings and enable external imports

https://github.com/yearn/brownie-strategy-mix/blob/681320b3c0a8c90fe89f3c2dfce1b9824fa0e113/brownie-config.yml#L19

unparalleled-js:
- [ ] Once it makes sense, create issues in the compiler plugin repos to have configs for remappings

PR #284, opened 2021-12-03
feat: support Github dependencies

unparalleled-js:
### What I did

fixes: #149 

### How I did it

When deps specified in Github format, download them directory in userspace and allow deps to be referneced from a project.

### How to verify it

Add this to your `ape-config.yaml`:

```yaml
dependencies:
  open_zeppelin: OpenZeppelin/openzeppelin-contracts@4.4.0
```

When the package does not exist in `.ape/packages/open_zeppelin`, it will grab it from Github. Note that `open_zeppelin` is a user-defined key in this example: it can be anything, such as `open_zeppelin_0.4.0` if you were trying to support multiple versions.

### Checklist


src/ape/managers/project.py
```diff
@@ -26,17 +27,31 @@ def __post_init__(self):
...
+                raise ProjectError("Dependencies must have a name.")
 
-        return PackageManifest.from_dict(manifest_dict)
+            return PackageManifest.from_dict(manifest_dict)
+        else:
+            # Github dependency
+            try:
+                path, version = dependency_id.split("@")
+            except ValueError:
+                raise ValueError("Invalid Github ID. Must be given as <org>/<repo>@<version>")
```
fubuloubu, line 45: It would be sweet to be able to snag branches as well, although that might not be the most repeatable build process

src/ape/managers/project.py
```diff
@@ -26,17 +27,31 @@ def __post_init__(self):
...
+
+            target_path = self._cache_folder / "contracts" / dependency_name
```
fubuloubu, line 47: Don't need "contracts" in cache folder, it is only for build files

Also was wanting to use a cache folder in data folder to store the raw files, and to compile them into a manifest file that would be referenced as a single JSON file

fubuloubu (commented):
Looking great!

src/ape/managers/project.py
unparalleled-js: I think you can have tags on other branches? The version just matches to a tag I think. I am not sure if that works or not but it might.

src/ape/utils.py
```diff
@@ -222,6 +243,39 @@ def available_plugins(self) -> Set[str]:
             if repo.name.startswith("ape-")
         }
 
+    def download_package(self, repo_path: str, version: str, target_path: Path):
```
unparalleled-js, line 246: Kinda neat - this works if you don't have the Github auth token, since it's public repos

fubuloubu: Wait, we probably don't need the Auth token for detected 2nd class either?

unparalleled-js: Wow you are right! We are totally erroring out early for no reason

unparalleled-js: https://github.com/ApeWorX/ape/pull/287/files

src/ape/utils.py
```diff
@@ -222,20 +269,77 @@ def available_plugins(self) -> Set[str]:
...
+        dir_b: file_c
+
+      and you provide the path to `dir_a`, it will return a list containing
+      the Paths to `file_a`, `file_b` and `file_c`.
+    """
+    if not directory.is_dir():
+        # Was given a file, return a list of just that file.
+        return [directory]
+
+    all_files = []
```
unparalleled-js, line 320: Change to `rglob("*.*")`

evanblank3 (approved):
Thanks for the walkthrough!

PR #287, opened 2021-12-03
fix: remove unneeded github auth check when displaying available plugins

unparalleled-js:
### What I did

Removed uneeded auth token check. We can get this info anon because it's public

Personal Access Tokens are only required when accessing private github information. Public repos are not private, so it makes sense that it is not needed here (https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)

### How I did it

Realized from other work and comments from @fubuloubu that it is not needed.

### How to verify it

Make sure you don't have the token set and run 

```bash
ape plugins list --all
```

You should see the available plugins

### Checklist


tests/integration/cli/test_misc.py
```diff
@@ -13,10 +15,23 @@
...
+    assert result.exit_code == 0
+
+
+# Only run these tests once (limited to single arbitrary project).
+# This is to prevent Github from rate limiting us and causing test failures.
+@skip_projects_except(["empty-config"])
+@pytest.mark.parametrize(
+    "args",
+    (
         ["plugins", "list"],
```
unparalleled-js, line 31: If these give us troubles down the line, we should perhaps add a `GITHUB_ACCESS_TOKEN` env var to the repo secrets or something.

evanblank3 (approved):
Thanks for the walkthrough!

Issue #290, opened 2021-12-06
Initializing a contract from an address


unparalleled-js:
### Overview

Make it better UX to implement a contract from address from its type.

### Specification

The follow example uses `ape console --network ::hardhat`

___

This is what it looks like to deploy a contract. This looks great! We want something like this for initializing a contract from address:

(here, assume I have a contract named `Fund.sol`)
```python
account = accounts.test_accounts[0]
contract = account.deploy(project.Fund)  # Deploying a contract looks nice! This returns a contract we can make calls on.
```

outputs:

```
INFO: Submitted 0xa63f2b37ae85774a40c365b6a2988721101513f52c3cd313458228c82d2809f2 (gas_used=527861)
SUCCESS: Contract 'Fund' deployed to: 0x274b028b03A250cA03644E6c578D81f019eE1323
```

The contract is now deployed to **0x274b028b03A250cA03644E6c578D81f019eE1323**. Great!
Let's say we come back tomorrow with a new `console` session, we want to be able to initialize the contract from that address `0x274b028b03A250cA03644E6c578D81f019eE1323` without having to re-deploy it.

Currently, this is how it works (this is bad! This is what we want to change and is the reason for this ticket):

```python
from ape.api.contracts import ContractInstance

contract = ContractInstance(
    _address="0x274b028b03A250cA03644E6c578D81f019eE1323",
    _contract_type=project.Fund
)
```

Notice how we have to import some weird class `ContractInstance` and use a protected kwarg initialization....
But, the `deploy` feature, we can literally use the name of the contract.

so, this is how we _want_ it to look:

```python
contract = project.Fund.at("0x274b028b03A250cA03644E6c578D81f019eE1323")
```

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #291, opened 2021-12-07
feat: ability to initialize contract objects with an address

evanblank3:
### What I did

fixes: #290 

Update the process for retrieving/initializing a  previously deployed contract to be more UX friendly.

### How I did it

Modify  `__getattr__(...)` in `ProjectManager` to return a `ContractContainer` instead of a `ContractType` and refactored `deploy(...)`

### How to verify it

```python
contract = project.Fund("0x274b028b03A250cA03644E6c578D81f019eE1323")
```

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -5,13 +5,7 @@ import click
 
 from ape.exceptions import AccountsError, AliasAlreadyInUseError, SignatureError
 from ape.logging import logger
-from ape.types import (
-    AddressType,
-    ContractType,
-    MessageSignature,
-    SignableMessage,
-    TransactionSignature,
-)
+from ape.types import AddressType, MessageSignature, SignableMessage, TransactionSignature
 from ape.utils import cached_property
 
 from .address import AddressAPI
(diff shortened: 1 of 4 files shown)


src/ape/api/accounts.py
```diff
@@ -137,26 +131,22 @@ def transfer(
...
+        txn = contract(*args, **kwargs)
         txn.sender = self.address
         receipt = self.call(txn)
 
         if not receipt.contract_address:
             raise AccountsError(f"'{receipt.txn_hash}' did not create a contract.")
 
         address = click.style(receipt.contract_address, bold=True)
-        logger.success(f"Contract '{contract_type.contractName}' deployed to: {address}")
+        logger.success(f"Contract '{contract._contract_type.contractName}' deployed to: {address}")
```
unparalleled-js, line 144: Maybe we should make `_contract_type` public on the `ContractContainer` class

src/ape/api/contracts.py
```diff
@@ -234,8 +234,8 @@ def get_handler(abi_type: str) -> Any:
 
 @dataclass
 class ContractContainer:
-    _provider: ProviderAPI
     _contract_type: ContractType
+    _provider: Optional[ProviderAPI]
```
unparalleled-js, line 238: Maybe we should add a comment here saying like `"This is only None the user is not connected to a provider."`

unparalleled-js (approved):

Issue #262, opened 2021-11-24
Addresses returned from contracts are not checksum addresses


unparalleled-js:
### Environment information

latest ape

### What went wrong?

when a contract returns the `address` type in `ape`, it is not a `ChecksumAddress` and so that makes comparison checking awkward in ape.

### How can it be fixed?

If we notice the `internType` is `address`, wrap it in a `ChecksumAddress` class.

PR #294, opened 2021-12-08
feat: return checksummed addresses from contract calls

unparalleled-js:
### What I did

fixes: #262 

### How I did it

call `to_checksum_address()` on the data being return so long that it is of an address type.

### How to verify it

Two things:

1. Verify that addresses returned from contracts are now checksummed
2. Verify that normal calls unrelated to addresses still work.

Test that tests this new feature: https://github.com/unparalleled-js/ape-demo-project/blob/main/tests/test_fund.py#L17-L21, 
The other tests pass as well in that demo project.
Thus, this addresses the issue.

**NOTE: eth-abi python package naturally returns HexAddress as "normalized". It is not much effort to change that, but to fix their test suite against that is actually kind of challenging believe it or not and I am not sure they would want that.. Though I have started the process to see what it's like: https://github.com/ethereum/eth-abi/pull/152**

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from eth_abi import decode_abi as abi_decode
 from eth_abi import encode_abi as abi_encode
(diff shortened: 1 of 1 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -236,10 +236,25 @@ def encode_calldata(self, abi: ABI, *args) -> bytes:
...
-    def decode_calldata(self, abi: ABI, raw_data: bytes) -> Any:
+    def decode_calldata(self, abi: ABI, raw_data: bytes) -> Tuple[Any, ...]:
         output_types = [o.canonical_type for o in abi.outputs]
         try:
-            return abi_decode(output_types, raw_data)
+            vm_return_values = abi_decode(output_types, raw_data)
+            if not vm_return_values:
+                return vm_return_values
+
+            if not isinstance(vm_return_values, (tuple, list)):
```
unparalleled-js, line 246: From what I could tell, the return values here were always a tuple anyway, but I added this part just to not make any assumptions about it

fubuloubu (approved):

PR #295, opened 2021-12-09
docs: document the CLI utilities

unparalleled-js:
### What I did

Add missing doc strs to the `ape.cli` namespace.

### How I did it

Lots of revising.

### How to verify it

Lots of reading (audience for these particular docs would be plugin writers probably).

### Checklist

### DIFF
--- a/src/ape/cli/arguments.py
+++ b/src/ape/cli/arguments.py
@@ -19,10 +19,22 @@ def _require_non_existing_alias(value):
 
 
 def existing_alias_argument(account_type: Optional[Type[AccountAPI]] = None):
+    """
+    A ``click.argument`` for an existing account alias.
+
+    Args:
+        account_type (type[:class:`~ape.api.accounts.AccountAPI`], optional):
+          If given, limits the type of account the user may choose from.
+    """
+
     return click.argument("alias", type=Alias(account_type=account_type))
 
 
 def non_existing_alias_argument():
+    """
+    A ``click.argument`` for an account alias that does not yet exist in ape.
+    """
+
     return click.argument(
         "alias", callback=lambda ctx, param, value: _require_non_existing_alias(value)
     )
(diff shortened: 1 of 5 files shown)


src/ape/cli/arguments.py
```diff
@@ -19,10 +19,22 @@ def _require_non_existing_alias(value):
 
 
 def existing_alias_argument(account_type: Optional[Type[AccountAPI]] = None):
+    """
+    A ``click.argument`` for an existing account alias.
+
+    Args:
+        account_type (type[:class:`~ape.api.accounts.AccountAPI`], optional):
+          If given, limits the type of account the user may choose from.
```
fubuloubu, line 27: What do you think of this?
```suggestion
          If given, limits the account plugin the user may choose accounts from.
```

src/ape/cli/choices.py
```diff
@@ -143,15 +171,27 @@ def get_metavar(self, param):
 
 
 class OutputFormat(Enum):
+    """
+    Various output format types to use for outputting data
+    to the user.
```
fubuloubu, line 176: Not particularly descriptive

fubuloubu (approved):

src/ape/cli/arguments.py
unparalleled-js: I wanted to incorporate plugins in there somehow, because that is the common use-case.  However, a plugin could have multiple account API implementations and this method can actually wall-off to just one of those too if it wanted.

src/ape/cli/choices.py
unparalleled-js: What do you think of 

```
  An enum representing output formats, such as ``TREE`` or ``YAML``.
  Use this to select a subset of common output formats to use
  when creating a :meth:`~ape.cli.choices.output_format_choice`.
```

fubuloubu: Better. Is there a way to decorate specific options in an enum to describe what they mean?

unparalleled-js: Ya we can document individual enums values, I just added a commit that does that.

fubuloubu (approved):

PR #300, opened 2021-12-10
docs: document AddressAPI

unparalleled-js:
### What I did

Add method docs strs to `AddressAPI`

### How I did it

Played a "Guess the Doc Str" game with @Ninjagod1251 

### How to verify it

Read through 'em

### Checklist

### DIFF
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -103,7 +103,7 @@ class AddressAPI:
         The total balance of the account.
 
         Returns:
-            `int`
+            int
         """
 
         return self.provider.get_balance(self.address)
@@ -145,7 +145,9 @@ class AddressAPI:
 
 class Address(AddressAPI):
     """
-    A generic blockchain address. Typically used when we do not know the contract type at a given address, or to refer to an EOA the user doesn't personally control.
+    A generic blockchain address.
+    Typically, this is used when we do not know the contract type at a given address,
+    or to refer to an EOA the user doesn't personally control.
     """
 
     _address: AddressType
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -9,10 +9,23 @@ from .providers import ProviderAPI
 
 @abstractdataclass
 class AddressAPI:
+    """
+    A base address API class.
+    """
+
     _provider: Optional[ProviderAPI] = None
 
     @property
     def provider(self) -> ProviderAPI:
+        """
+        The current active provider if connected to one.
+        If there is no active provider at runtime, then this raises an
+        :class:`~ape.exceptions.AddressError`.
+
+        Returns:
+            :class:`~ape.api.providers.ProviderAPI`
+        """
+
         if not self._provider:
             raise AddressError(
                 f"Incorrectly implemented provider API for class {type(self).__name__}"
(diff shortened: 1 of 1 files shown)


src/ape/api/address.py
```diff
@@ -9,10 +9,23 @@
 
 @abstractdataclass
 class AddressAPI:
+    """
+    A base address API class.
```
fubuloubu, line 13: Add something like "...all account types subclass this type (Externally-Owned Accounts and Contract Accounts)"

src/ape/api/address.py
```diff
@@ -22,15 +35,32 @@ def provider(self) -> ProviderAPI:
...
+        """
+
         self._provider = value
 
     @property
     @abstractmethod
     def address(self) -> AddressType:
-        ...
+        """
+        The raw address type.
```
fubuloubu, line 51: The address of this account
```suggestion
        The address of this account. Subclasses must override and provide this value.
```

src/ape/api/address.py
```diff
@@ -22,15 +35,32 @@ def provider(self) -> ProviderAPI:
...
-        # This displays methods to IPython on `a.[TAB]` tab completion
+        """
+        Display methods to IPython on ``a.[TAB]`` tab completion.
+
+        Returns:
+            list[str]: Method names that IPython uses for tab completion.
+        """
         return [
             "address",
             "balance",
```
fubuloubu, line 66: tbh would like to add better integration with this via the base class, but can only do that once pydantic upgrade happens

src/ape/api/address.py
```diff
@@ -45,33 +75,86 @@ def __repr__(self) -> str:
...
+
         return self.provider.get_nonce(self.address)
 
     @property
     def balance(self) -> int:
+        """
+        The total balance of the account.
+
+        Returns:
+            `int`
```
fubuloubu, line 104: are the ticks on purpose here? the others don't have it

src/ape/api/address.py
```diff
@@ -45,33 +75,86 @@ def __repr__(self) -> str:
...
+        Returns:
+            bool
+        """
+
         return len(self.code) > 0
 
 
 class Address(AddressAPI):
+    """
+    A blockchain address.
```
fubuloubu, line 146:
```suggestion
    A generic blockchain address. Typically used when we do not know the contract type at a given address, or to refer to an EOA the user doesn't personally control.
```

src/ape/api/address.py
unparalleled-js: Should we use the word `account` here though?
The `AddressAPI` can also represent a contract, is that considered an account?

src/ape/api/address.py
unparalleled-js: You mean the feature for IPython or the docs for it?
If the feature, we should not add docs for this.

src/ape/api/address.py
unparalleled-js: Accident!

src/ape/api/address.py
unparalleled-js: Ha that is insightful

src/ape/api/address.py
unparalleled-js: Ah, well based on the comment above, I see that contracts must be commonly also referred to as an accounts

src/ape/api/address.py
unparalleled-js: btw, I have started looking into pydantic work and it is on my list! I have rebased the branch at least.

fubuloubu: > You mean the feature for IPython or the docs for it? If the feature, we should not add docs for this.

oh I mean the feature of integrating with IPython, but looking at methods and fields of the dataclass

src/ape/api/address.py
```diff
@@ -9,10 +9,25 @@
 
 @abstractdataclass
 class AddressAPI:
+    """
+    A base address API class. All account-types subclass
+    this type, including externally owned accounts as well
+    as contract accounts.
```
fubuloubu, line 15: actually there isn't any other type of account than these two lol :sweat:

fubuloubu (approved):

src/ape/api/address.py
unparalleled-js: Okay cool. Well if we delete this method / approach, that is no big deal.

src/ape/api/address.py
unparalleled-js: ha so maybe like `which includes both ...`

unparalleled-js: I don't think this do necessarily suggests there are other potential types

unparalleled-js: Maybe we just don't need the `including` part ?

fubuloubu: yeah, sounds good

fubuloubu (approved):

Issue #310, opened 2021-12-14
Encode AddressAPI as address string for ABI


evanblank3:
### Overview

Currently you have to explicitly send an address (str) for address types in the ABI. We want to be able to pass an `AddressAPI` object, similar to the logic of `sender=<AccountAPI>` currently in place.

Old behavior:
```python
contract.mint(receiver.address, 0, sender=owner)
```

New behavior:
```python
contract.mint(receiver, 0, sender=owner)
```

### Specification

Create a private method to encode `AddressAPI` as an underlying address/str type, with hooks in the contract.

PR #311, opened 2021-12-14
feat: add string address encoding

evanblank3:
### What I did

fixes: #310
Update logic to encode `AddressAPI` objects as their underlying address value [str]

### How I did it

Private helper method called in several `Contract encode` methods

### How to verify it

Old behavior:
```python
contract.mint(receiver.address, 0, sender=owner)
```

New behavior:
```python
contract.mint(receiver, 0, sender=owner)
```

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,6 @@ version.py
 
 # Ape stuff
 .build/
+
+# VS Code
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -119,6 +119,3 @@ version.py
 
 # Ape stuff
 .build/
-
-# VS Code
-.vscode
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -13,6 +13,23 @@ if TYPE_CHECKING:
     from ape.managers.networks import NetworkManager
 
 
+def _encode_address_args(*args):
+    """
+    Helper for encoding Ape AddressAPI to underlying address for ABI use
+    """
+    return [arg.address if isinstance(arg, AddressAPI) else arg for arg in args]
+
+
+def _encode_address_kwargs(**kwargs):
+    """
+    Helper for encoding Ape AddressAPI to underlying address for ABI use
+    """
+    return {
+        key: value.address if isinstance(value, AddressAPI) else value
+        for key, value in kwargs.items()
+    }
+
+
 @dataclass
 class ContractConstructor:
     deployment_bytecode: bytes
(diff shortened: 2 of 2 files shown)


evanblank3:
Not addressed in this PR, but a question I had working in this area, should sender be allowed to pass a str for the address? 
i.e. `contract.mint(receiver, 0, sender=owner.address)' 

That would align the behavior with `receiver` in the example?

If it should not, we will need to do some type checking around the value passed in for `sender`.
It currently breaks when attempting to pass `sender-owner.address`

.gitignore
```diff
@@ -119,3 +119,6 @@ version.py
 
 # Ape stuff
 .build/
+
+# VS Code
```
unparalleled-js, line 123: I was told that IDE specific entries should be put in our global `.gitignore` files

src/ape/contracts/base.py
```diff
@@ -13,6 +13,23 @@
     from ape.managers.networks import NetworkManager
 
 
+def _encode_address_args(*args):
+    """
+    Helper for encoding Ape AddressAPI to underlying address for ABI use
```
unparalleled-js, line 18: Maybe we can make this more generic, like `_encode_input_args()`, just in case we have more of these type-mappings and stuff?

And `_encode_input_kwargs()`

Also, I think these doc strs a little distracting, maybe just comment like `# Convert higher level address types to 'str'`

evanblank3: I went back and forth about that, but thought that if this was a one-off maybe it should be specific? If we wanted to do a more general solution it might be better to see what other issues need to be addressed and make a design change as well?

.gitignore
evanblank3: oops

fubuloubu:
> Not addressed in this PR, but a question I had working in this area, should sender be allowed to pass a str for the address? i.e. `contract.mint(receiver, 0, sender=owner.address)`
> 
> That would align the behavior with `receiver` in the example?
> 
> If it should not, we will need to do some type checking around the value passed in for `sender`. It currently breaks when attempting to pass `sender-owner.address`

the `sender` has to sign, so I don't think this should be changed

src/ape/contracts/base.py
```diff
@@ -13,6 +13,19 @@
     from ape.managers.networks import NetworkManager
 
 
+def _encode_address_args(*args):
+    # Convert higher level address types to str
+    return [arg.address if isinstance(arg, AddressAPI) else arg for arg in args]
+
+
+def _encode_address_kwargs(**kwargs):
```
fubuloubu, line 21: What kwargs need to be encoded? Assuming we disallow `sender` from being a simple address

evanblank3: This was to capture the previous behavior, i.e.:
```python
# ContractConstructor.__call__
kwargs["sender"] = sender.address
...
# ContractCall.__call__
# ContractTransaction.__call__
txn.sender = kwargs["sender"].address
```

fubuloubu: Might have to rethink this later, but this is fine for now

fubuloubu (approved):

unparalleled-js (approved):
lgtm

Issue #312, opened 2021-12-14
Unable to deploy contracts


unparalleled-js:
### Environment information

Latest ape

### What went wrong?

```
    contract = deploy(sender=owner)
  File "/Users/jules/ApeProjects/ape-demo-project/scripts/utils.py", line 11, in deploy
    return account.deploy(project.Fund, *args, **kwargs)
  File "/Users/jules/PycharmProjects/ape/src/ape/api/accounts.py", line 150, in deploy
    return ContractInstance(  # type: ignore
NameError: name 'ContractInstance' is not defined
```

https://github.com/ApeWorX/ape/blob/main/src/ape/api/accounts.py#L150

^ Looks like when we changed the `contract` namespace, we broke this part

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

PR #313, opened 2021-12-14
fix: resolve import error preventing deploying contracts from working

unparalleled-js:
### What I did

fixes: #312

### How I did it

Do a local import for now. There is a cyclic dependency because between accounts and contracts.

### How to verify it

Example:

```python
account.deploy(project.Fund)
```

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -147,6 +147,8 @@ class AccountAPI(AddressAPI):
         address = click.style(receipt.contract_address, bold=True)
         logger.success(f"Contract '{contract.contract_type.contractName}' deployed to: {address}")
 
+        from ape.contracts import ContractInstance
+
         return ContractInstance(  # type: ignore
             _provider=self.provider,
             _address=receipt.contract_address,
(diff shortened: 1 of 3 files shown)


src/ape/api/accounts.py
```diff
@@ -147,6 +147,8 @@ def deploy(self, contract: "ContractContainer", *args, **kwargs) -> "ContractIns
         address = click.style(receipt.contract_address, bold=True)
         logger.success(f"Contract '{contract.contract_type.contractName}' deployed to: {address}")
 
+        from ape.contracts import ContractInstance
```
fubuloubu, line 150: Can this be put at the top? or does that create a cycle?

fubuloubu (approved):

src/ape/api/accounts.py
unparalleled-js: That creates the cycle... I tried looking into other ways to fix, but they just need each other because a contract _is_ an address type and the `deploy` method returns stuff related to contracts. It is very cyclic

evanblank3 (approved):

Issue #261, opened 2021-11-24
Equality Handling between AccountType and AccountAPI (and str)


unparalleled-js:
### Overview

When making assertions in ape test, it would be nice to be able to assert an account is an address regardless of whether it is an `AddressAPI` subclass or an address literal (or convertible to an address type). This will make writing tests easier and read more intuitively.

### Specification

```python
# NOTE: `contract.owner()` returns an address like "0xAbCd...1234"
assert contract.owner() == owner  # is `AccountAPI("0xAbCd...1234")`
assert contract.owner() == "0xAbCd...1234"
assert contract.owner() == "my-name.eth"
```

PR #317, opened 2021-12-14
feat: compare accounts and address str

evanblank3:
### What I did

fixes: #261

Allow address string comparison with accounts.

### How I did it

Override `__eq__` for `AddressAPI` to allow comparison with address str.

### How to verify it

Before:
```python
contract = owner.deploy(project.ticket)
receipt = contract.mint(receiver, 0, sender=owner)

assert contract.ownerOf(0) == receiver.address
```
After:
```python
assert contract.ownerOf(0) == receiver == '0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7'
```

### Checklist


evanblank3:
I didn't include work for comparing ENS strings with AddressAPI but can either open a second issue or wait to merge and add it to this PR. Thoughts?

src/ape/api/address.py
```diff
@@ -54,6 +54,18 @@ def address(self) -> AddressType:
             :class:`~ape.types.AddressType`
         """
 
+    def __eq__(self, __o: object) -> bool:
+        """
+        Compares AddressAPI/str objs
```
unparalleled-js, line 59:
```suggestion
        Compares :class:`~ape.api.AddressAPI`/``str`` objects.
```

unparalleled-js: I think this needs to be expanded also to say that it does a string comparison on the address or something.

src/ape/api/address.py
```diff
@@ -72,6 +84,13 @@ def __dir__(self) -> List[str]:
         ]
 
     def __repr__(self) -> str:
+        """
```
unparalleled-js, line 87: I don't think we need to document `repr`, or at least I haven't been

src/ape/api/address.py
```diff
@@ -54,6 +54,18 @@ def address(self) -> AddressType:
...
+    def __eq__(self, __o: object) -> bool:
+        """
+        Compares AddressAPI/str objs
+
+        Returns:
+            bool: comparison result
+        """
+
+        if isinstance(__o, str):
+            return self.address == __o
```
unparalleled-js, line 66:
```suggestion
            return super().__eq__(self.address, __o)
```

Would this make sense?

src/ape/api/address.py
```diff
@@ -54,6 +54,18 @@ def address(self) -> AddressType:
...
+        return super.__eq__(self, __o)
```
unparalleled-js, line 67:
```suggestion
        return super().__eq__(self, __o)
```

I thought we had to call `super`, no?

src/ape/api/address.py
evanblank3: Because we're overriding it by comparing it to an item that isn't the same type we can't use the super method, it results in this:
```python
TypeError: super(type, obj): obj must be an instance or subtype of type
```

src/ape/api/address.py
evanblank3: No prob, I'll pull it out.

src/ape/api/address.py
evanblank3: I'll put it back to a more conventional implementation but these behave the same (I think/doesn't break, lol):
```python
super().__eq__(__o)

super.__eq__(self, __o)
```

unparalleled-js: ^ the syntax highlighting is suspicious

fubuloubu: question, why not just `return convert(self, AddressType) == convert(__o, AddressType)`?

fubuloubu: also, `other` would be better than `__o`

fubuloubu:
> I didn't include work for comparing ENS strings with AddressAPI but can either open a second issue or wait to merge and add it to this PR. Thoughts?

2nd PR

fubuloubu:
Not sure where we were talking about this, but the conversion of `"1 gwei"` to `int` would be valuable for things like specifying gas prices, as well as comparisions to account balances e.g. `assert contract.balance == "1 ether"`

src/ape/api/address.py
evanblank3: Much cleaner, I originally used the `__o` from vscode code stub auto-generation for the override.
03075bf

fubuloubu (approved):

PR #320, opened 2021-12-15
docs: add documentation to the 'ape.managers' classes

unparalleled-js:
### What I did

Doc strs for the managers!

### How I did it

Lots of revising.

### How to verify it

build the docs!

You should see stuff like this.

Review the words of the docs.

Screen Shot 2021-12-15 at 16 14 29

Screen Shot 2021-12-15 at 16 14 36

### Checklist


src/ape/managers/accounts.py
```diff
@@ -56,6 +95,13 @@ def get_accounts_by_type(self, type_: Type[AccountAPI]) -> List[AccountAPI]:
         return accounts_with_type
 
     def __len__(self) -> int:
+        """
+        The number of account containers managed by ``ape``.
```
fubuloubu, line 99: I believe it's just the number of accounts

src/ape/managers/converters.py
```diff
@@ -17,33 +17,107 @@
 
 # NOTE: This utility converter ensures that all bytes args can accept hex too
 class HexConverter(ConverterAPI):
+    """
+    A converter that converts ``str`` to ``HexBytes``.
+    """
+
     def is_convertible(self, value: str) -> bool:
+        """
```
fubuloubu, line 25: One thing I think works is that `abstractmethod`s with docstrings still appear if the subclass does not override them. So in cases where a generic do string would do, it would make sense not to update each subclass as that is overly verbose and unnecessary.

Please check that this is the case, but could reduce a lot of work!

src/ape/managers/networks.py
```diff
@@ -133,13 +206,37 @@ def parse_network_choice(
...
         return ProviderContextManager(self, provider)
 
     @property
     def default_ecosystem(self) -> EcosystemAPI:
+        """
+        The default ecosystem. Call
+        :meth:`~ape.managers.networks.NetworkManager.set_default_ecosystem` to
+        change the default ecosystem. If a default is not set and there is
+        only a single ecosystem installed, such as Ethereum, then returns
+        that ecosystem.
```
fubuloubu, line 237: Note: probably want a way to actually specify a default in `ape-config.yaml`

src/ape/managers/project.py
```diff
@@ -29,6 +29,19 @@ def _create_source_dict(contracts_paths: Collection[Path]) -> Dict[str, Source]:
 
 @dataclass
 class ProjectManager:
+    """
+    A singleton in ``ape`` for accessing contract-types and dependencies.
```
fubuloubu, line 33: Not really used as a singleton now that I think about it. There is `project` (the local project) and `Project` (alias to `ProjectManager`)

src/ape/managers/networks.py
unparalleled-js: I just created an issue: https://github.com/ApeWorX/ape/issues/321

src/ape/managers/converters.py
unparalleled-js: - [x] `is_convertible()` now comes from the API class

src/ape/managers/project.py
unparalleled-js: Oh good point!

src/ape/managers/accounts.py
```diff
@@ -105,6 +168,18 @@ def __getitem__(self, account_id) -> AccountAPI:
 
     @__getitem__.register
     def __getitem_int(self, account_id: int) -> AccountAPI:
+        """
+        Get an account by index. For example, when you do the CLI command
+        ``ape accounts list --all``, you will see a list of enumerated accounts
+        by their indices. Use this method as a quicker, ad-hoc way to get an
+        account from that index. **NOTE**: It is generally preferred to use
+        :meth:`~ape.managers.accounts.AccountManager.load` or
+        :meth:`~ape.managers.accounts.AccountManager.__getitem_str__`.
```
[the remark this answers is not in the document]
unparalleled-js: thanks, this also made me realize that the special methods were not showing up, so I fixed that too

fubuloubu (approved):

PR #338, opened 2021-12-20
docs: use raises keyword in method docs where apt

unparalleled-js:
### What I did

Uses `Raises:` sphinx keyword whenever possible.
This makes things look more nice and organized on the site and makes use of the sphinx feature.

### How I did it

Find places where we mention what a method raises and replace it with this syntax.
Find other places where we don't mention it and add it.

### How to verify it

read through docs
build and see what I mean by the sytnax.

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -75,6 +75,9 @@ class AccountAPI(AddressAPI):
         """
         Make a transaction call.
 
+        Raises:
+            :class:`~ape.exceptions.AccountsError`: When the nonce is invalid.
+
         Args:
             txn (:class:`~ape.api.providers.TransactionAPI`): The transaction to submit in a call.
             send_everything (bool): ``True`` will send the value difference from balance and fee.
@@ -239,6 +242,9 @@ class AccountContainerAPI:
         """
         Get an account by address.
 
+        Raises:
+            IndexError: When there is no account, locally, with the given address.
+
         Returns:
             :class:`~ape.api.accounts.AccountAPI`
         """
@@ -251,6 +257,9 @@ class AccountContainerAPI:
     def append(self, account: AccountAPI):
         """
         Add an account to the container.
+
+        Raises:
+            :class:`~ape.exceptions.AccountsError`: When the account already is in the container.
 
         Args:
             account (:class:`~ape.api.accounts.AccountAPI`): The account to add.
@@ -270,7 +279,9 @@ class AccountContainerAPI:
     def remove(self, account: AccountAPI):
         """
         Delete an account.
-        Must be known to ``ape`` or else raises :class:`~ape.exceptions.AccountsError`.
+
+        Raises:
+            :class:`~ape.exceptions.AccountsError`: When the account is not known to ``ape``.
 
         Args:
             account (:class:`~ape.accounts.AccountAPI`): The account to remove.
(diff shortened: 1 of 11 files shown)


src/ape/api/providers.py
```diff
@@ -65,18 +65,17 @@ def max_fee(self) -> int:
...
+            NotImplementedError: When setting in a class that did not override the setter.
+
+        Returns:
+            int
         """
         return 0
 
     @max_fee.setter
     def max_fee(self, value: int):
-        """
```
unparalleled-js, line 73: turns out you can't document setters

src/ape/managers/compilers.py
```diff
@@ -51,11 +51,8 @@ def registered_compilers(self) -> Dict[str, CompilerAPI]:
             compiler = compiler_class(config=config)
 
             for extension in extensions:
-
-                if extension in registered_compilers:
```
unparalleled-js, line 55: I am not sure this could ever happen. And if it did, I think this would still confuse the user

src/ape/api/accounts.py
```diff
@@ -285,7 +296,9 @@ def remove(self, account: AccountAPI):
     def __delitem__(self, address: AddressType):
         """
         Delete an account.
-        Must be overriden or else raises ``NotImplementedError``.
+
+        Raises:
+            NotImplementError: When this is not overridden in a plugin.
```
evanblank3, line 301:
```suggestion
            NotImplementError: When not overridden within a plugin.
```

src/ape_ethereum/ecosystem.py
```diff
@@ -152,12 +152,19 @@ def as_dict(self):
 class Receipt(ReceiptAPI):
     def raise_for_status(self, txn: TransactionAPI):
         """
-        Raises :class`~ape.exceptions.OutOfGasError` when the
-        transaction failed and consumed all the gas.
+        Raise an error for the given transaction if the transaction has failed.
```
evanblank3, line 155:
```suggestion
        Raise an error for the given transaction, if the transaction has failed.
```

src/ape_ethereum/ecosystem.py
```diff
@@ -152,12 +152,19 @@ def as_dict(self):
...
 
-        Raises :class:`~ape.exceptions.TransactionError`
-        when the transaction has a failing status otherwise.
+        Args:
+            txn (:class:`ape.api.providers.TransactionAPI`): The sent-transaction to
+              check the status of.
```
evanblank3, line 159:
```suggestion
            txn (:class:`ape.api.providers.TransactionAPI`): The status of the sent-transaction to
              check.
```

evanblank3: Was trying to not end in 'of', but I think I changed the meaning/intention of the comment so please disregard.

src/ape_ethereum/ecosystem.py
unparalleled-js: Hmm I am not sure a comma is needed here.

src/ape_ethereum/ecosystem.py
unparalleled-js: Ah, okay, well maybe something like:

```
The transaction to check
```

fubuloubu (approved):
Love this change!

PR #366, opened 2021-12-21
fix: ensure the correct type when signing a message using eth-account

fubuloubu:
### What I did
Noticed that when you sign a message using eth-account, it does not return `r` and `s` in the signature the same as with signing a transaction

### DIFF
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -4,6 +4,7 @@ from typing import Iterator, Optional
 
 import click
 from eth_account import Account as EthAccount  # type: ignore
+from eth_utils import to_bytes
 
 from ape.api import AccountAPI, AccountContainerAPI, TransactionAPI
 from ape.convert import to_address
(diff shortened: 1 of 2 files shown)


unparalleled-js (approved):
Looks good! I reviewed on my phone. Is there any way we can test this?

fubuloubu:
> Looks good! I reviewed on my phone. Is there any way we can test this?

we ran into this bug with `ape-flashbots` when we had signed a message and tried to convert it to a string and got a failure because `r` and `s` were `int` instead of `bytes`

could probably unit test thing fairly easily :eyes:

src/ape_accounts/accounts.py
```diff
@@ -131,11 +132,19 @@ def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
             return None
 
         signed_msg = EthAccount.sign_message(msg, self.__key)
-        return MessageSignature(v=signed_msg.v, r=signed_msg.r, s=signed_msg.s)  # type: ignore
+        return MessageSignature(  # type: ignore
+            v=signed_msg.v,
```
unparalleled-js, line 136: Kind of weird, I just noticed we do the opposite of this in `ape.types.signatures`

fubuloubu: I think what you see is that `v` is converted to `bytes`

this logic is somewhat messed up, yeah

unparalleled-js:
I did some work on `ape-keyring` over this break and I ran into this issue there as well.

fubuloubu:
> I did some work on `ape-keyring` over this break and I ran into this issue there as well.

Let's definitely work on this Monday. It's a blocker

unparalleled-js:
For consistency, we should fix this in the `TestAccount` (`ape_test.accounts.py`) as well.
That returns `int` instead of bytes here

fubuloubu:
> For consistency, we should fix this in the `TestAccount` (`ape_test.accounts.py`) as well. That returns `int` instead of bytes here

fixed in b0dc7ce

unparalleled-js (approved):

PR #367, opened 2021-12-22
docs: rename to method-docs TOC header to include "python"

unparalleled-js:
### What I did

I think the last TOC section is not 100% clear on what it is.
This rename makes things super obvious I think.

### How I did it

Noticed this section was vague and thought about what a new user might think if we they were only looking for the python docs.

### How to verify it

Does this make sense to you? (Attachment)

Screen Shot 2021-12-22 at 12 53 41 

See how it matches the section above "CLI Reference"

### Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,7 +24,7 @@
 
 ```{eval-rst}
 .. toctree::
-   :caption: User API Reference
+   :caption: Python Reference
    :maxdepth: 1
 
    methoddocs/ape.md


evanblank3 (approved):

fubuloubu (approved):
Brownie uses `API Reference` but I agree this is better. Also `API` might become confusing with our plugin classes (which I've thought about renaming)

unparalleled-js:
> Brownie uses `API Reference` but I agree this is better. Also `API` might become confusing with our plugin classes (which I've thought about renaming)

Yeah, I noticed this too. I also noticed that a lot of javascript projects do this t.. Ape / brownie seem more like SDKs than APIs to me though.

My theory is that Javascript projects overload the term API a lot because of HTTP influences and brownie was following suit?

PR #370, opened 2021-12-26
docs: general revisions and fixes

unparalleled-js:
### What I did

* Switched `:meth:` to `:py:attr:` for `@property` based methods (removes `()` from HTML link output)
* Switch to using `typing` modules types instead of `sphinx` native types where applicable (so `Iterator[str]` instead of `iter[str]`)
* Removes docs from `Web3Provider` so that it uses the base `ProviderAPI` class docs (removes a lot duplication)
* General documentation revising, format fixes, and additions!

### Checklist

### DIFF
--- a/docs/methoddocs/utils.md
+++ b/docs/methoddocs/utils.md
@@ -4,4 +4,5 @@
 .. automodule:: ape.utils
     :members:
     :show-inheritance:
+    :exclude-members: abstractmethod, dataclass
 ```
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -29,7 +29,7 @@ class AccountAPI(AddressAPI):
         Display methods to IPython on ``a.[TAB]`` tab completion.
 
         Returns:
-            list[str]: Method names that IPython uses for tab completion.
+            List[str]: Method names that IPython uses for tab completion.
         """
         return list(super(AddressAPI, self).__dir__()) + [
             "alias",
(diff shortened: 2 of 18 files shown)


docs/methoddocs/utils.md
```diff
@@ -4,4 +4,5 @@
 .. automodule:: ape.utils
     :members:
     :show-inheritance:
+    :exclude-members: abstractmethod, dataclass
```
unparalleled-js, line 7: These borrowed methods docs don't load correctly for us and trying to overwrite them causes other issues

src/ape/api/accounts.py
```diff
@@ -29,7 +29,7 @@ def __dir__(self) -> List[str]:
         Display methods to IPython on ``a.[TAB]`` tab completion.
 
         Returns:
-            list[str]: Method names that IPython uses for tab completion.
+            List[str]: Method names that IPython uses for tab completion.
```
unparalleled-js, line 32: An example of switching to the `typing` module types instead.
I think this will make docs easier to work on and cause less confusions for contributors.

src/ape/api/providers.py
```diff
@@ -659,47 +648,18 @@ def get_block(self, block_id: BlockID) -> BlockAPI:
...
 
     def send_call(self, txn: TransactionAPI) -> bytes:
-        """
-        Execute a new message call immediately without creating a
-        transaction on the block chain.
-        """
         return self._web3.eth.call(txn.as_dict())
 
     def get_transaction(self, txn_hash: str, required_confirmations: int = 0) -> ReceiptAPI:
-        """
```
unparalleled-js, line 687: Deleting all web3 provider docs because they are just copied from the `ProviderAPI` and the output will use the bae class docs if we don't set any, so it is less we have to manager by just deleting these.

fubuloubu (dismissed):

fubuloubu (approved):

PR #371, opened 2021-12-26
refactor: remove incorrectly used property 'is_valid' from the transaction API

unparalleled-js:
### What I did

Remove unneeded method and check.

The `is_valid` method always returns `False` and is not used correctly. The "check" that happens verifies the truthiness of the method existing, it does not call the method or anything, and thus the checked result is always `True` (false positive).

At the same time, there is no implementation or explanation of this property. It should be removed if it not needed or else it should be implemented properly. Since we are getting by without it, I opted for removing here. Let me know otherwise!

### How I did it

I was documenting these methods and noticed the docs did not really say what `is_valid` means. Then, I noticed it was not getting called correctly. Then, I noticed it was never actually getting set o

### How to verify it

This PR should have no effect on anything.

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -50,9 +50,6 @@ class EthereumConfig(ConfigItem):
 
 
 class BaseTransaction(TransactionAPI):
-    def is_valid(self) -> bool:
-        return False
-
     def as_dict(self) -> dict:
         data = super().as_dict()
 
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -50,9 +50,6 @@ class EthereumConfig(ConfigItem):
 
 
 class BaseTransaction(TransactionAPI):
-    def is_valid(self) -> bool:
```
unparalleled-js, line 53: Since this did not have `@property`, it was not getting called correctly.

fubuloubu (approved):

PR #373, opened 2021-12-27
docs: fix path versioning issue / filter out betas

unparalleled-js:
### What I did

* Fix path issue causing versioning system not to work right
* Filter out beta and alpha from versioning system
* Keep CLI TOC nested to 1 level to match the other and look cleaner.

### How I did it

Using path lib to sanitize the path (a leading slash was causing issues)

### How to verify it

Build docs locally
Copy "latest" folder and create fake folders next to it for "stable", "v0.1.0.beta.1" and "v0.1.0". The beta should be filtered out.

### Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -77,7 +77,7 @@ def fixpath(path: str) -> str:
     new = f"/{project}/latest/_static"
 
     if suffix:
-        new = str(Path(new) / suffix)
+        new = str(Path(new) / suffix.lstrip("/"))
 
     return new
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -88,8 +88,14 @@ def get_versions() -> List[str]:
         return []
 
     versions = [
-        d.name for d in build_dir.iterdir() if d.is_dir and re.match(r"v\d+.?\d?.?\d?", d.stem)
+        d.name
+        for d in build_dir.iterdir()
+        if d.is_dir
+        and re.match(r"v\d+.?\d?.?\d?", d.stem)
+        and "beta" not in d.name
```
unparalleled-js, line 95: I guess we could also just prevent the docs from getting built and pushed

fubuloubu (approved):

Issue #321, opened 2021-12-15
Allow specifying default network ecosystem, network, and provider in ape-config


unparalleled-js:
### Overview

Allow us to set a default value for ecosystem, network, and provider using the `ape-config.yaml` file.

### Specification

Maybe

```yaml
ethereum:
  development:
    default_provider: hardhat
```

but I am not sure.

unparalleled-js:
realized it is way easier to just the ecosystem plugin for the config. Reason: This is the only way to set defaults per other ecosystems besides Ethereum because they can set them in their own config rather than trying to dynamically generate these in the `networks` plugin (I tried that it was not working anyway)

PR #375, opened 2021-12-29
feat: set and honor default provider in config

unparalleled-js:
### What I did

fixes: #321

* Add way to set default providers per network in `ape-config.yaml`.
* Causes `test` to be the default default provider for the `development` network (which is what prompted me to make this PR)
* To get the tests to work, I had to force a reload of the ape module in `conftest`.

### How I did it

Use the `ethereum` config.

### How to verify it

Try changing the default provider in the config and see that it works:

```yaml
ethereum:
  development:
    default_provider: geth
```

After the change above, it should start up and use the dev geth:

```bash
ape run <script>
```

Comment out the yaml and see that `test` is the default and run the command again and notice it uses the `test` by  default.

### Checklist

### DIFF
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -13,6 +13,7 @@ from ape.cli.utils import Abort
 from ape.exceptions import ContractError
 from ape.logging import LogLevel, logger
 from ape.managers.config import ConfigManager
+from ape.managers.networks import NetworkManager
 from ape.managers.project import ProjectManager
 from ape.types import ContractType
 
@@ -27,6 +28,7 @@ class ApeCliContextObject:
     def __init__(self):
         self.logger = logger
         self._project = None
+        self._networks = None
 
     @property
     def project(self) -> ProjectManager:
@@ -48,6 +50,15 @@ class ApeCliContextObject:
     @property
     def config(self) -> ConfigManager:
         return self._project.config
+
+    @property
+    def networks(self) -> NetworkManager:
+        if not self._networks:
+            from ape import networks
+
+            self._networks = networks
+
+        return self._networks
 
     @staticmethod
     def abort(msg: str, base_error: Exception = None):
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -12,6 +12,7 @@ from ape.cli.choices import (
 from ape.cli.utils import Abort
 from ape.exceptions import ContractError
 from ape.logging import LogLevel, logger
+from ape.managers.config import ConfigManager
 from ape.managers.project import ProjectManager
 from ape.types import ContractType
 
@@ -43,6 +44,10 @@ class ApeCliContextObject:
             self._project = project
 
         return self._project
+
+    @property
+    def config(self) -> ConfigManager:
+        return self._project.config
 
     @staticmethod
     def abort(msg: str, base_error: Exception = None):
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -12,6 +12,7 @@ from ape.cli.choices import (
 from ape.cli.utils import Abort
 from ape.exceptions import ContractError
 from ape.logging import LogLevel, logger
+from ape.managers.accounts import AccountManager
 from ape.managers.config import ConfigManager
 from ape.managers.networks import NetworkManager
 from ape.managers.project import ProjectManager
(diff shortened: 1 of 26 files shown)


fubuloubu (dismissed):

src/ape/plugins/__init__.py
```diff
@@ -120,7 +120,11 @@ def __init__(self):
         for _, name, ispkg in pkgutil.iter_modules():
             if name.startswith("ape_") and ispkg:
                 try:
-                    plugin_manager.register(importlib.import_module(name))
+                    module = importlib.import_module(name)
+                    if plugin_manager.is_registered(module):
+                        plugin_manager.unregister(module)
+
+                    plugin_manager.register(module)
```
fubuloubu, line 127: Why not this? I don't get why unregistering the module here
```suggestion
                    if not plugin_manager.is_registered(module):
                        plugin_manager.register(module)
```

unparalleled-js: If you switch projects in the same python session, the plugin configs don't reload otherwise.
This was causing issues in the tests.

unparalleled-js: Like it will still have the plugin configs from the first original project. I am not sure the best way to handle this. It has been really difficult to try and test this.

fubuloubu: Hmm, yes this would seem to be a big issue then

fubuloubu: Wait, why are you trying to load configs from another project?

src/ape/cli/options.py
```diff
@@ -116,7 +116,7 @@ def network_option(default: str = networks.default_ecosystem.name):
           use as the default. Defaults to how ``ape`` normally
           selects a default network.
     """
-
+    default = default or networks.default_ecosystem.name
```
unparalleled-js, line 119: Step 1 of New Years laziness:  Calculate the default runtime more lazily for this CLI option (this was required to get the tests to work)

src/ape/managers/config.py
```diff
@@ -37,17 +37,21 @@ class ConfigManager:
...
-    dependencies: List[str] = []
+    dependencies: Dict[str, str] = {}
     plugin_manager: PluginManager
-    _plugin_configs: Dict[str, ConfigItem] = dict()
 
-    def __post_init__(self):
-        config_file = self.PROJECT_FOLDER / CONFIG_FILE_NAME
+    _project_plugin_configs: Dict[str, Dict[str, ConfigItem]] = {}
+
+    @property
```
unparalleled-js, line 45: In order to persist laziness throughout the New Year, we had to

1. Make this a "property" instead of happening at `__init__` time.
2. Make it cached per project. This way, if you change directories to another project (as the tests have to do), this can grab the new project's `ape-config.yaml` file changes. Yay!

src/ape/managers/networks.py
```diff
@@ -31,27 +30,46 @@ class NetworkManager:
     plugin_manager: PluginManager
     active_provider: Optional[ProviderAPI] = None
     _default: Optional[str] = None
+    _project_ecosystems: Dict[str, Dict[str, EcosystemAPI]] = {}
 
     def __repr__(self):
         return f"<NetworkManager, active_provider={self.active_provider}>"
 
-    @cached_property
+    @property
```
unparalleled-js, line 38: Don't be fooled! The laziness goes on here still. Only now, we have to re-initialize the Ecosystem classes when we change project directories (as we do in the tests). This way, the new config file changes can get applied to the providers (such as the default provider!)

src/ape/plugins/__init__.py
unparalleled-js: This was no longer needed thank to your suggestions for trying to load these things more lazily. So far, it seems like this is a success!

fubuloubu:
Seems like you are entering the new year with a new, lazy attitude!

src/ape/managers/config.py
```diff
@@ -37,17 +37,21 @@ class ConfigManager:
     PROJECT_FOLDER: Path
     name: str = ""
     version: str = ""
-    dependencies: List[str] = []
+    dependencies: Dict[str, str] = {}
```
unparalleled-js, line 40: This type was incorrect - I am just fixing it here

fubuloubu (approved):

Issue #376, opened 2021-12-30
APE doesn't handle lists


SwapOperator:
### Environment information

* `ape` and plugin versions: 0.1.0b1

```
$ ape --version

0.1.0b1

$ ape plugins list

Installed Plugins:
  etherscan     0.1.0a2
  hardhat     0.1.0a3
  alchemy     0.1.0a1
  vyper     0.1.0a8
  infura     0.1.0a4

```

* Python Version: Python 3.8.6
* OS: macOS

### What went wrong?

I have tested my contract func that works with lists and got error.
**Func Example:**
```
def foo(_tokens: address[25]) -> bool:
```
**Token list and testing func:**
```
tokens = [token1, token2, token3, token4, ZERO_ADDRESS, ZERO_ADDRESS,ZERO_ADDRESS...] 
assert deployer.foo(tokens, sender=owner)
```

**Error:**
```
eth_abi.exceptions.EncodingTypeError: Value `<Token 0x6418E...`  of type <class 'ape.contracts.base.ContractInstance'> cannot be encoded by AddressEncoder
```

Please include information like:

* what command you ran: **ape run deploy_test  --network ethereum:rinkeby:infura**
* the code that caused the failure (see [this link](https://help.github.com/articles/basic-writing-and-formatting-syntax/) for help with formatting code)
* full output of the error you received

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

fubuloubu:
@SwapOperator sorry it took so long, we fixed a host of other conversion issues now too!

SwapOperator:
> @SwapOperator sorry it took so long, we fixed a host of other conversion issues now too!

Super cool! I should update my version :)

fubuloubu:
> > @SwapOperator sorry it took so long, we fixed a host of other conversion issues now too!
> 
> Super cool! I should update my version :)

Newest version not released yet! You can try installing from `main` if you like though
```bash
$ pip install git+https://github.com/ApeWorX/ape.git@main
```

PR #377, opened 2021-12-31
feat: handle iterable args

fubuloubu:
### What I did
Added support to `ConversionManager` for handling `tuple` and `list` conversions (e.g. recursively applying conversions)

fixes: #376

### Checklist

- [x] Add integration support to `ape.contracts` ABI arg processing

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -36,6 +36,7 @@ class ContractConstructor:
     deployment_bytecode: bytes
     abi: Optional[ABI]
     provider: ProviderAPI
+    converter: "ConversionManager"
 
     def __post_init__(self):
         if len(self.deployment_bytecode) == 0:
@@ -67,6 +68,7 @@ class ContractCall:
     abi: ABI
     address: AddressType
     provider: ProviderAPI
+    converter: "ConversionManager"
 
     def __repr__(self) -> str:
         return self.abi.signature
@@ -97,6 +99,7 @@ class ContractCall:
 @dataclass
 class ContractCallHandler:
     provider: ProviderAPI
+    converter: "ConversionManager"
     address: AddressType
     abis: List[ABI]
 
@@ -130,6 +133,7 @@ class ContractTransaction:
     abi: ABI
     address: AddressType
     provider: ProviderAPI
+    converter: "ConversionManager"
 
     def __repr__(self) -> str:
         return self.abi.signature
@@ -151,6 +155,7 @@ class ContractTransaction:
 @dataclass
 class ContractTransactionHandler:
     provider: ProviderAPI
+    converter: "ConversionManager"
     address: AddressType
     abis: List[ABI]
 
@@ -179,6 +184,7 @@ class ContractLog:
 @dataclass
 class ContractEvent:
     provider: ProviderAPI
+    converter: "ConversionManager"
     address: str
     abis: List[ABI]
     cached_logs: List[ContractLog] = []
@@ -199,6 +205,7 @@ class ContractInstance(AddressAPI):
     """
 
     _address: AddressType
+    _converter: "ConversionManager"
     _contract_type: ContractType
 
     def __repr__(self) -> str:
@@ -256,6 +263,7 @@ class ContractInstance(AddressAPI):
 
             kwargs = {
                 "provider": self.provider,
+                "converter": self._converter,
                 "address": self.address,
                 "abis": selected_abis,
             }
@@ -298,6 +306,8 @@ class ContractContainer:
     _provider: Optional[ProviderAPI]
     # _provider is only None when a user is not connected to a provider.
 
+    _converter: "ConversionManager"
+
     def __repr__(self) -> str:
         return f"<{self.contract_type.contractName}>"
 
@@ -324,6 +334,7 @@ class ContractContainer:
         return ContractInstance(  # type: ignore
             _address=address,
             _provider=self._provider,
+            _converter=self._converter,
             _contract_type=self.contract_type,
         )
 
@@ -347,6 +358,7 @@ class ContractContainer:
         constructor = ContractConstructor(  # type: ignore
             abi=self.contract_type.constructor,
             provider=self._provider,
+            converter=self._converter,
             deployment_bytecode=self._deployment_bytecode,
         )
         return constructor.encode(*args, **kwargs)
@@ -388,6 +400,7 @@ def _Contract(
         return ContractInstance(  # type: ignore
             _address=converted_address,
             _provider=provider,
+            _converter=converters,
             _contract_type=contract_type,
         )
 
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -54,6 +54,7 @@ class ContractConstructor:
         )
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        args = self.converter.convert(args, tuple)
         if "sender" in kwargs:
             sender = kwargs["sender"]
             txn = self.encode(*args, **kwargs)
@@ -74,11 +75,13 @@ class ContractCall:
         return self.abi.signature
 
     def encode(self, *args, **kwargs) -> TransactionAPI:
+        args = self.converter.convert(args, tuple)
         return self.provider.network.ecosystem.encode_transaction(
             self.address, self.abi, *_encode_address_args(*args), **_encode_address_kwargs(**kwargs)
         )
 
     def __call__(self, *args, **kwargs) -> Any:
+        args = self.converter.convert(args, tuple)
         txn = self.encode(*args, **kwargs)
         txn.chain_id = self.provider.network.chain_id
 
@@ -108,6 +111,7 @@ class ContractCallHandler:
         return abis[-1].signature
 
     def __call__(self, *args, **kwargs) -> Any:
+        args = self.converter.convert(args, tuple)
         selected_abi = _select_abi(self.abis, args)
         if not selected_abi:
             raise ArgumentsLengthError()
@@ -139,11 +143,13 @@ class ContractTransaction:
         return self.abi.signature
 
     def encode(self, *args, **kwargs) -> TransactionAPI:
+        args = self.converter.convert(args, tuple)
         return self.provider.network.ecosystem.encode_transaction(
             self.address, self.abi, *_encode_address_args(*args), **_encode_address_kwargs(**kwargs)
         )
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        args = self.converter.convert(args, tuple)
         if "sender" in kwargs:
             sender = kwargs["sender"]
             txn = self.encode(*args, **kwargs)
@@ -164,6 +170,7 @@ class ContractTransactionHandler:
         return abis[-1].signature
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        args = self.converter.convert(args, tuple)
         selected_abi = _select_abi(self.abis, args)
         if not selected_abi:
             raise ArgumentsLengthError()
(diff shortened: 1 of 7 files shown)


src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
         return to_checksum_address(value)
 
 
-hex_address_converter = HexAddressConverter(None, None)  # type: ignore
+hex_address_converter = HexAddressConverter(None, None, None)  # type: ignore
+
+
+class ListTupleConverter(ConverterAPI):
+    """
+    A converter that converts any tuple or list of items
```
unparalleled-js, line 100:
```suggestion
    A converter that converts any tuple or list of items to a list of
```

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, list) or isinstance(value, tuple)
+
+    def convert(self, value: Union[list, tuple]) -> Union[list, tuple]:
+        """
+        Convert the items inside the given list or tuple.
+
+        Args:
+            value (``list`` or ``tuple``): The collection to convert.
```
unparalleled-js, line 112:
```suggestion
            value (Union[List, Tuple]): The collection to convert.
```

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+class ListTupleConverter(ConverterAPI):
+    """
+    A converter that converts any tuple or list of items
+    :class:`~ape.types.AddressType`.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, list) or isinstance(value, tuple)
+
+    def convert(self, value: Union[list, tuple]) -> Union[list, tuple]:
```
unparalleled-js, line 107:
```suggestion
    def convert(self, value: Union[lList, Tuple]) -> Union[List, Tuple]:
```

Don't these have to be the `typing` classes?

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+        """
+        Convert the items inside the given list or tuple.
+
+        Args:
+            value (``list`` or ``tuple``): The collection to convert.
+
+        Returns:
+            ``list`` or ``tuple`` (depending on input)
```
unparalleled-js, line 115: Union[list, tuple]: Depending on the input

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+
+
+class ListTupleConverter(ConverterAPI):
+    """
+    A converter that converts any tuple or list of items
+    :class:`~ape.types.AddressType`.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, list) or isinstance(value, tuple)
```
unparalleled-js, line 105:
```suggestion
        return isinstance(value, (list, tuple))
```

src/ape/managers/converters.py
fubuloubu: Hmm, this statement is just way off entirely, isn't it?

src/ape/managers/converters.py
```diff
@@ -119,14 +180,18 @@ def __repr__(self):
 
     @cached_property
     def _converters(self) -> Dict[Type, List[ConverterAPI]]:
+        list_tuple_converter.converter = self  # Splice this in (no need for the rest)
         converters: Dict[Type, List[ConverterAPI]] = {
             AddressType: [address_api_converter, hex_address_converter],
             bytes: [hex_converter],
             int: [],
+            Decimal: [],
+            list: [list_tuple_converter],
```
unparalleled-js, line 189: Should these also be `List` and `Tuple` (from `typing` module)?

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+                    raise ConversionError(f"Unsupported ABI Type: {v.__class__}")
+
+            # Try all of them to see if one converts it over (only use first one)
+            for check_fn, convert_fn in map(lambda c: (c.is_convertible, c.convert), converters):
+                if check_fn(v):
+                    converted_value.append(convert_fn(v))
+                    conversion_found = True
+                    break
+
+                if conversion_found:
```
unparalleled-js, line 145: Does this ever get execute?
I would think the break from line 143 would always happen first

src/ape/managers/converters.py
fubuloubu: these are concrete types, e.g. `convert(v, list)`

Tried to think of what `convert(v, List[int, str, ..., bytes])` might look like, but it seemed bad UX

src/ape/managers/converters.py
fubuloubu: Oh whoops, leftover from when I flattened a `for` loop

src/ape/managers/converters.py
```diff
@@ -91,7 +92,67 @@ def convert(self, value: str) -> AddressType:
...
+                    break
```
fubuloubu, line 146:
```suggestion
                    break
```

src/ape/managers/converters.py
```diff
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Type
+from decimal import Decimal
+from typing import Any, Dict, List, Optional, Type, Union
```
fubuloubu, line 2:
```suggestion
from typing import Any, Dict, List, Optional, Tuple, Type, Union
```

src/ape/managers/converters.py
unparalleled-js: Yeah that makes sense.

unparalleled-js (changes requested):
This PR does not actually fix the bug it says!
lists and tuples already work, it is the fact they are passing `ContractInstance` classes as "addresses" that is causing their bug (the bug ticket wording should be changed to reflect this - I will add a comment).

We need to support `ContractInstance` in the conversion and that should resolve their bug.

fubuloubu:
It works now!
```py
In [1]: a = accounts.load("my-alias")

In [2]: convert([a], list)
Out[2]: ['0xe868a852342Be8a43C4Ce8fC4642aaA7b857Bb82']
```

unparalleled-js:
> It works now!
> 
> ```python
> In [1]: a = accounts.load("my-alias")
> 
> In [2]: convert([a], list)
> Out[2]: ['0xe868a852342Be8a43C4Ce8fC4642aaA7b857Bb82']
> ```

I am not still not able to pass in a `ContractInstance`

fubuloubu:
> Deploy a smart contract containing a method that accepts an array of addresses

Sorry, this still doesn't work yet :sweat_smile: 

Still working on ABI encoding step where `args = convert(args, tuple)` will be used for `*args`

unparalleled-js:
> > Deploy a smart contract containing a method that accepts an array of addresses
> 
> Sorry, this still doesn't work yet 😅
> 
> Still working on ABI encoding step where `args = convert(args, tuple)` will be used for `*args`
> 
> Edit: prior, the list conversion wasn't even working, e.g. this test: https://github.com/ApeWorX/ape/pull/377/files#diff-265642a4b634858b0c1bf2c7325eda296e978fa772db3b8f648365bc571ecf6fR45

Oh ok no worries! I will check out this list conversion fix though, brb

unparalleled-js (approved):

fubuloubu:
@unparalleled-js 4cba949 should contain the fix, although I need to add a test for it

fubuloubu:
Also realizing we'd probably want support for converting Transaction `kwargs` as well

unparalleled-js:
The Project Manager needs to be updated, I am getting this error when trying to deploy:

```
    374             raise AttributeError(f"{self.__class__.__name__} has no attribute '{attr_name}'.")
    375 
--> 376         return ContractContainer(  # type: ignore
    377             contract_type=contract_type, _provider=self.networks.active_provider
    378         )

TypeError: __init__() missing 1 required positional argument: '_converter'
```

fubuloubu:
> The Project Manager needs to be updated, I am getting this error when trying to deploy:
> 
> ```
>     374             raise AttributeError(f"{self.__class__.__name__} has no attribute '{attr_name}'.")
>     375 
> --> 376         return ContractContainer(  # type: ignore
>     377             contract_type=contract_type, _provider=self.networks.active_provider
>     378         )
> 
> TypeError: __init__() missing 1 required positional argument: '_converter'
> ```

fixed in 873263d

unparalleled-js:
Okay, so now I can't call my method that takes a list anymore. I always get the `ArgumentsLengthError()`

unparalleled-js:
Fo

> Okay, so now I can't call my method that takes a list anymore. I always get the `ArgumentsLengthError()`
> 
> ArgumentsLengthError: The number of the given arguments do not match what is defined in the ABI.

Setting breaking points, it seems to be doubling my args. Like if I give it a list with two addresses in it (one an account and one a contract) it will turn it into two lists in `ContractTransactionHandler` line 173.

```
In [8]: c.fund([c, a], value=100, sender=a2)
> /ape/src/ape/contracts/base.py(174)__call__()
-> selected_abi = _select_abi(self.abis, args)
(Pdb) args
self = fund(address[] test)
args = ([<Fund 0x274b028b03A250cA03644E6c578D81f019eE1323>, <TestAccount 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C>],)
(Pdb) c
> /ape/src/ape/contracts/base.py(176)__call__()
-> raise ArgumentsLengthError()
(Pdb) args
self = fund(address[] test)
args = (['0x274b028b03A250cA03644E6c578D81f019eE1323', '0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C'], ['0x274b028b03A250cA03644E6c578D81f019eE1323', '0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C'])
(Pdb) 
```

src/ape/managers/converters.py
```diff
@@ -91,7 +92,52 @@ def convert(self, value: str) -> AddressType:
...
+        """
+
+        converted_value: List[Any] = []
+
+        for v in value:
+            # Try all of them to see if one converts it over (only use first one)
+            conversion_found = False
+            # NOTE: Double loop required because we might not know the exact type of the inner
+            #       items. The UX of having to specify all inner items seemed poor as well.
+            for typ in self.converter._converters:
```
unparalleled-js, line 124: I think we may need to filter out `list` and `tuple` here - that could explain why my lists are doubling.

src/ape/contracts/base.py
```diff
@@ -18,11 +18,6 @@
     from ape.managers.networks import NetworkManager
 
 
-def _encode_address_args(*args):
-    # Convert higher level address types to str
-    return [arg.address if isinstance(arg, AddressAPI) else arg for arg in args]
```
unparalleled-js, line 23: I am guessing you are going to add a converter that does this?

src/ape/managers/converters.py
fubuloubu: naw, fixed this in c7ddce6

src/ape/contracts/base.py
fubuloubu: we already have one!

fubuloubu: This was there as a stopgap because we didn't do the full conversion step inside of `ape/contracts` methods yet

fubuloubu: Next would come the elimination of `_encode_address_kwargs` as well

fubuloubu: `_encode_address_kwargs` removed in 875cd73

unparalleled-js (approved):
It works!

Issue #374, opened 2021-12-28
Caching: Deployment Map


unparalleled-js:
### Overview

It is a bit of hurdle trying to figure out how to manage and share deployment addresses for projects. This way, we can run automations from team-mate to team-mate.

### Specification

Similar to this feature in brownie: https://eth-brownie.readthedocs.io/en/stable/deploy.html#the-deployment-map

For sharing, maybe we could have config items automatically populate the deployment map. (Does brownie do this?)

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

fubuloubu:
You could have it populate items in the `ape-config.yaml` that could be used in a script to give access to these addresses in a much faster way in order to load your contract system fast

note that ethPM has support for these keys (not in the most streamlined way)

unparalleled-js:
> You could have it populate items in the `ape-config.yaml` that could be used in a script to give access to these addresses in a much faster way in order to load your contract system fast
> 
> note that ethPM has support for these keys (not in the most streamlined way)

Would `ape` be changing the `ape-config.yaml` or would my script?

unparalleled-js:
Right now, I just have a script that creates a `map.json` file which is controlled in `git`.

fubuloubu:
> > You could have it populate items in the `ape-config.yaml` that could be used in a script to give access to these addresses in a much faster way in order to load your contract system fast
> > 
> > note that ethPM has support for these keys (not in the most streamlined way)
> 
> Would `ape` be changing the `ape-config.yaml` or would my script?

Ape shouldn't change the config, mutable Co fig osunds awful. However, we could read those keys and load those contracts at the given addresses very easily. For a deployed project, it tends not to change very often e.g. Yearn Vaults has a registry that indexes all the vaults, I just need the address of the registry available on every deployed network/ecosystem

unparalleled-js:
Okay, yes I thought so, but was just making sure.
So in essence, this feature has two parts:

1. Allow the user to set addresses in the config file and allow them to access / initialize `ContractInstance`s of those addresses
2. Cache active deployments (in `.build` directory?) and also allow user to create `ContractInstance`s of those.

I am trying to think of syntax for this:

```python
my_contract = project.MyContract.at(-1)
```

where `-1` indicates the latest (most recent) deployment, including both the config file and active ones.

Else, we can have like a map in the config file?

```python
contract_address = config.get_config("deployments").token_v2
token_v2 = project.My
```

___

[...]

PR #378, opened 2022-01-02
feat: allow saving deployments to config file

unparalleled-js:
### What I did

Partially addresses: #374 
Allows you to save your deployments to your config file.

### How I did it

Add another base key `deployments` to config manager user config logic

### How to verify it

Have this in your config: 

```yaml
deployments:
  ethereum:
    rinkeby: 
    - contract_type: MyContract
      address: 0xc111222333444555666777888999AaABbBCcCDdD
```

Then create a script that prints the address:

```python
from ape import config, networks

def main():
    network_name = networks.active_provider.network.name
    network_deployments = config.deployments["ethereum"][network_name]
    contract_address = [d for d in network_deployments if d["contract_type"] == "MyContract"][0]["address"]
    print(contract_address)
```

Then run that script in the same network from the config:

```bash
ape run print_address --network :rinkeby:infura
```

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -1,0 +1,46 @@
+# Configure Ape
+
+A `ape-config.yaml` file allows you to configure ape. This guide serves as an index of the settings you can include 
+in an `ape-config.yaml` file.
+
+## Deployments
+
+Share deployments with teammates.
+
+```yaml
+deployments:
+  ethereum:
+    mainnet:
+      MyContract:
+      - 0xc123aAacCcbBbaAa444777000111222111222111
+    ropsten:
+      MyContract:
+      - 0xc222000cCcbBbaAa444777000111222111222222
+```
+
+## Networks
+
+Set default networks.
+
+```yaml
+ethereum:
+  development:
+    default_provider: hardhat
+```
+
+## Plugins
+
+Set which plugins you want to always use:
+
+```yaml
+plugins:
+  - name: solidity
+    version: 0.1.0b2
+  - name: ens
+```
+
+Install these plugins by running command:
+
+```bash
+ape plugins install
+```
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -44,3 +44,13 @@ Install these plugins by running command:
 ```bash
 ape plugins install
 ```
+
+## Testing
+
+Configure your test accounts:
+
+```yaml
+test:
+  mnemonic: test test test test test test test test test test test junk
+  number_of_accounts: 5
+```
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -5,7 +5,7 @@ in an `ape-config.yaml` file.
 
 ## Deployments
 
-Share deployments with teammates:
+Share import deployments to public networks with your teammates:
 
 ```yaml
 deployments:
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -1,11 +1,11 @@
 # Configure Ape
 
-A `ape-config.yaml` file allows you to configure ape. This guide serves as an index of the settings you can include 
+An `ape-config.yaml` file allows you to configure ape. This guide serves as an index of the settings you can include 
 in an `ape-config.yaml` file.
 
 ## Deployments
 
-Share deployments with teammates.
+Share deployments with teammates:
 
 ```yaml
 deployments:
@@ -20,7 +20,7 @@ deployments:
 
 ## Networks
 
-Set default networks.
+Set default networks:
 
 ```yaml
 ethereum:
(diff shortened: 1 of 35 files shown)


src/ape/managers/config.py
```diff
@@ -38,8 +38,8 @@ class ConfigManager:
     name: str = ""
     version: str = ""
     dependencies: Dict[str, str] = {}
+    deployments: Dict[str, Dict[str, str]] = {}
```
unparalleled-js, line 41: Maybe a good time to start a guide that is solely devoted to the config file?

fubuloubu (approved):

fubuloubu:
Still think deployments need to be expanded a bit:
```yaml
deployments:
  ethereum:  # ecosystem
    rinkeby:  # network
      MyContract:  # ContractType[deployments]
      - 0x1122C33728208F55e11AC24536b6AB207A4a9E03  # No string necessary?
```

unparalleled-js:
> Still think deployments need to be expanded a bit:
> 
> ```yaml
> deployments:
>   ethereum:  # ecosystem
>     rinkeby:  # network
>       MyContract:  # ContractType[deployments]
>       - 0x1122C33728208F55e11AC24536b6AB207A4a9E03  # No string necessary?
> ```

Good point! Especially about including the ecosystem and using real contract types. Let me add in these revisions; I want the schema to be solid enough to not have to change too much to lessen breaking changes

src/ape/managers/config.py
unparalleled-js: I initialized one and added a few settings, including this one.

unparalleled-js:
Oh I just now saw your comment on #374  regarding the syntax for `.at(-1)`.

docs/userguides/config.md
```diff
@@ -0,0 +1,56 @@
+# Configure Ape
+
+An `ape-config.yaml` file allows you to configure ape. This guide serves as an index of the settings you can include 
+in an `ape-config.yaml` file.
+
+## Deployments
+
+Share deployments with teammates:
```
fubuloubu, line 8:
```suggestion
Share import deployments to public networks with your teammates:
```

src/ape/managers/config.py
```diff
@@ -1,17 +1,23 @@
...
 from ape.api import ConfigDict, ConfigItem
+from ape.convert import to_address
 from ape.exceptions import ConfigError
 from ape.plugins import PluginManager
 from ape.utils import load_config
 
 CONFIG_FILE_NAME = "ape-config.yaml"
 
 
+class DeploymentConfig(ConfigItem):
```
fubuloubu, line 16: Love this! So easy to add support for a new key!

src/ape/managers/config.py
```diff
@@ -58,6 +64,15 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
         self.version = user_config.pop("version", "")
         self.dependencies = user_config.pop("dependencies", {})
 
+        # Sanitize deployment addresses.
+        deployments = user_config.pop("deployments", {})
+        for ecosystem_name, networks in deployments.items():
+            for network_name, contract_deployments in networks.items():
```
fubuloubu, line 70: What happens if `ecosystem_name` or `network_name` are not valid?

src/ape/managers/config.py
```diff
@@ -58,6 +64,15 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
...
+                for deployment in [d for d in contract_deployments if "address" in d]:
```
fubuloubu, line 71: I feel like `if "address" in d` should be filtered out by the config item framework somehow (e.g. if neither is in there, the config should fail to parse it raising an error to fix it)

fubuloubu (dismissed):

src/ape/managers/config.py
unparalleled-js: Yeah this is probably a remnant for when I had it as a dict

unparalleled-js: Dug in more...

The case where the user leaves off the the address like this:

```
deployments:
  ethereum:
    rinkeby: 
    - contract_type: MyContract

```
 
currently makes it past the parsing stage. Leaving off `if "address" in d` causes a `KeyError` at that part (maybe that is ok?). If we want this to be an error-state, perhaps it is best, like you said, to fail earlier in the config parsing stage.

src/ape/managers/config.py
unparalleled-js: Nothing was happening before, but now, it will raise a `ConfigError` if they are invalid.

src/ape/managers/config.py
unparalleled-js: Okay, I made it raise better errors now for this situation!

PR #382, opened 2022-01-03
feat: added uninstall command and made install have confirmations

Ninjagod1251:
### What I did

added an uninstall command so we could uninstall all the plugins from the config.yaml

fixes: #

### How I did it

added a block of code in the _cli.py to make the command

### How to verify it

install and uninstall the config list of plugins in your own project!

just remember the schema for the config file.

### Checklist

### DIFF
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -211,9 +211,7 @@ def uninstall(cli_ctx, skip_confirmation):
             any_uninstall_failed = True
 
         # if plugin is installed and 2nd class. We should uninstall it
-        if is_plugin_installed(module_name) and (
-            module_name in github_client.available_plugins or skip_confirmation
-        ):
+        if installed_plugin and (available_plugin or skip_confirmation):
             cli_ctx.logger.info(f"Uninstalling {package_name}...")
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -194,17 +194,20 @@ def uninstall(cli_ctx, skip_confirmation):
             any_uninstall_failed = True
 
         module_name, package_name = extract_module_and_package_install_names(plugin)
-        # check for installed check the config.yaml
-        if not is_plugin_installed(module_name):
-            cli_ctx.logger.warning(f"Plugin '{plugin}' is not installed.")
-            any_uninstall_failed = True
+
+        available_plugin = module_name not in github_client.available_plugins
+        installed_plugin = is_plugin_installed(plugin)
 
         # if plugin is installed but not a 2nd class. It must be a third party
-        if module_name not in github_client.available_plugins and is_plugin_installed(plugin):
+        if not available_plugin and installed_plugin:
             cli_ctx.logger.warning(
                 f"Plugin '{module_name}' is installed but not in available plugins."
                 f" Please uninstall outside of Ape."
             )
+            any_uninstall_failed = True
+        # check for installed check the config.yaml
+        elif not installed_plugin:
+            cli_ctx.logger.warning(f"Plugin '{plugin}' is not installed.")
             any_uninstall_failed = True
 
         # if plugin is installed and 2nd class. We should uninstall it
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -168,7 +168,40 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
+            )
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been added.")
+            else:
+                cli_ctx.logger.error(f"Failed to add '{package_name}'.")
+                sys.exit(1)
+
+
+@cli.command(short_help="Uninstall all plugins in the local config file")
+@ape_cli_context()
+@skip_confirmation_option("Don't ask for confirmation to install the plugins")
+def uninstall(cli_ctx, skip_confirmation):
+    plugins = config.get_config("plugins") or []
+    for plugin in plugins:
+        module_name, package_name = extract_module_and_package_install_names(plugin)
+        if is_plugin_installed(module_name) and (
+            module_name in github_client.available_plugins
+            or skip_confirmation
+            or click.confirm(f"Uninstall unknown 3rd party plugin '{package_name}'?")
+        ):
+            cli_ctx.logger.info(f"Uninstalling {package_name}...")
+            # NOTE: Be *extremely careful* with this command, as it modifies the user's
+            #       installed packages, to potentially catastrophic results
+            # NOTE: This is not abstracted into another function *on purpose*
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "uninstall", "--quiet", f"{package_name}", "-y"]
+            )
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been removed.")
+            else:
+                cli_ctx.logger.error(f"Failed to remove '{package_name}'.")
+                sys.exit(1)
 
 
 @cli.command(short_help="Uninstall an ape plugin")
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -156,6 +156,7 @@ def add(cli_ctx, plugin, version, skip_confirmation):
 @ape_cli_context()
 @skip_confirmation_option("Don't ask for confirmation to install the plugins")
 def install(cli_ctx, skip_confirmation):
+    is_insall_fail = False
     plugins = config.get_config("plugins") or []
     for plugin in plugins:
         module_name, package_name = extract_module_and_package_install_names(plugin)
@@ -173,12 +174,18 @@ def install(cli_ctx, skip_confirmation):
             )
             if result != 0:
                 cli_ctx.logger.error(f"Failed to add '{package_name}'.")
+                is_insall_fail = True
+    if is_insall_fail:
+        sys.exit(1)
+    else:
+        sys.exit(0)
 
 
 @cli.command(short_help="Uninstall all plugins in the local config file")
 @ape_cli_context()
 @skip_confirmation_option("Don't ask for confirmation to install the plugins")
 def uninstall(cli_ctx, skip_confirmation):
+    is_uninstall_fail = False
     plugins = config.get_config("plugins") or []
     for plugin in plugins:
         module_name, package_name = extract_module_and_package_install_names(plugin)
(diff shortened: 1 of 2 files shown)


src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,40 @@ def install(cli_ctx, skip_confirmation):
...
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
+            )
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been added.")
+            else:
+                cli_ctx.logger.error(f"Failed to add '{package_name}'.")
+                sys.exit(1)
```
unparalleled-js, line 178: We should not exit if a single plugin does not install

src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,40 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
+            )
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been added.")
```
unparalleled-js, line 175: Dont need to log success for every plugin installed

src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,40 @@ def install(cli_ctx, skip_confirmation):
...
+@ape_cli_context()
+@skip_confirmation_option("Don't ask for confirmation to install the plugins")
+def uninstall(cli_ctx, skip_confirmation):
+    plugins = config.get_config("plugins") or []
+    for plugin in plugins:
+        module_name, package_name = extract_module_and_package_install_names(plugin)
+        if is_plugin_installed(module_name) and (
+            module_name in github_client.available_plugins
+            or skip_confirmation
+            or click.confirm(f"Uninstall unknown 3rd party plugin '{package_name}'?")
```
unparalleled-js, line 191: Probably don't need to have this `click.confirm` here - uninstalling is less risky than installing

src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,40 @@ def install(cli_ctx, skip_confirmation):
...
+            else:
+                cli_ctx.logger.error(f"Failed to add '{package_name}'.")
+                sys.exit(1)
+
+
+@cli.command(short_help="Uninstall all plugins in the local config file")
+@ape_cli_context()
+@skip_confirmation_option("Don't ask for confirmation to install the plugins")
+def uninstall(cli_ctx, skip_confirmation):
+    plugins = config.get_config("plugins") or []
```
unparalleled-js, line 185: So here, we can just add a `confirmed = click.confirm(f"Uninstall the following plugins? {plugins}")`

And then just pass `-y` into the `subprocess.call`

unparalleled-js: Ah nevermind, let's not do thhis

Ninjagod1251: I would need a walkthrough on how to prompt the user and let them pass in a -y

src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,38 @@ def install(cli_ctx, skip_confirmation):
...
+                [sys.executable, "-m", "pip", "uninstall", "--quiet", f"{package_name}"]
+            )
+            is_uninstall_fail = False
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been removed.")
+            else:
+                cli_ctx.logger.error(f"Failed to remove '{package_name}'.")
+                is_uninstall_fail = True
+            if is_uninstall_fail is True:
+                sys.exit(1)
```
unparalleled-js, line 202: Should be outside of the for loop

Also you just do `if is_uninstall_fail` (don't have to `is True`)

src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,38 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
+            )
+            if result != 0:
```
unparalleled-js, line 174: We don't want to exit per plugin, we want to exit at the very end (outside the for loop) with status of 1 if any of them have failed

src/ape_plugins/_cli.py
```diff
@@ -168,7 +169,45 @@ def install(cli_ctx, skip_confirmation):
...
+                cli_ctx.logger.error(f"Failed to add '{package_name}'.")
+                is_insall_fail = True
+    if is_insall_fail:
+        sys.exit(1)
+    else:
+        sys.exit(0)
```
unparalleled-js, line 181: You don't need to do `sys.exit(0)`, that happens by default

src/ape_plugins/_cli.py
```diff
@@ -156,6 +156,7 @@ def add(cli_ctx, plugin, version, skip_confirmation):
 @ape_cli_context()
 @skip_confirmation_option("Don't ask for confirmation to install the plugins")
 def install(cli_ctx, skip_confirmation):
+    is_insall_fail = False
```
unparalleled-js, line 159: Can we rename this to something like `install_failures_occurred` or `failures_occurred`.

It should kind-of read like english

src/ape_plugins/_cli.py
```diff
@@ -168,7 +169,62 @@ def install(cli_ctx, skip_confirmation):
...
+            any_uninstall_failed = True
+
+        module_name, package_name = extract_module_and_package_install_names(plugin)
+        # check for installed check the config.yaml
+        if not is_plugin_installed(module_name):
+            cli_ctx.logger.warning(f"Plugin '{plugin}' is not installed.")
+            any_uninstall_failed = True
+
+        # if plugin is installed but not a 2nd class. It must be a third party
+        if module_name not in github_client.available_plugins and is_plugin_installed(plugin):
```
unparalleled-js, line 203: I think we should make this condition happen first? That way, we don't both checking if the plugin is installed or not if is not an available plugin. And then, make this `if` an `elif`.

src/ape_plugins/_cli.py
```diff
@@ -168,7 +169,62 @@ def install(cli_ctx, skip_confirmation):
...
+            cli_ctx.logger.warning(
+                f"Plugin '{module_name}' is installed but not in available plugins."
+                f" Please uninstall outside of Ape."
+            )
+            any_uninstall_failed = True
+
+        # if plugin is installed and 2nd class. We should uninstall it
+        if is_plugin_installed(module_name) and (
```
unparalleled-js, line 211: We are calling `is_plugin_installed` twice.
We are also checking if the `module_name` is in the available plugins list twice.

Can we just store the result of those checks as `boolean` variables and re-use them?

src/ape_plugins/_cli.py
```diff
@@ -168,7 +183,58 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+            result = subprocess.call(
+                [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
```
unparalleled-js, line 187: If the user has said to skip confirmation, we should pass in `"-y"` here

src/ape_plugins/_cli.py
```diff
@@ -168,7 +183,58 @@ def install(cli_ctx, skip_confirmation):
...
+            any_uninstall_failed = True
+            pass
+
+        # if plugin is installed and 2nd class. We should uninstall it
+        if installed_plugin and (available_plugin or skip_confirmation):
+            cli_ctx.logger.info(f"Uninstalling {package_name}...")
+            # NOTE: Be *extremely careful* with this command, as it modifies the user's
+            #       installed packages, to potentially catastrophic results
+            # NOTE: This is not abstracted into another function *on purpose*
+            result = subprocess.call(
```
unparalleled-js, line 228: If the user has passed in to skip confirmation, we should pass in `"-y"` here!

src/ape_plugins/_cli.py
```diff
@@ -168,7 +183,58 @@ def install(cli_ctx, skip_confirmation):
...
+                [sys.executable, "-m", "pip", "uninstall", "--quiet", f"{package_name}"]
+            )
+            if result == 0:
+                cli_ctx.logger.success(f"Plugin '{package_name}' has been removed.")
```
unparalleled-js, line 232: So since we can't tell in some situations that the packages was actually removed, causing this success message to be wrong, we can just check that if the plugin is installed and only echo success based on that.

src/ape_plugins/_cli.py
```diff
@@ -168,7 +183,58 @@ def install(cli_ctx, skip_confirmation):
...
+            any_uninstall_failed = True
+            pass
+        # check for installed check the config.yaml
+        elif not installed_plugin:
+            cli_ctx.logger.warning(f"Plugin '{module_name}' is not installed.")
+            any_uninstall_failed = True
+            pass
+
+        # if plugin is installed and 2nd class. We should uninstall it
+        if installed_plugin and (available_plugin or skip_confirmation):
```
unparalleled-js, line 223: I don't understand why `skip_confirmation` is here ... That seems wrong to me.

unparalleled-js: same with the `install` command

src/ape_plugins/_cli.py
```diff
@@ -168,7 +187,70 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-            subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"])
+
+            args = [sys.executable, "-m", "pip", "install", "--quiet", f"{package_name}"]
```
unparalleled-js, line 191:
```suggestion
            args = [sys.executable, "-m", "pip", "install", "--quiet", package_name]
```

src/ape_plugins/_cli.py
```diff
@@ -168,7 +187,70 @@ def install(cli_ctx, skip_confirmation):
...
+            result = subprocess.call(args)
+            plugin_got_installed = is_plugin_installed(module_name)
+            if result != 0 and not plugin_got_installed:
```
unparalleled-js, line 194: Put back success log?

src/ape_plugins/_cli.py
```diff
@@ -168,7 +187,70 @@ def install(cli_ctx, skip_confirmation):
...
+            #       installed packages, to potentially catastrophic results
+            # NOTE: This is not abstracted into another function *on purpose*
+
+            args = [sys.executable, "-m", "pip", "uninstall", "--quiet"]
+            if skip_confirmation:
+                args.append("-y")
+            args.append(package_name)
+
+            result = subprocess.call(args)
+            plugin_still_installed = is_plugin_installed(module_name)
```
unparalleled-js, line 244: do same check for `add` and `remove` ?

unparalleled-js (approved):

PR #385, opened 2022-01-04
feat: chain manager

unparalleled-js:
### What I did

fixes: #129
^ Note there are several other related to chain manager, such adding mining support. This work is just to initialize the chain manager as well as swap out the snapshotting login in `ape_test` using the chain manager rather than the `ProviderAPI` alone.

### How I did it

* Create new class ChainManager, a top-level object that takes the `NetworksManager`.
* The chain manager raises an error when not connected to a provider.
* The chain manager keeps track of a "snapshot stack"

### How to verify it

See the new test module `tests/integration/test_chain.py`.

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -54,7 +54,7 @@ Useful for development purposes, such as controlling the state of the blockchain
 Also handy for querying data about the chain and managing local caches.
 """
 
-accounts = _AccountManager(config, _converters, plugin_manager, networks, chain)  # type: ignore
+accounts = _AccountManager(config, _converters, plugin_manager, networks)  # type: ignore
 """Manages accounts for the current project. See :class:`ape.managers.accounts.AccountManager`."""
 
 Project = _partial(
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -47,7 +47,14 @@ networks = _NetworkManager(config, plugin_manager)  # type: ignore
 
 _converters = _ConversionManager(config, plugin_manager, networks)  # type: ignore
 
-accounts = _AccountManager(config, _converters, plugin_manager, networks)  # type: ignore
+chain = _ChainManager(networks)  # type: ignore
+"""
+The current connected blockchain; requires an active provider.
+Useful for development purposes, such as controlling the state of the blockchain.
+Also handy for querying data about the chain and managing local caches.
+"""
+
+accounts = _AccountManager(config, _converters, plugin_manager, networks, chain)  # type: ignore
 """Manages accounts for the current project. See :class:`ape.managers.accounts.AccountManager`."""
 
 Project = _partial(
@@ -68,13 +75,6 @@ Contract = _partial(_Contract, networks=networks, converters=_converters)
 
 convert = _converters.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
-
-chain = _ChainManager(networks)  # type: ignore
-"""
-The current connected blockchain; requires an active provider.
-Useful for development purposes, such as controlling the state of the blockchain.
-Also handy for querying data about the chain and managing local caches.
-"""
 
 
 __all__ = [
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -44,9 +44,6 @@ class AccountAPI(AddressAPI):
     def alias(self) -> Optional[str]:
         """
         A shortened-name for quicker access to the account.
-
-        Returns:
-            str (optional)
         """
         return None
 
@@ -56,7 +53,7 @@ class AccountAPI(AddressAPI):
         Sign a message.
 
         Args:
-          msg (SignableMessage): The message to sign.
+          msg (:class:`~ape.types.signatures.SignableMessage`): The message to sign.
             See these
             `docs <https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.messages.SignableMessage>`__  # noqa: E501
             for more type information on this type.
@@ -137,9 +134,7 @@ class AccountAPI(AddressAPI):
         if not txn.signature:
             raise SignatureError("The transaction was not signed.")
 
-        receipt = self.provider.send_transaction(txn)
-        self.chain.account_history.append(receipt)
-        return receipt
+        return self.provider.send_transaction(txn)
 
     @cached_property
     def _convert(self) -> Callable:
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -137,7 +137,9 @@ class AccountAPI(AddressAPI):
         if not txn.signature:
             raise SignatureError("The transaction was not signed.")
 
-        return self.provider.send_transaction(txn)
+        receipt = self.provider.send_transaction(txn)
+        self._chain.account_history.append_transaction(txn)
+        return receipt
 
     @cached_property
     def _convert(self) -> Callable:
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -39,6 +39,7 @@ class AddressAPI:
     def provider(self, value: ProviderAPI):
         """
         Set the active provider if connected to one.
+
         Args:
             value (:class:`~ape.api.providers.ProviderAPI`): The provider to set.
         """
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -1,9 +1,9 @@
 from typing import List, Optional
 
+from ape.exceptions import AddressError
 from ape.types import AddressType
 from ape.utils import abstractdataclass, abstractmethod
 
-from ..exceptions import AddressError
 from .providers import ProviderAPI
 
 
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -35,6 +35,16 @@ class AddressAPI:
 
         return self._provider
 
+    @provider.setter
+    def provider(self, value: ProviderAPI):
+        """
+        Set the active provider if connected to one.
+        Args:
+            value (:class:`~ape.api.providers.ProviderAPI`): The provider to set.
+        """
+
+        self._provider = value
+
     @property
     @abstractmethod
     def address(self) -> AddressType:
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -1,10 +1,12 @@
-from typing import List, Optional
+from typing import TYPE_CHECKING, List, Optional
 
-from ape.exceptions import AddressError
 from ape.types import AddressType
 from ape.utils import abstractdataclass, abstractmethod
 
 from .providers import ProviderAPI
+
+if TYPE_CHECKING:
+    from ape.managers.chain import ChainManager
 
 
 @abstractdataclass
@@ -13,7 +15,7 @@ class AddressAPI:
     A base address API class. All account-types subclass this type.
     """
 
-    _provider: Optional[ProviderAPI] = None
+    _chain: Optional["ChainManager"] = None
 
     @property
     def provider(self) -> ProviderAPI:
@@ -28,23 +30,7 @@ class AddressAPI:
             :class:`~ape.api.providers.ProviderAPI`
         """
 
-        if not self._provider:
-            raise AddressError(
-                f"Incorrectly implemented provider API for class {type(self).__name__}"
-            )
-
-        return self._provider
-
-    @provider.setter
-    def provider(self, value: ProviderAPI):
-        """
-        Set the active provider if connected to one.
-
-        Args:
-            value (:class:`~ape.api.providers.ProviderAPI`): The provider to set.
-        """
-
-        self._provider = value
+        return self._chain.provider
 
     @property
     @abstractmethod
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -19,9 +19,11 @@ class AddressAPI:
     def provider(self) -> ProviderAPI:
         """
         The current active provider if connected to one.
+
         Raises:
             :class:`~ape.exceptions.AddressError`: When there is no active
                provider at runtime.
+
         Returns:
             :class:`~ape.api.providers.ProviderAPI`
         """
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -1,13 +1,10 @@
-from typing import TYPE_CHECKING, List, Optional
+from typing import List, Optional
 
 from ape.types import AddressType
 from ape.utils import abstractdataclass, abstractmethod
 
 from ..exceptions import AddressError
 from .providers import ProviderAPI
-
-if TYPE_CHECKING:
-    from ape.managers.chain import ChainManager
 
 
 @abstractdataclass
@@ -16,34 +13,25 @@ class AddressAPI:
     A base address API class. All account-types subclass this type.
     """
 
-    _chain: Optional["ChainManager"] = None
-
-    @property
-    def chain(self):
-        """
-        The current active blockchain, if connected to one.
-
-        Raises:
-            :class:`~ape.exceptions.AddressError`: When there is no active
-               chain at runtime.
-        """
-
-        if not self._chain:
-            raise AddressError(
-                f"Incorrectly implemented provider API for class {type(self).__name__}. "
-                "Provider missing access to chain manager."
-            )
+    _provider: Optional[ProviderAPI] = None
 
     @property
     def provider(self) -> ProviderAPI:
         """
         The current active provider if connected to one.
-
         Raises:
             :class:`~ape.exceptions.AddressError`: When there is no active
                provider at runtime.
+        Returns:
+            :class:`~ape.api.providers.ProviderAPI`
         """
-        return self.chain.provider
+
+        if not self._provider:
+            raise AddressError(
+                f"Incorrectly implemented provider API for class {type(self).__name__}"
+            )
+
+        return self._provider
 
     @property
     @abstractmethod
--- a/src/ape/api/explorers.py
+++ b/src/ape/api/explorers.py
@@ -1,10 +1,10 @@
 from typing import Iterator, Optional
 
-from ape.api import ReceiptAPI
 from ape.types import AddressType, ContractType
 from ape.utils import abstractdataclass, abstractmethod
 
 from . import networks
+from .providers import ReceiptAPI
 
 
 @abstractdataclass
--- a/src/ape/api/explorers.py
+++ b/src/ape/api/explorers.py
@@ -1,5 +1,6 @@
-from typing import Optional
+from typing import Iterator, Optional
 
+from ape.api import ReceiptAPI
 from ape.types import AddressType, ContractType
 from ape.utils import abstractdataclass, abstractmethod
 
@@ -48,8 +49,20 @@ class ExplorerAPI:
         Get the contract type for a given address if it has been published in an explorer.
 
         Args:
-            address (str): The contract address.
+            address (:class:`~ape.types.AddressType`): The contract address.
 
         Returns:
-            :class:`~ape.contracts.ContractType` if published, else ``None``.
+            Optional[:class:`~ape.contracts.ContractType`]: If not published, returns ``None``.
+        """
+
+    @abstractmethod
+    def get_account_transactions(self, address: AddressType) -> Iterator[ReceiptAPI]:
+        """
+        Get a list of list of transactions performed by an address.
+
+        Args:
+            address (:class:`~ape.types.AddressType`): The account address.
+
+        Returns:
+            Iterator[:class:`~ape.api.providers.ReceiptAPI`]
         """
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -345,6 +345,15 @@ class NetworkAPI:
 
         return self.config_manager.get_config(self.ecosystem.name)
 
+    @cached_property
+    def network_config(self) -> ConfigItem:
+        """
+        The configuration of the network. See :class:`~ape.managers.config.ConfigManager`
+        for more information on plugin configurations.
+        """
+
+        return self.config.get(self.name, {})  # type: ignore
+
     @property
     def chain_id(self) -> int:
         """
@@ -391,11 +400,25 @@ class NetworkAPI:
         Returns:
             int
         """
-        try:
-            return self.config[self.name]["required_confirmations"]
-        except KeyError:
-            # Is likely a 'development' network.
-            return 0
+        return self.network_config.get("required_confirmations", 0)  # type: ignore
+
+    @property
+    def approximate_block_time(self) -> int:
+        """
+        The approximate amount of time it takes for a new block to get mined to the chain.
+        Configure in your ``ape-config.yaml`` file via.
+
+        Config example::
+
+            ethereum:
+              mainnet:
+                approximxate_block_time: 15
+
+        Returns:
+            int
+        """
+
+        return self.network_config.get("approximate_block_time", 0)  # type: ignore
 
     @cached_property
     def explorer(self) -> Optional["ExplorerAPI"]:
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -346,12 +346,7 @@ class NetworkAPI:
         return self.config_manager.get_config(self.ecosystem.name)
 
     @cached_property
-    def network_config(self) -> ConfigItem:
-        """
-        The configuration of the network. See :class:`~ape.managers.config.ConfigManager`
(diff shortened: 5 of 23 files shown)


src/ape/types/__init__.py
```diff
@@ -21,6 +21,14 @@
 as well as a block number or hash (HexBytes).
 """
 
+SnapshotID = Union[str, int, bytes]
+"""
+An ID representing a point in time on a blockchain, as used in the
+:meth:`~ape.managers.chain.ChainManager.snapshot` and
+:meth:`~ape.managers.chain.ChainManager.snapshot` methods. Can be a ``str`` or  an ``int``.
```
unparalleled-js, line 28:
```suggestion
:meth:`~ape.managers.chain.ChainManager.snapshot` methods. Can be a ``str``, ``int``, or ``bytes``.
```

src/ape/__init__.py
```diff
@@ -68,8 +69,17 @@
 convert = _converters.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
 
+chain = _ChainManager(networks)  # type: ignore
+"""
+The current connected blockchain; requires an active provider.
+Useful for development purposes, such as controlling the state of the blockchain.
+Also handy for querying data and managing local caches.
```
fubuloubu, line 76:
```suggestion
Also handy for querying data about the chain and managing local caches.
```

src/ape/exceptions.py
```diff
@@ -1,4 +1,9 @@
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
+
+from eth_utils import humanize_hash
```
fubuloubu, line 3: hmm... probably could use this a lot more in the codebase

src/ape/types/__init__.py
```diff
@@ -21,6 +21,14 @@
 as well as a block number or hash (HexBytes).
 """
 
+SnapshotID = Union[str, int, bytes]
+"""
+An ID representing a point in time on a blockchain, as used in the
+:meth:`~ape.managers.chain.ChainManager.snapshot` and
+:meth:`~ape.managers.chain.ChainManager.snapshot` methods. Can be a ``str``, ``int``, or ``bytes``.
+Providers will expect and handle snapshot IDs differently.
```
fubuloubu, line 29: Add "There shouldn't be a need to change providers when using this feature, so there should not be confusion over this type in practical use cases"

src/ape/managers/chain.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            snapshot_index = self._snapshots.index(snapshot_id)
+            self._snapshots = self._snapshots[:snapshot_index]
+
+        provider.revert(snapshot_id)
+
+    def _get_test_provider(self) -> TestProviderAPI:
+        provider = self.provider
+        if not isinstance(provider, TestProviderAPI):
+            raise NotImplementedError(f"Provider '{provider.name}' does not support snapshotting.")
```
fubuloubu, line 101: Hmm, I was picturing letting the lower-level error from `ProviderAPI` bubble up where `NotImplementedError` just tells you right when you use the feature that it's unsupported.

src/ape_test/fixtures.py
```diff
@@ -3,30 +3,36 @@
...
-
-        if active_provider is None:
-            raise ProviderError("Provider is not set.")
+    def chain(self) -> ChainManager:
+        return self._chain
 
-        return active_provider
+    @pytest.fixture
+    def provider(self) -> ProviderAPI:
+        return self._chain.provider
```
fubuloubu, line 35: Probably don't even need this fixture anymore with `chain`

fubuloubu (commented):
Really excited about this one!

src/ape/managers/chain.py
unparalleled-js: `ProviderAPI` does not have development methods though, so it would be like `method not found ` or something.

And `geth` - which _is_ a `TestProviderAPI` (has a `snapshot()` method) will raise it's `NotImplementedError` and that will bubble up normally.

src/ape/exceptions.py
unparalleled-js: Yeah, TIL

tests/integration/conftest.py
```diff
@@ -1,9 +1,42 @@
...
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_protocol(item, nextitem):
+    snapshot_id = chain.snapshot()
+    yield
+
+    try:
+        chain.restore(snapshot_id)
+    except (HeaderNotFound, ChainError):
```
unparalleled-js, line 21: This seems like a bug and I am hoping to figure this out more and remove this part...

The tests work locally but I am having troubles in CI.

src/ape/managers/chain.py
unparalleled-js: The only one someone could get here is if there were using `infura` or something. They'd be having other issues most likely if that were the case anyway. The real `NotImplentedError` _do_ bubble, such as with `geth`.

src/ape/managers/chain.py
```diff
@@ -0,0 +1,143 @@
...
+
+        latest_block = chain.blocks[-1
+
+    """
+
+    _provider: ProviderAPI
+
+    def __getitem__(self, block_number: int) -> BlockAPI:
+        """
+        Get a block by number.
```
fubuloubu, line 29: Probably explain the behavior of a negative number here (e.g. -1 is head/chain tip, -2 is one behind, etc.)

src/ape/managers/chain.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass
+class BlockContainer:
+    """
+    A list of blocks on the chain.
+
+    Usages example::
+
+        from ape import chain
+
+        latest_block = chain.blocks[-1
```
fubuloubu, line 21:
```suggestion
        latest_block = chain.blocks[-1]
```

src/ape/managers/chain.py
```diff
@@ -0,0 +1,143 @@
...
+class ChainManager:
+    """
+    A manager for controlling and manipulating development blockchains
+    and querying chain-data.
+    """
+
+    _networks: NetworkManager
+    _snapshots: List[SnapshotID] = []
+
+    blocks: BlockContainer = None  # type: ignore
```
fubuloubu, line 64: do you even have to put this here with `__post_init__`?
```suggestion
    blocks: BlockContainer  # type: ignore
```

src/ape/managers/chain.py
```diff
@@ -0,0 +1,143 @@
...
+    """The list of blocks on the chain."""
+
+    def __post_init__(self):
+        self.blocks = BlockContainer(self._networks.active_provider)
+
+    @property
+    def provider(self) -> ProviderAPI:
+        """
+        The active :class:`~ape.api.providers.ProviderAPI`. Most often, this should
+        be a :class:`~ape.api.providers.TestProviderAPI`.
```
fubuloubu, line 74: Not necessarily true, you would use this in scripts and querying public chains too
```suggestion
        The active :class:`~ape.api.providers.ProviderAPI`.
```

src/ape/managers/chain.py
fubuloubu: Ah. Well just saying that "chain does not support X feature" will probably grow to more than just snapshots.

src/ape/managers/chain.py
unparalleled-js: Yeah it seems like it does

src/ape/managers/chain.py
```diff
@@ -0,0 +1,144 @@
...
+
+    """
+
+    _provider: ProviderAPI
+
+    def __getitem__(self, block_number: int) -> BlockAPI:
+        """
+        Get a block by number. Negative numbers start at the chain head and
+        move backwards. For example, ``-1`` would be the latest block and
+        ``-2`` would be the penultimate block, and so on.
```
fubuloubu, line 31: Penultimate is a little too fancy for some readers lol
```suggestion
        ``-2`` would be the block prior to that one, and so on.
```

fubuloubu (approved):

src/ape/managers/chain.py
unparalleled-js: ya penultimate almost sounds like it is the second to last block EVER too

src/ape/managers/chain.py
unparalleled-js: Yes, this is just a helper method I made, I can make it more subject to change though by allowing a custom message given to it.

fubuloubu (dismissed):

src/ape/exceptions.py
```diff
@@ -184,3 +198,22 @@ class AddressError(ApeException):
...
+    """
+    Raised when given an unknown snapshot ID.
+    """
+
+    def __init__(self, snapshot_id: "SnapshotID"):
+        if isinstance(snapshot_id, bytes):
+            # Is block hash
+            snapshot_id = humanize_hash(snapshot_id)  # type: ignore
+
+        super().__init__(f"Unknown snapshot ID '{str(snapshot_id)}'.")
```
unparalleled-js, line 219:
```suggestion
        super().__init__(f"Unknown snapshot ID '{snapshot_id}'.")
```

unparalleled-js: needed for mypy

unparalleled-js:
- [ ] Add test showing we can change the networks in chain manager
- [ ] pending_timestamp()
- [ ] history - get history per account()
- [ ] chain_id
- [ ] gas_price
- [ ] base_fee

src/ape/api/address.py
```diff
@@ -1,59 +1,55 @@
...
     """
     A base address API class. All account-types subclass this type.
     """
 
-    _provider: Optional[ProviderAPI] = None
+    _chain: Optional["ChainManager"] = None
 
     @property
-    def provider(self) -> ProviderAPI:
+    def chain(self):
```
fubuloubu, line 22:
```suggestion
    def chain(self) -> "ChainManager":
```

src/ape/api/address.py
```diff
@@ -1,59 +1,55 @@
...
+               chain at runtime.
         """
 
-        if not self._provider:
+        if not self._chain:
             raise AddressError(
-                f"Incorrectly implemented provider API for class {type(self).__name__}"
+                f"Incorrectly implemented provider API for class {type(self).__name__}. "
+                "Provider missing access to chain manager."
             )
```
fubuloubu, line 35: Nothing is returned here?

src/ape/utils.py
```diff
@@ -479,11 +479,19 @@ class AbstractDataClassMeta(DataClassMeta, ABCMeta):
 errors will occur. This class cannot be instantiated on its own.
 """
 
+
+def convert(*args, **kwargs):
+    from ape import convert
+
+    return convert(*args, **kwargs)
+
+
```
fubuloubu, line 488: I don't know if I like this being here, since you have to load a dependency from the top

src/ape/api/address.py
unparalleled-js: yeah sorry, this is very WIP right now

src/ape/api/accounts.py
```diff
@@ -137,7 +137,9 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
         if not txn.signature:
             raise SignatureError("The transaction was not signed.")
 
-        return self.provider.send_transaction(txn)
+        receipt = self.provider.send_transaction(txn)
+        self._chain.account_history.append_transaction(txn)
```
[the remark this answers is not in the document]
unparalleled-js: Yeah, this is another circular dependency

src/ape/api/address.py
```diff
@@ -1,19 +1,21 @@
...
 class AddressAPI:
     """
     A base address API class. All account-types subclass this type.
     """
 
-    _provider: Optional[ProviderAPI] = None
+    _chain: Optional["ChainManager"] = None
 
     @property
     def provider(self) -> ProviderAPI:
```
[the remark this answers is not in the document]
unparalleled-js: Ah yes, okay I just put the chain stuff only in `AccountAPI` and left `AddressAPI` alone for now

src/ape/api/address.py
unparalleled-js: outdated now!

src/ape/utils.py
unparalleled-js: removed, I only needed this in `chain`, so I just moved it there

src/ape/api/address.py
```diff
@@ -148,9 +132,6 @@ def codesize(self) -> int:
     def is_contract(self) -> bool:
         """
         ``True`` when there is code associated with the address.
-
-        Returns:
```
unparalleled-js, line 152: `Returns` looks back for `property` docs, makes it look like you have to call it

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+    Usages example::
+
+        from ape import chain
+
+        latest_block = chain.blocks[-1]
+
+    """
+
+    networks: NetworkManager
+    _poll_wait_interval = 2  # Seconds
```
fubuloubu, line 28: Config?

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+        return self._get_block("latest")
+
+    @property
+    def height(self) -> int:
+        """
+        The latest block number.
+        """
+
+        return self.head.number
+
```
fubuloubu, line 82: Can we move these above `__getitem__`?

fubuloubu: Also `_get_block` and `_provider`, etc.

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+    @property
+    def _provider(self) -> ProviderAPI:
+        if not self.networks.active_provider:
+            raise ProviderNotConnectedError()
+
+        return self.networks.active_provider
+
+    @property
+    def _network_confirmations(self) -> int:
```
fubuloubu, line 91: Should this be a private property?

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+class AccountHistory:
+    """
+    A container mapping account addresses to the transaction from the active session.
+    """
+
+    _map: Dict[AddressType, List[ReceiptAPI]] = {}
+
+    def __getitem__(self, address: Union[AddressAPI, AddressType, str]) -> List[ReceiptAPI]:
```
fubuloubu, line 174: Can this method query `ProviderAPI` (via `ExplorerAPI` to get transactions indexed by account's nonce via TBD query method) to get the set of `ReceiptAPI` from those transactions?

fubuloubu: Obviously, this query method would not function for chains without an Explorer (or where it is not configured) since it is a non-trivial problem to identify an account's past transactions without indexing them.

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+              to a provider.
+        """
+
+        provider = self._networks.active_provider
+        if not provider:
+            raise ProviderNotConnectedError()
+
+        return provider
+
+    @property
```
fubuloubu, line 285: `cached_property`? so as not to continue initializing this class every time it is called

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+                f"Provider '{provider.name}' does not support method '{method_name}'."
+            )
+
+        return provider
+
+
+def _convert(*args, **kwargs):
+    from ape import convert
+
+    return convert(*args, **kwargs)
```
fubuloubu, line 400: Can this be a `cached_property` on `AccountHistory`?

unparalleled-js:
- [ ] I am still working on `pending_timestamp`

src/ape/managers/chain.py
fubuloubu: Or just set it inside of `__post_init__`

src/ape/managers/chain.py
```diff
@@ -0,0 +1,360 @@
...
+    @property
+    def chain_id(self) -> int:
+        """
+        The blockchain ID.
+        See `ChainList <https://chainlist.org/>`__ for a comprehensive list of IDs.
+        """
+
+        network_name = self.provider.network.name
+        if network_name not in self._chain_id_map:
+            self._chain_id_map[network_name] = self.provider.chain_id
```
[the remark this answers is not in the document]
unparalleled-js: yeah there is a ticket for that!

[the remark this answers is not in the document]
unparalleled-js: This is more a caching trick right here.
This way, we only make the request to get the chain ID once per network we connect to.

src/ape/managers/chain.py
unparalleled-js: It is cached by network right now.
So if we have already made the request, the ID is already present in `self._chain_id_map`. Otherwise, we make the provider call. It is like `@cached_proeprty` but we are controlling more how we cache

src/ape/managers/chain.py
unparalleled-js: Yes!

But, I am pretty sure lazy imports are automatically cached. At least it seems like it

fubuloubu: Oh yes, that is true. `importlib` caches imports

src/ape/managers/chain.py
unparalleled-js: Private like `__network_confirmations` ? I tend to only use those in extreme cases, like `__lock` or `__private_key`.

However, maybe you are suggesting to make this public? In which case, yeah I agree, I just pushed that change. Please let me know if you meant something else

src/ape/api/explorers.py
```diff
@@ -48,8 +49,20 @@ def get_contract_type(self, address: AddressType) -> Optional[ContractType]:
...
-            address (str): The contract address.
+            address (:class:`~ape.types.AddressType`): The contract address.
 
         Returns:
-            :class:`~ape.contracts.ContractType` if published, else ``None``.
+            Optional[:class:`~ape.contracts.ContractType`]: If not published, returns ``None``.
+        """
+
+    @abstractmethod
+    def get_account_transactions(self, address: AddressType) -> Iterator[ReceiptAPI]:
```
unparalleled-js, line 59: Making this an `Iterator` because who the heck knows how many transactions one has.
In the `ape-etherscan`, we have to paginate through requests (the API token is going to be necessary, so that is next after https://github.com/ApeWorX/ape-etherscan/pull/12).

src/ape/managers/chain.py
unparalleled-js: how about we just make this is a param on the `poll` method and let it default to 2?

src/ape/managers/chain.py
unparalleled-js: Done!

src/ape/managers/chain.py
unparalleled-js: Done!

src/ape/managers/chain.py
unparalleled-js: I made it `public` ? Let me know if that is not what you meant!

src/ape/api/explorers.py
fubuloubu: Like this, can handle pagination via iterator to smooth it out

src/ape/managers/chain.py
```diff
@@ -0,0 +1,424 @@
...
+    def poll_blocks(
+        self,
+        start: Optional[int] = None,
+        required_confirmations: Optional[int] = None,
+        poll_wait_interval: int = 2,
+    ) -> Iterator[BlockAPI]:
+        """
+        Poll new blocks. Optionally set a start block to include historical blocks.
+        **NOTE**: This is a deamon method; it does not terminate unless an exception
+        occurrs.
```
fubuloubu, line 138: typo

src/ape/managers/chain.py
```diff
@@ -0,0 +1,424 @@
...
+        Usage example::
+
+            from ape import chain
+
+            for new_block in chain.blocks.poll_blocks():
+                print(f"New block found: number={new_block.number}")
+
+        Args:
+            start (Optional[int]): The block number to start with. Defaults to the block
+              of the ``+1`` the current block number.
```
fubuloubu, line 149:
```suggestion
              of the ``+1`` the current block number when this method was called.
```

src/ape/managers/chain.py
```diff
@@ -0,0 +1,424 @@
...
+            required_confirmations (Optional[int]): The amount of confirmations to wait
+              before yielding the block.
```
fubuloubu, line 151:
```suggestion
              before yielding the block. The more confirmations, the less likely a reorg will occur.
```

src/ape/managers/chain.py
```diff
@@ -0,0 +1,424 @@
...
+            poll_wait_interval (int): The amount of seconds to wait before checking for
+              new blocks. Defaults to ``2``.
```
fubuloubu, line 153: can probably default to a portion of expected block time (which is different chain!)

src/ape/managers/chain.py
```diff
@@ -0,0 +1,424 @@
...
+        if network_name not in self._chain_id_map:
+            self._chain_id_map[network_name] = self.provider.chain_id
+
+        return self._chain_id_map[network_name]
+
+    @property
+    def gas_price(self) -> int:
+        """
+        The price for what it costs to transact
+        (pre-`EIP-1559 <https://eips.ethereum.org/EIPS/eip-1559>`__).
```
fubuloubu, line 335: Post EIP-1559 I tihnk this is just some sort of summary of base_fee + miner tip

fubuloubu (approved):
Looking good!

src/ape/managers/chain.py
unparalleled-js: Oh that is smart

src/ape/managers/chain.py
unparalleled-js: ?

unparalleled-js: Oh lol I see it now

unparalleled-js:
Testing the polling feature! Looks like we are getting all the blocks!

Screen Shot 2022-01-07 at 15 32 53

unparalleled-js:
CLI instant new-block event server:

```
@cli.command(cls=NetworkBoundCommand)
@network_option()
def poll_blocks(network):
    for new_block in chain.blocks.poll_blocks():
        click.echo(
            f"New block found: number={new_block.number}, timestamp={new_block.timestamp}, size={new_block.size}"
        )
```

src/ape/api/networks.py
```diff
@@ -391,11 +395,25 @@ def required_confirmations(self) -> int:
...
+    def approximate_block_time(self) -> int:
+        """
+        The approximate amount of time it takes for a new block to get mined to the chain.
+        Configure in your ``ape-config.yaml`` file.
+
+        Config example::
+
+            ethereum:
+              mainnet:
+                approximxate_block_time: 15
```
fubuloubu, line 410: would `block_time` be just as right? these things are parameters that different chains target

src/ape_ethereum/ecosystem.py
```diff
@@ -40,17 +40,26 @@
...
+        required_confirmations=12, approximate_block_time=15
+    )  # type: ignore
+    kovan: NetworkConfig = NetworkConfig(
+        required_confirmations=3, approximate_block_time=4
+    )  # type: ignore
+    rinkeby: NetworkConfig = NetworkConfig(
+        required_confirmations=3, approximate_block_time=15
+    )  # type: ignore
+    goerli: NetworkConfig = NetworkConfig(
+        required_confirmations=10, approximate_block_time=15
```
fubuloubu, line 60: all PoA networks can be 2 confs, they don't ever reorg

fubuloubu (approved):

src/ape/api/networks.py
unparalleled-js: Okay, yeah if that is something generally understood.

src/ape/api/networks.py
```diff
@@ -391,11 +395,25 @@ def required_confirmations(self) -> int:
...
+    def block_time(self) -> int:
+        """
+        The approximate amount of time it takes for a new block to get mined to the chain.
+        Configure in your ``ape-config.yaml`` file.
+
+        Config example::
+
+            ethereum:
+              mainnet:
+                approximxate_block_time: 15
```
fubuloubu, line 410: Oopsie!
```suggestion
                block_time: 15
```

fubuloubu (approved):

PR #386, opened 2022-01-04
fix: bug preventing str-type when specifying transaction types.

unparalleled-js:
### What I did

Fixes regression where we were not longer able to specify `type="0x0"` for static fee transactions.
Adds Transactions guide.

### How I did it

handle hexbytes differently than bytes

### Checklist

### DIFF
--- a/docs/userguides/transactions.md
+++ b/docs/userguides/transactions.md
@@ -21,7 +21,7 @@ def deploy():
     return account.deploy(project.MyContract)
 ```
 
-## Dynamic Fee Transactions
+## Dynamic-Fee Transactions
 
 Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), all transactions used a `gas_price`.
 After the London fork of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`.
@@ -51,7 +51,11 @@ the value returned from the
 [ProviderAPI.base_fee](../methoddocs/api.html?highlight=accountapi#ape.api.providers.ProviderAPI.base_fee) method
 property.
 
-## Static Fee Transactions
+## Static-Fee Transactions
+
+Static-fee transactions are the transactions that Ethereum used before the London-fork
+(before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)).
+**However, some applications may still require using static-fee transactions.**
 
 One way to use a static-fee transaction is by specifying the `gas_price` as a key-value argument:
 
@@ -59,9 +63,12 @@ One way to use a static-fee transaction is by specifying the `gas_price` as a ke
 contract.fundMyContract(value="1 gwei", gas_price="100 gwei")
 ```
 
-Another way to use a static-fee transaction (without having to provide `gas_price`) is to set the key-value
+**NOTE**: Miners prioritize static-fee transactions based on the highest `gas_price`.
+
+Another way to use a static-fee transactions (without having to provide `gas_price`) is to set the key-value
 argument `type` equal to `0x00`.
 
 ```python
 contract.fundMyContract(value="1 gwei", type="0x0")
 ```
+
--- a/docs/userguides/transactions.md
+++ b/docs/userguides/transactions.md
@@ -23,7 +23,7 @@ def deploy():
 
 ## Dynamic Fee Transactions
 
-Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transaction used a `gas_price`. After the London fork
+Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transactions used a `gas_price`. After the London fork
 of Etheruem, the `gas_price` got broken up into two values `max_fee` and `max_priority_fee`. The `ape` platform supports
 both types of transactions. By default, transactions use the dynamic fee model. Making contract calls without specifying
 any additional `kwargs` will use a dynamic-fee transaction.
--- a/docs/userguides/transactions.md
+++ b/docs/userguides/transactions.md
@@ -24,7 +24,7 @@ def deploy():
 ## Dynamic Fee Transactions
 
 Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transactions used a `gas_price`. After the London fork
-of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`. The `ape` platform
+of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`. The `ape` framework
 supports both types of transactions. By default, transactions use the dynamic fee model. Making contract calls without
 specifying any additional `kwargs` will use a dynamic-fee transaction.
 
--- a/docs/userguides/transactions.md
+++ b/docs/userguides/transactions.md
@@ -46,7 +46,7 @@ The `max_priority_fee` cannot exceed the `max_fee`, as the `max_fee` includes bo
 The `max_priority_fee`, when omitted, defaults to the return value from the
 [ProviderAPI.priority_fee](../methoddocs/api.html?highlight=accountapi#ape.api.providers.ProviderAPI.priority_fee)
 method property.
-The `max_fee`, when omitted, defaults to the `priority_fee` (which get it's default applied beforehand) plus the latest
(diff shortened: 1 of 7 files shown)


docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+def deploy():
+    account = accounts.load("MyAccount")
+    return account.deploy(project.MyContract)
+```
+
+## Dynamic Fee Transactions
+
+Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transaction used a `gas_price`. After the London fork
```
unparalleled-js, line 26:
```suggestion
Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transactions used a `gas_price`. After the London fork
```

docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+Calling certain methods on a deployed-contract is a way to transact.
+
+```python
+contract = deploy()  # Example from above, that returns a contract instance.
+contract.fundMyContract(value="1 gwei")  # Assuming there is a method named 'fundMyContract' on MyContract.
+```
+
+In the example above, the call to `fundMyContract()` invokes a dynamic-fee transaction. To have more control of the 
+fee-values, you can specify the `max_fee` and the `max_priority_fee`.
```
fubuloubu, line 39: you can do just one or the other too

docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+```python
+contract = deploy()  # Example from above, that returns a contract instance.
+contract.fundMyContract(value="1 gwei", max_priorty_fee=50000000000, max_fee=100000000000)
```
fubuloubu, line 43: This should work now too:
```suggestion
contract.fundMyContract(value="1 gwei", max_priority_fee="50 gwei", max_fee="1.2 gwei")
```

docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+## Static Fee Transactions
+
+To use the original static fee transactions, you can merely specify a `type="0x0"`.
```
fubuloubu, line 48:
```suggestion
To use the original static fee transactions without having to specify `gas_price`, you just have to specify `type="0x0"`.
```

docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+```python
+contract = deploy()  # Example from above, that returns a contract instance.
+contract.fundMyContract(value="1 gwei", type="0x0")
+```
+
+Also, specifying a `gas_price` will automatically trigger the usage of a static-fee transaction.
+
+```python
+contract = deploy()  # Example from above, that returns a contract instance.
+contract.fundMyContract(value="1 gwei", gas_price=100000000000)
```
fubuloubu, line 59:
```suggestion
contract.fundMyContract(value="1 gwei", gas_price="100 gwei")
```

docs/userguides/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+## Static Fee Transactions
+
+To use the original static fee transactions, you can merely specify a `type="0x0"`.
+
+```python
+contract = deploy()  # Example from above, that returns a contract instance.
+contract.fundMyContract(value="1 gwei", type="0x0")
+```
+
+Also, specifying a `gas_price` will automatically trigger the usage of a static-fee transaction.
```
fubuloubu, line 55: This sentence being before the previous is a little confusing. One of them should change so it reads better.

fubuloubu (approved):

docs/userguides/transactions.md
```diff
@@ -0,0 +1,59 @@
...
+
+def deploy():
+    account = accounts.load("MyAccount")
+    return account.deploy(project.MyContract)
+```
+
+## Dynamic Fee Transactions
+
+Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transactions used a `gas_price`. After the London fork
+of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`. The `ape` platform
```
fubuloubu, line 27:
```suggestion
of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`. The `ape` framework
```

fubuloubu: Also, I usually put different sentences on new lines. Markdown will shove them together.

docs/userguides/transactions.md
```diff
@@ -0,0 +1,59 @@
...
+contract.fundMyContract(value="1 gwei", gas_price="100 gwei")
+```
+
+Another way to use a static-fee transaction (without having to provide `gas_price`) is to set the key-value
+argument `type` equal to `0x00`.
+
+```python
+contract.fundMyContract(value="1 gwei", type="0x0")
+```
+
```
fubuloubu, line 59: Add a note that `max_priority_fee` cannot exceed `max_fee`, and that `base_fee = max_fee - max_priority_fee`, which much match network conditions in order to be mined.

Lastly, what happens if `max_priority_fee` is omitted or `max_fee` is omitted in terms of the value of the unspecified fee?

I would think that specifying `max_priority_fee` means `max_fee = ProviderAPI.base_fee + max_priority_fee`, and that specifying `max_fee` without `max_priority_fee` means `max_priority_fee` will be some average over the last N blocks of that value, and `base_fee` is set to `max_fee - max_priority_fee` (and might take longer to confirm based on the current value of `base_fee`, which we can create a warning about)

fubuloubu (dismissed):

docs/userguides/transactions.md
fubuloubu: Also I should specify that we have to do this lmao

unparalleled-js: I think your expectations of what happens is correct. Here is where the defaults get set: https://github.com/ApeWorX/ape/blob/main/src/ape/api/accounts.py#L109-L115

* If you omit the `max_priority_fee` and and include the `max_fee`, the `max_priority_fee` just gets set from the provider.

* If you omit the `max_fee` and include the `max_priority_fee`, the `max_fee` still just becomes the `base_fee` plus the `max_priority_fee`.

As far as I know, there is no way to set the base fee. I don't know how `priority_fee` gets calculated by the provider.

docs/userguides/transactions.md
```diff
@@ -0,0 +1,67 @@
...
+def deploy():
+    account = accounts.load("MyAccount")
+    return account.deploy(project.MyContract)
+```
+
+## Dynamic Fee Transactions
+
+Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)), all transactions used a `gas_price`.
+After the London fork of Etheruem, the `gas_price` got broken up into two values, `max_fee` and `max_priority_fee`.
+The `ape` framework supports both types of transactions. By default, transactions use the dynamic fee model.
```
unparalleled-js, line 28:
```suggestion
The `ape` framework supports both types of transactions. By default, transactions use the dynamic-fee model.
```

docs/userguides/transactions.md
fubuloubu: > As far as I know, there is no way to set the base fee.

That is correct, it is essentially a "protocol parameter" set by the previous block and how full/empty it is.

> And I don't know how `priority_fee` gets calculated by the provider.

The formula `max_fee = base_fee + priority_fee` is probably used.

docs/userguides/transactions.md
```diff
@@ -0,0 +1,73 @@
...
+One way to use a static-fee transaction is by specifying the `gas_price` as a key-value argument:
+
+```python
+contract.fundMyContract(value="1 gwei", gas_price="100 gwei")
+```
+
+**NOTE**: Miners prioritize static-fee transactions based on the highest `gas_price`.
+
+Another way to use a static-fee transaction (without having to provide `gas_price`) is to set the key-value
+argument `type` equal to `0x00`.
```
fubuloubu, line 69: Add "By default, the gas_price is set using the provider's estimation" or something similar.

fubuloubu (dismissed):

docs/userguides/transactions.md
unparalleled-js: - [x] Done!

fubuloubu (approved):

Issue #387, opened 2022-01-04
Bug: Docker will not build


sabotagebeats:
### What went wrong?

Please include information like:

https://hub.docker.com/repository/docker/apeworx/ape/builds
Docker builds for ape are failing to build 

* full output of the error you received
```
#8 166.2 Searching for pycparser
#8 166.2 Reading https://pypi.org/simple/pycparser/
#8 166.5 Downloading https://files.pythonhosted.org/packages/62/d5/5f610ebe421e85889f2e55e33b7f9a6795bd982198517d912eb1c76e1a53/pycparser-2.21-py2.py3-none-any.whl#sha256=8ee45429555515e1f6b185e78100aea234072576aa43ab53aefcae078162fca9
#8 167.2 Best match: pycparser 2.21
#8 167.2 Processing pycparser-2.21-py2.py3-none-any.whl
#8 167.2 Installing pycparser-2.21-py2.py3-none-any.whl to /usr/local/lib/python3.8/dist-packages
#8 167.2 Adding pycparser 2.21 to easy-install.pth file
#8 167.2 
#8 167.2 Installed /usr/local/lib/python3.8/dist-packages/pycparser-2.21-py3.8.egg
#8 167.2 error: typing-extensions 4.0.1 is installed but typing-extensions<4,>=3.7.4 is required by {'trie'}
------
executor failed running [/bin/sh -c python3 ./setup.py install]: exit code: 1
```

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

Issue #316, opened 2021-12-14
Bug: bdist_wheel fail to build on ubuntu 20.04 fresh install with pyenv 


sabotagebeats:
### Environment information

* `ape` and plugin versions:
```
$ ape --version
Traceback (most recent call last):
  File "/home/doge/.pyenv/versions/apeworx/lib/python3.8/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/home/doge/.pyenv/versions/apeworx/lib/python3.8/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/doge/.pyenv/versions/apeworx/lib/python3.8/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (typing-extensions 4.0.1 (/home/doge/.pyenv/versions/apeworx/lib/python3.8/site-packages), Requirement.parse('typing-extensions<4,>=3.7.4'), {'trie'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/doge/.pyenv/versions/apeworx/bin/ape", line 33, in <module>
    sys.exit(load_entry_point('eth-ape', 'console_scripts', 'ape')())
  File "/home/doge/.pyenv/versions/apeworx/bin/ape", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
[...]

PR #388, opened 2022-01-04
fix: Dockerfile not building for typing-extensions incorrect verions

sabotagebeats:
### What I did

fix an error with incorrect version of `typing-extensions` for a required dependency `'trie'`

fixes: #387

### How I did it
added the line `RUN pip install typing-extensions==3.10.0.2` to the Dockerfile 

### How to verify it
Dockerfile now builds.

```
$ docker build .
$ docker images
REPOSITORY           TAG       IMAGE ID       CREATED          SIZE
<none>               <none>    07f16cf5e34a   47 minutes ago   586MB
$ docker run 07f16cf5e34a plugins list
Installed Plugins:
  vyper     0.1.0b1
  ens     0.1.0b2
  infura     0.1.0b2
  etherscan     0.1.0b1
  solidity     0.1.0b1
$ docker run -it 07f16cf5e34a console
In [1]:
```

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:20.04
 RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip --yes
 COPY . .
 # need to use typing-extensions<4 to deal with issue 387 bug: Docker will not build
-# todo: remove this line when 'trie' supports typing-extensions>4
+# TODO: Figure out a better solution or wait for it to resolve itself.
 RUN pip install typing-extensions==3.10.0.2
 RUN python3 ./setup.py install
 RUN ape plugins add solidity --yes
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04
 RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip --yes
 COPY . .
+RUN pip install typing-extensions==3.10.0.2
 RUN python3 ./setup.py install
 RUN ape plugins add solidity --yes
 RUN ape plugins add vyper --yes
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM ubuntu:20.04
 RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip --yes
 COPY . .
+# need to use typing-extensions<4 to deal with issue 387 bug: Docker will not build
+# todo: remove this line when 'trie' supports typing-extensions>4
 RUN pip install typing-extensions==3.10.0.2
 RUN python3 ./setup.py install
 RUN ape plugins add solidity --yes


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04
 RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip --yes
 COPY . .
+RUN pip install typing-extensions==3.10.0.2
```
fubuloubu, line 4: Add a comment why this is necessary

fubuloubu (dismissed):

Dockerfile
unparalleled-js: yeah or a TODO to remove once this stuff we are able to us typing-extensions>=4.0.0

sabotagebeats: ok cool I added a note and a todo to remove it

Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM ubuntu:20.04
 RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.8 python3-pip --yes
 COPY . .
+# need to use typing-extensions<4 to deal with issue 387 bug: Docker will not build
+# todo: remove this line when 'trie' supports typing-extensions>4
```
unparalleled-js, line 5: This is a little misleading: https://github.com/ethereum/py-trie/blob/master/setup.py#L54
I am not exactly sure what the whole deal is though..

fubuloubu: 2.0.0 isn't released yet

https://pypi.org/project/py-trie/

fubuloubu: v2.0.0 is released... weird
https://pypi.org/project/trie/

unparalleled-js: Maybe just say `TODO: Figure out a better solution or wait for it to resolve itself`.
I think this is a good patch for an immediate fix but we still ned to figure out what exactly is going on.

sabotagebeats: @fubuloubu it could be that the dependency that we are using which uses `trie` as a dependency does not have their pin updated?

Dockerfile
sabotagebeats: I updated the TODO so that it is more descriptive and to figure out a better solution.

unparalleled-js (approved):

Dockerfile
fubuloubu: That's definitely the case!

This is the dependency that uses it, however they updated it for v0.5.0-alpha.2:
https://github.com/ethereum/py-evm/commit/c8e5358cc33602f777037d64f94ec7081c98123c

`eth-tester` is not up to date yet tho:
https://github.com/ethereum/eth-tester/blob/master/setup.py#L26

sabotagebeats: ok so I'll make an issue that has as a blocker that eth-tester updates

sabotagebeats: ok so I'll make an issue that has as a blocker that eth-tester updates

fubuloubu: Created this to track: https://github.com/ethereum/eth-tester/pull/229

fubuloubu (approved):

Dockerfile
sabotagebeats: @fubuloubu this one is also tracking in our repo so I updated it https://github.com/ApeWorX/ape/issues/316

PR #389, opened 2022-01-05
feat: check signature function on AccountAPI

unparalleled-js:
### What I did

Was having a hard time figuring out now to sign a message. I thought there was a bug (I was using `vrs` when the `eth-account` module was asking for `rsv`). Nonetheless, I felt we needed a simple test for this, especially as there are no places where `sign_message()` is used.

### How I did it

* Test that the signature is expected
* Test that the sender address is expected

### How to verify it

test passses

### Checklist


fubuloubu (dismissed):
Should we add a method to recover accounts from signatures?

Or maybe a way to to determine if an account signed something e.g.:

```py
from ape.types.signatures import _Signature  # NOTE: baseclass of both signature types

...

class AccountAPI:
    ...

    def check_signature(self,
        data: Union[SignableMessage, TransactionAPI],
        signature: _Signature=None,  # TransactionAPI doesn't need it
    ) -> bool:
        if isinstance(data, SignableMessage):
            return self.address == Account.recover_message(data, vrs=signature)
        elif isinstance(data, TransactionAPI):
            return self.address == Account.recover_transaction(data.encode())
        else:
            raise Exception(f"Unsupported Message type: {type(data)}")
```

unparalleled-js:
> Should we add a method to recover accounts from signatures?
> 
> Or maybe a way to to determine if an account signed something e.g.:
> 
> ```python
> from ape.types.signatures import _Signature  # NOTE: baseclass of both signature types
> 
> ...
> 
> class AccountAPI:
>     ...
> 
>     def check_signature(self,
>         data: Union[SignableMessage, TransactionAPI],
>         signature: _Signature=None,  # TransactionAPI doesn't need it
>     ) -> bool:
>         if isinstance(data, SignableMessage):
>             return self.address == Account.recover_message(data, vrs=signature)
>         elif isinstance(data, TransactionAPI):
>             return self.address == Account.recover_transaction(data.encode())
>         else:
>             raise Exception(f"Unsupported Message type: {type(data)}")
> ```

Yes. I much rather call an `ape` method for this.

src/ape/api/accounts.py
```diff
@@ -208,6 +210,23 @@ def deploy(self, contract: "ContractContainer", *args, **kwargs) -> "ContractIns
...
         )
 
+    def check_signature(
+        self,
+        data: Union[SignableMessage, TransactionAPI],
+        signature: Optional[_Signature] = None,  # TransactionAPI doesn't need it
+    ) -> bool:
+        if isinstance(data, SignableMessage):
+            if signature:
+                return self.address == Account.recover_message(data, vrs=signature)
```
unparalleled-js, line 220: We don't need to call `encode_vrs()` because the `vrs=` kwarg on `recover_message()` expects a tuple.
Whereas if we use the `signature=` kwarg on `recover_message()`, we have to call `encode_rsv()` because it expects a concatenated string.

unparalleled-js:
If this merges in, I can create 2 non-urgent tickets in `ape-ledger` and `ape-trezor` (and `ape-keyring`) to use the new method in its tests

src/ape/types/signatures.py
```diff
@@ -39,4 +39,4 @@ class TransactionSignature(_Signature):
     """
 
 
-_ = SignableMessage
+__all__ = ["MessageSignature", "TransactionSignature", "SignableMessage"]
```
unparalleled-js, line 42: had to do this so that `SignableMessage` shows up the docs (it wasn't before)

fubuloubu (approved):

PR #394, opened 2022-01-06
feat: add upgrade option to plugins 'add' and 'install' commands

Ninjagod1251:
### What I did

Add upgrade to the options in adding or installing plugins

fixes: #

### How I did it

added the upgrade_confirmation_option to cli, options, and init to make it viable

### How to verify it

use the `ape plugins add ens --upgrade` to try it 

### Checklist


src/ape/cli/options.py
```diff
@@ -145,6 +145,24 @@ def skip_confirmation_option(help=""):
     )
 
 
+def upgrade_confirmation_option(help=""):
```
unparalleled-js, line 148: I am not sure we should put this here. This option seems only really specific to the `ape_plugins` plugin, so maybe we just define in the normal `_cli.py` there.

unparalleled-js, line 148: We should rename this to jus `upgrade_option`

Ninjagod1251: so i should only define this method in the _cli?

src/ape/cli/options.py
```diff
@@ -145,6 +145,24 @@ def skip_confirmation_option(help=""):
...
+    A ``click.option`` for upgrade plugin confirmation (``--upgrade``).
+
+    Args:
+        help (str): CLI option help text. Defaults to ``""``.
+    """
+
+    return click.option(
+        "-U",
+        "--upgrade",
+        "upgrade_confirmation",
```
unparalleled-js, line 159: We don't need this part - `upgrade_confirmation`.
If you exclude this, you can call the parameter `upgrade` inside the CLI command method

src/ape_plugins/_cli.py
```diff
@@ -122,12 +122,31 @@ def _list(cli_ctx, display_all):
...
+    )
+
+
 @cli.command(short_help="Install an ape plugin")
 @click.argument("plugin")
 @click.option("-v", "--version", help="Specify version (Default is latest)")
 @skip_confirmation_option(help="Don't ask for confirmation to add the plugin")
 @ape_cli_context()
-def add(cli_ctx, plugin, version, skip_confirmation):
+@upgrade_confirmation(help="Specify if user wants to update plugin")
```
unparalleled-js, line 148:
```suggestion
@upgrade_confirmation(help="Upgrade plugin to to the newest available version")
```

src/ape_plugins/_cli.py
```diff
@@ -122,12 +122,31 @@ def _list(cli_ctx, display_all):
...
+def add(cli_ctx, plugin, version, skip_confirmation, upgrade_confirmation):
```
unparalleled-js, line 149:
```suggestion
def add(cli_ctx, plugin, version, skip_confirmation, upgrade):
```

src/ape_plugins/_cli.py
```diff
@@ -152,7 +171,12 @@ def add(cli_ctx, plugin, version, skip_confirmation):
         # NOTE: Be *extremely careful* with this command, as it modifies the user's
         #       installed packages, to potentially catastrophic results
         # NOTE: This is not abstracted into another function *on purpose*
-        result = subprocess.call([sys.executable, "-m", "pip", "install", "--quiet", plugin])
+
+        args = [sys.executable, "-m", "pip", "install", "--quiet"]
+        if upgrade_confirmation:
```
unparalleled-js, line 176:
```suggestion
        if upgrade:
```

src/ape_plugins/_cli.py
```diff
@@ -164,7 +188,8 @@ def add(cli_ctx, plugin, version, skip_confirmation):
 @cli.command(short_help="Install all plugins in the local config file")
 @ape_cli_context()
 @skip_confirmation_option("Don't ask for confirmation to install the plugins")
-def install(cli_ctx, skip_confirmation):
+@upgrade_confirmation(help="Specify if user wants to update plugin")
```
unparalleled-js, line 191:
```suggestion
@upgrade_confirmation(help="Upgrade the plugins to their newest available versions")
```

src/ape_plugins/_cli.py
```diff
@@ -194,7 +219,10 @@ def install(cli_ctx, skip_confirmation):
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
 
-            args = [sys.executable, "-m", "pip", "install", "--quiet", package_name]
+            args = [sys.executable, "-m", "pip", "install", "--quiet"]
+            if upgrade_confirmation:
```
unparalleled-js, line 223: I don't think this ever gets here though?
On line 212, we are checking if the plugin is not installed. If we are upgrading the plugin, then it will already be installed, which mean we will never get into this loop

src/ape_plugins/_cli.py
```diff
@@ -122,12 +122,31 @@ def _list(cli_ctx, display_all):
...
+    A ``click.option`` for upgrade plugin confirmation (``--upgrade``).
+
+    Args:
+        help (str): CLI option help text. Defaults to ``""``.
+    """
+
+    return click.option(
+        "-U",
+        "--upgrade",
+        "upgrade_option",
```
unparalleled-js, line 136: you don't need to put `option` in the name.

every parameter to the commands are options.

just calling it `upgrade` is more expected, which is what it will default to
```suggestion
```

unparalleled-js: I know it seems weird, but this is just a `click`-ism

fubuloubu (approved):
Like this change! Very smart addition

unparalleled-js (changes requested):
First off, nice work! Thanks for letting me nitpick clickisms.

But, let's make sure we can upgrade packages that are already installed (we can't yet). That is the main use-case for this feature.

Thanks!

src/ape_plugins/_cli.py
```diff
@@ -145,8 +144,8 @@ def upgrade_option(help=""):
 @click.option("-v", "--version", help="Specify version (Default is latest)")
 @skip_confirmation_option(help="Don't ask for confirmation to add the plugin")
 @ape_cli_context()
-@upgrade_option(help="Specify if user wants to update plugin")
-def add(cli_ctx, plugin, version, skip_confirmation, upgrade_option):
+@upgrade_option(help="Upgrade the plugin to to the newest available version")
```
evanblank3, line 147:
```suggestion
@upgrade_option(help="Upgrade the plugin to the newest available version")
```

src/ape_plugins/_cli.py
```diff
@@ -188,8 +187,8 @@ def add(cli_ctx, plugin, version, skip_confirmation, upgrade_option):
 @cli.command(short_help="Install all plugins in the local config file")
 @ape_cli_context()
 @skip_confirmation_option("Don't ask for confirmation to install the plugins")
-@upgrade_option(help="Specify if user wants to update plugin")
-def install(cli_ctx, skip_confirmation, upgrade_option):
+@upgrade_option(help="Upgrade the plugin to to the newest available version")
```
evanblank3, line 190:
```suggestion
@upgrade_option(help="Upgrade the plugin to the newest available version")
```

fubuloubu (dismissed):

src/ape_plugins/_cli.py
Ninjagod1251: resolved with the latest push

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,66 @@ def _list(cli_ctx, display_all):
...
-
     if version:
         plugin = f"{plugin}=={version}"
 
     if plugin in CORE_PLUGINS:
         cli_ctx.abort(f"Cannot add 1st class plugin '{plugin}'")
 
     elif is_plugin_installed(plugin):
-        cli_ctx.abort(f"Plugin '{plugin}' already installed")
+        cli_ctx.logger.info(f"Plugin '{plugin}' already installed.")
```
unparalleled-js, line 161: Maybe let's only log this if they did not say `--upgrade`,

So an `else` after the `if` on line 162?

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,66 @@ def _list(cli_ctx, display_all):
...
+        if upgrade:
+            cli_ctx.logger.info(f"Updating {plugin} plugin.")
+            args = [sys.executable, "-m", "pip", "install", "--quiet", "--upgrade", plugin]
```
unparalleled-js, line 164: Is there a way to not have to copy all of this again?

Like could we just build up a list of args?

If the user did not say `--upgrade` (so in an `else`), we can just return early (after logging).

fubuloubu (approved):

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,68 @@ def _list(cli_ctx, display_all):
             click.echo()
 
 
+def upgrade_option(help=""):
```
unparalleled-js, line 125:
```suggestion
def upgrade_option(help: str=""):
```

src/ape_plugins/_cli.py
unparalleled-js: This still does not seem resolved.

We should move this line to an `else` follow the `if upgrade` line, no?

If the user wants to upgrade, there is no need to tell them it is already installed I think, they likely are aware of that

unparalleled-js: And then it can be made a warning and we can also include something like `Use --upgrade if you want to install the latest version"

src/ape_plugins/_cli.py
```diff
@@ -181,8 +213,17 @@ def install(cli_ctx, skip_confirmation):
...
-            cli_ctx.logger.warning(f"Plugin '{module_name}' is trusted and already installed.")
-            pass
+            cli_ctx.logger.info(f"Plugin '{module_name}' is trusted and already installed.")
+            if upgrade:
+                cli_ctx.logger.info(f"Updating {module_name} plugin.")
+                args.append("--upgrade")
+                args.append(module_name)
+                result = subprocess.call(args)
+                if result == 0 and is_plugin_installed(module_name):
+                    cli_ctx.logger.success(f"Plugin '{module_name}' has been added.")
```
unparalleled-js, line 223:
```suggestion
                    cli_ctx.logger.success(f"Plugin '{module_name}' has been upgraded.")
```

src/ape_plugins/_cli.py
```diff
@@ -181,8 +213,17 @@ def install(cli_ctx, skip_confirmation):
             any_install_failed = True
         # check for installed check the config.yaml
         elif installed_plugin and available_plugin:
-            cli_ctx.logger.warning(f"Plugin '{module_name}' is trusted and already installed.")
-            pass
+            cli_ctx.logger.info(f"Plugin '{module_name}' is trusted and already installed.")
```
unparalleled-js, line 216: Same thing here - we should move this to the other side of the `if upgrade` in an `else` block so it only happens when the user forgot to include `--upgrade` and the package is already installed. We should also warn them to use `--upgrade` if they wish to upgrade, in case they mistakenly forgot `--uprade`

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,73 @@ def _list(cli_ctx, display_all):
...
+            result = subprocess.call(args)
+
+            if result == 0 and is_plugin_installed(plugin):
+                cli_ctx.logger.success(f"Plugin '{plugin}' has been upgraded.")
+            else:
+                cli_ctx.logger.error(f"Failed to add '{plugin}'.")
+                sys.exit(1)
+        else:
+            cli_ctx.logger.warning(
+                f"{plugin} has already installed."
```
unparalleled-js, line 175:
```suggestion
                f"{plugin} is already installed."
```

unparalleled-js: If you don't end in a space, it will all be together without it

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,73 @@ def _list(cli_ctx, display_all):
...
+                f"Please type --upgrade if you want to update {plugin}"
```
unparalleled-js, line 176:
```suggestion
                f"Use the '--upgrade' option if you want to update '{plugin}'."
```

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,22 @@ def install(cli_ctx, skip_confirmation):
             any_install_failed = True
         # check for installed check the config.yaml
         elif installed_plugin and available_plugin:
-            cli_ctx.logger.warning(f"Plugin '{module_name}' is trusted and already installed.")
-            pass
+            cli_ctx.logger.info(f"Plugin '{module_name}' is trusted and already installed.")
```
unparalleled-js, line 220: Can you remove this log line?

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,22 @@ def install(cli_ctx, skip_confirmation):
...
+            if upgrade:
+                cli_ctx.logger.info(f"Updating {module_name} plugin.")
+                args.append("--upgrade")
+                args.append(module_name)
+                result = subprocess.call(args)
+                if result == 0 and is_plugin_installed(module_name):
+                    cli_ctx.logger.success(f"Plugin '{module_name}' has been upgraded.")
+                else:
+                    cli_ctx.logger.error(f"Failed to add '{module_name}'.")
+                    sys.exit(1)
```
unparalleled-js, line 230: Hmmm... Do we want to fail the whole command if a single plugin fails to install properly? Were we doing this before?

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,22 @@ def install(cli_ctx, skip_confirmation):
...
+            cli_ctx.logger.info(f"Plugin '{module_name}' is trusted and already installed.")
+            if upgrade:
+                cli_ctx.logger.info(f"Updating {module_name} plugin.")
+                args.append("--upgrade")
+                args.append(module_name)
+                result = subprocess.call(args)
+                if result == 0 and is_plugin_installed(module_name):
+                    cli_ctx.logger.success(f"Plugin '{module_name}' has been upgraded.")
+                else:
+                    cli_ctx.logger.error(f"Failed to add '{module_name}'.")
```
unparalleled-js, line 229:
```suggestion
                    cli_ctx.logger.error(f"Failed to upgrade '{module_name}'.")
```

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,22 @@ def install(cli_ctx, skip_confirmation):
...
+                    sys.exit(1)
+            else:
+                cli_ctx.logger.warning(
+                    f"{module_name} has already installed."
+                    f"Please type --upgrade if you want to update {module_name}"
```
unparalleled-js, line 234:
```suggestion
                    f"Use the '--upgrade' option if you want to update '{plugin}'."
```

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,22 @@ def install(cli_ctx, skip_confirmation):
...
+                args.append(module_name)
+                result = subprocess.call(args)
+                if result == 0 and is_plugin_installed(module_name):
+                    cli_ctx.logger.success(f"Plugin '{module_name}' has been upgraded.")
+                else:
+                    cli_ctx.logger.error(f"Failed to add '{module_name}'.")
+                    sys.exit(1)
+            else:
+                cli_ctx.logger.warning(
+                    f"{module_name} has already installed."
```
unparalleled-js, line 233:
```suggestion
                    f"{module_name} is already installed. "
```

src/ape_plugins/_cli.py
unparalleled-js: I think we can remove the `sys.exit(1)` here do what we do below where we track if any of them failed and then exit with a status of `1` at the very end

src/ape_plugins/_cli.py
```diff
@@ -193,8 +243,9 @@ def install(cli_ctx, skip_confirmation):
             # NOTE: Be *extremely careful* with this command, as it modifies the user's
             #       installed packages, to potentially catastrophic results
             # NOTE: This is not abstracted into another function *on purpose*
-
-            args = [sys.executable, "-m", "pip", "install", "--quiet", package_name]
+            if upgrade:
```
unparalleled-js, line 246: I am not sure it really matters to include `--upgrade` here because the plugin is being installed for the first time and it will install the latest

Ninjagod1251: I am thinking if the user puts in a specific version number and it installs. 

What happens when they run a different script to upgrade? Is that possible?

src/ape_plugins/_cli.py
Ninjagod1251: I thin you are right and we should try and install as many as we can and just show error loggers and sys exit at the end
I removed the sys exit and place a flag so it can continue run and just exit with 1 at the end

src/ape_plugins/_cli.py
unparalleled-js: I just tried it on `pip`. I looks like the `--upgrade` flag gets ignored. I think it an error would more helpful though.

That is probably out of scope for this PR, but we can make the options incompatible at the CLI-option level. Do want to  add that to the grocery list? I know a trick for this.

src/ape_plugins/_cli.py
```diff
@@ -122,39 +122,73 @@ def _list(cli_ctx, display_all):
...
     if version:
         plugin = f"{plugin}=={version}"
 
     if plugin in CORE_PLUGINS:
         cli_ctx.abort(f"Cannot add 1st class plugin '{plugin}'")
 
     elif is_plugin_installed(plugin):
-        cli_ctx.abort(f"Plugin '{plugin}' already installed")
+        if upgrade:
+            cli_ctx.logger.info(f"Updating {plugin} plugin.")
```
unparalleled-js, line 163:
```suggestion
            cli_ctx.logger.info(f"Updating '{plugin}'...")
```

^ just to match line 184.
I love this btw, this is looking great

src/ape_plugins/_cli.py
unparalleled-js: Like `--upgrade` should be allowed with `--version` because they are incompatible
but that can be later on

src/ape_plugins/_cli.py
```diff
@@ -181,8 +217,21 @@ def install(cli_ctx, skip_confirmation):
             any_install_failed = True
         # check for installed check the config.yaml
         elif installed_plugin and available_plugin:
-            cli_ctx.logger.warning(f"Plugin '{module_name}' is trusted and already installed.")
-            pass
+            if upgrade:
+                cli_ctx.logger.info(f"Updating {module_name} plugin.")
```
unparalleled-js, line 221:
```suggestion
                cli_ctx.logger.info(f"Updating '{module_name}'...")
```

unparalleled-js (approved):
LGTM

unparalleled-js (approved):

PR #402, opened 2022-01-17
feat: allow setting the pending timestamp from top-level 'chain' class

unparalleled-js:
### What I did

Allows setting the pending timestamp

### How I did it

Copied brownie and used the `eth-tester` method `time_travel()`

### How to verify it

Does my test make sense?

Usage example:

```python
from ape import chain

chain.pending_timestamp += 3600
```

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -589,12 +589,15 @@ class TestProviderAPI(ProviderAPI):
         """
 
     @abstractmethod
-    def increase_time(self, seconds: int):
+    def fast_forward_time(self, new_timestamp: int) -> int:
         """
         Increase the time into the future for testing purposes.
 
         Args:
-            seconds (int): The amount of seconds to increase time.
+            new_timestamp (int): The timestamp to set.
+
+        Returns:
+            int: The new timestamp.
         """
 
 
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -588,6 +588,15 @@ class TestProviderAPI(ProviderAPI):
             snapshot_id (str): The snapshot ID.
         """
 
+    @abstractmethod
+    def increase_time(self, seconds: int):
+        """
+        Increase the time into the future for testing purposes.
+
+        Args:
+            seconds (int): The amount of seconds to increase time.
+        """
+
 
 class Web3Provider(ProviderAPI):
     """
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -297,6 +297,7 @@ class ChainManager(_ConnectedChain):
     _chain_id_map: Dict[str, int] = {}
     _block_container_map: Dict[int, BlockContainer] = {}
     _account_history_map: Dict[int, AccountHistory] = {}
+    _time_offset = 0
 
     @property
     def blocks(self) -> BlockContainer:
@@ -359,9 +360,22 @@ class ChainManager(_ConnectedChain):
     def pending_timestamp(self) -> int:
         """
         The current epoch time of the chain, as an ``int``.
+        Set the timestamp to fast-forward time into the future.
+
+        Usage example::
+
+            from ape import chain
+
+            chain.pending_timestamp += 3600
         """
 
-        return int(time.time())
+        return int(time.time() + self._time_offset)
+
+    @pending_timestamp.setter
+    def pending_timestamp(self, new_value: int):
+        provider = self._get_test_provider(TestProviderAPI.fast_forward_time.__name__)
+        new_timestamp = provider.fast_forward_time(new_value)
+        self._time_offset = new_timestamp - self.pending_timestamp
 
     def __repr__(self) -> str:
         props = f"id={self.chain_id}" if self._networks.active_provider else "disconnected"
--- a/src/ape_test/providers.py
+++ b/src/ape_test/providers.py
@@ -73,10 +73,8 @@ class LocalNetwork(TestProviderAPI, Web3Provider):
             if current_hash != snapshot_id:
                 return self._tester.revert_to_snapshot(snapshot_id)
 
-    def increase_time(self, seconds: int):
-        current_timestamp = self.get_block("latest").timestamp
-        requested_timestamp = current_timestamp + seconds
-        self._tester.time_travel(requested_timestamp)
+    def fast_forward_time(self, new_timestamp: int) -> int:
+        return self._tester.time_travel(new_timestamp)
 
 
 def _get_vm_err(web3_err: TransactionFailed) -> ContractLogicError:
--- a/src/ape_test/providers.py
+++ b/src/ape_test/providers.py
@@ -73,6 +73,11 @@ class LocalNetwork(TestProviderAPI, Web3Provider):
             if current_hash != snapshot_id:
                 return self._tester.revert_to_snapshot(snapshot_id)
 
+    def increase_time(self, seconds: int):
+        current_timestamp = self.get_block("latest").timestamp
+        requested_timestamp = current_timestamp + seconds
+        self._tester.time_travel(requested_timestamp)
+
 
 def _get_vm_err(web3_err: TransactionFailed) -> ContractLogicError:
     err_message = str(web3_err).split("execution reverted: ")[-1] or None
--- a/tests/integration/test_chain.py
+++ b/tests/integration/test_chain.py
@@ -105,3 +105,10 @@ def test_blocks_range_too_high_stop(chain_at_block_5):
         f"'stop={len_plus_1}' cannot be greater than the chain length (6). "
         f"Use 'poll_blocks()' to wait for future blocks."
     )
+
+
+def test_set_pending_timestamp(chain):
+    start_timestamp = chain.pending_timestamp
+    chain.pending_timestamp += 3600
+    new_timestamp = chain.pending_timestamp
+    assert new_timestamp - start_timestamp == 3600


src/ape/managers/chain.py
```diff
@@ -359,9 +360,22 @@ def base_fee(self) -> int:
...
+            chain.pending_timestamp += 3600
         """
 
-        return int(time.time())
+        return int(time.time() + self._time_offset)
+
+    @pending_timestamp.setter
+    def pending_timestamp(self, new_value: int):
+        provider = self._get_test_provider(TestProviderAPI.fast_forward_time.__name__)
+        new_timestamp = provider.fast_forward_time(new_value)
```
unparalleled-js, line 377: When you do `chain.pending_timestamp += 3600`, it first add the current pending timestamp (which is the OS time + the existing offset which starts at `0`). Then, it calls `time_travel()` from `eth-tester` with the new timestamp (or the evm method in hardhat, not implemented yet)

src/ape/managers/chain.py
```diff
@@ -359,9 +360,22 @@ def base_fee(self) -> int:
...
+
+            chain.pending_timestamp += 3600
         """
 
-        return int(time.time())
+        return int(time.time() + self._time_offset)
+
+    @pending_timestamp.setter
+    def pending_timestamp(self, new_value: int):
+        provider = self._get_test_provider(TestProviderAPI.fast_forward_time.__name__)
```
fubuloubu, line 376: What is going on here? Can't this be `self.provider`?

src/ape/managers/chain.py
```diff
@@ -359,9 +360,22 @@ def base_fee(self) -> int:
...
+
+        Usage example::
+
+            from ape import chain
+
+            chain.pending_timestamp += 3600
         """
 
-        return int(time.time())
+        return int(time.time() + self._time_offset)
```
fubuloubu, line 372: Just noticed that you could ask for `self.provider.get_block("pending").timestamp` to get this value from the provider

src/ape/managers/chain.py
```diff
@@ -359,9 +360,22 @@ def base_fee(self) -> int:
...
+
+    @pending_timestamp.setter
+    def pending_timestamp(self, new_value: int):
+        provider = self._get_test_provider(TestProviderAPI.fast_forward_time.__name__)
+        new_timestamp = provider.fast_forward_time(new_value)
+        self._time_offset = new_timestamp - self.pending_timestamp
```
fubuloubu, line 378: If you get timestamp from the provider, we no longer need this (sorry for making this suggestion before, using `time.time()` seems like not the right way to be doing this)

src/ape/api/providers.py
```diff
@@ -588,6 +588,18 @@ def revert(self, snapshot_id: SnapshotID):
             snapshot_id (str): The snapshot ID.
         """
 
+    @abstractmethod
+    def fast_forward_time(self, new_timestamp: int) -> int:
```
fubuloubu, line 592: Seems better to just do `set_timestamp` and let the provider plugin specify what range of timestamps it supports

src/ape/managers/chain.py
unparalleled-js: The `_get_test_provider` helper private method just ensures that `self.provider` is indeed a `TestProviderAPI` implementation. The `TestProviderAPI` is the mixin that has the snapshotting and other development-only related methods that normal `ProviderAPI` implementations won't have.

src/ape/managers/chain.py
unparalleled-js: Oh nice!

src/ape/managers/chain.py
unparalleled-js: Oh also, `mypy` is happier when we know it is a `TestProviderAPI`, that is kind of what drove this

fubuloubu: outside the scope of this PR, but really don't like how this works

down below, I see that the mechanics are working the way I expected, but was hoping somehow it would be a lot more ergonomic rather than having to pass `TestProviderAPI.fast_forward_time.__name__` to it...

maybe this can be hacked by wrapping `self.provider.fast_forward_time` in a `try..except` that catches `AttributeError` and converts to `NotImplementedError` instead e.g.:

```py
@pending_timestamp.setter
    def pending_timestamp(self, new_value: int):
        try:
            self.provider.fast_forward_time(new_value)
        except AttributeError as e:
            raise NotImplementedError from e
```

It would be even neater if fields that are in `TestProviderAPI` but not `ProviderAPI` would instead just raise this error directly so we don't have to hack it so much e.g.:

```py
@pending_timestamp.setter
    def pending_timestamp(self, new_value: int):
        self.provider.fast_forward_time(new_value)
        # raises `NotImplementedError` if only `ProviderAPI`
```

src/ape/managers/chain.py
```diff
@@ -359,9 +360,23 @@ def base_fee(self) -> int:
...
 
-        return int(time.time())
+        pending_block = self.provider.get_block("pending")
+        return int(pending_block.timestamp + self._time_offset)
+
+    @pending_timestamp.setter
+    def pending_timestamp(self, new_value: int):
+        provider = self._get_test_provider(TestProviderAPI.set_timestamp.__name__)
+        new_timestamp = provider.set_timestamp(new_value)
+        self._time_offset = new_timestamp - self.pending_timestamp
```
fubuloubu, line 379: Shouldn't actually need to track this anymore now that you are using `self.provider.get_block("pending").timestamp`!

src/ape/managers/chain.py
```diff
@@ -359,9 +360,23 @@ def base_fee(self) -> int:
     def pending_timestamp(self) -> int:
         """
         The current epoch time of the chain, as an ``int``.
+        Set the timestamp to fast-forward time into the future.
```
fubuloubu, line 363: For some reason, if this supports traveling backwards in limited circumstances (say during the genesis block), this docstring might not be right

src/ape/managers/chain.py
```diff
@@ -359,9 +360,23 @@ def base_fee(self) -> int:
...
+        Usage example::
+
+            from ape import chain
+
+            chain.pending_timestamp += 3600
         """
 
-        return int(time.time())
+        pending_block = self.provider.get_block("pending")
+        return int(pending_block.timestamp + self._time_offset)
```
fubuloubu, line 373:
```suggestion
        pending_block = self.provider.get_block("pending")
        return self.provider.get_block("pending").timestamp
```

src/ape/api/providers.py
```diff
@@ -588,6 +588,18 @@ def revert(self, snapshot_id: SnapshotID):
             snapshot_id (str): The snapshot ID.
         """
 
+    @abstractmethod
+    def set_timestamp(self, new_timestamp: int) -> int:
+        """
+        Increase the time into the future for testing purposes.
```
fubuloubu, line 594: Same here re: "into the future"

fubuloubu (dismissed):

src/ape/managers/chain.py
fubuloubu: This might be inside the scope of this PR if this is the only instance where this occurs :thinking:

fubuloubu: Ah ha! Okay, maybe we solve it like this:

1. All fields/methods are in `ProviderAPI` that are destined for `TestProviderAPI`
2. Those fields all raise `NotImplementedError` in `ProviderAPI`
3. Override those fields in `TestProviderAPI` to instead be decorated with `@apimethod`

Now you can initialize normal `ProviderAPI` classes without issue, but if a `TestProviderAPI` subclass does not implement the methods, it will throw. Most importantly, mypy is happy lol.

src/ape/managers/chain.py
```diff
@@ -369,14 +368,12 @@ def pending_timestamp(self) -> int:
             chain.pending_timestamp += 3600
         """
 
-        pending_block = self.provider.get_block("pending")
-        return int(pending_block.timestamp + self._time_offset)
+        return int(self.provider.get_block("pending").timestamp)
```
fubuloubu, line 371: is this not an int by default?

src/ape/managers/chain.py
```diff
@@ -369,14 +368,12 @@ def pending_timestamp(self) -> int:
...
 
     @pending_timestamp.setter
     def pending_timestamp(self, new_value: int):
         provider = self._get_test_provider(TestProviderAPI.set_timestamp.__name__)
-        new_timestamp = provider.set_timestamp(new_value)
```
fubuloubu: No longer need to have `TestProviderAPI.set_timestamp` return anything either

fubuloubu (approved):

src/ape/managers/chain.py
unparalleled-js: it is a float, I don't get why though because I think it actually is just always an int

unparalleled-js: Okay, yeah this should be an int: https://ethereum.stackexchange.com/questions/7853/is-the-block-timestamp-value-in-solidity-seconds-or-milliseconds/7854

fubuloubu (approved):

src/ape/managers/chain.py
fubuloubu: NOTE: Come back to this later and fix

fubuloubu:
Just want to note this: https://github.com/ApeWorX/ape/pull/402#discussion_r785645065

PR #405, opened 2022-01-17
fix: better support for initializing past receipts

unparalleled-js:
### What I did

Working on the final implementation for `get_account_transactions()` in `ape-etherscan` and noticed a few things here that are off.

1. Uses `block_time` for waiting for receipt confirmations
2. Only logs that a transaction was submitted when submitting a transaction, not when initializing a `ReceiptAPI` class after the fact.
3. Removes a warning that always happens when converting to a `HexAddress`. I want to use the `convert` API to ensure the value is a checksum address, so it seems like we should not warn about this since we are converting it right after.
4. Includes explorer URL in receipt where apt
5. prettifies txn hash on receipt

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -276,7 +276,9 @@ class ReceiptAPI:
             # the user is aware of this. Or, this is a development environment.
             return self
 
-        confirmations_occurred = 0
+        confirmations_occurred = self._get_confirmations_occurred()
+        if confirmations_occurred >= self.required_confirmations:
+            return self
 
         # If we get here, that means the transaction has been recently submitted.
         log_message = f"Submitted {self.txn_hash}"
@@ -289,8 +291,7 @@ class ReceiptAPI:
 
         with ConfirmationsProgressBar(self.required_confirmations) as progress_bar:
             while confirmations_occurred < self.required_confirmations:
-                latest_block = self.provider.get_block("latest")
-                confirmations_occurred = latest_block.number - self.block_number  # type: ignore
+                confirmations_occurred = self._get_confirmations_occurred()
                 progress_bar.confs = confirmations_occurred
 
                 if confirmations_occurred == self.required_confirmations:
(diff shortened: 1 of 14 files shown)


src/ape/api/providers.py
```diff
@@ -689,6 +688,7 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
             else self.network.required_confirmations
         )
         receipt = self.get_transaction(txn_hash.hex(), required_confirmations=req_confs)
+        logger.info(f"Submitted {receipt.txn_hash} (gas_used={receipt.gas_used})")
```
fubuloubu, line 691: Note that `web3.eth.wait_for_transaction_receipt` is blocking, so that you will have to wait a decent number of confs to see `gas_used` from the receipt.

Probably should be seperated into two info log events:
1. f"Submitted {txn_hash.hex()} ({explorer_txn_url})
2. f"Confirmed {receipt.txn_hash} (gas_used={receipt.gas_used})

Number (2) might best live in `ReceiptAPI` come to think of it

unparalleled-js: Oh that makes sense.

unparalleled-js:
I can't properly test some of this without first merging in https://github.com/ApeWorX/ape/pull/408

src/ape/managers/project.py
```diff
@@ -285,11 +285,15 @@ def load_contracts(
         # If a file is deleted from ``sources`` but is in
         # ``cached_sources``, remove its corresponding ``contract_types`` by
         # using ``ContractType.source_id`` and ``ContractType.sourcePath``
-        deleted_sources = cached_sources.keys() - set(map(str, sources))
+        deleted_source_ids = cached_sources.keys() - set(
```
unparalleled-js, line 288: The changes in this file should have been part of the PR but I didn't realize it had broken. I added a test that fails now without this. Basically, the contracts were not getting returning properly because `ape` thought were always deleted.

fubuloubu (approved):

PR #406, opened 2022-01-17
refactor: raise NotImplementedError for calling test API on normal provider

unparalleled-js:
### What I did

Get rid of `_get_test_provider()` from `ChainManager` and replace with a smarter `ProviderAPI` that first checks if the missing member is a member of the `TestProviderAPI` and raises an appropriate error at that point, allowing us to always use `ProviderAPI` as if it were a `TestProviderAPI`.

### How I did it

Using `__getattr__` on `ProviderAPI` and checking for members that are in `TestProviderAPI` and not `ProviderAPI`.

### How to verify it

Try calling a test member api from the chain manager when connected to a non-development chain

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -382,14 +382,6 @@ class ProviderAPI:
     request_header: str
     """A header to set on HTTP/RPC requests."""
 
-    def __getattr__(self, item):
-        # Check if trying to use a 'TestProvierAPI' method.
-        test_provider_api_members = [m for m in dir(TestProviderAPI) if m not in dir(ProviderAPI)]
-        if item in test_provider_api_members:
-            raise NotImplementedError(f"{self.name} is not a 'TestProviderAPI' implementation.")
-
-        raise NotImplementedError(f"ProviderAPI has no member '{item}'.")
-
     @abstractmethod
     def connect(self):
         """
@@ -564,9 +556,24 @@ class ProviderAPI:
             Iterator[dict]: A dictionary of events.
         """
 
+    def snapshot(self) -> SnapshotID:
+        raise self._get_not_test_api_error(ProviderAPI.snapshot.__name__)
+
+    def revert(self, snapshot_id: SnapshotID):
+        raise self._get_not_test_api_error(ProviderAPI.revert.__name__)
+
+    def set_timestamp(self, new_timestamp: int):
+        raise self._get_not_test_api_error(ProviderAPI.set_timestamp.__name__)
+
     def _try_track_receipt(self, receipt: ReceiptAPI):
         if self._chain:
             self._chain.account_history.append(receipt)
+
+    def _get_not_test_api_error(self, method_name: str) -> NotImplementedError:
+        return NotImplementedError(
+            f"Propvider '{self.name}' is not a 'TestProviderAPI' "
+            f"and thus method '{method_name}' is not implemented."
+        )
 
 
 class TestProviderAPI(ProviderAPI):
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -354,6 +354,13 @@ class BlockAPI:
         """
 
 
+def not_implemented(fn):
+    def inner(*args, **kwargs):
+        raise NotImplementedError(fn.__name__)
+
+    return inner
+
+
 @abstractdataclass
 class ProviderAPI:
     """
@@ -556,24 +563,21 @@ class ProviderAPI:
             Iterator[dict]: A dictionary of events.
         """
 
+    @not_implemented
     def snapshot(self) -> SnapshotID:
-        raise self._get_not_test_api_error(self.snapshot.__name__)
+        pass
 
+    @not_implemented
     def revert(self, snapshot_id: SnapshotID):
-        raise self._get_not_test_api_error(self.revert.__name__)
+        pass
 
+    @not_implemented
     def set_timestamp(self, new_timestamp: int):
-        raise self._get_not_test_api_error(self.set_timestamp.__name__)
+        pass
 
     def _try_track_receipt(self, receipt: ReceiptAPI):
         if self._chain:
             self._chain.account_history.append(receipt)
-
-    def _get_not_test_api_error(self, method_name: str) -> NotImplementedError:
-        return NotImplementedError(
-            f"Propvider '{self.name}' is not a 'TestProviderAPI' "
-            f"and thus method '{method_name}' is not implemented."
-        )
 
 
 class TestProviderAPI(ProviderAPI):
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -557,13 +557,13 @@ class ProviderAPI:
         """
 
     def snapshot(self) -> SnapshotID:
-        raise self._get_not_test_api_error(ProviderAPI.snapshot.__name__)
+        raise self._get_not_test_api_error(self.snapshot.__name__)
 
     def revert(self, snapshot_id: SnapshotID):
-        raise self._get_not_test_api_error(ProviderAPI.revert.__name__)
+        raise self._get_not_test_api_error(self.revert.__name__)
 
     def set_timestamp(self, new_timestamp: int):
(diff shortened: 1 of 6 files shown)


src/ape/managers/chain.py
```diff
@@ -427,14 +424,5 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
             snapshot_index = self._snapshots.index(snapshot_id)
             self._snapshots = self._snapshots[:snapshot_index]
 
-        provider.revert(snapshot_id)
+        self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
-
-    def _get_test_provider(self, method_name: str) -> TestProviderAPI:
```
unparalleled-js, line 433: Now, we are able to delete this method and no longer have to do stuff like ` self._get_test_provider(TestProviderAPI.revert.__name__)` because it happens at the provider level via `__getattr__`

src/ape/api/providers.py
```diff
@@ -382,6 +382,14 @@ class ProviderAPI:
     request_header: str
     """A header to set on HTTP/RPC requests."""
 
+    def __getattr__(self, item):
+        # Check if trying to use a 'TestProvierAPI' method.
+        test_provider_api_members = [m for m in dir(TestProviderAPI) if m not in dir(ProviderAPI)]
```
fubuloubu, line 387: wonder if you can do a set operation here

fubuloubu:
```suggestion
        test_provider_api_members = set(dir(TestProviderAPI)) - set(dir(ProviderAPI))
```

src/ape/api/providers.py
```diff
@@ -382,6 +382,14 @@ class ProviderAPI:
...
+        if item in test_provider_api_members:
+            raise NotImplementedError(f"{self.name} is not a 'TestProviderAPI' implementation.")
+
+        raise NotImplementedError(f"ProviderAPI has no member '{item}'.")
```
fubuloubu, line 391: The only trick to `__getattr__` is that it *has* to raise `AttributeError` otherwise it gets into an infinite loop

Also, should probably call the `super` method
```suggestion
            raise AttributeError(f"{self.name} is not a 'TestProviderAPI' implementation.")

        raise AttributeError(f"ProviderAPI has no member '{item}'.")
        
        return super().__getattr__(item)
```

Handle the attribute error whenever you try to use a `TestProviderAPI` method, and just `raise NotImplementedError from attr_err`

fubuloubu: Perhaps a more simpler method is still this though: https://github.com/ApeWorX/ape/pull/402#discussion_r785619966

A bit more work to set up, as you'll have to maintain the methods in both places (where one is raising `NotImplementedError` and the other is `abstractmethod`) but perhaps you can filter the set of method names in `TestProviderAPI` that are `abstractmethod` and not in `ProviderAPI`, and add those as methods to `ProviderAPI` where they all just raise `NotImplementedError` instead.

unparalleled-js: Yeah I want to avoid having to remember to add the methods to two separate spaces or else we will likely forget to do that in the future

unparalleled-js: I will try to keep it dynamic without using the `__getattr__` approach

src/ape/api/providers.py
```diff
@@ -382,6 +382,23 @@ class ProviderAPI:
     request_header: str
     """A header to set on HTTP/RPC requests."""
 
+    def __post_init__(self):
```
unparalleled-js, line 385: Sooo we will have to update `infura` and `alchemy` to call `super().__post_init__()` to get this to fully work.

Altso, `mypy` still don't know about these newly added members, I am not sure if there is a hack for that but I am cool with `# type: ignore` comments because _we_ know these methods exist.

unparalleled-js: If we like this approach, I will get the PRs ready for those 2 plugins so we remember to fix them after we release

fubuloubu: Another nice thing is that pydantic update will fix some of these issues

src/ape/api/providers.py
```diff
@@ -382,6 +382,23 @@ class ProviderAPI:
...
+                f"Propvider '{self.name}' is not a 'TestProviderAPI' "
+                f"and thus method '{method_name}' is not implemented."
+            )
+
+        for test_api_method in abstract_test_api_methods:
+            try:
+                getattr(self, test_api_method) is None
+            except AttributeError:
+                setattr(self, test_api_method, lambda: raise_not_implemented(test_api_method))
+
```
fubuloubu, line 401: Couldn't this all be done by adding it to the class externally? Then we don't need `__post_init__`

fubuloubu (approved):

src/ape/api/providers.py
unparalleled-js: Sure. `setattr` needs an instance to work though, and it is a bit weird to have code for a class happening in some other class.

unparalleled-js: Lol, I just realized I don't need to use `setattr` and then we can probably do this at the module level, I like that

unparalleled-js: I can just duplicate the methods in both if you want.

src/ape/api/providers.py
```diff
@@ -556,10 +556,25 @@ def get_events(self, **filter_params) -> Iterator[dict]:
             Iterator[dict]: A dictionary of events.
         """
 
+    def snapshot(self) -> SnapshotID:
```
unparalleled-js, line 559: We just need to remember to add new test API methods here.
I think `mypy` will actually help us there, so I am less worried than I was, and this solution is cleaner than the post init idea.

src/ape/api/providers.py
```diff
@@ -354,6 +354,13 @@ def decode(cls, data: Dict) -> "BlockAPI":
         """
 
 
+def not_implemented(fn):
```
unparalleled-js, line 357: Should we move this to `utils`?

fubuloubu: Yes, was thinking this. Maybe call it `raises_not_implemented`

src/ape/api/providers.py
```diff
@@ -556,25 +563,22 @@ def get_events(self, **filter_params) -> Iterator[dict]:
             Iterator[dict]: A dictionary of events.
         """
 
+    @not_implemented
```
fubuloubu, line 566: Add a quick note about why we're doing this

fubuloubu (approved):

src/ape/utils.py
```diff
@@ -460,6 +460,23 @@ class AbstractDataClassMeta(DataClassMeta, ABCMeta):
...
+    A decorator to use over API methods that you want to be callable
+    but that are not actually implemented.
+
+    An example is when calling abstracts methods from the
+    :class:`~ape.api.providers.TestProviderAPI` mixin but when using a
+    provider that is not for testing.
+    """
+
+    def inner(*args, **kwargs):
+        raise NotImplementedError(fn.__name__)
```
fubuloubu, line 475:
```suggestion
        raise NotImplementedError(f"Attempted to call method '{fn.__name__}' in 'ProviderAPI', which is only available in 'TestProviderAPI'.")
```

fubuloubu: But then it's not really a util, is it? lol

unparalleled-js: I just moved it back, I like the longer messaging

tests/integration/cli/test_compile.py
```diff
@@ -50,6 +50,7 @@ def test_compile(ape_cli, runner, project):
 )
 def test_compile_specified_contracts(ape_cli, runner, project, contract_path, clean_cache):
     result = runner.invoke(ape_cli, ["compile", contract_path])
+    breakpoint()
```
fubuloubu, line 53: remove this!

fubuloubu (approved):

fubuloubu (approved):

PR #408, opened 2022-01-17
fix: adhere to ethpm manifest compilation expectations

unparalleled-js:
### What I did

Fix ethpm-type manifest related validation errors.
The source IDs were full OS paths and should be relevant to the project `contracts/` directory.
This PR  fixes that!

What it was before:

```json
{"sources": {"Users/Jules/Projects/ApeProject/contracts/Fund.sol": ...}, {"Users/Jules/Projects/ApeProject/contracts/nested/Foo.sol": ...}}}
```

What it is now:

```json
{"sources": {"Fund.sol": ...}, {"nested/Foo.sol": ...}}
```

### How I did it

Adjust PM compiler to set source ID as relevant from the `contracts/` directory. Since we don't have access to the `ProjectManager` in the `CompilerAPI`, I made the `base_path` a config property in the `ape-pm` plugin that defaults to `Path("contracts/"). Some other general tweaking to handle this.

### How to verify it

tests passing, make sure this makes sense too.

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -106,6 +106,6 @@ class CompilerManager:
 
     def _get_contract_path(self, path: Path):
         try:
-            return path.relative_to(self.config.PROJECT_FOLDER)
+            return path.relative_to(self.config.PROJECT_FOLDER / "contracts")
         except ValueError:
             return path
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -81,7 +81,10 @@ class CompilerManager:
             for path in paths_to_compile:
                 logger.info(f"Compiling '{self._get_contract_path(path)}'.")
 
-            compiled_contracts = self.registered_compilers[extension].compile(paths_to_compile)
+            base_path = self.config.PROJECT_FOLDER / Path("contracts")
+            compiled_contracts = self.registered_compilers[extension].compile(
+                paths_to_compile, base_path=base_path
+            )
 
             for contract_type in compiled_contracts:
 
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -10,7 +10,7 @@ from ethpm_types.utils import compute_checksum
 from ape.contracts import ContractContainer
 from ape.exceptions import ProjectError
 from ape.managers.networks import NetworkManager
-from ape.utils import get_all_files_in_directory, github_client
+from ape.utils import get_all_files_in_directory, get_relative_path, github_client
 
 from .compilers import CompilerManager
 from .config import ConfigManager
@@ -187,8 +187,12 @@ class ProjectManager:
         files: List[Path] = []
         base_path = self.contracts_folder.parent
         for extension in self.compilers.registered_compilers:
-            src_files = self.contracts_folder.rglob("*" + extension)
-            files.extend([Path(str(s).split(str(base_path))[-1].strip("/")) for s in src_files])
+            files.extend(
+                map(
+                    lambda f: get_relative_path(f, base_path),
+                    self.contracts_folder.rglob("*" + extension),
+                )
+            )
 
         return files
 
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -17,18 +17,6 @@ from .config import ConfigManager
 from .converters import ConversionManager
 
 
-def _create_source_dict(contracts_paths: Collection[Path]) -> Dict[str, Source]:
-    return {
-        str(source): Source(  # type: ignore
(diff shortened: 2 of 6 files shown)


src/ape/managers/project.py
```diff
@@ -185,8 +185,10 @@ def sources(self) -> List[Path]:
             List[pathlib.Path]: A list of a source file paths in the project.
         """
         files: List[Path] = []
+        base_path = self.contracts_folder.parent
         for extension in self.compilers.registered_compilers:
-            files.extend(self.contracts_folder.rglob("*" + extension))
+            src_files = self.contracts_folder.rglob("*" + extension)
+            files.extend([Path(str(s).split(str(base_path))[-1].strip("/")) for s in src_files])
```
fubuloubu, line 191: https://github.com/ApeWorX/ape/blob/main/src/ape/utils.py#L75

fubuloubu: so maybe
```py
files.extend(
    map(
        lambda f: get_relative_path(f, base_path),
        self.contracts_folder.rglob("*" + extension),
    )
)
```

src/ape/managers/project.py
```diff
@@ -10,25 +10,13 @@
...
 from ape.managers.networks import NetworkManager
-from ape.utils import get_all_files_in_directory, github_client
+from ape.utils import get_all_files_in_directory, get_relative_path, github_client
 
 from .compilers import CompilerManager
 from .config import ConfigManager
 from .converters import ConversionManager
 
 
-def _create_source_dict(contracts_paths: Collection[Path]) -> Dict[str, Source]:
```
unparalleled-js, line 20: moved into class because needed the `self.contract_path`

src/ape_pm/compiler.py
```diff
@@ -5,11 +5,17 @@
 from ethpm_types import ABI, ContractType
 from pydantic import parse_obj_as
 
-from ape.api import CompilerAPI
+from ape.api import CompilerAPI, ConfigItem
 from ape.exceptions import CompilerError
 
 
+class InterfaceCompilerConfig(ConfigItem):
+    base_path: Path = Path("contracts")
```
fubuloubu, line 13: I like this addition, but it might need to be in core to look right w/ the config e.g.

```yaml
pm:
  base_path: contracts
```

versus maybe something more like this:

```yaml
project:
  contracts_folder: ./contracts
```

fubuloubu (dismissed):

src/ape_pm/compiler.py
unparalleled-js: Yeah that definitely looks better. However, we will need a nicer way to get at the `config_manager` in `CompilerAPI`. We have to do this same thing in `ape-solidity` and `ape-vyper`.

Also, is there ever a chance that a user will want to use a different base path for `ape-pm` versus `ape-solidity` / `ape-vyper`?

src/ape/managers/compilers.py
```diff
@@ -81,7 +81,10 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
             for path in paths_to_compile:
                 logger.info(f"Compiling '{self._get_contract_path(path)}'.")
 
-            compiled_contracts = self.registered_compilers[extension].compile(paths_to_compile)
+            base_path = self.config.PROJECT_FOLDER / Path("contracts")
```
unparalleled-js, line 84: If we make the contracts path also config drive, we can get this from there and it an default to this

fubuloubu: Turn this into an issue? to add config for this

src/ape_pm/compiler.py
```diff
@@ -19,20 +20,27 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
...
         contract_types: List[ContractType] = []
         for path in filepaths:
             with path.open() as f:
                 data = json.load(f)
 
+            source_id = (
+                str(get_relative_path(path, base_path))
+                if base_path and path.is_absolute()
+                else str(path)
+            )
```
fubuloubu, line 35: so this will need to be upstreamed to `ape-solidity` and `ape-vyper`?

fubuloubu (approved):

src/ape_pm/compiler.py
unparalleled-js: Yes, along with the new API parameter

src/ape/managers/compilers.py
unparalleled-js: Done! https://github.com/ApeWorX/ape/issues/409

fubuloubu (approved):

Issue #400, opened 2022-01-12
Unprocessed Config Items are an Error, not a Warning


fubuloubu:
If there are unrecognized keys in the config file, `ape` currently bails out during `ape plugins install`:

```bash
$ ape plugins install
ERROR: (ConfigError) Unprocessed config items.
```

This is the config file I was trying to use. `solidity:import_remappings` is the unrecognized key.

```yaml
plugins:
  - name: solidity

dependencies:
  OpenZeppelin: OpenZeppelin/openzeppelin-contracts@3.1.0

solidity:
    import_remappings:
      - "@openzeppelin=OpenZeppelin/openzeppelin-contracts@3.1.0"
```

What should instead happen is that the install proceeds, perhaps by noting the unrecognized keys during install, and then later on the keys are fine after the install because the plugin can now process them.

unparalleled-js:
If they key are not recognized by the end of the install though, we still want to raise?
(Like actual unrecognized keys)

fubuloubu:
> If they key are not recognized by the end of the install though, we still want to raise? (Like actual unrecognized keys)

Naw, just warnings until it's all clean. I don't see how we could handle those two separately without also checking the install status of plugins and whether they exist. Way easier to just raise on misspelled keys

PR #413, opened 2022-01-20
fix: warn when unprocessed config items instead of error

unparalleled-js:
### What I did

fixes: #400 
Fixes issue preventing you to use an `ape-config.yaml` to install plugins when that config file contains keys that those plugins use.

Additionally made some small fixes I noticed in `ape plugins li

### How I did it

Change error to logger warning

### How to verify it

`ape plugins install`
`ape plugins uninstall`

### Checklist

### DIFF
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -110,11 +110,13 @@ class ConfigManager:
 
             configs[plugin_name] = config
 
-        if len(user_config.keys()) > 0:
+        remaining_keys = list(user_config.keys())
+
+        if len(remaining_keys) > 0:
+            remaining_keys_str = ", ".join(remaining_keys)
             logger.warning(
-                "Not all config items have been processed yet. "
-                "If warning persists after 'ape plugins install', "
-                "there are unrecognized config keys."
+                f"Unprocessed plugin config(s): {remaining_keys_str}. "
+                "Plugins may not be installed yet or may be mis-spelled."
             )
 
         self._plugin_configs_by_project[project_name] = configs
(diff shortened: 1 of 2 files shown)


src/ape_plugins/_cli.py
```diff
@@ -168,7 +168,7 @@ def add(cli_ctx, plugin, version, skip_confirmation, upgrade):
             if result == 0 and is_plugin_installed(plugin):
                 cli_ctx.logger.success(f"Plugin '{plugin}' has been upgraded.")
             else:
-                cli_ctx.logger.error(f"Failed to add '{plugin}'.")
```
unparalleled-js, line 171: changed because sometimes the plugin still gets installed / updated even though it was not completely successful.

src/ape/managers/config.py
```diff
@@ -110,7 +111,11 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
             configs[plugin_name] = config
 
         if len(user_config.keys()) > 0:
-            raise ConfigError("Unprocessed config items.")
+            logger.warning(
+                "Not all config items have been processed yet. "
+                "If warning persists after 'ape plugins install', "
+                "there are unrecognized config keys."
```
fubuloubu, line 117: would it make sense to display the keys?

fubuloubu (approved):

src/ape/managers/config.py
unparalleled-js: Yes, I just improved this a bit

fubuloubu (approved):

Issue #396, opened 2022-01-08
Bug: Unable to have script with same name as python-framework module-name, such as "site"


unparalleled-js:
### Environment information

Latest ape, 0.1.0b3

* Python Version: 3.9.7
* OS: macOS

### What went wrong?

1. I created a script named `site.py` which is intended to just launch a simple forms site.
2. I tried to run the script using `ape run site`. It was finding a `main` method and calling it, however it was not _my_ main method.

`ape` is not really intended to run python framework module `main` methods, even incidentally, so this seems like a bug.

### How can it be fixed?

Force import of local module rather than blanket import.

unparalleled-js:
For example, if forever some reason `python` creates a standard framework named `deploy`, we would no longer be able to have a script named `deploy`.

Should be an easy fix though? Just need to think a little about ramifications.

fubuloubu:
Bronwie does it really differently:
https://github.com/eth-brownie/brownie/blob/master/brownie/_cli/run.py
https://github.com/eth-brownie/brownie/blob/master/brownie/project/scripts.py

unparalleled-js:
They do call `import_module` like we do at some point: https://github.com/eth-brownie/brownie/blob/master/brownie/project/scripts.py#L149

unparalleled-js:
After digging in more to have brownie works, I have figured out that brownie will actually register your whole absolute path, import the full path to the script via `import_module`. So very similar to ours. I have implemented this in ape and it works (including a test).

unparalleled-js:
I think maybe we could just do this via `scripts.name` ? That is a little better perhaps.

PR #415, opened 2022-01-20
fix: issue preventing scripts from having the same name as a package

unparalleled-js:
### What I did

fixes: #396 

Fix bug preventing scripts to have names like `site.py` because `site` is a built-in python package.

### How I did it

Copied brownie.
`brownie` seems to register the user's full path, cache imports, and import the module via its full-path.

### How to verify it

Are able to run scripts with names such as `site.py`.
Are able to run any scripts you could before.

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -1,4 +1,6 @@
 import json
+import sys
+from importlib import import_module
 from pathlib import Path
 from typing import Collection, Dict, List, Optional, Union
 
@@ -12,6 +14,7 @@ from ape.exceptions import ProjectError
 from ape.managers.networks import NetworkManager
 from ape.utils import get_all_files_in_directory, get_relative_path, github_client
 
+from ..logging import logger
 from .compilers import CompilerManager
 from .config import ConfigManager
 from .converters import ConversionManager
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -6,8 +6,8 @@
 import ape
 
 # NOTE: Ensure that we don't use local paths for these
-ape.config.DATA_FOLDER = Path(mkdtemp())
-ape.config.PROJECT_FOLDER = Path(mkdtemp())
+ape.config.DATA_FOLDER = Path(mkdtemp()).resolve()
```
unparalleled-js, line 9: ran into crazy weirdness because this is a symlink but resolving the symlink to the real path helped

src/ape/managers/project.py
```diff
@@ -441,6 +444,69 @@ def _create_source_dict(self, contracts_paths: Collection[Path]) -> Dict[str, So
...
+
+        script_path = get_relative_path(script_file, Path.cwd())
+        script_parts = script_path.parts[:-1]
+
+        if any(p == ".." for p in script_parts):
+            raise ProjectError("Cannot execute script from outside current directory")
+
+        # Add to Python path so we can search for the given script to import
+        root_path = Path(".").resolve().root
+        sys.path.append(root_path)
```
unparalleled-js, line 479: So now we register the full absolute path.
Then, we import the full path (line 483).

This is how brownie does it and it seems to work 🤷🏻‍♀️

fubuloubu (approved):

Issue #150, opened 2021-09-17
bug: Exception: Not a valid ABI interface JSON file


sabotagebeats:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
# ...copy and paste result of above command here...

$ ape plugins list
# ...copy and paste result of above command here...
```

* Python Version: x.x.x
* OS: osx/linux/win

### What went wrong?
ape looking for json files in compile makes it crash if the json is not abi

Please include information like:

* what command you ran
`ape compile`
* the code that caused the failure (see [this link](https://help.github.com/articles/basic-writing-and-formatting-syntax/) for help with formatting code)
* full output of the error you received
`Exception: Not a valid ABI interface JSON file`

### How can it be fixed?
make jsons be ignored if they do not compile to abi

fubuloubu:
Solution: when attempting to compile a JSON file that is not the correct format, skip compiling

PR #420, opened 2022-01-23
fix: bug preventing from solidity dependencies to work properly

unparalleled-js:
### What I did

fixes: #150 (not pushed yet)
Issue preventing dependencies from working correctly.
https://github.com/ApeWorX/ape-solidity/pull/20

A change that is part of this and something we wanted is that now, when we download a dependency from Github, instead of caching all the files, we only cache the manifest.

In `ape-solidity` (PR Linked above), we then de-compile the files from the manifest and put them in the `contracts/.cache` directory. Since only `solidity` needs this, it makes sense that it happens in the plugin.

### How I did it

Moves dependencies to `<project>/contracts/.dependencies` folder. Reason: This way, are in the base-path, which seems to be needed.  We still do the mapping (basically ignoring `.dependencies` part). This feels a lot like `remix`.### How to verify it

`simple-nft` project

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -68,8 +68,14 @@ class ProjectManager:
 
     def _extract_dependency_manifest(self, name: str, download_path: str) -> PackageManifest:
         target_path = self.config.packages_folder / name
-        target_path.mkdir(exist_ok=True, parents=True)
         manifest_file_path = target_path / "manifest.json"
+
+        # Handles migrating older ape when we cached the entire project
+        # rather than just the manifest file.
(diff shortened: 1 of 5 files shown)


src/ape/managers/project.py
```diff
@@ -54,17 +56,17 @@ def __post_init__(self):
         if isinstance(self.path, str):
             self.path = Path(self.path)
 
+        config = self.config.load()
```
unparalleled-js, line 59: bug 1: top-level configs were no longer loading in time. This forces it :/

fubuloubu (dismissed):

fubuloubu (approved):

Issue #423, opened 2022-01-24
Empty Contract Check Not Performed Before Contract Call


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.0b5.dev8+g7bb0af8
$ ape plugins list
Installed Plugins:
  ens          0.1.0b2
  tokens       0.1.0b3
  infura       0.1.0b2
```

* Python Version: 3.9.9
* OS: linux

### What went wrong?

When querying a contract that doesn't actually exist, `ape` attempts the call but then gets a decoder error when the contract returns 0 bytes.

Example:

```py
$ ape console --network :ropsten:infura

In [1]: from ape_tokens import tokens

In [2]: yfi = tokens["YFI"]

In [3]: yfi.balanceOf("ychad.eth")
---------------------------------------------------------------------------
InsufficientDataBytes                     Traceback (most recent call last)
File /site-packages/ape_ethereum/ecosystem.py:257, in Ethereum.decode_calldata(self, abi, raw_data)
    256 try:
--> 257     vm_return_values = abi_decode(output_types, raw_data)
    258     if not vm_return_values:

File /site-packages/eth_abi/codec.py:181, in ABIDecoder.decode_abi(self, types, data)
    179 stream = self.stream_class(data)
--> 181 return decoder(stream)

File /site-packages/eth_abi/decoding.py:127, in BaseDecoder.__call__(self, stream)
    126 def __call__(self, stream: ContextFramesBytesIO) -> Any:
--> 127     return self.decode(stream)

File /site-packages/eth_utils/functional.py:45, in apply_to_return_value.<locals>.outer.<locals>.inner(*args, **kwargs)
     43 @functools.wraps(fn)
     44 def inner(*args, **kwargs) -> T:  # type: ignore
---> 45     return callback(fn(*args, **kwargs))

File /site-packages/eth_abi/decoding.py:173, in TupleDecoder.decode(self, stream)
    172 for decoder in self.decoders:
--> 173     yield decoder(stream)

File /site-packages/eth_abi/decoding.py:127, in BaseDecoder.__call__(self, stream)
    126 def __call__(self, stream: ContextFramesBytesIO) -> Any:
--> 127     return self.decode(stream)

File /site-packages/eth_abi/decoding.py:198, in SingleDecoder.decode(self, stream)
    197 def decode(self, stream):
--> 198     raw_data = self.read_data_from_stream(stream)
[...]

PR #426, opened 2022-01-24
fix: check if contract prior to transacting

unparalleled-js:
### What I did

fixes: #423 

### How I did it

When initializing a `ContractInstance`, make the check then.

### How to verify it

Using the follow code:

```python
from ape_tokens import tokens
yfi = tokens["YFI"]
yfi.decimals()
```

Should work:

```bash
ape console --network :mainnet:infura
```

Should fail:

```bash
ape console --network :ropsten:infura
```

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -106,7 +106,7 @@ class ContractCall:
 class ContractCallHandler:
     provider: ProviderAPI
     converter: "ConversionManager"
-    address: "ContractInstance"
+    contract: "ContractInstance"
     abis: List[ABI]
 
     def __repr__(self) -> str:
@@ -117,9 +117,9 @@ class ContractCallHandler:
         return self.converter.convert(v, tuple)
 
     def __call__(self, *args, **kwargs) -> Any:
-        if not self.address.is_contract:
+        if not self.contract.is_contract:
             network = self.provider.network.name
-            raise _get_non_contract_error(self.address.address, network)
+            raise _get_non_contract_error(self.contract.address, network)
 
         args = self._convert_tuple(args)
         selected_abi = _select_abi(self.abis, args)
@@ -128,7 +128,7 @@ class ContractCallHandler:
 
         return ContractCall(  # type: ignore
             abi=selected_abi,
-            address=self.address.address,
+            address=self.contract.address,
             provider=self.provider,
             converter=self.converter,
         )(*args, **kwargs)
@@ -183,7 +183,7 @@ class ContractTransaction:
 class ContractTransactionHandler:
     provider: ProviderAPI
     converter: "ConversionManager"
-    address: "ContractInstance"
+    contract: "ContractInstance"
     abis: List[ABI]
 
     def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


src/ape/contracts/base.py
```diff
@@ -231,6 +236,13 @@ class ContractInstance(AddressAPI):
     _converter: "ConversionManager"
     _contract_type: ContractType
 
+    def __post_init__(self):
+        if not self.is_contract:
```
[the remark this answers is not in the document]
unparalleled-js: K that makes sense but currently `ContractCall` and `ContractTransaction` do not have access to the `AddressAPI`. They get initialized with `AddressType`. We could instead pass along the `ContractInstance` in place of the `AddressType` as a means of refactor to handle this situation. Else, we can make the exception happen in `ContractInstance.__getattr__()`, which may work but is a little strange I think because the error would raise without actually calling the method.

src/ape/contracts/base.py
```diff
@@ -271,6 +276,13 @@ def __getattr__(self, attr_name: str) -> Any:
             any: The return value from the contract call, or a transaction receipt.
         """
 
+        if not self.is_contract:
```
unparalleled-js, line 279: Note: Making this raise in `__getattr__` also causes the error to occur if just doing `MyContract.method` (without calling it). That may not be ideal. But I am struggling to find a way to get this to work with the call/transaction handler classes.

We could add this to the provider level but then we have to add it to basically all the providers.

fubuloubu: I mean on the actual `Contract*Handler.__call__(*args, **kwargs)` method have it raise if it is not a contract.

https://github.com/ApeWorX/ape/blob/main/src/ape/contracts/base.py#L114
https://github.com/ApeWorX/ape/blob/main/src/ape/contracts/base.py#L187

fubuloubu: Like, the interface is defined just fine, but when you make the call (e.g. `my_contract.myMethod(sender=a)`) have it raise there

Also consider adding a `kwarg` named `skip_contract_check=False` to those methods, so you can override this behavior for... reasons

unparalleled-js: `ContractCallHandler` and `ContractTransactionHandler` do have a reference to the `AddressAPI` version.

Maybe I can refactor so that they do.

src/ape/contracts/base.py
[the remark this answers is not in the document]
fubuloubu: Ah, I see what you mean now.

How bad would it be to pass the full `ContractInstance` object as a backwards ref?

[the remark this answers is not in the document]
unparalleled-js: I just did and it works okay, I am just keeping the type `AddressAPI` since all we need is the `is_contract` method

src/ape/contracts/base.py
```diff
@@ -106,7 +106,7 @@ def __call__(self, *args, **kwargs) -> Any:
 class ContractCallHandler:
     provider: ProviderAPI
     converter: "ConversionManager"
-    address: AddressType
+    address: AddressAPI
```
fubuloubu, line 109: At this point, should it be: `contract_instance: ContractInstance` instead? We actually know the link backwards is 100% always a `ContractInstance`

fubuloubu (dismissed):

src/ape/contracts/base.py
unparalleled-js: I suppose. It is just weird because even though it is technically a contract instance, is not necessarily a contract. And the API we need is only in the `AddressAPI` so imo `ContractInstance` is unnecessarily specific.

But yeah, since it is always that (currently), I can make that change.

It is easier to make types more specific than it is to do the opposite, so if this ever needs to just be `AddressAPI`, going backwards could be a breaking change

fubuloubu (dismissed):

fubuloubu (approved):

Issue #409, opened 2022-01-18
enhancement: allow configuring contracts/ directory 


unparalleled-js:
### Overview

Right now, a project directory can be configured. The `contracts/` directory is assumed to be `<project-dir>/contracts`.

We should allow configuring the contracts directory but defaulting to `<project-dir>/contracts`.

### Specification

* Add `contracts_folder` property to `ConfigManager`
* Replace all usages of a path ending in `contracts/` off the project, (`ProjectManager`, `CompilerManager`, etc.)

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

unparalleled-js:
Wondering if we should allow a list:

```yaml
contracts:
- src/contracts
- src/other-contracts-dir
- ~/global-contracts
```

fubuloubu:
> Wondering if we should allow a list:
> 
> ```yaml
> contracts:
> - src/contracts
> - src/other-contracts-dir
> - ~/global-contracts
> ```

That's interesting. I haven't seen a project with more than one contracts directory though, although I suppose foundry repos typically use git submodules to link to ds-test and other dependencies

PR #429, opened 2022-01-24
feat: allow configuring a global contracts path

unparalleled-js:
### What I did

fixes: #409 

### How I did it

Grab / create the path using the value from the conifg `contracts_folder:` property.
If the value does not exist, default to what it was before (`self.config.PROJECT_FOLDER / "contracts"`)

### How to verify it

* Create a project. Make things work as they did before.
* Set `contracts_folder: src` in your `ape-config.yaml` and rename `contracts/` to `src/`. Check that things still work. Recompile, etc.
* Renamed `src` back to `contracts`. Now, things should fail and it should not find your contracts.
* Create a global directory, like `$HOME/globalContracts`. Add contracts there. Set in your config the path to there (absolute). Notice things work - it compiles the global contracts.

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -3,6 +3,22 @@
 An `ape-config.yaml` file allows you to configure ape. This guide serves as an index of the settings you can include 
 in an `ape-config.yaml` file.
 
+## Contracts Folder
+
+Specify a different path to your `contracts/` directory.
+This is useful when using a different naming convention, such as `src/` rather than `contracts/`.
+
+```yaml
+contracts_folder: src
+```
+
+You can also use an absolute path.
+This is useful for projects that compile contracts outside their directory.
+
+```yaml
+contracts_folder: "~/GlobalContracts"
+```
+
 ## Deployments
 
 Share import deployments to public networks with your teammates:
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -75,9 +75,8 @@ class ConfigManager:
             contracts_folder_value = Path(user_config.pop("contracts_folder")).expanduser()
 
             # Attempt to resolve the path in the case it is relative to the project directory.
+            # NOTE: It is okay for this directory not to exist at this point.
             contracts_folder = Path(contracts_folder_value).resolve()
-            if not contracts_folder.exists() or not list(contracts_folder.iterdir()):
-                logger.warning(f"No sources file found in '{contracts_folder}'.")
         else:
             contracts_folder = self.PROJECT_FOLDER / "contracts"
 
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -76,8 +76,8 @@ class ConfigManager:
 
             # Attempt to resolve the path in the case it is relative to the project directory.
             contracts_folder = Path(contracts_folder_value).resolve()
-            if not contracts_folder.exists():
-                raise ConfigError(f"Contracts path '{contracts_folder}' does not exist.")
+            if not contracts_folder.exists() or not list(contracts_folder.iterdir()):
+                logger.warning(f"No sources file found in '{contracts_folder}'.")
         else:
             contracts_folder = self.PROJECT_FOLDER / "contracts"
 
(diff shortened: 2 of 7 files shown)


docs/userguides/config.md
```diff
@@ -3,6 +3,22 @@
...
+
+```yaml
+contracts_folder: src
+```
+
+You can also use an absolute path.
+This is useful for projects that compile contracts outside their directory.
+
+```yaml
+contracts_folder: "~/GlobalContracts"
```
fubuloubu, line 19: Not sure I've seen a use case for this (global contracts dir), although it is probably very useful for referencing other projects locally as well

src/ape_compile/_cli.py
```diff
@@ -35,7 +35,8 @@ def cli(cli_ctx, file_paths, use_cache, display_size):
     a project is loaded. You do not have to manually trigger a recompile.
     """
     if not file_paths and cli_ctx.project.sources_missing:
-        cli_ctx.logger.warning("No 'contracts/' directory detected")
+        contracts_dir_name = cli_ctx.project.config.contracts_folder.name
+        cli_ctx.logger.warning(f"No '{contracts_dir_name}/' directory detected")
```
fubuloubu, line 39: Doesn't this happen here now?
https://github.com/ApeWorX/ape/pull/429/files#diff-140ab16c725a530ed14f1900c24c98f2f3fc4b9ec4b96a0c439f0daeaf02de26R74

fubuloubu (approved):

src/ape_compile/_cli.py
unparalleled-js: Yeah, this is always `"contracts"` from what I can tell for that reason. It does not hurt to be safe but am fine with removing it too.

docs/userguides/config.md
unparalleled-js: Yeah, you can quickly "extend" a project this way. Maybe for like a set of experimental or custom scripts you have.

fubuloubu: Sure! I don't see why not to allow it, just have to see what people do with this feature lol

src/ape_compile/_cli.py
fubuloubu: yeah seems redundant now

unparalleled-js: Okay, I was wrong, so this is needed because you also get this error if the directory exists but is empty.

Could refactor so we check if it empty as config-load time though... Hmmmm...

unparalleled-js: I put it back. I think it is ok for it to be empty at other points ape but just not here.

src/ape/managers/config.py
```diff
@@ -69,6 +70,18 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
...
         self.name = user_config.pop("name", "")
         self.version = user_config.pop("version", "")
+
+        if "contracts_folder" in user_config:
+            contracts_folder_value = Path(user_config.pop("contracts_folder")).expanduser()
+
+            # Attempt to resolve the path in the case it is relative to the project directory.
+            contracts_folder = Path(contracts_folder_value).resolve()
+            if not contracts_folder.exists():
+                raise ConfigError(f"Contracts path '{contracts_folder}' does not exist.")
```
unparalleled-js, line 80: Note: this does not ensure there are actually contracts in the directory - it could be empty. It seems better to wait until the user tries to compile before doing that, which is why we still have a check i `ape_compile._cli.py::compile`

unparalleled-js: HMMMMMMM

maybe it is just most consistent to _not_ have this check....

unparalleled-js:
NEW PLAN

* I think it is best to be okay with the contracts directory not existing a config load time. It is only critical that it exists at compile time. Reason: This maintain consistent behavior when the user _does not_ specify a contracts directory in their config file... Like, `ape` never warned or cared before when your project had no `contracts/` directory so maybe it should not now either, even when using a different naming convention, etc.

fubuloubu (approved):

Issue #398, opened 2022-01-12
Dependencies: Refactor dependencies config into key-value object


fubuloubu:
Currently the keys are like:

```yaml
dependencies:
  MyDependency: github-org/github-repo@1.2.3
```

Which is not very scaleable to other types of dependencies from other locations

Suggestion to break it out a bit:
```yaml
dependencies:
  - name: MyDependency
    github: github-org/github-repo
    version: 1.2.3
```

At first, only `github` and `version` keys are specified. Eventually relax the requirement of `version` key to allow alternatives like `branch`, and then eventually add other types of dependencies such as `gitlab`, `ipfs`, `ipns`, etc.

unparalleled-js:
Talked elsewhere but the 'type' is not really needed, the key is the type. yaml.

fubuloubu:
> Talked elsewhere but the 'type' is not really needed, the key is the type. yaml.

some "types" might not be excplicit, like github vs. gitlab, IPFS links, etc.

although I guess write up more what you mean in the resulting yaml?

unparalleled-js:
> although I guess write up more what you mean in the resulting yaml?

I was only referring to how we use the schema to know the type, thinking about our convo regarding the pydantic config utility. We should be able to migrate to that easily if needed! I have PR in my local fork that resolves this issue (will open after #420)

PR #431, opened 2022-01-25
refactor: plugin system for projects and dependencies

unparalleled-js:
### What I did

fixes: #398

### Checklist


src/ape/managers/config.py
```diff
@@ -69,7 +82,27 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
...
+        if not isinstance(dependencies, list):
+            raise ConfigError("'dependencies' config item must be a list of dicts.")
+
+        decoded_deps = []
+        dep_config_cls: Union[Type[IPFSDependency], Type[GithubDependency]]
+        for dep in dependencies:
+            if "ipfs" in dep:
+                dep_config_cls = IPFSDependency
+            elif "github" in dep:
+                dep_config_cls = GithubDependency
```
fubuloubu, line 96: check out this: https://pydantic-docs.helpmanual.io/usage/model_config/#smart-union

src/ape/managers/project.py
```diff
@@ -59,14 +60,16 @@ def __post_init__(self):
...
-            for n, dep_id in config.dependencies.items()
+            d.name: self._extract_dependency_manifest(d) for d in config.dependencies
         }
 
     def __repr__(self):
         return "<ProjectManager>"
 
-    def _extract_dependency_manifest(self, name: str, download_path: str) -> PackageManifest:
+    def _extract_dependency_manifest(
+        self, dependency_item: Union[IPFSDependency, GithubDependency]
```
fubuloubu, line 70: See? I knew this would come in handy somewhere else!

Maybe our new union type should live in `ape.types` :thinking:

src/ape/managers/project.py
```diff
@@ -86,30 +89,28 @@ def _extract_dependency_manifest(self, name: str, download_path: str) -> Package
...
+            manifest_dict = self._download_manifest(dependency_item, manifest_file_path)
 
         if "name" not in manifest_dict:
             manifest_dict["name"] = name.replace("_", "-")
 
         return PackageManifest(**manifest_dict)
 
-    def _download_manifest(self, name: str, download_path: str, manifest_target_path: Path) -> Dict:
+    def _download_manifest(
+        self, dependency_item: Union[IPFSDependency, GithubDependency], manifest_target_path: Path
```
fubuloubu, line 100: hehe, yes def `Dependency`

src/ape/managers/project.py
```diff
@@ -86,30 +89,28 @@ def _extract_dependency_manifest(self, name: str, download_path: str) -> Package
...
-            except ValueError:
-                raise ValueError("Invalid Github ID. Must be given as <org>/<repo>@<version>")
-
-            # Download manifest
+        elif isinstance(dependency_item, GithubDependency):
             with tempfile.TemporaryDirectory() as temp_dir:
                 temp_project_path = Path(temp_dir) / name
                 temp_project_path.mkdir(exist_ok=True, parents=True)
+                path = dependency_item.github
+                version = dependency_item.version
```
fubuloubu, line 113: Wonder if this could instead be some sort of `abstractmethod` in `DependencyConfig`, and then each dependency config type must implement the same method for fetching and extracting their content to a manfiest file

https://github.com/ApeWorX/ape/pull/431/files#diff-140ab16c725a530ed14f1900c24c98f2f3fc4b9ec4b96a0c439f0daeaf02de26R26

src/ape/managers/config.py
```diff
@@ -19,6 +19,19 @@ class DeploymentConfig(ConfigItem):
...
+    name: str
+
+
+class IPFSDependency(DependencyConfig):
+    ipfs: str
+
+
+class GithubDependency(DependencyConfig):
+    github: str
+    version: str
```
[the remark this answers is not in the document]
unparalleled-js: This downloads tar-bars though (releases), in which case the version is required I think.

Would you prefer a `tag` + `clone` approach?

src/ape/managers/project.py
unparalleled-js: Yes, I really do like this typing strategy

src/ape/managers/config.py
```diff
@@ -19,6 +33,65 @@ class DeploymentConfig(ConfigItem):
     contract_type: str
 
 
+class DependencyConfig(ConfigItem):
+    name: str
+    _compilers: "CompilerManager" = None  # type: ignore
```
unparalleled-js, line 38: any dependencies that are not in the PM format require the compilers so they can put that way

(looking at you Github)

Not sure if there is a better way?

src/ape/managers/config.py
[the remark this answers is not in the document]
unparalleled-js: Maybe if they do version, use the release. Otherwise, do clone.

docs/userguides/config.md
```diff
@@ -3,6 +3,36 @@
...
+## Dependencies
+
+Configure dependencies for your ape project.
+Ape downloads and caches dependencies in the `.ape/packages/<name>` directory where `<name>` refers to the name of the dependency.
+
+For example:
+
+```yaml
+dependencies:
+  - name: open_zeppelin
```
fubuloubu, line 15: Is this a big deal? lol
```suggestion
  - name: OpenZeppelin
```

docs/userguides/config.md
```diff
@@ -3,6 +3,36 @@
...
+```
+
+will download the [Open Zeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts) package with version `4.4.0`.
+
+To use your dependencies in the `ape-solidity` plugin, configure `import_remappings`:
+
+```yaml
+solidity: 
+  import_remapping:
+    - "@openzeppelin=open_zeppelin"
```
fubuloubu, line 27:
```suggestion
    - "@OpenZeppelin=OpenZeppelin"
```

docs/userguides/config.md
```diff
@@ -3,6 +3,36 @@
...
+```
+
+Now, in your solidity files, import `openzeppelin` sources via:
```
fubuloubu, line 30:
```suggestion
Now, in your solidity files, import `OpenZeppelin` sources via:
```

docs/userguides/config.md
```diff
@@ -3,6 +3,36 @@
...
+
+```solidity
+import "@openzeppelin/token/ERC721/ERC721.sol";
```
fubuloubu, line 33:
```suggestion
import "@OpenZeppelin/token/ERC721/ERC721.sol";
```

fubuloubu (commented):
Just me being OCD with CamelCase lol

src/ape/managers/config.py
```diff
@@ -19,6 +33,65 @@ class DeploymentConfig(ConfigItem):
...
+        Returns:
+            PackageManifest
+        """
+
+
+class IPFSDependency(DependencyConfig):
+    ipfs: str
+
+    def extract_manifest(self) -> PackageManifest:
+        response = requests.get(self.ipfs)
```
fubuloubu, line 54: this probably won't work lol... maybe we skip this one for now, and do `GitLab` instead (since it's basically the same)

src/ape/managers/config.py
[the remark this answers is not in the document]
fubuloubu: yes that sounds right

src/ape/managers/config.py
```diff
@@ -19,6 +33,65 @@ class DeploymentConfig(ConfigItem):
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_project_path = Path(temp_dir) / self.name
+            temp_project_path.mkdir(exist_ok=True, parents=True)
+            github_client.download_package(self.github, self.version, temp_project_path)
+
+            temp_contracts_path = temp_project_path / "contracts"
+            if not temp_contracts_path.exists():
+                raise ConfigError(
+                    "Dependency does not have a supported file structure. "
+                    "Expecting 'contracts/' path."
```
fubuloubu, line 73: a lot of dapptools/foundry projects will use `src/`

src/ape/types/__init__.py
```diff
@@ -39,7 +41,13 @@
...
 """
 
-AddressType = ChecksumAddress  # type: ignore
+DependencyType = Union["IPFSDependency", "GithubDependency"]
+"""
+A project dependency, such as a :class:`ape.managers.config.IPFSDependency`
+or a :class:`ape.managers.config.GithubDependency`.
+"""
+
+AddressType = NewType("AddressType", HexAddress)  # type: ignore
```
fubuloubu, line 50: I like using the `ChecksumAddress` assumption though, at least for processing user input, to ensure they didn't make a mistake typing the address

docs/userguides/config.md
unparalleled-js: HA! I appreciate a fellow OCD-er TBH.

src/ape/types/__init__.py
unparalleled-js: `mypy` just really hates this lol I keep fighting it

src/ape/managers/config.py
unparalleled-js: Good call! Let me find an example make the switch

src/ape/managers/config.py
unparalleled-js: Yeah I was wondering this.

This should be configurable on the `DependencyConfig`...
It can mirror our own `contracts_folder` perhaps

src/ape/managers/config.py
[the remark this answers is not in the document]
unparalleled-js: This is done and works great!
Also fixes a bug preventing you from storing multiple versions of a dependecny.

So now:

```yaml
dependencies:
  - name: open_zeppelin
```

with clone `main` and still wok and I still have my `4.4.0` as well so I can switch back and forth without having to redownload. Feeling nifty.

docs/userguides/config.md
```diff
@@ -19,6 +19,46 @@ This is useful for projects that compile contracts outside their directory.
...
+```
+
+will download the [Open Zeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts) package with version `4.4.0`.
+
+To use your dependencies in the `ape-solidity` plugin, configure `import_remappings`:
+
+```yaml
+solidity: 
+  import_remapping:
+    - "@penzeppelin=OpenZeppelin/4.4.0"
```
fubuloubu, line 43: lulz
```suggestion
    - "@openzeppelin=OpenZeppelin/4.4.0"
```

fubuloubu: p.s. love the version number used there to fight dependency conflicts

docs/userguides/config.md
```diff
@@ -19,6 +19,46 @@ This is useful for projects that compile contracts outside their directory.
...
+```
+
+You can also set the branch and name of the dependency's contract path, e.g.:
+
+```yaml
+dependencies:
+  - name: DappToolsERC20
+    github: dapphub/erc20
+    branch: dappnix
+    contracts_folder: src
```
fubuloubu, line 59: :heart_eyes:

src/ape/managers/config.py
fubuloubu: was picturing this would be `ipfs: Qm....` so just the content id. Can probably use a public resolver for now

fubuloubu: Actually, infura has one too

fubuloubu (dismissed):

src/ape/utils.py
```diff
@@ -368,14 +377,17 @@ def available_plugins(self) -> Set[str]:
...
         """
         Get a release from Github.
 
         Args:
             repo_path (str): The path on Github to the repository,
                             e.g. ``OpenZeppelin/openzeppelin-contracts``.
             version (str): The version of the release to get.
+
+        Returns:
+            GitRelease: The
```
evanblank3, line 390: The?

src/ape/api/projects.py
```diff
@@ -0,0 +1,50 @@
...
+
+from ape.types import DependencyType
+from ape.utils import abstractdataclass, abstractmethod
+
+if TYPE_CHECKING:
+    from ape.managers.compilers import CompilerManager
+
+
+@abstractdataclass
+class DependencyAPI:
```
unparalleled-js, line 13: This is not the `ProjectAPI` we talked about - this refers to like Github, Gitlab, etc.
The Github implementation is found in `ape.managers.project` currently.

src/ape/logging.py
```diff
@@ -102,6 +102,7 @@ def __init__(self):
         self._logger = _logger
         self._web3_request_manager_logger = _get_logger("web3.RequestManager")
         self._web3_http_provider_logger = _get_logger("web3.providers.HTTPProvider")
+        _logger.setLevel(LogLevel.INFO.value)
```
unparalleled-js, line 105: fixes race condition

unparalleled-js:
OKAY this PR has reached the point of massive-ness so I am going to open smaller PRs to help with this.

src/ape/utils.py
```diff
@@ -384,6 +396,32 @@ def get_release(self, repo_path: str, version: str):
...
+                            e.g. ``OpenZeppelin/openzeppelin-contracts``.
+            target_path (Path): The local path to store the repo.
+            branch (Optional[str]): The branch to clone. Defaults to the default branch.
+        """
+
+        repo = self.get_repo(repo_path)
+        branch = branch or repo.default_branch
+        logger.info(f"Cloning branch '{branch}' from '{repo.name}'.")
+
+        clone = pygit2.clone_repository(repo.git_url, str(target_path), checkout_branch=branch)
```
unparalleled-js, line 422: https://github.com/libgit2/pygit2/issues/1119

^ I tried making a progress bar for cloning but was not successful

src/ape/managers/project.py
```diff
@@ -24,11 +25,98 @@
...
 
 from .compilers import CompilerManager
-from .config import ConfigManager
+from .config import ConfigManager, GithubDependencyConfig
 from .converters import ConversionManager
+from .networks import NetworkManager
 
 
-class ProjectManager:
+class BaseProject:
```
unparalleled-js, line 33: So the dependencies and the `ProjectManager` share some things, such as the cache file related logic.
The difference between the two is that the `ProjectManager` has access to the source files whereas the `Dependency` just downloads the source files in a temporary directory and builds the manifest file from that.
Everything related to that is in this file, there a few classes:

`BaseManager`
`ProjectManager`
`DependencyProject` - a base class that any dependency type should be able to use, such as `GitLab`, `IPFS`, etc.
`GithubDependency` - the main implementation we have the downloads or clones from github to build the manifest

unparalleled-js: I am looking for feedback on the design of this.

I have considered we might want to init an `ape.api.projects` module where we can put the `BaseProject` and the `DependencyProject`.

src/ape/managers/compilers.py
```diff
@@ -78,15 +85,18 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
             Dict[str, ``ContractType``]: A mapping of contract names to their type.
         """
 
+        # NOTE: Must cache 'base_path' before checking registered compilers.
```
fubuloubu, line 88: Could do `lrucache` using project_folder?

src/ape/managers/compilers.py
```diff
@@ -95,6 +105,14 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
 
                 contract_types_dict[contract_type.name] = contract_type
 
+        # Fix source IDs for when compiler did not account for base_path,
+        # such as solidity<0.6.9.
```
fubuloubu, line 109: Is there a way for this to become contained in the solidity plugin?

src/ape/managers/project.py
```diff
@@ -249,9 +136,108 @@ def sources_missing(self) -> bool:
...
+
+        contracts = self.load_contracts()
+        if attr_name in contracts:
+            contract_type = contracts[attr_name]
+        elif attr_name in self.dependencies:
+            contract_type = self.dependencies[attr_name]  # type: ignore
+        else:
+            # Fixes anomaly when accessing non-ContractType attributes.
+            # Returns normal attribute if exists. Raises 'AttributeError' otherwise.
+            return self.__getattribute__(attr_name)  # type: ignore
```
fubuloubu, line 233: yes, this was the solution @evanblank3

src/ape/api/projects.py
```diff
@@ -0,0 +1,224 @@
...
+if TYPE_CHECKING:
+    from ape.managers.compilers import CompilerManager
+
+
+class ProjectAPI:
+    """
+    An abstract base-class for working with projects. The top-level
+    :class:`ape.managers.project.ProjectManager` implements this class. This class is
+    also a base-class for dependencies (see :class:`~ape.api.projects.DependencyAPI`,
+    such as from Github, that get cached as manifest files.
```
fubuloubu, line 29: Add note about "this can be extended in a plugin to work with non-Ape projects"

src/ape/managers/project.py
```diff
@@ -64,150 +54,47 @@ class ProjectManager:
...
     """
 
-    def __init__(
-        self,
-        path: Path,
-    ) -> None:
-        self.path = Path(path) if isinstance(path, str) else path
-
     def __repr__(self):
         return "<ProjectManager>"
```
fubuloubu, line 65: Move this to `ProjectAPI` as `f"<{self.__class__.__name__}>"`? Maybe add a field about where it came from or what it's name is or whatever.

src/ape/managers/compilers.py
unparalleled-js: `project_folder` is not an argument. We just need to grab the value (`self.config.contracts_folder`) before calling `registered_compilers`. Because at that point, if we are compiling a dependency, it will reload the config file and we will lose the path set by the `dependencies:` stanza in our root project config. This is only a problem when the dependency has a different named contracts folder than the default.

src/ape/managers/compilers.py
unparalleled-js: yes I found a way

src/ape/managers/config.py
```diff
@@ -83,6 +151,13 @@ def _plugin_configs(self) -> Dict[str, ConfigItem]:
         self.name = user_config.pop("name", "")
         self.version = user_config.pop("version", "")
 
+        dependencies = user_config.pop("dependencies", []) or []
+        if not isinstance(dependencies, list):
+            raise ConfigError("'dependencies' config item must be a list of dicts.")
+
+        # TODO: Handle decoding multiple dependency types
```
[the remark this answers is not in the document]
unparalleled-js: This might be part of the work for getting multiple dependency APIs working together? We will have to make `ConfigManager` a `BaseModel`, which I think makes a lot of sense since it is configuration-file driven.

I will create a ticket for this once we are happy with this design, e.g. the new `ProjectAPI` + `DependencyAPI(ProjectAPI)`.

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+    get_all_files_in_directory,
+    get_relative_path,
+    injected_before_use,
+)
+
+if TYPE_CHECKING:
+    from ape.managers.compilers import CompilerManager
+
+
+class ProjectAPI:
```
unparalleled-js, line 24: At this time, I am not sure what to do as far as making this some sort of BaseModel. Thoughts?

fubuloubu: Seems like `path` would be a good thing to do that for. Right now `path` is a class variable, and this will probably not work too well when you load multiple projects.

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """
+        The name of the project.
+        """
+
+    @property
+    @abstractmethod
+    def cache_folder(self) -> Path:
```
fubuloubu, line 63: Does this need to be a part of the API?

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def __init__(
+        self,
+        path: Path,
+    ) -> None:
+        self.path = Path(path) if isinstance(path, str) else path
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
```
fubuloubu, line 56: This field is in the manifest, wonder if we can just use that

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+    @property
+    @abstractmethod
+    def cache_folder(self) -> Path:
+        """
+        The folder storing the manifest.
+        """
+
+    @property
+    @abstractmethod
+    def version(self) -> str:
```
fubuloubu, line 70: This field is in the manifest, wonder if we can just use that

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        The version of the project.
+        """
+
+    @property
+    def manifest_cachefile(self) -> Path:
```
fubuloubu, line 76: This seems sort of like an internal optimization (to eliminate having to reload every time)

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Returns:
+            pathlib.Path
+        """
+
+        file_name = self.name or "__local__"
+        return self.cache_folder / (file_name + ".json")
+
+    @property
+    def cached_manifest(self) -> Optional[PackageManifest]:
```
fubuloubu, line 90: This seems internal

src/ape/api/projects.py
```diff
@@ -0,0 +1,229 @@
...
+                    hash=compute_checksum(source.read_bytes()),
+                ),
+                urls=[],
+                content=source.read_text(),
+            )
+            for source in contracts_paths
+        }
+
+
+class DependencyAPI(ProjectAPI):
```
fubuloubu, line 160: Was hoping this could be separate from `ProjectAPI`

fubuloubu (commented):
Hmm, in general I was thinking about it like this, because these concepts seem orthogonal:

`DependencyAPI`: parse a config item for a dependency and somehow fetch that project

`ProjectAPI`: parse a project and turn it into a manifest somehow (fetching other dependencies, maybe compiling it, etc.)

`ApeProject(ProjectAPI)` would be our native implementation that handles our types of projects (maybe triggered off of `ape-config.yaml` being in the project?)

`GithubDependency(DependencyAPI)` would be supported out of the box too, might be easy to add a few others

Our `ProjectManager` would parse the local project as an `ApeProject` and know how to resolve their dependencies by using `DependencyAPI` methods to fetch and `ProjectAPI` methods to parse them to a manifest, so all of the dependencies are standardized as a set of manifest files for long-term storage between sessions (and eventually deployment, as we want to deploy all manifests upstream of our local project's)

src/ape/api/projects.py
unparalleled-js: If we make them separate, we have to be able to re-create new compilers, config object, etc. Like a whole new ape instance is needed or else nothing works.

unparalleled-js: this is why I had to make the `compile_external_project` method. Because otherwise, we need to re-initialize a new `CompilerManager` object

src/ape/managers/config.py
unparalleled-js: Looks like smart union does not support compound types yet, and since this is a `List[DependencyType]` it might not work yet. Looks Pydantic plans to support that eventually though.

src/ape/managers/config.py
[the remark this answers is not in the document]
unparalleled-js: It looks like pydantic does not support compound types yet for their smart union.

src/ape/api/projects.py
unparalleled-js: Will get this to work the way you want.

unparalleled-js:
Dependencies drawio

unparalleled-js:
Currently, we don't support downloading from private repos on GitHub. Maybe a `ape-github` first class repo would make sense to keep that out of core ape?

unparalleled-js:
@fubuloubu As far as I can tell, it is about ready, though I should probably come back to it with fresh eyes later on.

Dockerfile
```diff
@@ -4,7 +4,7 @@ COPY . .
 # need to use typing-extensions<4 to deal with issue 387 bug: Docker will not build
 # TODO: Figure out a better solution or wait for it to resolve itself.
 RUN pip install typing-extensions==3.10.0.2
-RUN python3 ./setup.py install
+RUN pip install .
```
fubuloubu, line 7: undo this change, we were playing in `main`

src/ape/api/projects.py
```diff
@@ -0,0 +1,259 @@
...
+        return project_manifest
+
+
+def _load_manifest_from_file(file_path: Path) -> Optional[PackageManifest]:
+    if not file_path.exists():
+        return None
+
+    def handle_corrupt_manifest():
+        logger.warning(f"Existing manifest file '{file_path}' corrupted (re-building).")
+        file_path.unlink()
```
fubuloubu, line 243: I feel like the caller of this should handle deletion, so we don't end up with a weird silent bug

src/ape/api/projects.py
```diff
@@ -0,0 +1,259 @@
...
+
+    except json.JSONDecodeError:
+        handle_corrupt_manifest()
+        return None
+
+    try:
+        return PackageManifest(**manifest_dict)
+    except ValidationError:
+        handle_corrupt_manifest()
+        return None
```
fubuloubu, line 259:
```suggestion
    try:
        manifest_dict = json.loads(file_path.read_text())
        assert isinstance(manifest_dict, dict) and "manifest" in manifest_dict
        return PackageManifest(**manifest_dict)
    except (ValidationError, json.JSONDecodeError, AssertionError):
        logger.warning(f"Existing manifest file '{file_path}' corrupted (re-building).")
        file_path.unlink()
        return None
```

src/ape/managers/project.py
```diff
@@ -4,28 +4,257 @@
...
+    By default, the `:class:`~ape.managers.project.ProjectManager` uses an
+    ``ApeProject`` at the current-working directory.
+    """
+
+    @property
+    def is_valid(self) -> bool:
+        if (self.path / CONFIG_FILE_NAME).exists():
+            return True
+
+        logger.warning(f"'{self.path.name}' is not an 'ApeProject'.")
```
fubuloubu, line 38:
```suggestion
        logger.warning(f"'{self.path.name}' is not an 'ApeProject', but attempting to process as one.")
```

src/ape/managers/project.py
```diff
@@ -4,28 +4,257 @@
...
+        }
+
+        def file_needs_compiling(source: Path) -> bool:
+            path = str(get_relative_path(source, self.contracts_folder))
+
+            if path not in cached_sources:
+                return True  # New file added
+
+            cached = cached_sources[path]
+            cached.compute_checksum(algorithm="md5")
```
fubuloubu, line 105: Hmm, don't really like this (but I know `ethpm-types` would have to be updated)

Feel like it shoudl be `cached.checksum = compute_checksum(cached, algorithm="md5")`

src/ape/managers/project.py
```diff
@@ -4,28 +4,257 @@
...
+
+    plugin_manager: ClassVar[PluginManager] = injected_before_use()  # type: ignore
+    project_manager: ClassVar["ProjectManager"] = injected_before_use()  # type: ignore
+
+    def __init__(self, data_folder: Path):
+        self.DATA_FOLDER = data_folder
+
+    @cached_property
+    def dependency_types(self) -> Dict[str, Type[DependencyAPI]]:
+        dependency_classes = {}
```
fubuloubu, line 238:
```suggestion
        dependency_classes = {
            "github": GithubDependency,
            "local": LocalDependency,
        }
```

src/ape/managers/project.py
```diff
@@ -4,28 +4,257 @@
...
+
+        for _, (config_key, dependency_class) in self.plugin_manager.dependencies:
+            dependency_classes[config_key] = dependency_class
+
+        dependency_classes["github"] = GithubDependency
+        dependency_classes["local"] = LocalDependency
```
fubuloubu, line 244:
```suggestion
```

src/ape/managers/project.py
```diff
@@ -4,28 +4,257 @@
...
+        return dependency_classes
+
+    def decode_dependency(self, config_dependency_data: Dict) -> DependencyAPI:
+        for key, dependency_cls in self.dependency_types.items():
+            if key in config_dependency_data:
+                return dependency_cls(
+                    **config_dependency_data,
+                    project_manager=self.project_manager,
+                    _data_folder=self.DATA_FOLDER,
+                )  # type: ignore
```
fubuloubu, line 254: When pydantic refactor, look into more smartly processing this with `Union` on `ConfigItem`

fubuloubu (approved):
A lot going on here, but looking really good!

src/ape/api/projects.py
unparalleled-js: `assert` statements is only for debug / testing, they really shouldn't be used in python production code.
If you set `__debug__ = False` (like via `python -O`), the `assert` statements get stripped out of there.

unparalleled-js: I will try to refactor this to a single except block nonetheless!

src/ape/managers/project.py
unparalleled-js: I have looked into this already struggled due to the fact that it is list of the union type, and pydantic does not yet support that but they are planning on it

src/ape/managers/project.py
unparalleled-js: We can do:

```python
cached.checksum = Checksum(  # type: ignore
    hash=compute_checksum(cached.content.encode("utf8"), algorithm="md5"),
    algorithm="md5",
)
```
without waiting for an update

unparalleled-js: However, `cachced.content` is optional and `mypy` does not like that

src/ape/api/projects.py
unparalleled-js: Don

src/ape/api/projects.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@abstractdataclass
+class ProjectAPI:
+    """
+    An abstract base-class for working with projects.
+    This class can also be extended to a plugin for supporting non-ape projects.
+    """
+
+    compilers: "CompilerManager"
```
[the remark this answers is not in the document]
unparalleled-js: There is no base API model yet and `injected_before_use()` does not seem to work with `abstractdataclass`.
I am trying to do this but struggling for reasons like this.

Since all the API classes work with way right now, can we do this at the same time as those, as more research is required there.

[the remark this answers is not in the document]
fubuloubu: for sure that works

src/ape/managers/project.py
fubuloubu: Can work on this later, just seems like a method that should be more pure. The in-place mutation always skeeves me out

fubuloubu (dismissed):

src/ape/managers/project.py
```diff
@@ -4,28 +4,271 @@
...
+            return files
+
+        for extension in self.compilers.registered_compilers:
+            files.extend(self.contracts_folder.rglob(f"*{extension}"))
+
+        return files
+
+    def create_manifest(
+        self, file_paths: Optional[List[Path]] = None, use_cache: bool = True
+    ) -> PackageManifest:
```
evanblank3, line 67: Does this need a doc string?

src/ape/managers/project.py
```diff
@@ -452,85 +553,116 @@ def __getattr__(self, attr_name: str) -> ContractContainer:
...
-        return self.path / "scripts"
+        For example, give it ``HelloWorld`` and it returns
+        ``<absolute-project-path>/contracts/HelloWorld.sol``.
 
-    @property
-    def tests_folder(self) -> Path:
-        """
-        The path to the ``tests/`` directory of the project.
+        Another example is to give it ``contracts/HelloWorld.sol`` and it also
+        returns ``<absolute-project-path>/contracts/HelloWorld.sol``.
```
evanblank3, line 596: Would it be more helpful to change the example here to be a different folder than contracts? This allows for something like `another_contract_folder/HelloWorld.sol` and it would return `<absolute-project-path>/another_contract_folder/HelloWorld.sol`?

evanblank3 (approved):

src/ape/managers/project.py
unparalleled-js: By not including one, it will automatically used the doc str from the abstractmethod in `ProjectAPI`, which seems desirable. It is not intuitive that is happening but it cuts down on repetitive documentation

src/ape/managers/project.py
unparalleled-js: Yeah, maybe if it said `<contracts-folder>`, it would be more obvious that it is a variable to the project. Good catch!

src/ape/managers/project.py
evanblank3: Missed that this was at the abstract level, thanks!

evanblank3 (approved):

PR #432, opened 2022-01-25
feat: added timestamp conversion feature

johnson2427:
### What I did
Added `TimestampConversions` class to _conversions and added to __init__.py under plugins

fixes: #

### How I did it
Added the feature, wanted some feedback, have to think of other instances that this may fail
### How to verify it
Testing and help from friends
### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -6,10 +6,9 @@ from dataclassy import dataclass
 from ape.api import AddressAPI, BlockAPI, ProviderAPI, ReceiptAPI
 from ape.exceptions import ChainError, ProviderNotConnectedError, UnknownSnapshotError
 from ape.logging import logger
+from ape.managers.networks import NetworkManager
 from ape.types import AddressType, BlockID, SnapshotID
 from ape.utils import cached_property
-
-from .networks import NetworkManager
 
 
 @dataclass
@@ -196,6 +195,7 @@ class AccountHistory(_ConnectedChain):
 
     _map: Dict[AddressType, List[ReceiptAPI]] = {}
 
+    # TODO: Inject ConversionManager into AccountHistory
     @cached_property
     def _convert(self) -> Callable:
         from ape import convert
@@ -370,8 +370,11 @@ class ChainManager(_ConnectedChain):
         return self.provider.get_block("pending").timestamp
 
     @pending_timestamp.setter
-    def pending_timestamp(self, new_value: int):
-        self.provider.set_timestamp(new_value)
+    def pending_timestamp(self, new_value: str):
+        # TODO: inject ConversionManager into ChainManager
+        from ape import convert
+
+        self.provider.set_timestamp(convert(value=new_value, type=int))
 
     def __repr__(self) -> str:
         props = f"id={self.chain_id}" if self._networks.active_provider else "disconnected"
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -143,7 +143,8 @@ list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 class TimestampConverter(ConverterAPI):
     """
-    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp
+    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp.
+    Must be UTC timezone
     """
 
     def is_convertible(self, value: str) -> bool:
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 from decimal import Decimal
 from typing import Any, Dict, List, Tuple, Type, Union
 
@@ -158,7 +158,11 @@ class TimestampConverter(ConverterAPI):
         return True
 
     def convert(self, value: str) -> int:
-        return int(datetime.strptime(value, "%m-%d-%Y %H:%M:%S").strftime("%s"))
+        return int(
+            datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
+            .replace(tzinfo=timezone.utc)
+            .strftime("%s")
+        )
 
 
 timestamp_converter = TimestampConverter(None, None, None)  # type: ignore
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from decimal import Decimal
 from typing import Any, Dict, List, Tuple, Type, Union
 
@@ -140,6 +141,29 @@ class ListTupleConverter(ConverterAPI):
 list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 
+class TimestampConverter(ConverterAPI):
+    """
+    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp
+    """
+
+    def is_convertible(self, value: str) -> bool:
+        if not isinstance(value, str):
+            return False
+        if " " not in value or len(value.split(" ")) > 2:
+            return False
+        try:
+            datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
+        except ValueError:
+            return False
+        return True
+
+    def convert(self, value: str) -> int:
+        return int(datetime.strptime(value, "%m-%d-%Y %H:%M:%S").strftime("%s"))
+
+
+timestamp_converter = TimestampConverter(None, None, None)  # type: ignore
+
+
 @dataclass
 class ConversionManager:
     """
@@ -169,7 +193,7 @@ class ConversionManager:
         converters: Dict[Type, List[ConverterAPI]] = {
             AddressType: [address_api_converter, hex_address_converter],
             bytes: [hex_converter],
-            int: [],
+            int: [timestamp_converter],
             Decimal: [],
             list: [list_tuple_converter],
             tuple: [list_tuple_converter],
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -159,9 +159,7 @@ class TimestampConverter(ConverterAPI):
 
     def convert(self, value: str) -> int:
         return int(
-            datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
-            .replace(tzinfo=timezone.utc)
-            .strftime("%s")
+            datetime.strptime(value, "%m-%d-%Y %H:%M:%S").replace(tzinfo=timezone.utc).timestamp()
         )
 
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,11 +4,6 @@ repos:
     hooks:
     -   id: check-yaml
 
--   repo: https://github.com/asottile/seed-isort-config
-    rev: v2.2.0
-    hooks:
-    -   id: seed-isort-config
-
 -   repo: https://github.com/pre-commit/mirrors-isort
     rev: v5.9.3
     hooks:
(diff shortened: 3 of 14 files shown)


johnson2427:
Tests failed, I wasn't sure I should add this the the __init__.py file. And there are still some other cases I wanted to work through.

src/ape_ethereum/__init__.py
```diff
@@ -15,6 +15,11 @@ def converters():
     yield int, WeiConversions
 
 
+@plugins.register(plugins.ConversionPlugin)
+def time_converters():
+    yield str, TimestampConversions
+
+
```
fubuloubu, line 22: So, for the easy conversion things, where accessing config manager, network manager, or other conversions (via conversion manager) wasn't necessary, I "cheated" a bit and defined them directly in the conversion manager's file:
https://github.com/ApeWorX/ape/blob/main/src/ape/managers/converters.py#L92

Everything is set None there because they didn't link all of their dependencies like a normal conversion should do

fubuloubu: Also, note for future knowledge, you can only register all of your converters 1 time, so the way to add it here (for other type of converters where dependencies are needed, or where it makes more sense for it to live in the plugin, like with `WeiConversions`) is to add another yield line to `converters` above, not duplicate the function twice e.g.:

```diff
  @plugins.register(plugins.ConversionPlugin)
  def converters():
      yield int, WeiConversions
+     yield str, TimestampConversions
```

src/ape_ethereum/_converters.py
```diff
@@ -37,3 +38,15 @@ def convert(self, value: str) -> int:
...
+class TimestampConversions(ConverterAPI):
+    """Converts seconds to date and time string value"""
+
+    def is_convertible(self, value: int) -> bool:
+        if not isinstance(value, int):
+            return False
+        return True
+
+    def convert(self, value: int) -> str:
+        return datetime.fromtimestamp(value).strftime('%m-%d-%Y %H:%M:%S')
```
fubuloubu, line 52: I was picturing the reverse direction, since timestamps in EVM are always in seconds, but for users they can write their timestamps as ISO strings for arguments that take timestamps:

```py
chain.pending_timestamp = "02-01-2022 12:00:00"
# Converts to an integer value inside of `pending_timestamp.setter`
```

fubuloubu: Note that you actually have to add this `convert(new_value, int)` statement to the chain manager [here](https://github.com/ApeWorX/ape/blob/main/src/ape/managers/chain.py#L373)

This means you have to inject the `converters` Conversion Manager inside of the `chain` Chain Manager when it is created [here](https://github.com/ApeWorX/ape/blob/main/src/ape/__init__.py#L50)

(See [here](https://github.com/ApeWorX/ape/blob/main/src/ape/__init__.py#L57) for an example of this injection)

src/ape_ethereum/__init__.py
johnson2427: I was stuck between this option and what I did. I reverted to this now

src/ape_ethereum/_converters.py
johnson2427: I will reverse these

unparalleled-js: Would it make sense to move `TimestampConversions` to the core-ape converters?

johnson2427: Should I move this from ape_ethereum to managers/converters.py?

We went over this and we were looking at the WeiConverter class, but converters looks like a better place for this

johnson2427: @unparalleled-js I didn't reload the page so I wrote the same question, but I believe we should

fubuloubu: yes, see my other comment, please mv to `src/ape/managers/converters.py`

src/ape/managers/chain.py
```diff
@@ -370,8 +370,12 @@ def pending_timestamp(self) -> int:
         return self.provider.get_block("pending").timestamp
 
     @pending_timestamp.setter
-    def pending_timestamp(self, new_value: int):
-        self.provider.set_timestamp(new_value)
+    def pending_timestamp(self, new_value: str):
+        """
+        Converts date string to timestamp
+        """
+        from ape import convert
```
fubuloubu, line 377: inject `ConversionManager` into `ChainManager`

src/ape/managers/chain.py
```diff
@@ -370,8 +370,12 @@ def pending_timestamp(self) -> int:
...
+        convert(value=new_value, type=int)
```
fubuloubu, line 378: You deleted the line that was doing the work lol
```suggestion
        convert(value=new_value, type=int)
        self.provider.set_timestamp(new_value)
```

src/ape/managers/converters.py
```diff
@@ -170,6 +192,7 @@ def _converters(self) -> Dict[Type, List[ConverterAPI]]:
             AddressType: [address_api_converter, hex_address_converter],
             bytes: [hex_converter],
             int: [],
+            str: [],
```
fubuloubu, line 195: We index here by the output type, not input type. Also, we want to add our new converter here
```suggestion
            int: [timestamp_converter],
```

src/ape_ethereum/__init__.py
```diff
@@ -13,6 +13,7 @@ def config_class():
 @plugins.register(plugins.ConversionPlugin)
 def converters():
     yield int, WeiConversions
+    yield str, TimestampConversions
```
fubuloubu, line 16: remove all changes in this file

fubuloubu (commented):
Almost there!

src/ape/managers/chain.py
fubuloubu: or, leave a `TODO` to do this. @evanblank3 is working on this anyways

src/ape/managers/chain.py
johnson2427: Does this change new_value to a timestamp, or does this return a value?
I think it should be `self.provider.set_timestamp(convert(value=new_value, type=int))`

src/ape/managers/chain.py
johnson2427: I'll put a TODO, but is that what we did in AccountHistory? Line 200 of chain.py?

johnson2427:
What do we do about this linting issue?

src/ape/managers/chain.py
```diff
@@ -370,8 +373,14 @@ def pending_timestamp(self) -> int:
         return self.provider.get_block("pending").timestamp
 
     @pending_timestamp.setter
-    def pending_timestamp(self, new_value: int):
-        self.provider.set_timestamp(new_value)
+    def pending_timestamp(self, new_value: str):
+        """
+        Converts date string to timestamp
```
unparalleled-js, line 378: Note that doc strs in property setters do not show up.
This is probably better as a comment somewhere.

src/ape/managers/converters.py
```diff
@@ -140,6 +141,27 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
 list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 
+class TimestampConverter(ConverterAPI):
+    """Converts seconds to date and time string value"""
```
unparalleled-js, line 145: I think we should include the string format(s) required in the doc here so users know what is possible.

src/ape/managers/converters.py
```diff
@@ -140,6 +141,27 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
...
+
+    def is_convertible(self, value: str) -> bool:
+        if not isinstance(value, str):
+            return False
+        if " " not in value or len(value.split(" ")) > 2:
+            return False
+        try:
+            datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
```
unparalleled-js, line 153: I wish we also supported `%Y-%m-%dT%H:%M:%S.%f`, as this is the format I see all over server timestamps

unparalleled-js: I guess we would lose precision though... hmm...

src/ape/managers/converters.py
```diff
@@ -140,6 +141,27 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
 list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 
+class TimestampConverter(ConverterAPI):
```
unparalleled-js, line 144: Can we make one that also accepts `datetime.datetime` objects?

src/ape/managers/chain.py
fubuloubu: oh, yeah. put a TODO there too

fubuloubu (approved):

fubuloubu:
don't forget to run `pre-commit` before your commits

src/ape/managers/converters.py
johnson2427: Instead of a string? I don't see why not

src/ape/managers/chain.py
johnson2427: not sure you meant to delete 374 and add those lines, so I didn't include that change
I deleted that docstring and left the #TODO

src/ape/managers/chain.py
johnson2427: done

src/ape/managers/converters.py
johnson2427: I didn't add this to this issue though, I think we should add a separate issue for this, what do you think?

johnson2427:
I forgot pre-commit every time. I'll add that to the notes, sorry about that

unparalleled-js:
> I forgot pre-commit every time. I'll add that to the notes, sorry about that

You can make it happen automatically via `pre-commit install` in the repo

src/ape/managers/converters.py
unparalleled-js: Yes a separate issue is just fine.
This way though, we don't have to worry about trying to cover every possible string format because the user can just parse their own `datetime` objects so I think it is helpful.

src/ape/managers/converters.py
```diff
@@ -140,6 +141,29 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
...
+    """
+    Converts string of %m-%d-%Y %H:%M:%S format to a timestamp
+    """
+
+    def is_convertible(self, value: str) -> bool:
+        if not isinstance(value, str):
+            return False
+        if " " not in value or len(value.split(" ")) > 2:
```
unparalleled-js, line 152: Check this out:  https://dateparser.readthedocs.io/en/latest/

^ Not saying we need to use this right now... but this seems like a tool that does what we would ideally like, we can parse any str to a `datetime` object that `dateparser` supports

src/ape/managers/converters.py
johnson2427: completely agree

src/ape/managers/converters.py
johnson2427: The problem with this is it does run into some issues (sometimes 11/12 will be recorded as November 12th, and sometimes its recorded as December 11th). Not a frequent issue, but when making smart contracts, I think this has to be more reliable?

src/ape/managers/converters.py
```diff
@@ -140,6 +141,29 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
 list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 
+class TimestampConverter(ConverterAPI):
+    """
+    Converts string of %m-%d-%Y %H:%M:%S format to a timestamp
```
evanblank3, line 146:
```suggestion
    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp
```

unparalleled-js (dismissed):
LGTM

unparalleled-js (dismissed):

unparalleled-js (approved):

evanblank3 (approved):

PR #434, opened 2022-01-26
refactor: dependency injection for managers

evanblank3:
### What I did

fixes: #

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -22,6 +22,7 @@ from .utils import USER_AGENT
 plugin_manager = _PluginManager()
 """Manages plugins for the current project. See :class:`ape.plugins.PluginManager`."""
 
+_ConfigManager.plugin_manager = plugin_manager
 config = _ConfigManager(  # type: ignore
     # Store all globally-cached files
     DATA_FOLDER=_Path.home().joinpath(".ape"),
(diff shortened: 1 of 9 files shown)


src/ape/utils.py
```diff
@@ -444,6 +444,16 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
     return [path]
 
 
+class dependency(property):
```
unparalleled-js, line 447: I feel like `dependency` is starting to become an overloaded term.

Users may think this has something to do with project dependencies, like Open Zeppelin?

Maybe a better name for this is `placeholder(property)` or something?

src/ape/managers/config.py
```diff
@@ -19,7 +17,6 @@ class DeploymentConfig(ConfigItem):
     contract_type: str
 
 
-@dataclass
```
unparalleled-js, line 22: We either need an `init` method or to make this a data class again because we are failing it initialize it

src/ape/__init__.py
```diff
@@ -31,39 +32,47 @@
...
-    config=config,
-    compilers=compilers,
-    networks=networks,
-    converter=_converters,
-)
+_ProjectManager.config = config
+_ProjectManager.compilers = compilers
+_ProjectManager.networks = networks
+_ProjectManager.converter = _converters
+Project = _partial(_ProjectManager)
```
fubuloubu, line 75: at this point, just import `ProjectManager as Project`

fubuloubu: I also like how this removes partial

fubuloubu: p.s. `Contract` down below (which also is a partial` should probably be a full class, and then we can remove the partial as well through DI

src/ape/utils.py
fubuloubu: `injected_property` is good!

fubuloubu: or how about `to_be_injected` or `injected_before_use`? To show how to use it

src/ape/managers/project.py
```diff
@@ -48,27 +46,27 @@ class ProjectManager:
...
+    compilers: ClassVar[CompilerManager] = dependency()  # type: ignore
     """
     The group of compiler plugins for compiling source files. See
     :class:`~ape.managers.compilers.CompilerManager` for more information.
     Call method :meth:`~ape.managers.project.ProjectManager.load_contracts` in this class
     to more easily compile sources.
     """
 
-    networks: NetworkManager
+    networks: ClassVar[NetworkManager] = dependency()  # type: ignore
```
fubuloubu, line 69: mypy was something I had not considered

src/ape/managers/converters.py
```diff
@@ -176,7 +174,13 @@ def _converters(self) -> Dict[Type, List[ConverterAPI]]:
         }
 
         for plugin_name, (conversion_type, converter_class) in self.plugin_manager.converters:
-            converter = converter_class(self.config.get_config(plugin_name), self.networks, self)
+
+            # dependency injection
+            converter_class.config = self.config.get_config(plugin_name)
+            converter_class.networks = self.networks
+            converter_class.converter = self
```
fubuloubu, line 181: should probably rename this var to `conversion_manager` for clarity, hmm

src/ape/managers/config.py
fubuloubu: Well, now everything around where the injected dependency is also a class var lol. They should be set as instance vars now

unparalleled-js:
@evanblank3 is this PR still needed? Looks like we merged another PR with a lot of this work in it

evanblank3:
> @evanblank3 is this PR still needed? Looks like we merged another PR with a lot of this work in it

Yeah, this was a first attempt at the DI/pydantic work but I'll close it out. Originally I was hoping to finish up the work and then tag this one before closing but it just needs to go. lol

PR #435, opened 2022-01-26
refactor: remove dataclassy from managers

evanblank3:
### What I did
 Removes `dataclassy` from all the managers and makes them regular python classes. Uses `injected_before_use` class to act as a property placeholder for the injected properties

### How I did it
Adds `injected_before_use` class 
 Adds `__init__()` methods to pass the required non-injected properties by keyword.

### How to verify it
All tests run

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -22,6 +22,7 @@ from .utils import USER_AGENT
 plugin_manager = _PluginManager()
 """Manages plugins for the current project. See :class:`ape.plugins.PluginManager`."""
 
+_ConfigManager.plugin_manager = plugin_manager
 config = _ConfigManager(
     # Store all globally-cached files
     data_folder=_Path.home().joinpath(".ape"),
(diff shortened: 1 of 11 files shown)


unparalleled-js (dismissed):
Looks good as a first step to deleting `dataclassy`.

Next step is converting the API classes to pydantic models.

evanblank3:
After discussion with @fubuloubu moving back to the Class level injection of properties:
- only have to set them once
- setting them in __init__() doesn't make them class vars anymore
- fault detection, it raises if it's not set and you try to access it

unparalleled-js:
> After discussion with @fubuloubu moving back to the Class level injection of properties:
> 
> * only have to set them once
> * setting them in **init**() doesn't make them class vars anymore
> * fault detection, it raises if it's not set and you try to access it

I don't really understand these benefits!

* You only have to set them once in `init` as well?
* What is the benefit of them being class vars? Since we are passing in references to the init anyway
* It also raises when you don't properly initialize the manager with one?

src/ape/__init__.py
```diff
@@ -22,52 +22,65 @@
...
+_ProjectManager.compilers = compilers
+_ProjectManager.networks = networks
+_ProjectManager.converter = _converters
 Project = _partial(
     _ProjectManager,
-    config=config,
-    compilers=compilers,
-    networks=networks,
-    converter=_converters,
 )
```
fubuloubu, line 79: let's remove this partial! just change the import to `from .managers.project import ProjectManager as Project`

src/ape/__init__.py
```diff
@@ -22,52 +22,65 @@
...
 """User-facing class for instantiating Projects (in addition to the currently
 active ``project``). See :class:`ape.managers.project.ProjectManager`."""
 
-project = Project(config.PROJECT_FOLDER)
+project = Project(path=config.PROJECT_FOLDER)
 """The currently active project. See :class:`ape.managers.project.ProjectManager`."""
 
 Contract = _partial(_Contract, networks=networks, converters=_converters)
```
fubuloubu, line 86: I'd like to get rid of this partial too, by turning `_Contract` into an object, or somehow just use `ContractInstance` (but I don't think that'd work quite as well.

src/ape/managers/chain.py
```diff
@@ -195,12 +194,9 @@ class AccountHistory(_ConnectedChain):
 
     _map: Dict[AddressType, List[ReceiptAPI]] = {}
 
-    # TODO: Inject ConversionManager into AccountHistory
     @cached_property
     def _convert(self) -> Callable:
-        from ape import convert
-
-        return convert
+        return self._converters.convert
```
fubuloubu, line 199: Nice

src/ape/managers/chain.py
```diff
@@ -314,7 +312,9 @@ def account_history(self) -> AccountHistory:
         A mapping of transactions from the active session to the account responsible.
         """
         if self.chain_id not in self._account_history_map:
-            history = AccountHistory(self._networks)  # type: ignore
+            AccountHistory._networks = self._networks
+            AccountHistory._converters = self._converters
```
fubuloubu, line 316: Should actually be able to do this injection inside of `__init__`! So you don't have to think about it later.

src/ape/managers/config.py
```diff
@@ -47,9 +44,20 @@ class ConfigManager:
     contracts_folder: Path = None  # type: ignore
     dependencies: Dict[str, str] = {}
     deployments: Dict[str, Dict[str, List[DeploymentConfig]]] = {}
-    plugin_manager: PluginManager
+    plugin_manager: ClassVar[PluginManager] = injected_before_use()  # type: ignore
     _plugin_configs_by_project: Dict[str, Dict[str, ConfigItem]] = {}
 
+    def __init__(
+        self,
+        *,
```
fubuloubu, line 52: Why is this here?

fubuloubu:
> * You only have to set them once in `init` as well?

For some of the classes that are not singletons, you would have to do this every time you init. It gets annoying. Class vars are set once and forget!

> * What is the benefit of them being class vars? Since we are passing in references to the init anyway

The managers are singletons, so you don't notice the benefit as much, but here is a good example of how it helps: https://github.com/ApeWorX/ape/pull/435#discussion_r793780770

> * It also raises when you don't properly initialize the manager with one?

This is good fault handling that you forgot something

src/ape/managers/config.py
evanblank3: The * ensures that all params passed after it are done by keyword and not position.

fubuloubu: ah okay. but won't this ignore any args provided?

src/ape/__init__.py
fubuloubu: Also, here is one example where we made ugly code cleaner w/ dependency injection!

src/ape/managers/config.py
evanblank3: It will get very angry about the args, do we want to support *arg, **kwargs on these init methods?

src/ape/__init__.py
evanblank3: I'll need to discuss the requirements/workflow around this change, whenever you have a moment.

src/ape/managers/config.py
fubuloubu: No, I would just remove it if that's the case

src/ape/__init__.py
evanblank3: Opening a ticket specific to this refactor work, removing partial here and moving _Contract creation inside ContractContainer

unparalleled-js (approved):
LGTM

PR #438, opened 2022-01-27
docs: fix issue with address type

unparalleled-js:
### What I did

There is a lot of buggy-ness with trying to link to the docs for `AddressType`, sometimes intermittently and when you least expect.

### How I did it

Instead, just have a few spots where were link to [ChecksumAddress](https://eth-typing.readthedocs.io/en/latest/types.html#checksumaddress)

Otherwise, put things kind of back to the way they were.

Most of this PR is removing dead links.

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -1,7 +1,7 @@
 import sys
 from typing import Union
 
-from eth_typing import ChecksumAddress
+from eth_typing import ChecksumAddress as AddressType
 from ethpm_types import (
     ABI,
     Bytecode,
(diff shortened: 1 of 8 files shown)


src/ape/types/__init__.py
```diff
@@ -39,9 +39,6 @@
 cases.
 """
 
-AddressType = ChecksumAddress  # type: ignore
```
unparalleled-js, line 42: fix = here

fubuloubu (approved):

evanblank3 (approved):

PR #439, opened 2022-01-27
feat: ability to compile external projects

unparalleled-js:
### What I did

*NOTE*: This PR is also a design proposal. I am trying to make comping dependency projects into manifest works better. I have everything working well in this PR: https://github.com/ApeWorX/ape/pull/431.  However, that PR is getting very large so this is part of my effort to try and break it down.

#### Added

* New method `compile_project()` in the `CompilerManager` that takes the project path and the contracts path as well as the sources. The intent is to compile cleanly without the active project's interference. Otherwise, we get weird issues with dependency recursion.
* New method content manager method `using_project` in the `ConfigManager` that temporarily changes the `PROJECT_FOLDER` and the `contract_folder` properties. This is useful for compiling dependencies as well in our tests (`conftest.py` changes).

### How I did it

Freshly re-registering compiling before compiling outsider projects. This way, the configs are all set to their defaults... This does feel hacky, I admit.

### How to verify it

Go outside your project directory and run the following in `ape console`:

```python
source_file = Path("user/ApeProjects/MyProject/contracts/MyContract.sol")
source_file.exists()  # <-- Ensure exists
compilers.compile_project([source_file], source_file.parent.parent, source_file.parent)
INFO: Compiling 'MyContract.sol'.
```

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -8,7 +8,6 @@ from ape.api import CompilerAPI
 from ape.exceptions import CompilerError
 from ape.logging import logger
 from ape.plugins import PluginManager
-from ape.utils import cached_property
 
 from .config import ConfigManager
 
@@ -29,11 +28,12 @@ class CompilerManager:
 
     config: ConfigManager
     plugin_manager: PluginManager
+    _registered_compilers_cache: Dict[Path, Dict[str, CompilerAPI]] = {}
 
     def __repr__(self):
         return f"<CompilerManager len(registered_compilers)={len(self.registered_compilers)}>"
 
-    @cached_property
+    @property
     def registered_compilers(self) -> Dict[str, CompilerAPI]:
         """
         Each compile-able file extension mapped to its respective
@@ -43,6 +43,10 @@ class CompilerManager:
             Dict[str, :class:`~ape.api.compiler.CompilerAPI`]: The mapping of file-extensions
             to compiler API classes.
         """
+
+        cache_key = self.config.PROJECT_FOLDER
+        if cache_key in self._registered_compilers_cache:
+            return self._registered_compilers_cache[cache_key]
 
         registered_compilers = {}
 
@@ -54,6 +58,7 @@ class CompilerManager:
                 if extension not in registered_compilers:
                     registered_compilers[extension] = compiler
 
+        self._registered_compilers_cache[cache_key] = registered_compilers
         return registered_compilers
 
     def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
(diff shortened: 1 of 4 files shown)


src/ape/managers/compilers.py
```diff
@@ -96,6 +93,31 @@ def compile(
...
+        if "registered_compilers" in self.__dict__:
+            compilers = self.__dict__.pop("registered_compilers")
+
+        try:
+            with self.config.using_project(project_path, contracts_path):
+                return self.compile(contract_filepaths)
+        finally:
+            if compilers:
+                self.__dict__["registered_compilers"] = compilers
+
```
fubuloubu, line 120: This is kind of wild. What is the concern about the compilers being already registered as plugins for other projects?

I would think that the plugins shouldn't really have to change, e.g. if you are a Vyper-only project, but you use a dependency that is Solidity, then you should have to have both plugins installed (otherwise it doesn't compile the Solidity dependencies). But I guess I should understand more how this is working first.

src/ape/managers/config.py
```diff
@@ -183,6 +184,41 @@ def serialize(self) -> Dict:
...
+
+            project_path = Path("path/to/project")
+            contracts_path = project_path / "contracts"
+
+            with config.using_project(project_path, contracts_path):
+                my_project = Project(project_path)
+
+        Args:
+            project_folder (pathlib.Path): The path of the context's project.
+            contracts_folder (pathlib.Path): The path to the context's source files.
```
fubuloubu, line 205: Is this ever not using something that can be set by the config in `project_folder`?

src/ape/managers/config.py
```diff
@@ -183,6 +184,41 @@ def serialize(self) -> Dict:
...
+            Generator
+        """
+
+        initial_project_path = self.PROJECT_FOLDER
+        initial_contracts_path = self.contracts_folder
+
+        self.PROJECT_FOLDER = project_folder
+        self.contracts_folder = contracts_folder
+
+        yield
```
fubuloubu, line 217: curious if this might make more sense
```suggestion
        yield Project(project_folder)
```

e.g.

```py
with config.using_project(project_path, contracts_path) as project:
    ...
```

fubuloubu: With this, hopefully works well with the fixture:

```py
@pytest.fixture(params=project_names)
def project(request, config):
    project_source_dir = projects_directory / request.param
    project_dest_dir = config.PROJECT_FOLDER / project_source_dir.name
    copy_tree(project_source_dir.as_posix(), project_dest_dir.as_posix())
    with config.using_project(project_dest_dir, project_dest_dir / "contracts") as project:
        yield project

@pytest.fixture
def project_folder(project):
    return project.PROJECT_FOLDER
```

src/ape/managers/compilers.py
fubuloubu: hmm, perhaps the config can be reloaded only when a new project is loaded, and then we also make sure that values like `import_remappings` are always using the config value

src/ape/managers/config.py
fubuloubu: ah, so there is a contracts folder outside of the project?

src/ape/managers/config.py
unparalleled-js: I do like this idea as was able to get it to work, despite this being a cyclic dependency.

src/ape/managers/compilers.py
unparalleled-js: Let me experiment with that sentiment.

src/ape/managers/config.py
unparalleled-js: The whole point of this method is to not use the project config one

src/ape/managers/compilers.py
unparalleled-js: I was able to remove this method entirely!

Now, the `registered_compilers` is cached per project. So if you change project's, it will re-load.

src/ape/managers/config.py
unparalleled-js: This is how I am using it in the dependencies compilation:

```python
            with self._compilers.config.using_project(
                temp_project_path, contracts_folder=temp_contracts_path
            ):
                contract_types = self._compilers.compile(sources)


```

src/ape/managers/compilers.py
```diff
@@ -27,11 +27,12 @@ class CompilerManager:
 
     config: ClassVar[ConfigManager] = injected_before_use()  # type: ignore
     plugin_manager: ClassVar[PluginManager] = injected_before_use()  # type: ignore
+    _registered_compilers_cache: Dict[Path, Dict[str, CompilerAPI]] = {}
 
     def __repr__(self):
         return f"<CompilerManager len(registered_compilers)={len(self.registered_compilers)}>"
 
-    @cached_property
+    @property
```
unparalleled-js, line 35: Note: still cached, though now it is cached by project path. So if you change project paths, it will re-load.
This seems to work for what I need it for!

unparalleled-js:
This PR changed a lot since it was first opened! It feels wayy less hacky now.

src/ape/managers/compilers.py
fubuloubu: sick!

Issue #436, opened 2022-01-26
General: add functionality to convert datetime object to timestamp


johnson2427:
### Overview

we currently can convert a string in "MM-DD-YYYY HH:MM:SS" format. Want to add functionality of taking a datetime object and converting to a timestamp

### Specification

```python
date_time = datetime.datetime.fromisoformat("2020-10-15 12:00:00")
new_date_time_obj = date_time.replace(tzinfo=timezone.utc)
return int(new_date_time_obj.timestamp())
```

### Dependencies

None

fubuloubu:
You can add this to current timestamp converter via `Union[str, datetime]`

fubuloubu:
Also make sure that the chain manager method accepts this

fubuloubu:
timedelta might also be nice to add. I had thought about correcting the time for it, but that seems like it would lead to a lot of contextual errors. I think more directly just converting timedeltas (in string or `datetime.timedelta` object formats) to their integer equivalent in seconds:
```py
int(timedelta.total_seconds())
```

also parse `"1 weeks"` via:
```py
# "1 weeks"
# "1 weeks, 1 days"
# "1 weeks, 1 days, 1 seconds"
# Convert comma-separated string of `"{int} {kwarg}, ..."` to dict of `{"{kwarg}": "{int}", ...}`
int(timedelta(**timedelta_dict).total_seconds())
```

This might be tougher though.

johnson2427:
Oh, I can delete DatetimeConverter then

johnson2427:
I can add timedelta to TimestampConverter. I will make a new issue

PR #441, opened 2022-01-27
feat: added DatetimeConverter

johnson2427:
…ity issue with PyCharm/mypy

### What I did

Added DatetimeConverter
PyCharm and mypy failed to recognize the need for inputs in DatetimeConverter() on line 163 of src.ape.managers.converters.py
Solves issue #436

### How I did it

Very similar implementation to TimestampConverter. I wonder if we could have just added some functionality to TimestampConverter to read in either a datetime or string. Not necessarily needed, doesn't change anything else

### How to verify it

Test passed when there are `None` inputs in datetime_converter = DatetimeConverter() implementation

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,6 +94,7 @@ setup(
         "eth_abi==2.1.1",
         "eth-utils==1.10.0",
         "eth-rlp==0.2.1",
+        "python-dateutil>=2.8.2,<3.0",
     ],
     entry_points={
         "console_scripts": ["ape=ape._cli:cli"],
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -1,7 +1,8 @@
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from decimal import Decimal
 from typing import Any, ClassVar, Dict, List, Tuple, Type, Union
 
+from dateutil.parser import parse  # type: ignore
 from eth_utils import is_checksum_address, is_hex, is_hex_address, to_checksum_address
 from hexbytes import HexBytes
 
@@ -141,25 +142,32 @@ list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
 
 class TimestampConverter(ConverterAPI):
     """
-    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp.
-    Must be UTC timezone
+    Converts either a string, datetime object, or a timedelta object to a timestamp.
+    No timezone required, but should be formatted to UTC.
     """
 
-    def is_convertible(self, value: str) -> bool:
-        if not isinstance(value, str):
+    def is_convertible(self, value: Union[str, datetime, timedelta]) -> bool:
+        if not isinstance(value, (str, datetime, timedelta)):
             return False
-        if " " not in value or len(value.split(" ")) > 2:
-            return False
-        try:
-            datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
-        except ValueError:
-            return False
+        if isinstance(value, str):
+            if " " not in value and len(value.split(" ")) > 2:
+                return False
+            else:
+                try:
+                    parse(value)
+                except ValueError:
+                    return False
         return True
 
-    def convert(self, value: str) -> int:
-        return int(
-            datetime.strptime(value, "%m-%d-%Y %H:%M:%S").replace(tzinfo=timezone.utc).timestamp()
-        )
+    def convert(self, value: Union[str, datetime, timedelta]) -> int:
+        if isinstance(value, str):
+            return int(parse(value).replace(tzinfo=timezone.utc).timestamp())
+        elif isinstance(value, datetime):
+            return int(value.replace(tzinfo=timezone.utc).timestamp())
+        elif isinstance(value, timedelta):
+            return int((datetime.now(timezone.utc) + value).timestamp())
+        else:
+            raise ConversionError
 
 
 timestamp_converter = TimestampConverter(None, None, None)  # type: ignore
--- a/tests/functional/conversion/test_timestamp.py
+++ b/tests/functional/conversion/test_timestamp.py
@@ -1,6 +1,25 @@
+from datetime import datetime, timedelta, timezone
+
+import pytest
+
 from ape import convert
 
 
-def test_convert_string_timestamp():
-    time = "10-15-2021 12:15:12"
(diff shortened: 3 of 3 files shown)


johnson2427:
linting all passes, tests for macros and ubuntu will fail because the linter is not allowing me to put inputs into DatetimeConverter in converters.py

unparalleled-js:
See @fubuloubu 's comment in https://github.com/ApeWorX/ape/issues/436

^ We only need a single converter here, we can handle the union types. This way, we can share some of the logic a lot more easily too. Like just first check if it a str, if not convert it to the datetime and the rest of the logic flow is the same after that.

johnson2427:
Yep, just realized we can use Union

src/ape/managers/converters.py
```diff
@@ -140,6 +140,28 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
 list_tuple_converter = ListTupleConverter(None, None, None)  # type: ignore
```
unparalleled-js, line 140: @johnson2427  Notice the type ignore comment is required for these

src/ape/managers/converters.py
```diff
@@ -142,25 +142,37 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
...
+        if isinstance(value, str):
+            return int(
+                datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
+                .replace(tzinfo=timezone.utc)
+                .timestamp()
+            )
+        elif isinstance(value, datetime):
+            return int(value.timestamp())
+        else:
+            raise ConversionError
```
unparalleled-js, line 175: You can probably get rid of this `else` and just make the `elif` an else.

We should never get here with any other types because of the `is_convertible()` method.

I would say it is safe to assume that the given value is either a `str` or a `datetime`

src/ape/managers/converters.py
```diff
@@ -142,25 +142,37 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
...
-        except ValueError:
+        if isinstance(value, str):
+            if " " not in value or len(value.split(" ")) > 2:
+                return False
+            else:
+                try:
+                    datetime.strptime(value, "%m-%d-%Y %H:%M:%S")
+                except ValueError:
+                    return False
+        if isinstance(value, datetime) and str(value.tzinfo) != "UTC":
```
unparalleled-js, line 161: Is it possible to have a `datetime` object _without_ associated `tzinfo`? I am a little worried users are going to try and use `datetime` objects that do not have timezone information associated with them and their program will fail and they will assume the `TimestampConverter` just doesn't work for `datetime` objects.

src/ape/managers/converters.py
johnson2427: oooooooohhhhhhhhhhhhhhhhh

johnson2427: Thank you!

src/ape/managers/converters.py
johnson2427: I'm right there with you on this front. Do we force them to UTC or convert their timestamp to UTC?
If they use it and put in the time they want, and then we convert to UTC, wouldn't that mix up what they want?

unparalleled-js: Are we able to do this:

1. Check if there is not any tz info. If that is the case, it is still convertible and we assume it is UTC.
2. If there _is_ tzinfo and it is NOT UTC, then it is not convertible

So we handle the case where there isn't tz info but still don't convert if there is, if that makes sense.

evanblank3: I guess the end user won't see this class because of the abstraction, or I would recommend changing the class name to reflect the dependence on UTC like `UTCTimestampConverter` or something.

src/ape/managers/converters.py
```diff
@@ -142,25 +142,37 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
...
-    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp.
+    Converts either a string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp OR
+    converts a datetime object to a timestamp.
     Must be UTC timezone
     """
 
-    def is_convertible(self, value: str) -> bool:
-        if not isinstance(value, str):
+    def is_convertible(self, value: Union[str, datetime]) -> bool:
+        if not isinstance(value, str) and not isinstance(value, datetime):
```
fubuloubu, line 151:
```suggestion
        if not isinstance(value, (str, datetime)):
```

src/ape/managers/converters.py
fubuloubu: Warning, then implicitly convert to UTC?

unparalleled-js: We discussed in Discord and don't see a case where timezone should matter. We should just always convert to the non-localized timestamp or assume UTC when there is no timestamp.

A warning might be nice though.... Because maybe the user does not realize they are using localized dates.

johnson2427: I removed that, and we just tell the timestamp that this is UTC. The user has to know they have to hand over UTC time

src/ape/managers/converters.py
```diff
@@ -142,25 +143,33 @@ def convert(self, value: Union[List, Tuple]) -> Union[List, Tuple]:
 
 class TimestampConverter(ConverterAPI):
     """
-    Converts string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp.
+    Converts either a string of ``"%m-%d-%Y %H:%M:%S"`` format to a timestamp OR
```
unparalleled-js, line 146: Should we list the other formats if we are officially supporting them?

unparalleled-js (approved):
Looks great!

Considering those other date time formats to the method doc :)

setup.py
```diff
@@ -94,6 +94,7 @@
         "eth_abi==2.1.1",
         "eth-utils==1.10.0",
         "eth-rlp==0.2.1",
+        "python-dateutil==2.8.2",
```
unparalleled-js, line 97: This pin seems a little stringent

unparalleled-js: Just do `>=2.8.2,<3.0`

unparalleled-js (approved):

evanblank3 (approved):
LGTM

unparalleled-js (approved):

PR #445, opened 2022-01-28
refactor: api classes to pydantic conversion

evanblank3:
### What I did

fixes: #

### Checklist


src/ape/api/compiler.py
```diff
@@ -1,15 +1,14 @@
...
 
-from ape.utils import abstractdataclass, abstractmethod
+from ape.utils import AbstractBaseModel, abstractmethod, injected_before_use
 
 from .config import ConfigItem
 
 
-@abstractdataclass
-class CompilerAPI:
+class CompilerAPI(AbstractBaseModel):
```
fubuloubu, line 11: Love this name: `AbstractBaseModel`

src/ape/managers/chain.py
```diff
@@ -294,11 +294,12 @@ class ChainManager(_ConnectedChain):
...
-        BlockContainer._converters = self._converters
-
-        AccountHistory._networks = self._networks
-        AccountHistory._converters = self._converters
+        """
+        Handles injected properties for _ConnectedChain classes
+        """
+        for chain_class in (AccountHistory, BlockContainer):
+            chain_class._networks = self._networks
+            chain_class._converters = self._converters
```
fubuloubu, line 302: Actually, I like the previous way, it's a lot mor explicit.

src/ape/managers/compilers.py
```diff
@@ -45,8 +45,8 @@ def registered_compilers(self) -> Dict[str, CompilerAPI]:
         registered_compilers = {}
 
         for plugin_name, (extensions, compiler_class) in self.plugin_manager.register_compiler:
-            config = self.config.get_config(plugin_name)
-            compiler = compiler_class(config=config)
+            compiler_class.config = self.config.get_config(plugin_name)
+            compiler = compiler_class()
```
fubuloubu, line 49: I wonder if `config: ConfigItem` since it is contextual dependending on project (see #439) is better off staying as an instance arg

src/ape/managers/converters.py
```diff
@@ -184,23 +169,39 @@ class ConversionManager:
...
+            AddressAPIConverter,
+            HexAddressConverter,
+            HexConverter,
+            ListTupleConverter,
+            TimestampConverter,
+        ):
+            converter.config = self.config  # type: ignore
+            converter.networks = self.networks  # type: ignore
+            converter.converter = self  # type: ignore
+
```
fubuloubu, line 186: Note: none of the above actually needs any of these values to be set (which is why they were set to `None`), the dependency injection shouldn't cause the class to fail to work, only if the class tries to access that property will it raise.

tl;dr: delete this, it should still function just the same

src/ape/managers/converters.py
```diff
@@ -184,23 +169,39 @@ class ConversionManager:
...
-            list: [list_tuple_converter],
-            tuple: [list_tuple_converter],
+            list: [ListTupleConverter()],
+            tuple: [ListTupleConverter()],
         }
 
         for plugin_name, (conversion_type, converter_class) in self.plugin_manager.converters:
-            converter = converter_class(self.config.get_config(plugin_name), self.networks, self)
+
+            converter = converter_class()
```
fubuloubu, line 204: For this to work, I think you have to inject into `ConversionAPI` inside of `__init__`

src/ape/utils.py
```diff
@@ -444,6 +445,21 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    class Config:
+        keep_untouched = (cached_property,)
+        arbitrary_types_allowed = True
+
+    # NOTE: Due to https://github.com/samuelcolvin/pydantic/issues/1241
+    #       we have to add this cached property workaround in order to avoid this error:
+    #
+    #           TypeError: cannot pickle '_thread.RLock' object
+
+    def __dir__(self) -> List[str]:
```
fubuloubu, line 458: Add note about how this integrates with IPython: https://ipython.readthedocs.io/en/stable/config/integrating.html

src/ape/managers/converters.py
fubuloubu: hold on, actually try this instead
```suggestion
    def __init__(self) -> None:
        """
        Handles the injected properties for ConverterAPI classes
        """
        ConverterAPI.config = self.config  # type: ignore
        ConverterAPI.networks = self.networks  # type: ignore
        ConverterAPI.converter = self  # type: ignore
        
```

evanblank3:
`refactor: ManagerBase` commit is in response to co-working with @fubuloubu on creating a Manager Base class that can handle injecting properties and passing them one time to all manager classes

src/ape/__init__.py
```diff
@@ -4,93 +4,23 @@
...
-accounts = _AccountManager()
-"""Manages accounts for the current project. See :class:`ape.managers.accounts.AccountManager`."""
-
-Project.config = config
-Project.compilers = compilers
-Project.networks = networks
-Project.converter = _converters
-"""User-facing class for instantiating Projects (in addition to the currently
-active ``project``). See :class:`ape.managers.project.ProjectManager`."""
+from .managers import Project, accounts, chain, config, converters, networks
```
fubuloubu, line 9: meant to say this before, but too much going on, wanted `converters` to be imported private

```suggestion
from .managers import Project, accounts, chain, config, networks
from .managers import converters as _converters
```

src/ape/__init__.py
```diff
@@ -4,93 +4,23 @@
...
 
-convert = _converters.convert
+convert = converters.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
 
 
 __all__ = [
     "accounts",
     "chain",
-    "compilers",
```
fubuloubu, line 93: Note: no need to export this, not really that helpful (but maybe it is? maybe I want to compile something ad-hoc?)

edit: actually I think I've done ad-hoc compilation before, it order to support "mutation testing" patterns where I've slightly modified source code of a contract and compiled it only for testing purposes. see: https://github.com/yearn/yearn-vaults/blob/main/tests/conftest.py#L79

src/ape/api/accounts.py
```diff
@@ -200,12 +200,12 @@ def deploy(self, contract: "ContractContainer", *args, **kwargs) -> "ContractIns
...
         logger.success(f"Contract '{contract_name}' deployed to: {address}")
 
-        from ape import _converters
+        from ape import converters
         from ape.contracts import ContractInstance
 
         return ContractInstance(  # type: ignore
             _provider=self.provider,
-            _converter=_converters,
+            _converter=converters,
```
fubuloubu, line 208: should try to DI `converters`, and we've talked about removing the need for `_provider=...` by injecting `network_manager` into `AddressAPI` instead and having `provider` either be a `property` (for `AccountAPI`) or `cached_property` (for `ContractInstance`) tied to `NetworkManager.active_provider`

src/ape/api/compiler.py
```diff
@@ -19,9 +16,6 @@ class CompilerAPI:
     this API.
     """
 
-    config: ConfigItem
-    """The :class:`ape.api.config.ConfigItem` for this compiler plugin."""
```
fubuloubu, line 23: Might want to keep this for now and initialized each instance with this property (still thinking about this)

src/ape/managers/__init__.py
```diff
@@ -0,0 +1,67 @@
+from pathlib import Path as _Path
+
+from ape.plugins import PluginManager as _PluginManager
```
fubuloubu, line 3: Don't need to keep this private anymore

src/ape/managers/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from ape.utils import USER_AGENT, ManagerAccessBase
+
+from .accounts import AccountManager as _AccountManager
+from .chain import ChainManager as _ChainManager
+from .compilers import CompilerManager as _CompilerManager
+from .config import ConfigManager as _ConfigManager
+from .converters import ConversionManager as _ConversionManager
+from .networks import NetworkManager as _NetworkManager
+from .project import ProjectManager as Project
```
fubuloubu, line 12: Don't need to keep these private

src/ape/managers/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+__all__ = [
+    "accounts",
+    "chain",
+    "config",
+    "converters",
+    "networks",
+    "Project",
+]
```
fubuloubu, line 67: It might help for other places (mostly type checking in `ape/api/*.py`) to also export the manager types, so this way imports are segregated to the top-level of this module

src/ape/managers/accounts.py
```diff
@@ -53,7 +44,9 @@ def containers(self) -> Dict[str, AccountContainerAPI]:
 
             accounts_folder = data_folder / plugin_name
             accounts_folder.mkdir(exist_ok=True)
-            containers[plugin_name] = container_type(accounts_folder, account_type, self.config)
+            containers[plugin_name] = container_type(
+                accounts_folder, account_type, self.config_manager
```
fubuloubu, line 48: Explore injecting `config_manager` into `AccountContainerAPI` via `ManagerAccessBase`

src/ape/managers/base.py
```diff
@@ -0,0 +1,5 @@
+from ape.utils import ManagerAccessBase
+
+
+class ManagerBase(ManagerAccessBase):
+    pass
```
fubuloubu, line 5: Note: this is to allow us to add other IPython integration features for managers

src/ape/managers/networks.py
```diff
@@ -1,16 +1,14 @@
-from typing import ClassVar, Dict, Iterator, Optional
+from typing import Dict, Iterator, Optional
 
 import yaml
-from pluggy import PluginManager  # type: ignore
```
fubuloubu, line 4: I think this was wrong anyways, `ape.plugins import PluginManager` (I wrote another class with same name)

src/ape/managers/project.py
```diff
@@ -52,33 +43,6 @@ class ProjectManager:
...
-    Call method :meth:`~ape.managers.project.ProjectManager.load_contracts` in this class
-    to more easily compile sources.
-    """
-
-    networks: ClassVar[NetworkManager] = injected_before_use()  # type: ignore
-    """
-    The manager of networks, :class:`~ape.managers.networks.NetworkManager`.
-    To get the active provide, use
-    :py:attr:`ape.managers.networks.NetworkManager.active_provider`.
-    """
```
fubuloubu, line 80: Actually move all of these docstrings to `ManagerAccessBase`

src/ape/utils.py
```diff
@@ -444,6 +454,45 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+
+
+class ManagerAccessBase:
+    account_manager: ClassVar["AccountManager"] = injected_before_use()  # type: ignore
+    chain_manager: ClassVar["ChainManager"] = injected_before_use()  # type: ignore
+    compiler_manager: ClassVar["CompilerManager"] = injected_before_use()  # type: ignore
+    config_manager: ClassVar["ConfigManager"] = injected_before_use()  # type: ignore
+    conversion_manager: ClassVar["ConversionManager"] = injected_before_use()  # type: ignore
+    network_manager: ClassVar["NetworkManager"] = injected_before_use()  # type: ignore
+    plugin_manager: ClassVar["PluginManager"] = injected_before_use()  # type: ignore
```
fubuloubu, line 474: note for another PR: we should be able to fix these types of issues by making `injected_before_use` take one argument, the expected type as a string, and fulfilling mypy type checking needs there.

```py
    account_manager: ClassVar["AccountManager"] = injected_before_use("AccountManager")
```

src/ape/utils.py
```diff
@@ -444,6 +454,45 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        arbitrary_types_allowed = True
+
+    # NOTE: Due to https://github.com/samuelcolvin/pydantic/issues/1241
+    #       we have to add this cached property workaround in order to avoid this error:
+    #
+    #           TypeError: cannot pickle '_thread.RLock' object
+
+    def __dir__(self) -> List[str]:
+        """
+        NOTE: Integrates with IPython
```
fubuloubu, line 489:
```suggestion
        NOTE: Integrates with IPython tab-completion
```

src/ape_test/fixtures.py
```diff
@@ -12,23 +12,23 @@
 class PytestApeFixtures:
```
fubuloubu, line 12: something evil maybe lol:
```suggestion
class PytestApeFixtures(ManagerAccessBase):
```

src/ape_test/fixtures.py
```diff
@@ -12,23 +12,23 @@
...
+        project_manager: ProjectManager,
+        chain_manager: ChainManager,
     ):
-        self._accounts = accounts
-        self._networks = networks
-        self._project = project
-        self._chain = chain
+        self.account_manager = account_manager
+        self.network_manager = network_manager
+        self._project = project_manager
```
fubuloubu, line 22: Should `project_manager` be in `ManagerAccessBase`?

src/ape_test/runners.py
```diff
@@ -13,30 +13,30 @@
 class PytestApeRunner:
```
fubuloubu, line 13: Same evil idea lol
```suggestion
class PytestApeRunner(ManagerAccessBase):
```

fubuloubu (commented):
Really like how this is coming out!

unparalleled-js:
Currently getting this error when trying to compile Solidity files: 

```
ERROR: Unable to load project. Reason: 1 validation error for SolidityCompiler
config
  field required (type=value_error.missing)
```

src/ape/utils.py
```diff
@@ -444,6 +454,45 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    """
+    Injected properties are injected class variables that must be set before use
+    NOTE: do not appear in a Pydantic model's set of properties
+    """
+
+    def __get__(self, *args):
+        raise ValueError("Value not set. Please inject this property before calling.")
+
+
+class ManagerAccessBase:
```
unparalleled-js, line 467: Maybe we should put this in `ape.managers.base.py` ?

src/ape/__init__.py
```diff
@@ -4,93 +4,23 @@
```
unparalleled-js, line 4: Because of the changes in this file, the docs no longer make a lot of sense.

We _have_ to document the top-level imports here...

So you may need to do like this:

```python
from .managers import acconts as _accounts

accounts = _accounts
"""
<TOP-LEVEL ACCOUNTS DOCS>
"""
```

unparalleled-js:
I am unable to use the `ape-ens` plugin from this PR:

I try  this:

```python
from ape_tokens import tokens

tokens["ENS"].balanceOf("juliya.eth")
```

and I get this error:

```
AttributeError: 'ENSConversions' object has no attribute 'networks'
```

src/ape/utils.py
fubuloubu: This would prevent it from being importable into the `ape.api` modules due to circular imports

fubuloubu:
> I am unable to use the `ape-ens` plugin from this PR:
> 
> I try this:
> 
> ```python
> from ape_tokens import tokens
> 
> tokens["ENS"].balanceOf("juliya.eth")
> ```
> 
> and I get this error:
> 
> ```
> AttributeError: 'ENSConversions' object has no attribute 'networks'
> ```

This is a breaking change for all imports, likely. We wanted to normalize the names to `*_manager` because we got into an issue due to confusing the different names

src/ape/utils.py
unparalleled-js: Ah, I see

src/ape/cli/options.py
```diff
@@ -41,7 +41,7 @@ def project(self) -> ProjectManager:
             from ape import project
 
             self._project = project
-            self._project.config.load()
```
unparalleled-js, line 44: I wonder if it would make sense to also have this class use the manager base class?
The class is `ApeCliContextObject` and is available fas a CLI automatic-pass-in option that basically contains managers, an `abort()` method and the `logger`.

fubuloubu: Can do! I am really loving this hack @evanblank3 is doing of injecting the managers into one place. Makes them really easy to access and reduce mutual dependencies!

src/ape_test/runners.py
```diff
@@ -1,35 +1,22 @@
...
-        self.network_manager = network_manager
-        self.chain_manager = chain_manager
+        self.project = project_manager
         self._warned_for_missing_features = False
         ape.reverts = RevertsContextManager  # type: ignore
 
-    @property
-    def _network_choice(self) -> str:
-        # The option the user providers via --network (or the default).
-        return self.config_manager.getoption("network")
```
evanblank3: Not sure about this, but when I updated I was getting a pre-commit error around `config_manager` not having a method `getoption()` but it doesn't appear to be used in the test. Please advise....

evanblank3: Found where this is used, added it back, and fixed the issue.

src/ape_test/plugin.py
```diff
@@ -38,11 +39,17 @@ def is_module(v):
     config.option.verbose = config.getoption("capture") == "no"
 
     # Inject the runner plugin (must happen before fixtures registration)
-    session = PytestApeRunner(config, project, networks, chain)
+    ManagerAccessBase.config_manager = config
```
unparalleled-js, line 42: Are the injections here still needed? I would think this would already be done at this point?

src/ape_test/runners.py
```diff
@@ -67,10 +58,10 @@ def pytest_sessionstart(self):
         so related assertions cannot be rewritten". The warning is not relevant
         for end users who are performing tests with ape.
         """
-        reporter = self.config.pluginmanager.get_plugin("terminalreporter")
```
unparalleled-js, line 70: **HEY**  this `config` is a `PytestConfig`, not a `ape.managers.ConfigManager`.
I feel like that will get confusing.

src/ape_test/runners.py
```diff
@@ -1,42 +1,33 @@
...
-        self.chain = chain
+class PytestApeRunner(ManagerAccessBase):
+    def __init__(self):
         self._warned_for_missing_features = False
         ape.reverts = RevertsContextManager  # type: ignore
 
     @property
     def _network_choice(self) -> str:
         # The option the user providers via --network (or the default).
-        return self.config.getoption("network")
```
unparalleled-js, line 31: I believe this was failing before because we were injecting the `config_manager` here?
The runner should get the `PytestConfig` the normal way probably?

src/ape_test/runners.py
fubuloubu: Now you see why I suggested name change to `*_manager` for everything

evanblank3: Yeah, I did get very confused about that. I now see that config wasn't the config_manager. LOL

src/ape/__init__.py
```diff
@@ -3,96 +3,57 @@
...
-    # Store all globally-cached files
-    data_folder=_data_folder,
-    # NOTE: For all HTTP requests we make
-    request_header={
-        "User-Agent": USER_AGENT,
-    },
-    # What we are considering to be the starting project directory
-    project_folder=_Path.cwd(),
-)
+config = _config_manager
```
fubuloubu, line 26: How about this:
```suggestion
config = ManagerAccessBase.config_manager
```

evanblank3: I think by pulling it from the `MangagerAccessBase` it possibly won't be initialized, vs. pulling it from the `managers/__init__.py` file?

fubuloubu: If you pull like `from managers import ManagerAccessBase as _ManagerAccessBase` and then add a note about the hack to ensure it is properly initialized, I think it will work.

src/ape/api/accounts.py
```diff
@@ -268,7 +263,7 @@ def __len__(self) -> int:
         """
 
     @abstractmethod
-    def __iter__(self) -> Iterator[AccountAPI]:
+    def __iter__(self):
```
evanblank3, line 266: mypy gets very angry about leaving this as `Iterator[AccountAPI]`. I tried a few type alternatives but the pydantic BaseModel seems to take precedent returning a type of `Generator[Tuple[str, Any], None, None]`.

unparalleled-js: Is it bad to be overriding this method? It sounds like the base model is already maybe doing something with it?

src/ape/types/signatures.py
```diff
@@ -1,11 +1,11 @@
 from typing import Iterator, Union
 
-from dataclassy import dataclass
 from eth_account.messages import SignableMessage
 from eth_utils import to_bytes
+from pydantic.dataclasses import dataclass
 
 
-@dataclass(frozen=True, slots=True, kwargs=True)
```
unparalleled-js, line 8: I think the pydantic `dataclass` still supports `frozen`.
https://pydantic-docs.helpmanual.io/usage/dataclasses/

^ Disregard if we are intentionally getting rid of that

src/ape_accounts/accounts.py
```diff
@@ -34,29 +34,27 @@ def aliases(self) -> Iterator[str]:
...
 
-    def __iter__(self) -> Iterator[AccountAPI]:
+    def __iter__(self):
         for keyfile in self._keyfiles:
-            yield KeyfileAccount(self, keyfile)  # type: ignore
+            yield KeyfileAccount(container=self, keyfile_path=keyfile)  # type: ignore
 
 
-# NOTE: `AccountAPI` is a dataclass
+# NOTE: `AccountAPI` is am AbstractBaseModel
```
unparalleled-js, line 42:
```suggestion
# NOTE: `AccountAPI` is an AbstractBaseModel
```

src/ape/api/accounts.py
fubuloubu: > Is it bad to be overriding this method? It sounds like the base model is already maybe doing something with it?

It is actaully, for serialization. We should probably consider an alternative method rather than overriding this

fubuloubu (commented):
This looks sooooo much cleaner than prior!

src/ape/contracts/base.py
```diff
@@ -12,29 +14,29 @@
...
     provider: ProviderAPI
-    converter: "ConversionManager"
 
-    def __post_init__(self):
-        if len(self.deployment_bytecode) == 0:
+    @validator("deployment_bytecode")
+    def check_empty_contract(cls, value):
+        if len(value) == 0:
             logger.warning("Deploying an empty contract (no bytecode)")
+        return value
```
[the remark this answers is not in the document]
evanblank3: I updated the logic, but I'm not so sure about the `Optional[bytes]` because it causes a trickle effect on the ecosystem/network that we might not want.

src/ape/utils.py
```diff
@@ -519,6 +532,80 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+class AbstractBaseModel(ManagerAccessBase, ABC, BaseModel):
+    class Config:
+        keep_untouched = (cached_property,)
+        arbitrary_types_allowed = True
+        underscore_attrs_are_private = True
+
+    # NOTE: Due to https://github.com/samuelcolvin/pydantic/issues/1241
+    #       we have to add this cached property workaround in order to avoid this error:
+    #
+    #           TypeError: cannot pickle '_thread.RLock' object
```
fubuloubu, line 598: Think this should be inside of `Config` above

tests/integration/test_ethereum.py
```diff
@@ -7,10 +7,10 @@
 @pytest.mark.parametrize("type_kwarg", (0, "0x0", b"\x00", "0", HexBytes("0x0"), HexBytes("0x00")))
 def test_create_static_fee_transaction(ethereum, type_kwarg):
     txn = ethereum.create_transaction(type=type_kwarg)
-    assert txn.type == TransactionType.STATIC.value
+    assert txn.type.value == TransactionType.STATIC.value
```
unparalleled-js, line 10: Can probably just do `txn.type == TransactionType.STATIC`

unparalleled-js: OH wait I think this should be the raw value though

src/ape/api/providers.py
```diff
@@ -59,34 +57,13 @@ class TransactionAPI:
...
     type: TransactionType = TransactionType.STATIC
+    max_fee: int = None  # type: ignore
 
     # If left as None, will get set to the network's default required confirmations.
     required_confirmations: Optional[int] = None
 
     signature: Optional[TransactionSignature] = None
 
-    @property
-    def max_fee(self) -> int:
```
fubuloubu, line 70: This doesn't work as a property? Is this the "computed property" bug?

src/ape/api/providers.py
```diff
@@ -59,34 +57,13 @@ class TransactionAPI:
...
-        """
-        The total amount in fees willing to be spent on a transaction.
-        Override this property as needed, such as for
-        `EIP-1559 <https://eips.ethereum.org/EIPS/eip-1559>`__ differences.
-
-        See :class:`~ape_ethereum.ecosystem.StaticFeeTransaction` and
-        :class:`~ape_ethereum.ecosystem.DynamicFeeTransaction` as examples.
-
-        Raises:
-            NotImplementedError: When setting in a class that did not override the setter.
```
fubuloubu, line 80: Also noting it doesn't actually raise an error.

src/ape/api/providers.py
```diff
@@ -110,15 +87,15 @@ def as_dict(self) -> dict:
...
 
     def __repr__(self) -> str:
-        data = as_dict(self)  # NOTE: `as_dict` could be overridden
+        data = self.as_dict()  # NOTE: `as_dict` could be overridden
         params = ", ".join(f"{k}={v}" for k, v in data.items())
         return f"<{self.__class__.__name__} {params}>"
 
     def __str__(self) -> str:
-        data = as_dict(self)  # NOTE: `as_dict` could be overridden
+        data = self.as_dict()  # NOTE: `as_dict` could be overridden
```
fubuloubu, line 98: just delete `as_dict` and use `dict` directly

src/ape/managers/config.py
```diff
@@ -23,7 +25,7 @@ class DeploymentConfig(ConfigItem):
     contract_type: str
 
 
-class ConfigManager:
+class ConfigManager(ManagerBase, BaseModel):
```
fubuloubu, line 28: Explore `BaseSettings` in another PR

evanblank3:
Closing this draft in favor of https://github.com/ApeWorX/ape/pull/509

Issue #449, opened 2022-01-29
`ContractContainer.__call__` function does not check `*args` length before deploying


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.0b4

$ ape plugins list
Installed Plugins:
  infura      0.1.0a4
  vyper       0.1.0b2
```

* Python Version:  3.8.9
* OS: linux

### What went wrong?

```py
../../../../.pyenv/versions/3.8.9/envs/nft-v1/lib/python3.8/site-packages/eth_abi/encoding.py:88: ValueOutOfBounds
_________________________________________________________________________ ERROR at setup of test_buyers_cant_mint _________________________________________________________________________

project = <ProjectManager>, owner = <TestAccount 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C>, artist = <TestAccount 0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7>

    @pytest.fixture
    def nft(project, owner, artist):
>       return owner.deploy(project.ApePiece, artist)

tests/conftest.py:26: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../.pyenv/versions/3.8.9/envs/nft-v1/lib/python3.8/site-packages/eth_abi/codec.py:87: in encode_abi
    return encoder(args)
../../../../.pyenv/versions/3.8.9/envs/nft-v1/lib/python3.8/site-packages/eth_abi/encoding.py:98: in __call__
    return self.encode(value)
../../../../.pyenv/versions/3.8.9/envs/nft-v1/lib/python3.8/site-packages/eth_abi/encoding.py:139: in encode
    self.validate_value(values)
../../../../.pyenv/versions/3.8.9/envs/nft-v1/lib/python3.8/site-packages/eth_abi/encoding.py:123: in validate_value
    self.invalidate_value(
[...]

PR #450, opened 2022-01-30
fix: raise ArgumentsLengthError when mismatched arguments length in contract constructor

unparalleled-js:
### What I did

fixes: #449 

Issue where error was confusing and inconsistent with other contract calls / txs when deploying a contract with the wrong number of constructor arguments.

### How I did it

Check the abi inputs in `ContractContainer` and raise `ArgumentsError` in that case

### How to verify it

Create a contract with a constructor that takes arguments:

```solidity
pragma solidity ^0.6.0;

contract Fund {

    uint secretNumber;

    constructor(uint _secretNumber) public {
        secretNumber = _secretNumber;
    }
}
```

Try to deploy the contract without giving it arguments and will arguments:

```python

accounts.test_accounts[0].deploy(project.Fund)   # <-- Fails with NEW-but-old custom exception 'ArgumentsError'
accounts.test_accounts[0].deploy(project.Fund, 5)   # <-- Works
```

### Checklist

### DIFF
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -52,8 +52,10 @@ class ArgumentsLengthError(ContractError):
             return f"({length}) " if length is not None else ""
 
         message = (
-            f"The number of the given arguments {_get_len_str(arguments_length)}"
-            f"do not match what is defined in the ABI {_get_len_str(inputs_length).strip()}."
+            f"The number of the given arguments "
+            f"{f'({arguments_length}) ' if arguments_length else ''}"
+            f"do not match what is defined in the "
+            f"ABI{f' ({inputs_length})' if inputs_length else ''}.".strip()
         )
         super().__init__(message)
 
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -47,7 +47,7 @@ class ArgumentsLengthError(ContractError):
     Raised when calling a contract method with the wrong number of arguments.
     """
 
-    def __init__(self, arguments_length: Optional[int] = None, inputs_length: Optional[int] = None):
+    def __init__(self, arguments_length: int, inputs_length: Optional[int] = None):
         message = (
             f"The number of the given arguments "
             f"{f'({arguments_length}) ' if arguments_length else ''}"
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -48,14 +48,11 @@ class ArgumentsLengthError(ContractError):
     """
 
     def __init__(self, arguments_length: Optional[int] = None, inputs_length: Optional[int] = None):
-        def _get_len_str(length: Optional[int]):
-            return f"({length}) " if length is not None else ""
-
         message = (
             f"The number of the given arguments "
             f"{f'({arguments_length}) ' if arguments_length else ''}"
             f"do not match what is defined in the "
-            f"ABI{f' ({inputs_length})' if inputs_length else ''}.".strip()
+            f"ABI{f' ({inputs_length})' if inputs_length else ''}."
         )
         super().__init__(message)
 
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -48,11 +48,11 @@ class ArgumentsLengthError(ContractError):
     """
 
     def __init__(self, arguments_length: int, inputs_length: Optional[int] = None):
+        abi_suffix = f" ({inputs_length})" if inputs_length else ""
         message = (
-            f"The number of the given arguments "
-            f"{f'({arguments_length}) ' if arguments_length else ''}"
+            f"The number of the given arguments ({arguments_length}) "
             f"do not match what is defined in the "
-            f"ABI{f' ({inputs_length})' if inputs_length else ''}."
+            f"ABI{abi_suffix}."
         )
         super().__init__(message)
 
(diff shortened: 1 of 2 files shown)


src/ape/exceptions.py
```diff
@@ -47,8 +47,14 @@ class ArgumentsLengthError(ContractError):
     Raised when calling a contract method with the wrong number of arguments.
     """
 
-    def __init__(self):
-        message = "The number of the given arguments do not match what is defined in the ABI."
+    def __init__(self, arguments_length: Optional[int] = None, inputs_length: Optional[int] = None):
+        def _get_len_str(length: Optional[int]):
+            return f"({length}) " if length is not None else ""
```
fubuloubu, line 52:
```suggestion
            return f" ({length}) " if length is not None else ""
```

src/ape/exceptions.py
```diff
@@ -47,8 +47,14 @@ class ArgumentsLengthError(ContractError):
...
+
+        message = (
+            f"The number of the given arguments {_get_len_str(arguments_length)}"
+            f"do not match what is defined in the ABI {_get_len_str(inputs_length).strip()}."
```
fubuloubu, line 56:
```suggestion
            f"do not match what is defined in the ABI{_get_len_str(inputs_length).strip()}."
```

src/ape/exceptions.py
```diff
@@ -47,8 +47,14 @@ class ArgumentsLengthError(ContractError):
...
     """
 
-    def __init__(self):
-        message = "The number of the given arguments do not match what is defined in the ABI."
+    def __init__(self, arguments_length: Optional[int] = None, inputs_length: Optional[int] = None):
+        def _get_len_str(length: Optional[int]):
+            return f"({length}) " if length is not None else ""
+
+        message = (
+            f"The number of the given arguments {_get_len_str(arguments_length)}"
```
fubuloubu, line 55:
```suggestion
            f"The number of the given arguments {_get_len_str(arguments_length)} "
```

fubuloubu (approved):
Just massaging some spacing

src/ape/exceptions.py
unparalleled-js: It is actually correct the way it is. The space is already there when the arg is not missing and line 52 adds a space at the end of it is

unparalleled-js:
```python
>>> from ape.exceptions import ArgumentsLengthError
>>> str(ArgumentsLengthError())
'The number of the given arguments do not match what is defined in the ABI .'
>>> str(ArgumentsLengthError(2))
'The number of the given arguments (2) do not match what is defined in the ABI .'
>>> str(ArgumentsLengthError(2, 1))
'The number of the given arguments (2) do not match what is defined in the ABI (1).'
```

unparalleled-js: Ope, I see one problem lol

unparalleled-js: Okay fixed now: 

```python
>>> from ape.exceptions import ArgumentsLengthError
>>> str(ArgumentsLengthError())
'The number of the given arguments do not match what is defined in the ABI.'
>>> str(ArgumentsLengthError(1))
'The number of the given arguments (1) do not match what is defined in the ABI.'
>>> str(ArgumentsLengthError(1, 2))
'The number of the given arguments (1) do not match what is defined in the ABI (2).'
```

src/ape/exceptions.py
```diff
@@ -47,8 +47,13 @@ class ArgumentsLengthError(ContractError):
...
     """
 
-    def __init__(self):
-        message = "The number of the given arguments do not match what is defined in the ABI."
+    def __init__(self, arguments_length: Optional[int] = None, inputs_length: Optional[int] = None):
+        message = (
+            f"The number of the given arguments "
+            f"{f'({arguments_length}) ' if arguments_length else ''}"
+            f"do not match what is defined in the "
+            f"ABI{f' ({inputs_length})' if inputs_length else ''}."
```
fubuloubu, line 55:
```suggestion
            "The number of the given arguments"
            f"{f' ({arguments_length})' if arguments_length else ' '}"
            " do not match what is defined in the "
            f" ABI{f' ({inputs_length})' if inputs_length else ''}."
```

fubuloubu (approved):
I'm being really pedantic now lol

src/ape/exceptions.py
unparalleled-js: I will trust that you tested this and it works.

unparalleled-js:
this does not work now that I committed your suggestion

fubuloubu (approved):

src/ape/exceptions.py
```diff
@@ -47,8 +47,13 @@ class ArgumentsLengthError(ContractError):
...
     """
 
-    def __init__(self):
-        message = "The number of the given arguments do not match what is defined in the ABI."
+    def __init__(self, arguments_length: int, inputs_length: Optional[int] = None):
+        abi_suffix = f" ({inputs_length})" if inputs_length else ""
+        message = (
+            f"The number of the given arguments ({arguments_length}) "
+            f"do not match what is defined in the "
+            f"ABI{abi_suffix}."
```
unparalleled-js, line 55: *NOTE*: The space **HAS** to be in the suffix because otherwise when the argument is not provider, you will have an extract space before the period, like `end ."` instead of `end."`

unparalleled-js:
Made `arguments_length` required since it always used anyway, that makes it look prettier.

Also, it still works, which is more important:

```python
>>> str(ArgumentsLengthError(3))
'The number of the given arguments (3) do not match what is defined in the ABI.'
>>> str(ArgumentsLengthError(3, 3))
```

fubuloubu (approved):

PR #459, opened 2022-02-02
chore: upgrade web.py dependency

unparalleled-js:
### What I did

* upgrade web3.py dependency to 5.27.0
* this required also updating eth-account to 0.5.7

### How to verify it

should maybe check if docker image builds first

### Checklist


fubuloubu (approved):

unparalleled-js:
Running into issue with this. Standby!

fubuloubu (approved):

evanblank3 (approved):

setup.py
```diff
@@ -94,7 +94,7 @@
         "rich>=10.14,<11",
         "tqdm>=4.62.3,<5.0",
         "typing-extensions ; python_version<'3.8'",
-        "web3[tester]==5.27.0",
+        "web3[tester]==5.26.0",
```
fubuloubu, line 97: I guess this doesn't work after all?

fubuloubu (commented):
Maybe should just close this for now until all things are working that we can loosen the pin a bit

setup.py
unparalleled-js: Oh oops I don't think I meant to undo that.
I had some weird issue where I had accidentally pushed this to my `origin/main`. I must have mistakenly reset this here.

unparalleled-js:
> Maybe should just close this for now until all things are working that we can loosen the pin a bit

From what I gather, this is the last release we can be on before having to figure out how to install `typing-extensions>=4.0`.
We need this version for `ape-hardhat` to work and be able to remove the pending-block hack

fubuloubu (approved):

Issue #465, opened 2022-02-03
bug: Allow raising Abort without providing a message


unparalleled-js:
### Environment information

0.1.0b4
macOS
python 3.9.7

### What went wrong?

Right now, if you do:

```python
raise Abort()
```

you get error:

```
TypeError: __init__() missing 1 required positional argument: 'message'
```

It would be nice if it defaulted to something generic so I can use in scripts without having to care. Like:

```
ERROR: Operation aborted.
```

### How can it be fixed?

Default argument

fubuloubu:
I think it would be nice that it contains some sort of info of what triggers the aborts, like file and line no

unparalleled-js:
> I think it would be nice that it contains some sort of info of what triggers the aborts, like file and line no

Since the exception is not on the stack at this point, the only way I know how to do that is to inspect the call frames

PR #466, opened 2022-02-03
fix: use a default message in Abort when user does not provide message argument

unparalleled-js:
### What I did

fixes: #465 

### How I did it

Remove period thing
Make argument optional.
Always coerce to `"Operation aborted."`

### How to verify it

```python
raise Abort()
```

### Checklist

### DIFF
--- a/src/ape/cli/utils.py
+++ b/src/ape/cli/utils.py
@@ -15,13 +15,12 @@ class Abort(click.ClickException):
     """
 
     def __init__(self, message: Optional[str] = None):
-        caller = getframeinfo(stack()[1][0])
-        file_path = Path(caller.filename)
-        location = file_path.name if file_path.exists() else caller.filename
-        message = (
-            message
-            or f"Operation aborted in {location}::{caller.function} on line {caller.lineno}."
-        )
+        if not message:
+            caller = getframeinfo(stack()[1][0])
+            file_path = Path(caller.filename)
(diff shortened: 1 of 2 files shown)


src/ape/cli/utils.py
```diff
@@ -10,10 +13,12 @@ class Abort(click.ClickException):
...
 
-    def __init__(self, message):
-        if not message.endswith("."):
-            message = f"{message}."
-
+    def __init__(self, message: Optional[str] = None):
+        caller = getframeinfo(stack()[1][0])
+        message = (
+            message
+            or f"Operation aborted in {caller.filename}::{caller.function} on line {caller.lineno}."
```
unparalleled-js, line 20: Do we want to just always append `caller.filename` and `caller.function` to the message somewhere or maybe save it as a property of the exception so it can be debugged?

fubuloubu: If you are really trying to debug the stack, this info is probably already tracked, but I often feel like our errors don't have enough context for me to determine if it's a problem with something I wrote, or an actual bug in ape

unparalleled-js: Have you tried `-v debug` ?

fubuloubu (approved):

evanblank3 (approved):

fubuloubu (approved):

Issue #467, opened 2022-02-04
bug: Handle deprecation of importlib SelectableGroups dict interface


unparalleled-js:
### Environment information

latest ape

* Python Version: 3.10
* OS: macOS

### What went wrong?

Deprecation warning occur when launching ape.

```
DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    if "ape_cli_subcommands" not in entry_points:
```

### How can it be fixed?

See this PR as an example: https://github.com/brechtm/rinohtype/commit/7bbf2afc45d0dee3fb0f7650878340ac6114e97e

I think we can change it to `entry_points = metadata.entry_points(group="ape_cli_command")`
May have to do that for python 3.10 and greater only, I am not sure.

PR #473, opened 2022-02-05
fix: deprecation warning when using entry points like dict in some python environments

unparalleled-js:
### What I did

fixes: #467

### How I did it

try/catch using groups.

### How to verify it

`ape`

### Checklist


unparalleled-js:
mypy failing because of https://github.com/ApeWorX/ape/pull/475

src/ape/_cli.py
```diff
@@ -72,15 +72,22 @@ def _suggest_cmd(usage_error):
...
+                )
 
-            if "ape_cli_subcommands" not in entry_points:
+            if not entry_points:
                 raise Abort("Missing registered cli subcommands")
 
             self._commands = {
-                clean_plugin_name(entry_point.name): entry_point.load
-                for entry_point in entry_points["ape_cli_subcommands"]
+                clean_plugin_name(entry_point.name): entry_point.load  # type: ignore
```
fubuloubu, line 89: There's a type error here now?

fubuloubu (approved):

src/ape/_cli.py
unparalleled-js: The type error is being introduced here when trying to use the non-deprecated python 3.10 syntax.

unparalleled-js: Since upgrading to python 3.10, I have just noticed a lot of weird things like this. i am trying to fix the ones I see.
Here is some more info on it: https://github.com/python/importlib_metadata/issues/298

unparalleled-js: The mypy type erros is because `entry_points()` method returns either `SelectableGroups` (deprecated) or `EntryPoints`. It is supposed to always return `EntryPoints` when usings kwargs, i.e. `group=`. However, `mypy` does not realize that. Also, I tried importing the `EntryPoints` type but it is not declared in their `__all__` so I couldn't really. `type: ignore` seems to be the way until `SelectableGroups` is removed entirely.

unparalleled-js: OKAY finally figured out what was going on for real - python 3.8 requires a separate library that works like the original way (dict stuff).

fubuloubu (approved):

PR #474, opened 2022-02-05
fix: raise when invalid passphrase on delete

unparalleled-js:
### What I did

Fix error message when trying to delete a key file account that did not exist.

### How I did it

Copy what was happening in other areas of the code when decrypting the keyfile.
Made extreme private decrypt method that they all share.

### How to verify it

`ape accounts delete this-does-not-exist`

### Checklist


unparalleled-js:
lint failing because of this: https://github.com/ApeWorX/ape/pull/475

fubuloubu (approved):

PR #475, opened 2022-02-05
fix: handle importing 'collections.abc.Mapping' in python 3.10

unparalleled-js:
### What I did

Delete method causing `ImportError` in python 3.10 from importing `collections.Mapping`.

### How I did it

Method is not used so deleting it.

### How to verify it

`mypy` should work again (causing errors in other PRs). **Thanks mypy for finding this bug btw.**

### Checklist

### DIFF
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -1,4 +1,3 @@
-import collections
 import json
 import os
 import re
@@ -7,6 +6,7 @@ import sys
 import tempfile
 import zipfile
 from abc import ABCMeta, abstractmethod
+from collections import namedtuple
 from copy import deepcopy
 from functools import lru_cache, partial
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


src/ape/utils.py
```diff
@@ -41,6 +41,11 @@
     from functools import singledispatchmethod  # type: ignore
 except ImportError:
     from singledispatchmethod import singledispatchmethod  # type: ignore
+try:
+    from collections.abc import Mapping as CollectionsMapping  # type: ignore
```
unparalleled-js, line 45: renaming because using `typing.Mapping` the top.

src/ape/utils.py
```diff
@@ -150,25 +148,6 @@ def get_package_version(obj: Any) -> str:
 USER_AGENT = f"Ape/{__version__} (Python/{_python_version})"
 
 
-def deep_merge(dict1, dict2) -> Dict:
```
unparalleled-js, line 153: realized this method is not used anywhere, so deleting it.
This method is causing mypy issues for `pytohn 3.10`

fubuloubu: Yeah I think I had this when I was still parsing all the config stuff manually. Pydantic config objects will help with this

fubuloubu (approved):

PR #476, opened 2022-02-05
fix: make version incompatible with upgrade in 'ape plugins add' command

unparalleled-js:
### What I did

Currently, when you do `ape plugins add <plugin>`, it lets you provide both the `--version` flag as well as the `--upgrade` flag. It seems like the `--upgrade` flag just doesn't do anything, but probably better to just raise / cut out early in this case because this is an ambiguous command.

### How I did it

add decorator that parses other objects in context to make sure option is used without. Otherwise raises `click.BadOptionUsage`.

### How to verify it

`ape plugins add solidity --version 0.1.0b2 --upgrade`

### Checklist


fubuloubu (approved):
This is neat. Might even want to consider upstreaming this to `click`!

Issue #383, opened 2022-01-03
Plugins: allow plugins add to add multiple plugins at the same time example: `ape plugins add etherscan infura`


sabotagebeats:
### Overview

Provide a simple overview of what you wish to see added. Please include:

* What you are trying to do
add multiple plugins at the same time
* Why Ape's current functionality is inadequate to address your goal
each plugin has to be added one at a time

### Specification

Describe the syntax and semantics of how you would like to see this feature implemented. The more detailed the better!
Would like to see it to be able to add multiple plugins at the same time for example:
`ape plugins add etherscan infura`

Remember, your feature is much more likely to be included if it does not involve any breaking changes.

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

fubuloubu:
I was just thinking about this today!

fubuloubu:
cc @Ninjagod1251

Issue #269, opened 2021-11-24
Bug: `ape plugins install` does not bump the versions


unparalleled-js:
### Environment information

Latest ape

### What went wrong?

When you increment the version in a `ape-config.yaml` for plugins to use to install from, it does not install the update version.

fubuloubu:
Also was thinking about version pinning, this could be the right behavior (if the `version` key is the config)

unparalleled-js:
Maybe we should do it setup.py style? Allow `>=1.0.0,2.0.0` for example

fubuloubu:
> Maybe we should do it setup.py style? Allow `>=1.0.0,<2.0.0` for example

yeah, I like this

PR #479, opened 2022-02-07
fix: consolidate 'ape plugins' commands and fix issues with upgrading

unparalleled-js:
### What I did

fixes: #269 
fixes: #383 

was experiencing lots of bugs and inconsistencies in `ape plugins` set of commands. This PR addresses a bunch of that.

### How I did it

Lots of refactoring.
Handle results in new class.
Use `ApePlugins` class to abstract all the forms of the plugin name.
**Combine 'add' with 'uninstall' and 'remove' with 'uninstall'**

### How to verify it

`ape plugins install solidity alchemy hardhat==0.1.0b1`
`ape plugins install hardhat --upgrade`
`ape plugins uninstall hardhat`
`ape plugins uninstall`
`ape plugins install`

Make sure `ape plugins install` installs the version in the config file, even if it is already installed.
Make sure `ape plugins uninstall` and `ape plugins install --upgrade` work.

^ All negative testing you can think of for this as well. e.g. adding plugins that are already added, trying to update to version that does not exist, etc.

### Checklist


fubuloubu (dismissed):

fubuloubu:
Also wondering if we should take the oppurtunity to merge `ape plugins add {name}` and `ape plugins install` into `ape plugins install NAMES...` (with default case of no plugins meaning "install from config file"), and `ape plugins uninstall` instead of `ape plugins remove` to more closely mirror `pip` install experiences

unparalleled-js:
> Also wondering if we should take the oppurtunity to merge `ape plugins add {name}` and `ape plugins install` into `ape plugins install NAMES...` (with default case of no plugins meaning "install from config file"), and `ape plugins uninstall` instead of `ape plugins remove` to more closely mirror `pip` install experiences

Right now, we have a `--version` flag on `ape plugins add`. I am trying to figure out how that would look when passing multiple names. We'd have to do something like `ape plugins install solidity==0.1.0b2 hardhat==0.1.0b2` or something?

fubuloubu:
> We'd have to do something like `ape plugins install solidity==0.1.0b2 hardhat==0.1.0b2` or something?

Yeah that sounds good to me

Dockerfile
```diff
@@ -7,9 +7,5 @@ WORKDIR /code
 # TODO: Figure out a better solution or wait for it to resolve itself.
 RUN pip install typing-extensions==3.10.0.2
 RUN python3 ./setup.py install
-RUN ape plugins add solidity --yes
-RUN ape plugins add vyper --yes
-RUN ape plugins add infura --yes
-RUN ape plugins add etherscan --yes
-RUN ape plugins add ens --yes
+RUN ape plugins install solidity vyper infura etherscan ens
```
unparalleled-js, line 10: I made sure this builds.

src/ape_plugins/_cli.py
```diff
@@ -122,221 +155,110 @@ def _list(cli_ctx, display_all):
...
-    if plugin in CORE_PLUGINS:
-        cli_ctx.abort(f"Cannot add 1st class plugin '{plugin}'")
-
-    elif is_plugin_installed(plugin):
-        if upgrade:
-            cli_ctx.logger.info(f"Updating '{plugin}'...")
-            args.append("--upgrade")
-            args.append(plugin)
-            result = subprocess.call(args)
+@plugins_argument()
```
fubuloubu, line 160: Wonder if the default value here can be "everything in `ape-config.yaml:plugins`" somehow

src/ape_plugins/_cli.py
```diff
@@ -122,221 +155,110 @@ def _list(cli_ctx, display_all):
...
-    cwd = getcwd()
-    config_path = Path(cwd) / CONFIG_FILE_NAME
-    cli_ctx.logger.info(f"Uninstalling plugins from config file at {config_path}")
+def uninstall(cli_ctx, plugins, skip_confirmation):
+    """Uninstall plugins"""
 
-    plugins = config.get_config("plugins") or []
+    failures_occurred = False
+    did_warn_about_version = False
     for plugin in plugins:
```
fubuloubu, line 228: If `plugins` is empty, what happens?

fubuloubu: Similar thought, would be nice to have a `-a/--all` option to uninstall everything back to a clean state.

unparalleled-js: Does `--all` uninstall _all_ plugins (including 3rd party)? Or does it just uninstall the ones from the config? Because just doing `ape plugins uninstall` with no arguments will already uninstall all the plugins in the config. I am a little nervous about offering a full plugin-nuke option. It seems like it is okay just to do `ape plugins uninstall plugin-not-in-config-a plugin-not-in-config-b` and maybe paired with a `ape plugins uninstall`

unparalleled-js: Also, would `--all` uninstall plugins you have installed in editable mode?

src/ape_plugins/_cli.py
unparalleled-js: It is!

src/ape_plugins/_cli.py
fubuloubu: I think it should exit early and do nothing if you don't give any arguments of plugins to uninstall. `--all` could have it pick whatever is in the config file. Seems like if you're doing manual/editable installs, you probably would want to use `pip` directly.

unparalleled-js: that makes it inconsistent with the behavior of `ape plugins install` though

unparalleled-js: If `ape plugins install` uses the config when no plugins are given, I would expect `ape plugins uninstall` to do the same thing, no?

src/ape_plugins/_cli.py
```diff
@@ -41,7 +35,54 @@ def _format_output(plugins_list: Collection[str]) -> Set:
...
 
 
-@cli.command(name="list", short_help="Display plugins")
+def plugins_argument():
+    """
+    An argument that is either the given list of plugins
+    or plugins loaded from the local config file.
+    """
+
+    def callback(ctx, value: Tuple[str]):
```
fubuloubu, line 44: Why not have this function take `params` and eliminate the lambda below?

fubuloubu (dismissed):

src/ape_plugins/_cli.py
fubuloubu: Naw, just meant essentially what you ended up doing anyways lol

src/ape_plugins/_cli.py
fubuloubu: I see how you did this. I am fine with the current behavior.

fubuloubu (approved):

PR #480, opened 2022-02-07
feat: add query base structure

evanblank3:
### What I did

Getting a base structure started for @johnson2427 and myself.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
         "ethpm-types>=0.1.0b5",
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
-        "pandas>=1.4.0,<2.0",
+        "pandas>=1.3.0,<2.0",
         "pandas-stubs>=1.2.0,<2.0",
         "pluggy>=0.13.1,<1.0",
         "pydantic>=1.9.0,<2.0",
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,7 @@ setup(
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
         "pandas>=1.4.0,<2.0",
+        "pandas-stubs>=1.2.0,<2.0",
         "pluggy>=0.13.1,<1.0",
         "pydantic>=1.9.0,<2.0",
         "PyGithub>=1.54,<2.0",
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
         "ethpm-types>=0.1.0b5",
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
+        "pandas>=1.4.0,<2.0",
         "pluggy>=0.13.1,<1.0",
         "pydantic>=1.9.0,<2.0",
         "PyGithub>=1.54,<2.0",
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ extras_require = {
         "flake8-breakpoint>=1.1.0,<2.0.0",  # detect breakpoints left in code
         "flake8-print>=4.0.0,<5.0.0",  # detect print statements left in code
         "isort>=5.9.3,<6.0",  # Import sorting linter
+        "pandas-stubs>=1.2.0,<2.0",  # NOTE: Needed due to mypy types
     ],
     "doc": [
         "myst-parser==0.16.0",  # Tools for parsing markdown files in the docs
@@ -81,7 +82,6 @@ setup(
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
         "pandas>=1.3.0,<2.0",
-        "pandas-stubs>=1.2.0,<2.0",
         "pluggy>=0.13.1,<1.0",
         "pydantic>=1.9.0,<2.0",
         "PyGithub>=1.54,<2.0",
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -33,6 +33,11 @@ class _BaseBlockQuery(_BaseQuery):
 
 
 class BlockQuery(_BaseBlockQuery):
+    """
+    A `QueryType` that collects properties of ``BlockAPI`` over a range of
+    blocks between ``start_block`` and ``stop_block``.
+    """
+
     type: Literal["blocks"]
 
 
@@ -52,17 +57,32 @@ class _BaseAccountQuery(BaseModel):
 
 
 class AccountQuery(_BaseAccountQuery):
+    """
+    A ``QueryType`` that collects properties of ``AccountAPI`` over a range of
+    nonces between ``start_nonce`` and ``stop_nonce``.
+    """
+
     type: Literal["accounts"]
     account: AddressType
 
 
 class ContractEventQuery(_BaseBlockQuery):
+    """
+    A ``QueryType`` that collects properties of ``EventABI`` over a range of
+    blocks between ``start_block`` and ``stop_block`` for a given ``Contract`` address.
+    """
+
     type: Literal["contract_events"]
     contract: AddressType
     event: EventABI
 
 
 class ContractMethodQuery(_BaseBlockQuery):
+    """
+    A ``QueryType`` that collects properties of ``MethodABI`` over a range of
+    blocks between ``start_block`` and ``stop_block`` for a given ``Contract`` address.
+    """
+
     type: Literal["contract_calls"]
     contract: AddressType
     method: MethodABI
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -25,8 +25,8 @@ class _BaseBlockQuery(_BaseQuery):
     def check_start_block_before_stop_block(cls, values):
         if values["stop_block"] < values["start_block"]:
             raise ValueError(
-                f"stop_block: {values['stop_block']} cannot be less than "
-                f"start_block: {values['start_block']}."
+                f"stop_block: '{values['stop_block']}' cannot be less than "
+                f"start_block: '{values['start_block']}'."
             )
 
         return values
@@ -49,8 +49,8 @@ class _BaseAccountQuery(BaseModel):
     def check_start_nonce_before_stop_nonce(cls, values):
         if values["stop_nonce"] < values["start_nonce"]:
             raise ValueError(
-                f"stop_nonce: {values['stop_nonce']} cannot be less than "
-                f"start_nonce: {values['start_nonce']}."
+                f"stop_nonce: '{values['stop_nonce']}' cannot be less than "
+                f"start_nonce: '{values['start_nonce']}'."
             )
 
         return values
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Literal, Optional, Union
+import sys
+from typing import Any, Dict, List, Optional, Union
 
 import pandas as pd
 from pydantic import BaseModel, PositiveInt, root_validator
@@ -6,6 +7,12 @@ from pydantic import BaseModel, PositiveInt, root_validator
 from ape.managers.networks import NetworkManager
 from ape.types import AddressType
 from ape.utils import abstractdataclass, abstractmethod
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 QueryType = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractMethodQuery"]
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -59,7 +59,7 @@ class _BaseAccountQuery(BaseModel):
 class AccountQuery(_BaseAccountQuery):
     """
     A ``QueryType`` that collects properties of ``TransactionAPI`` over a range
-    of transactions between ``start_nonce`` and ``stop_nonce``.
+    of transactions made by ``account`` between ``start_nonce`` and ``stop_nonce``.
     """
 
     type: Literal["accounts"]
@@ -68,8 +68,8 @@ class AccountQuery(_BaseAccountQuery):
 
 class ContractEventQuery(_BaseBlockQuery):
     """
-    A ``QueryType`` that collects properties of ``EventABI`` over a range of
-    logs between ``start_block`` and ``stop_block`` for a given ``Event``.
+    A ``QueryType`` that collects members from ``event`` over a range of
+    logs emitted by ``contract`` between ``start_block`` and ``stop_block``.
     """
 
     type: Literal["contract_events"]
@@ -79,8 +79,8 @@ class ContractEventQuery(_BaseBlockQuery):
 
 class ContractMethodQuery(_BaseBlockQuery):
     """
-    A ``QueryType`` that collects properties of ``MethodABI`` over a range of
-    logs between ``start_block`` and ``stop_block`` for a given ``Method``.
+    A ``QueryType`` that collects return values from calling ``method`` in ``contract``
+    over a range of blocks between ``start_block`` and ``stop_block``.
     """
 
     type: Literal["contract_calls"]
@@ -99,13 +99,13 @@ class QueryAPI:
         """
         Estimation of time needed to complete the query. The estimation is returned
         as an int representing milliseconds. A value of None indicates that the
-        query engine is not available for use.
+        query engine is not available for use or is unable to complete the query.
 
         Args:
-            query (``QueryType``): query to estimate
+            query (``QueryType``): Query to estimate execution time for.
 
         Returns:
-            Optional[int]: Represents milliseconds, returns ``None`` if unavailable.
+            Optional[int]: Represents milliseconds, returns ``None`` if unable to execute.
 
         """
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -1,19 +1,13 @@
-import sys
 from typing import Any, Dict, List, Optional, Union
 
 import pandas as pd
 from ethpm_types.abi import EventABI, MethodABI
 from pydantic import BaseModel, PositiveInt, root_validator
 
+from ape._compat import Literal
 from ape.managers.networks import NetworkManager
 from ape.types import AddressType
 from ape.utils import abstractdataclass, abstractmethod
-
-# We can remove this once we stop supporting python3.7.
-if sys.version_info >= (3, 8):
-    from typing import Literal
-else:
-    from typing_extensions import Literal
 
 QueryType = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractMethodQuery"]
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -12,12 +12,12 @@ QueryType = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractM
 
 class _BaseQuery(BaseModel):
     type: str  # Used as discriminator
-    columns: List[str]
+    columns: Union[str, List[str]]
 
 
 class _BaseBlockQuery(_BaseQuery):
     start_block: PositiveInt = 0
-    stop_block: Optional[PositiveInt] = None
+    stop_block: PositiveInt
 
     @root_validator(pre=True)
     def check_start_block_before_stop_block(cls, values):
@@ -36,7 +36,7 @@ class BlockQuery(_BaseBlockQuery):
 
 class _BaseAccountQuery(BaseModel):
     start_nonce: PositiveInt = 0
-    stop_nonce: Optional[PositiveInt] = None
+    stop_nonce: PositiveInt
 
     @root_validator(pre=True)
     def check_start_nonce_before_stop_nonce(cls, values):
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -58,8 +58,8 @@ class _BaseAccountQuery(BaseModel):
 
 class AccountQuery(_BaseAccountQuery):
     """
-    A ``QueryType`` that collects properties of ``AccountAPI`` over a range of
-    nonces between ``start_nonce`` and ``stop_nonce``.
+    A ``QueryType`` that collects properties of ``TransactionAPI`` over a range
+    of transactions between ``start_nonce`` and ``stop_nonce``.
     """
 
     type: Literal["accounts"]
@@ -69,7 +69,7 @@ class AccountQuery(_BaseAccountQuery):
 class ContractEventQuery(_BaseBlockQuery):
     """
     A ``QueryType`` that collects properties of ``EventABI`` over a range of
-    blocks between ``start_block`` and ``stop_block`` for a given ``Contract`` address.
+    logs between ``start_block`` and ``stop_block`` for a given ``Event``.
     """
 
     type: Literal["contract_events"]
@@ -80,7 +80,7 @@ class ContractEventQuery(_BaseBlockQuery):
 class ContractMethodQuery(_BaseBlockQuery):
     """
     A ``QueryType`` that collects properties of ``MethodABI`` over a range of
-    blocks between ``start_block`` and ``stop_block`` for a given ``Contract`` address.
+    logs between ``start_block`` and ``stop_block`` for a given ``Method``.
     """
 
     type: Literal["contract_calls"]
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -34,7 +34,7 @@ class _BaseBlockQuery(_BaseQuery):
 
 class BlockQuery(_BaseBlockQuery):
     """
-    A `QueryType` that collects properties of ``BlockAPI`` over a range of
+    A ``QueryType`` that collects properties of ``BlockAPI`` over a range of
     blocks between ``start_block`` and ``stop_block``.
     """
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -1,0 +1,29 @@
+import pandas as pd
+
+from ape.types import Query
+
+
+class QueryAPI:
+    def estimate_query(self, query: Query) -> pd.DataFrame:
+        """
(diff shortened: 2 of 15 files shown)


johnson2427 (commented):
I think we are missing pandas in the setup.py

src/ape/api/query.py
```diff
@@ -0,0 +1,29 @@
+import pandas as pd
+
+from ape.types import Query
+
+
+class QueryAPI:
+    def estimate_query(self, query: Query) -> pd.DataFrame:
```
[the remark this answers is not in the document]
fubuloubu: Or maybe
```suggestion
    def estimate_query(self, query: Query) -> Optional[int]:
```

where `None` means "can't perform this query", and `int` is milliseconds to perform it

src/ape/types/__init__.py
```diff
@@ -14,6 +14,8 @@
 )
 from hexbytes import HexBytes
 
+from ape.managers.query import AccountQuery, BlockQuery, EventQuery, MethodQuery
```
fubuloubu, line 17: Definitely don't import these here, this module must be "pure" of any classes from api or managers

src/ape/api/query.py
```diff
@@ -0,0 +1,79 @@
...
+    type: Literal["contract_calls"]
+    contract: AddressType
+    method_id: bytes
+    method_args: Dict[str, Any]
+
+
+@abstractdataclass
+class QueryAPI:
+
+    engine_to_use: Optional[str] = None
```
evanblank3, line 55: @unparalleled-js brought up a good suggestions to add this as a way of overriding the query_engine choice provided by performance. If you already know you want to use postgres this should allow you to short-circuit the logic to access it. I'm very open to renaming things as well, but figured this is just a first pass to get some basic structure in place.

src/ape/api/query.py
```diff
@@ -0,0 +1,79 @@
...
+from ape.types import AddressType
+from ape.utils import abstractdataclass, abstractmethod
+
+Query = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractMethodQuery"]
+
+
+class BaseQuery(BaseModel):
+    type: str  # Used as discriminator
+    columns: List[str]
+    filter_args: Dict[str, Any]
```
evanblank3, line 15: We will probably need to have a bigger discussion around how to handle this filter_args.

src/ape/api/query.py
```diff
@@ -0,0 +1,79 @@
...
+    type: Literal["contract_events"]
+    contract: AddressType
+    event_id: bytes
+
+
+class ContractMethodQuery(BaseBlockQuery):
+    type: Literal["contract_calls"]
+    contract: AddressType
+    method_id: bytes
+    method_args: Dict[str, Any]
```
evanblank3, line 49: Same for method_args

src/ape/managers/query.py
```diff
@@ -53,6 +22,48 @@ class QueryManager:
...
+        Get an engine by name.
+
+        Args:
+            engine_name (str): The name of the engine to get.
+
+        Returns:
+            :class:`~ape.api.query.QueryAPI`
+        """
+        if engine_name not in self.plugin_manager.query_engines:
+            raise QueryEngineException(engine_name)
```
fubuloubu, line 43: With the above being a dict, you shouldn't have to do this anymore

However, not sure this manager should be user-facing, so I don't think we even need this method for easy access

src/ape/managers/query.py
```diff
@@ -53,6 +22,48 @@ class QueryManager:
...
+        Raises: :class:`~ape.exceptions.QueryEngineException`: When the
+            query.engine_to_use is invalid or inaccessible
+
+        Returns:
+            pandas.DataFrame
+        """
+
+        # Get heuristics from all the query engines to perform this query
+        # NOTE: `self.query_engines` should be `cached_property` of loading
+        #       `[engine for _, engine in PluginManager.query_engines]`
```
fubuloubu, line 61: This was just a note from my previous comment, you can remove haha

src/ape/managers/query.py
```diff
@@ -53,6 +22,48 @@ class QueryManager:
...
+        estimates = map(lambda qe: (qe, qe.estimate_query(query)), self.query_engines)
```
fubuloubu, line 62: If changed to a dict
```suggestion
        estimates = map(lambda qe: (qe, qe.estimate_query(query)), self.query_engines.values())
```

src/ape/plugins/query.py
```diff
@@ -0,0 +1,26 @@
...
+    @hookspec
+    def query_engines(self) -> Iterator[Tuple[str, Type[QueryAPI]]]:
+        """
+        A hook that returns an iterator of types of a string query type and a ``QueryAPI`` subclass
+
+        Usage example::
+
+            @plugins.register(plugins.QueryPlugin)
+            def query_engines():
+                yield postgres, PostgresEngine
```
fubuloubu, line 22:
```suggestion
                yield "postgres", PostgresEngine
```

fubuloubu (commented):
Coming along!

src/ape/api/query.py
fubuloubu: I like this, just see comments I made to better integrate this feature

src/ape/api/query.py
fubuloubu: Yeah... lol

Tbh, the `start_block...end_block` and `start_nonce...end_nonce` params are fulfilling this role, but other things might be available to a dataframe (TBD research item post this PR)

Can probably remove for now!

src/ape/api/query.py
fubuloubu: These are specifically the setting for each arg in a call e.g.:

```py
my_contract.myMethod(arg1, arg2, ...)  # Call with `arg1, arg2, ...`
# SAME MEANING AS
my_contract.myMethod.query(arg1, arg2, ...)  # Query with `arg1, arg2, ...`
```

evanblank3:
Had to add pandas-stubs to get mypy to be quiet.

src/ape/api/query.py
```diff
@@ -1,48 +1,66 @@
...
-class BaseBlockQuery(BaseQuery):
-    start_block: int = 0
-    stop_block: Optional[int] = None
+class _BaseBlockQuery(_BaseQuery):
+    start_block: PositiveInt = 0
+    stop_block: Optional[PositiveInt] = None
 
+    @root_validator(pre=True)
+    def check_start_block_before_stop_block(cls, values):
+        if values["stop_block"] < values["start_block"]:
```
unparalleled-js, line 24: what about also verifying they are not negative?

unparalleled-js: Oh wait, I see `PositiveInt` type

src/ape/api/query.py
```diff
@@ -0,0 +1,100 @@
...
+    type: Literal["blocks"]
+
+
+class _BaseAccountQuery(BaseModel):
+    start_nonce: PositiveInt = 0
+    stop_nonce: Optional[PositiveInt] = None
+
+    @root_validator(pre=True)
+    def check_start_nonce_before_stop_nonce(cls, values):
+        if values["stop_nonce"] < values["start_nonce"]:
```
unparalleled-js, line 43: Since `stop_nonce` is optional, are we potentially comparing an `int` with `None` here?

src/ape/api/query.py
```diff
@@ -0,0 +1,100 @@
...
+@abstractdataclass
+class QueryAPI:
+
+    network_manager: NetworkManager
+
+    @abstractmethod
+    def estimate_query(self, query: QueryType) -> Optional[int]:
+        """
+        Estimation of time needed to complete the query. The estimation is returned
+        as an int representing milliseconds. A value of None indicates that the
```
unparalleled-js, line 79: Is the `int` still milliseconds or is it more like a priority number? Having a hard time picturing how we can know how many milliseconds, even approximately, a query will take

src/ape/exceptions.py
```diff
@@ -217,3 +217,9 @@ def __init__(self, snapshot_id: "SnapshotID"):
             snapshot_id = humanize_hash(snapshot_id)  # type: ignore
 
         super().__init__(f"Unknown snapshot ID '{str(snapshot_id)}'.")
+
+
+class QueryEngineError(ApeException):
+    """
+    Raised when a query engine is unavailable.
```
unparalleled-js, line 224: Seems like this doc should be more generic like `Raised when issues occur in a query engine.` That way, we can re-use tis exception in more places if we ever need.

src/ape/managers/chain.py
```diff
@@ -1,18 +1,23 @@
...
 from ape.managers.networks import NetworkManager
+from ape.managers.query import QueryManager
 from ape.types import AddressType, BlockID, SnapshotID
 from ape.utils import cached_property, injected_before_use
 
 
 class _ConnectedChain:
     _networks: ClassVar[NetworkManager] = injected_before_use()  # type: ignore
     _converters: ClassVar[ConversionManager] = injected_before_use()  # type: ignore
+    _query_manager: ClassVar[QueryManager] = cast(QueryManager, injected_before_use())
```
unparalleled-js, line 20: when do we use `cast()` versus `type: ignore`?

src/ape/managers/query.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class QueryManager:
+    """
+    A singelton that manages all query sources.
+
+    Args:
+        query (``QueryType``): query to execute
+
+    Returns:
```
unparalleled-js, line 21: This `Returns` block is empty.
Also, it has to go underneathe the `Usage example::` block.

src/ape/managers/query.py
```diff
@@ -0,0 +1,81 @@
...
+
+        return engines
+
+    def query(self, query: QueryType, engine_to_use: Optional[str] = None) -> pd.DataFrame:
+        """
+        Args:
+            query (``QueryType``): The type of query to execute
+            engine_to_use_ (str): short-circuit selection logic using a specific engine
+
+        Raises: :class:`~ape.exceptions.QueryEngineNotFoundError`: When the
```
unparalleled-js, line 56: I think you renamed this exception, perhaps, to `QueryEngineError`?

unparalleled-js (commented):
Neat, was thinking about what you said about the user mostly filtering the Dataframe on their own after the query and that makes a lot of sense to me actually.

Maybe at some point we can just have the user provider a callback that takes a dataframe and returns a dataframe, then they use the dataframe filtering techniques in the query engine. IDK just a though.

src/ape/api/query.py
fubuloubu: The estimate is loose. It definitely takes less than 1 second to query a single block via a Provider, so take a loose estimate of what that is (maybe it's 50ms) and multiply that by number of blocks (let's say you want to get 100 blocks, so that's 5 seconds total)

It doesn't have to be super accurate, just a good estimate to use for comparison's sake

src/ape/managers/chain.py
evanblank3: This will all be changed in the pydantic refactor, so I was just adding it the newer way but ultimately it won't be here. All the managers will be set in the `ManagerAccessBase` class.

setup.py
```diff
@@ -80,6 +80,8 @@
         "ethpm-types>=0.1.0b5",
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
+        "pandas>=1.4.0,<2.0",
+        "pandas-stubs>=1.2.0,<2.0",
```
unparalleled-js, line 84: are the stubs only used for type-checking? If yes, they can be moved to the `"lint"` section (line 20)

src/ape/api/query.py
evanblank3: Yeah, it will require the same treatment as we do the BlockContainer query method, except on the account:

```python
        if stop_block is None:
            stop_block = len(self)

        if stop_block > len(self):
            raise ChainError(
                f"'stop_block={stop_block}' cannot be greater than the chain length ({len(self)}). "
                f"Use '{self.poll_blocks.__name__}()' to wait for future blocks."
            )
```

setup.py
```diff
@@ -80,6 +80,8 @@
         "ethpm-types>=0.1.0b5",
         "hexbytes>=0.2.2,<1.0.0",
         "packaging>=20.9,<21.0",
+        "pandas>=1.4.0,<2.0",
```
unparalleled-js, line 83: need pandas 1.3 to support python 3.7

src/ape/api/query.py
```diff
@@ -0,0 +1,107 @@
...
+class ContractEventQuery(_BaseBlockQuery):
+    type: Literal["contract_events"]
+    contract: AddressType
+    event_id: bytes
+
+
+class ContractMethodQuery(_BaseBlockQuery):
+    type: Literal["contract_calls"]
+    contract: AddressType
+    method_id: bytes
```
fubuloubu, line 73: Wondering if it would be better to have a full `MethodABI` object here instead, so it can do validation of `method_args` as well as encoding :thinking:

src/ape/api/query.py
```diff
@@ -0,0 +1,107 @@
...
+
+class AccountQuery(_BaseAccountQuery):
+    type: Literal["accounts"]
+    account: AddressType
+
+
+class ContractEventQuery(_BaseBlockQuery):
+    type: Literal["contract_events"]
+    contract: AddressType
+    event_id: bytes
```
fubuloubu, line 67: Same with `EventABI`

fubuloubu: Also, should `event_filter: Dict[str, Any]` be another arg, to allow you to apply filters based on `indexed` args being specific values?

src/ape/_cli.py
```diff
@@ -72,15 +72,22 @@ def _suggest_cmd(usage_error):
 
     @property
     def commands(self) -> Dict:
+        group_name = "ape_cli_subcommands"
```
unparalleled-js, line 75: This part shouldn't be in the diff, may need to catch your origin main up or catch this branch up or both, not sure!

src/ape/managers/chain.py
```diff
@@ -92,15 +97,71 @@ def __iter__(self) -> Iterator[BlockAPI]:
...
+            :class:`~ape.exceptions.ChainError`: When ``stop_block`` is greater
+              than the chain length.
+
+        Args:
+            columns (Union[str, List[str]]): columns in the DataFrame to return
+            start_block (int): The first block, by number, to include in the
+              range. Defaults to 0.
+            stop_block (Optional[int]): The block number to stop before. Defaults
+              to the lastest block.
+            engine_to_use (Optional[QueryAPI]): query engine to use, bypasses performance estimation
```
fubuloubu, line 123:
```suggestion
            engine_to_use (Optional[QueryAPI]): query engine to use, bypasses query engine selection algorithm
```

src/ape/managers/chain.py
```diff
@@ -92,15 +97,71 @@ def __iter__(self) -> Iterator[BlockAPI]:
...
+
+        Returns:
+            pandas.DataFrame
+        """
+
+        if stop_block is None:
+            stop_block = len(self)
+
+        if stop_block > len(self):
```
fubuloubu, line 132: could probably be `elif`

src/ape/managers/query.py
```diff
@@ -0,0 +1,79 @@
...
+from ape.utils import cached_property
+
+if TYPE_CHECKING:
+    from ape.managers.networks import NetworkManager
+    from ape.plugins import PluginManager
+
+
+class QueryManager:
+    """
+    A singelton that manages all query sources.
```
fubuloubu, line 16:
```suggestion
    A singleton that manages query engines and performs queries.
```

src/ape/managers/query.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Args:
+        query (``QueryType``): query to execute
+
+    Usage example::
+
+        chain.blocks.query()
```
fubuloubu, line 23: Better example?
```suggestion
        chain.blocks.query("size")
```

src/ape/_cli.py
fubuloubu, line 75: This fix was from another PR right?

src/ape/managers/query.py
evanblank3: Yes, the syntax for a DataFrame is slightly different.
```python
biggest_block_size = chain.blocks.query("size").max()
```

src/ape/managers/query.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        A dict of all :class:`~ape.api.query.QueryAPI` instances across all
+        installed plugins.
+
+        Returns:
+            dict[str, :class:`~ape.api.query.QueryAPI`]
+        """
+
+        engines = {}
+        for plugin_name, (engine_name, engine_class) in self.plugin_manager.query_engines:
```
fubuloubu, line 43: Oh yeah, I forgot about this lol

src/ape/plugins/query.py
```diff
@@ -0,0 +1,27 @@
...
+    from ape.api import QueryAPI
+
+
+class QueryPlugin(PluginType):
+    """
+    A plugin for querying chains.
+    """
+
+    @hookspec
+    def query_engines(self) -> Iterator[Tuple[str, Type["QueryAPI"]]]:
```
fubuloubu, line 15: Since `plugin_name` is already given by the PluginManager, this would be redundant (e.g. `ape-postgres` already tells you it's `postgres`)
```suggestion
    def query_engines(self) -> Iterator[Type["QueryAPI"]]:
```

src/ape/api/query.py
```diff
@@ -0,0 +1,108 @@
...
+import pandas as pd
+from ethpm_types.abi import EventABI, MethodABI
+from pydantic import BaseModel, PositiveInt, root_validator
+
+from ape.managers.networks import NetworkManager
+from ape.types import AddressType
+from ape.utils import abstractdataclass, abstractmethod
+
+# We can remove this once we stop supporting python3.7.
+if sys.version_info >= (3, 8):
```
unparalleled-js, line 13: We should create a `_compat.py` file in the `ape` folder and import `Literal` from there both here and `ape.types` and where-ever else it's used. That way this mess is only in one spot.

unparalleled-js: Example: https://github.com/commaai/usr/blob/master/local/lib/python2.7/site-packages/service_identity/_compat.py

src/ape/api/query.py
```diff
@@ -0,0 +1,108 @@
+import sys
```
unparalleled-js, line 1: I think this whole file might currently be excluded from the docs and you may need to add some stuff there

src/ape/managers/chain.py
```diff
@@ -92,15 +97,72 @@ def __iter__(self) -> Iterator[BlockAPI]:
...
+        self,
+        columns: Union[str, List[str]],
+        start_block: int = 0,
+        stop_block: Optional[int] = None,
+        engine_to_use: Optional[QueryAPI] = None,
+    ) -> pd.DataFrame:
+        """
+        A method for querying blocks and returning a pandas DataFrame. If you
+        do not provide a starting block, the 0 block is assumed. If you do not
+        provide a stopping block, the last block is assumed. You can pass engine_to_use
```
unparalleled-js, line 110:
```suggestion
        provide a stopping block, the last block is assumed. You can pass ``engine_to_use``
```

src/ape/managers/chain.py
```diff
@@ -92,15 +97,72 @@ def __iter__(self) -> Iterator[BlockAPI]:
...
+        Raises:
+            :class:`~ape.exceptions.ChainError`: When ``stop_block`` is greater
+              than the chain length.
+
+        Args:
+            columns (Union[str, List[str]]): columns in the DataFrame to return
+            start_block (int): The first block, by number, to include in the
+              range. Defaults to 0.
+            stop_block (Optional[int]): The block number to stop before. Defaults
+              to the lastest block.
```
unparalleled-js, line 122:
```suggestion
              to the latest block. Defaults to the latest block.
```

src/ape/managers/chain.py
```diff
@@ -92,15 +97,72 @@ def __iter__(self) -> Iterator[BlockAPI]:
...
+            engine_to_use (Optional[QueryAPI]): query engine to use, bypasses query
+              engine selection algorithm.
+
+        Returns:
+            pandas.DataFrame
+        """
+
+        if stop_block is None:
+            stop_block = len(self)
```
unparalleled-js, line 131: Isn't `len(self)` technically the block number of the block after the last one?

src/ape/managers/query.py
```diff
@@ -0,0 +1,81 @@
...
+            engine_name = clean_plugin_name(plugin_name)
+            engines[engine_name] = engine_class(network_manager=self.network_manager)
+
+        return engines
+
+    def query(self, query: QueryType, engine_to_use: Optional[str] = None) -> pd.DataFrame:
+        """
+        Args:
+            query (``QueryType``): The type of query to execute
+            engine_to_use_ (str): short-circuit selection logic using a specific engine
```
unparalleled-js, line 54:
```suggestion
            engine_to_use_ (Optional[str]): Short-circuit selection logic using a specific engine. Defaults to None.
```

src/ape/managers/query.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Raises: :class:`~ape.exceptions.QueryEngineError`: When the engine_to_use
```
unparalleled-js, line 56:
```suggestion
        Raises: :class:`~ape.exceptions.QueryEngineError`: When given an ``engine_to_use`` that
```

src/ape/managers/query.py
```diff
@@ -0,0 +1,81 @@
...
+            is invalid or inaccessible
+
+        Returns:
+            pandas.DataFrame
+        """
+        if engine_to_use:
+            if engine_to_use not in self.engines:
+                raise QueryEngineError(f"Query engine {engine_to_use} not found.")
```
unparalleled-js, line 64:
```suggestion
                raise QueryEngineError(f"Query engine '{engine_to_use}' not found.")
```

The quotes really help me parse the error messages and makes it consistent with other errors in ape

unparalleled-js (commented):
The main thing I am wondering about is the comment using `len()`

src/ape/api/query.py
```diff
@@ -0,0 +1,122 @@
...
+                f"stop_block: '{values['stop_block']}' cannot be less than "
+                f"start_block: '{values['start_block']}'."
+            )
+
+        return values
+
+
+class BlockQuery(_BaseBlockQuery):
+    """
+    A `QueryType` that collects properties of ``BlockAPI`` over a range of
```
unparalleled-js, line 37:
```suggestion
    A ``QueryType`` that collects properties of ``BlockAPI`` over a range of
```

src/ape/managers/chain.py
evanblank3: @unparalleled-js thanks for catching this!

unparalleled-js (dismissed):
LGTM

src/ape/api/query.py
```diff
@@ -0,0 +1,122 @@
...
+                f"start_nonce: '{values['start_nonce']}'."
+            )
+
+        return values
+
+
+class AccountQuery(_BaseAccountQuery):
+    """
+    A ``QueryType`` that collects properties of ``TransactionAPI`` over a range
+    of transactions between ``start_nonce`` and ``stop_nonce``.
```
fubuloubu, line 62:
```suggestion
    of transactions made by ``account`` between ``start_nonce`` and ``stop_nonce``.
```

src/ape/api/query.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+
+    type: Literal["accounts"]
+    account: AddressType
+
+
+class ContractEventQuery(_BaseBlockQuery):
+    """
+    A ``QueryType`` that collects properties of ``EventABI`` over a range of
+    logs between ``start_block`` and ``stop_block`` for a given ``Event``.
```
fubuloubu, line 72:
```suggestion
    A ``QueryType`` that collects members from ``event`` over a range of
    logs emitted by ``contract`` between ``start_block`` and ``stop_block``.
```

src/ape/api/query.py
```diff
@@ -0,0 +1,122 @@
...
+
+    type: Literal["contract_events"]
+    contract: AddressType
+    event: EventABI
+
+
+class ContractMethodQuery(_BaseBlockQuery):
+    """
+    A ``QueryType`` that collects properties of ``MethodABI`` over a range of
+    logs between ``start_block`` and ``stop_block`` for a given ``Method``.
```
fubuloubu, line 83:
```suggestion
    A ``QueryType`` that collects return values from calling ``method``  in ``contract``
    over a range of blocks between ``start_block`` and ``stop_block``.
```

src/ape/api/query.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @abstractmethod
+    def estimate_query(self, query: QueryType) -> Optional[int]:
+        """
+        Estimation of time needed to complete the query. The estimation is returned
+        as an int representing milliseconds. A value of None indicates that the
+        query engine is not available for use.
+
+        Args:
+            query (``QueryType``): query to estimate
```
fubuloubu, line 105:
```suggestion
            query (``QueryType``): Query to estimate execution time for.
```

src/ape/managers/query.py
```diff
@@ -0,0 +1,82 @@
...
+        Raises: :class:`~ape.exceptions.QueryEngineError`: When given an
+            invalid or inaccessible ``engine_to_use`` value.
+
+        Returns:
+            pandas.DataFrame
+        """
+        if engine_to_use:
+            if engine_to_use not in self.engines:
+                raise QueryEngineError(f"Query engine `{engine_to_use}` not found.")
+            return self.engines[engine_to_use].perform_query(query)
```
fubuloubu, line 66: nitpick, but I find this easier to read with the space
```suggestion

            return self.engines[engine_to_use].perform_query(query)
```

fubuloubu (approved):
A few changes, and LGTM!

fubuloubu (approved):
Note: needs tests, but also needs 1 impl of `QueryAPI` first

PR #487, opened 2022-02-10
refactor: make range work more like python builtin

unparalleled-js:
### What I did

Made `chain.blocks.range` work even more like python built in range method by making it so if you provide a single value, that value is considered the end of the range instead of the beginning.

### How I did it

check if second arg is None, if so, set second arg to first arg and first arg to 0

### How to verify it

```python
for i in chain.blocks.range(5):
  print(chain.blocks[i])
```

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -90,24 +90,26 @@ class BlockContainer(_ConnectedChain):
             Iterator[:class:`~ape.api.providers.BlockAPI`]
         """
 
-        return self.range()
+        return self.range(len(self))
 
-    def range(self, start: int = 0, stop: Optional[int] = None) -> Iterator[BlockAPI]:
+    def range(self, start: int, stop: Optional[int] = None) -> Iterator[BlockAPI]:
         """
         Iterate over blocks. Works similarly to python ``range()``.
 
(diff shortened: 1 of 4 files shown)


evanblank3 (approved):

fubuloubu (approved):

unparalleled-js:
This already approved at one point but lost the approvals due to a merge conflict

src/ape/managers/chain.py
```diff
@@ -145,32 +145,40 @@ def query(
...
                 than 0.
-            :class:`~ape.exceptions.ChainError`: When ``start_block`` is less
+            :class:`~ape.exceptions.ChainError`: When ``start`` is less
                 than 0.
 
         Args:
-            start (int): The first block, by number, to include in the range.
-              Defaults to 0.
+            start_or_stop (int): When given just a single value, it is the stop.
+              Otherwise, it is the start.
```
fubuloubu, line 166:
```suggestion
              Otherwise, it is the start. This mimics the behavior of ``range`` built-in Python function.
```

fubuloubu (approved):

fubuloubu (approved):

PR #501, opened 2022-02-20
refactor: store results of query for plugins that cache

fubuloubu:
### What I did
Thought about how if a plugin is slow on first try, but quick after that (e.g. Caching design), there is no mechanism available for the database to update its cache if it is not called. This adds an option method that will be called on all plugins that did _not_ execute the query so that they can do this. 

### How I did it
Add an interface 

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -130,4 +130,3 @@ class QueryAPI:
             query (``QueryType``): query that was executed
             result (``pandas.DataFrame``): the result of the query
         """
-        pass
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -120,3 +120,14 @@ class QueryAPI:
         Returns:
             pandas.DataFrame
         """
+
+    def update_cache(self, query: QueryType, result: pd.DataFrame):
+        """
+        Allows a query plugin the chance to update any cache using the results obtained
+        from other query plugins. Defaults to doing nothing, override to store cache data.
+
+        Args:
+            query (``QueryType``): query that was executed
+            result (``pandas.DataFrame``): the result of the query
+        """
+        pass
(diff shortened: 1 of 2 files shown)


src/ape/api/query.py
```diff
@@ -120,3 +120,14 @@ def perform_query(self, query: QueryType) -> pd.DataFrame:
...
+    def update_cache(self, query: QueryType, result: pd.DataFrame):
+        """
+        Allows a query plugin the chance to update any cache using the results obtained
+        from other query plugins. Defaults to doing nothing, override to store cache data.
+
+        Args:
+            query (``QueryType``): query that was executed
+            result (``pandas.DataFrame``): the result of the query
+        """
+        pass
```
unparalleled-js, line 133:
```suggestion
```

Not needed with doc strs

src/ape/managers/query.py
```diff
@@ -64,20 +64,28 @@ def query(self, query: QueryType, engine_to_use: Optional[str] = None) -> pd.Dat
...
 
         # Go fetch the result from the engine
-        return engine.perform_query(query)
+        result = engine.perform_query(query)
+
+        # Update any caches
+        for engine in self.engines.values():
+            engine.update_cache(query, result)
+
+        # Finally, return the result
```
unparalleled-js, line 90:
```suggestion
```

unparalleled-js: The code does well at speaking for itself on this last part

unparalleled-js (approved):
Looks good! Couple small things.

src/ape/api/query.py
fubuloubu: How n00b of me lol

unparalleled-js (approved):

Issue #126, opened 2021-08-11
Querying: Migrate to Pydantic


fubuloubu:
### Elevator pitch

Pydantic is a better supported and easier to use library than dataclassy

### Value

To make maintenance and new features easier to add

### Dependencies 

#355

PR #509, opened 2022-02-24
refactor: breaking pydantic changes across all repos

NotPeopling2day:
### What I did

Adds Manager access to assist with dependency issues
Adds abstract base classes to pass manager access
Adds pydantic models for validation and serialization
Removes dataclassy from core repo 

Major thanks to @unparalleled-js and @fubuloubu 

fixes: #126 
fixes: #524 allowing warning for extra config fields

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -3,7 +3,7 @@ import signal
 signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
 
 import sys as _sys
-from functools import partial
+from functools import partial as _partial
 
 from ape.utils import ManagerAccessBase as _ManagerAccessBase
 
@@ -44,7 +44,7 @@ accounts = _ManagerAccessBase.account_manager
 project = _ManagerAccessBase.project_manager
 """The currently active project. See :class:`ape.managers.project.ProjectManager`."""
 
-Contract = partial(_Contract, networks=networks, converters=_converters)
+Contract = _partial(_Contract, networks=networks, converters=_converters)
 """User-facing class for instantiating contracts. See :class:`ape.contracts.base._Contract`."""
 
 convert = _ManagerAccessBase.conversion_manager.convert
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -13,7 +13,8 @@ from .managers.project import ProjectManager as Project
 # Wiring together the application
 
 config = _ManagerAccessMixin.config_manager
-"""The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
+"""
+The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
 """
 
 # Main types we export for the user
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -12,9 +12,6 @@ from .managers.project import ProjectManager as Project
 
 # Wiring together the application
 
-plugin_manager = _ManagerAccessMixin.plugin_manager
-"""Manages plugins for the current project. See :class:`ape.plugins.PluginManager`."""
-
 config = _ManagerAccessMixin.config_manager
 """The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
 """
@@ -27,8 +24,6 @@ compilers = _ManagerAccessMixin.compiler_manager
 networks = _ManagerAccessMixin.network_manager
 """Manages the networks for the current project. See
 :class:`ape.managers.networks.NetworkManager`."""
-
-_converters = _ManagerAccessMixin.conversion_manager
 
 chain = _ManagerAccessMixin.chain_manager
 """
@@ -43,7 +38,7 @@ accounts = _ManagerAccessMixin.account_manager
 project = _ManagerAccessMixin.project_manager
 """The currently active project. See :class:`ape.managers.project.ProjectManager`."""
 
-Contract = _partial(_Contract, networks=networks, converters=_converters)
+Contract = _partial(_Contract, networks=networks, converters=_ManagerAccessMixin.conversion_manager)
 """User-facing class for instantiating contracts. See :class:`ape.contracts.base._Contract`."""
 
 convert = _ManagerAccessMixin.conversion_manager.convert
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -39,7 +39,9 @@ accounts = _ManagerAccessMixin.account_manager
 project = _ManagerAccessMixin.project_manager
 """The currently active project. See :class:`ape.managers.project.ProjectManager`."""
 
-Contract = _partial(_Contract, networks=networks, converters=_ManagerAccessMixin.conversion_manager)
+Contract = _partial(
+    _Contract, networks=networks, conversion_manager=_ManagerAccessMixin.conversion_manager
+)
 """User-facing class for instantiating contracts. See :class:`ape.contracts.base._Contract`."""
 
 convert = _ManagerAccessMixin.conversion_manager.convert
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -5,48 +5,48 @@ signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
 import sys as _sys
 from functools import partial as _partial
 
-from ape.utils import ManagerAccessBase as _ManagerAccessBase
+from ape.utils import ManagerAccessMixin as _ManagerAccessMixin
 
 from .contracts import _Contract
 from .managers.project import ProjectManager as Project
 
 # Wiring together the application
 
-plugin_manager = _ManagerAccessBase.plugin_manager
+plugin_manager = _ManagerAccessMixin.plugin_manager
 """Manages plugins for the current project. See :class:`ape.plugins.PluginManager`."""
 
-config = _ManagerAccessBase.config_manager
+config = _ManagerAccessMixin.config_manager
 """The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
 """
 
 # Main types we export for the user
-compilers = _ManagerAccessBase.compiler_manager
+compilers = _ManagerAccessMixin.compiler_manager
 """Manages compilers for the current project. See
 :class:`ape.managers.compilers.CompilerManager`."""
 
-networks = _ManagerAccessBase.network_manager
+networks = _ManagerAccessMixin.network_manager
 """Manages the networks for the current project. See
 :class:`ape.managers.networks.NetworkManager`."""
 
-_converters = _ManagerAccessBase.conversion_manager
+_converters = _ManagerAccessMixin.conversion_manager
 
-chain = _ManagerAccessBase.chain_manager
+chain = _ManagerAccessMixin.chain_manager
 """
 The current connected blockchain; requires an active provider.
 Useful for development purposes, such as controlling the state of the blockchain.
 Also handy for querying data about the chain and managing local caches.
 """
 
-accounts = _ManagerAccessBase.account_manager
+accounts = _ManagerAccessMixin.account_manager
 """Manages accounts for the current project. See :class:`ape.managers.accounts.AccountManager`."""
 
-project = _ManagerAccessBase.project_manager
+project = _ManagerAccessMixin.project_manager
 """The currently active project. See :class:`ape.managers.project.ProjectManager`."""
 
 Contract = _partial(_Contract, networks=networks, converters=_converters)
 """User-facing class for instantiating contracts. See :class:`ape.contracts.base._Contract`."""
 
-convert = _ManagerAccessBase.conversion_manager.convert
+convert = _ManagerAccessMixin.conversion_manager.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
 
 __all__ = [
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -17,7 +17,6 @@ plugin_manager = _ManagerAccessBase.plugin_manager
 
 config = _ManagerAccessBase.config_manager
 """The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
-# NOTE: config is an injected property.
 """
 
 # Main types we export for the user
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -20,7 +20,6 @@ _DIFFLIB_CUT_OFF = 0.6
 
 
 def display_config(ctx, param, value):
-
     # NOTE: This is necessary not to interrupt how version or help is intercepted
     if not value or ctx.resilient_parsing:
         return
@@ -37,15 +36,11 @@ class ApeCLI(click.MultiCommand):
     _commands = None
 
     def invoke(self, ctx) -> Any:
-
         try:
             return super().invoke(ctx)
-
         except click.UsageError as err:
             self._suggest_cmd(err)
-
         except ApeException as err:
-
             if logger.level == LogLevel.DEBUG.value:
                 tb = traceback.format_exc()
                 err_message = tb or str(err)
@@ -57,7 +52,6 @@ class ApeCLI(click.MultiCommand):
 
     @staticmethod
     def _suggest_cmd(usage_error):
-
         if usage_error.message is None:
             raise usage_error
 
@@ -70,7 +64,6 @@ class ApeCLI(click.MultiCommand):
         suggested_commands = difflib.get_close_matches(
             bad_arg, list(usage_error.ctx.command.commands.keys()), cutoff=_DIFFLIB_CUT_OFF
         )
-
         if suggested_commands:
             if bad_arg not in suggested_commands:
                 usage_error.message = (
@@ -81,13 +74,10 @@ class ApeCLI(click.MultiCommand):
 
     @property
     def commands(self) -> Dict:
-
         group_name = "ape_cli_subcommands"
-
         if not self._commands:
             try:
                 entry_points = metadata.entry_points(group=group_name)  # type: ignore
-
             except TypeError:
                 entry_points = metadata.entry_points()
                 entry_points = (
@@ -108,11 +98,9 @@ class ApeCLI(click.MultiCommand):
         return list(sorted(self.commands))
 
     def get_command(self, ctx, name):
-
         if name in self.commands:
             try:
                 return self.commands[name]()
-
             except Exception as err:
                 logger.warn_from_exception(
                     err, f"Unable to load CLI endpoint for plugin 'ape_{name}'"
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -44,7 +44,6 @@ class ApeCLI(click.MultiCommand):
             if logger.level == LogLevel.DEBUG.value:
                 tb = traceback.format_exc()
                 err_message = tb or str(err)
-
             else:
                 err_message = str(err)
 
@@ -56,7 +55,6 @@ class ApeCLI(click.MultiCommand):
             raise usage_error
 
         match = re.match("No such command '(.*)'.", usage_error.message)
-
         if not match:
             raise usage_error
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -105,7 +105,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             if txn.max_priority_fee is None:  # type: ignore
                 txn.max_priority_fee = self.provider.priority_fee  # type: ignore
 
-            if txn.max_fee is None:
+            if not txn.max_fee:
                 txn.max_fee = self.provider.base_fee + txn.max_priority_fee
             # else: Assume user specified the correct amount or txn will fail and waste gas
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -105,7 +105,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             if txn.max_priority_fee is None:  # type: ignore
                 txn.max_priority_fee = self.provider.priority_fee  # type: ignore
 
-            if not txn.max_fee:
+            if txn.max_fee is None:
                 txn.max_fee = self.provider.base_fee + txn.max_priority_fee
             # else: Assume user specified the correct amount or txn will fail and waste gas
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, ClassVar, Iterator, List, Optional, Type, Union, cast
+from typing import TYPE_CHECKING, Callable, Iterator, List, Optional, Type, Union
 
 import click
 from eth_account import Account
@@ -8,21 +8,19 @@ from ape.exceptions import AccountsError, AliasAlreadyInUseError, SignatureError
 from ape.logging import logger
 from ape.types import AddressType, MessageSignature, SignableMessage, TransactionSignature
 from ape.types.signatures import _Signature
-from ape.utils import BaseInterfaceModel, abstractmethod, cached_property, injected_before_use
+from ape.utils import BaseInterfaceModel, abstractmethod, cached_property
 
-from .address import AddressBase
+from .address import BaseAddress
 from .providers import ReceiptAPI, TransactionAPI, TransactionType
 
 if TYPE_CHECKING:
     from ape.contracts import ContractContainer, ContractInstance
 
 
-class AccountAPI(BaseInterfaceModel, AddressBase):
+class AccountAPI(BaseInterfaceModel, BaseAddress):
     """
     An API class representing an account.
     """
-
-    container: ClassVar["AccountContainerAPI"] = cast("AccountContainerAPI", injected_before_use())
 
     def __dir__(self) -> List[str]:
         """
@@ -31,7 +29,7 @@ class AccountAPI(BaseInterfaceModel, AddressBase):
         Returns:
             List[str]: Method names that IPython uses for tab completion.
         """
-        return list(super(AddressBase, self).__dir__()) + [
+        return list(super(BaseAddress, self).__dir__()) + [
             "alias",
             "sign_message",
             "sign_transaction",
@@ -145,7 +143,7 @@ class AccountAPI(BaseInterfaceModel, AddressBase):
 
     def transfer(
         self,
-        account: Union[str, AddressType, AddressBase],
+        account: Union[str, AddressType, BaseAddress],
         value: Union[str, int, None] = None,
         data: Union[bytes, str, None] = None,
         **kwargs,
@@ -302,7 +300,7 @@ class AccountContainerAPI(BaseInterfaceModel):
         """
         self._verify_account_type(account)
 
-        if account.address in self.accounts:  # type: ignore
+        if account.address in self:
             raise AccountsError(f"Account '{account.address}' already in container.")
 
         self._verify_unused_alias(account)
@@ -324,7 +322,7 @@ class AccountContainerAPI(BaseInterfaceModel):
         """
         self._verify_account_type(account)
 
-        if account.address not in self.accounts:  # type: ignore
+        if account.address not in self:
             raise AccountsError(f"Account '{account.address}' not known.")
 
         self.__delitem__(account.address)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -229,7 +229,6 @@ class AccountAPI(AbstractBaseModel, AddressBase):
                 raise ValueError(
                     "Parameter 'signature' required when verifying a 'SignableMessage'."
                 )
-
         elif isinstance(data, TransactionAPI):
             return self.address == Account.recover_transaction(data.serialize_transaction())
         else:
@@ -371,7 +370,6 @@ class AccountContainerAPI(AbstractBaseModel):
             raise AccountsError(message)
 
     def _verify_unused_alias(self, account):
-
         if account.alias and account.alias in self.aliases:
             raise AliasAlreadyInUseError(account.alias)
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -8,7 +8,7 @@ from ape.exceptions import AccountsError, AliasAlreadyInUseError, SignatureError
 from ape.logging import logger
 from ape.types import AddressType, MessageSignature, SignableMessage, TransactionSignature
 from ape.types.signatures import _Signature
-from ape.utils import AbstractBaseModel, abstractmethod, cached_property, injected_before_use
+from ape.utils import BaseInterfaceModel, abstractmethod, cached_property, injected_before_use
 
 from .address import AddressBase
 from .providers import ReceiptAPI, TransactionAPI, TransactionType
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
     from ape.contracts import ContractContainer, ContractInstance
 
 
-class AccountAPI(AbstractBaseModel, AddressBase):
+class AccountAPI(BaseInterfaceModel, AddressBase):
     """
     An API class representing an account.
     """
@@ -235,7 +235,7 @@ class AccountAPI(AbstractBaseModel, AddressBase):
             raise ValueError(f"Unsupported Message type: {type(data)}.")
 
 
-class AccountContainerAPI(AbstractBaseModel):
+class AccountContainerAPI(BaseInterfaceModel):
     """
     An API class representing a collection of :class:`~ape.api.accounts.AccountAPI`
     instances.
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -27,6 +27,7 @@ class AccountAPI(AbstractBaseModel, AddressBase):
     def __dir__(self) -> List[str]:
         """
         Display methods to IPython on ``a.[TAB]`` tab completion.
+
         Returns:
             List[str]: Method names that IPython uses for tab completion.
         """
@@ -90,21 +91,18 @@ class AccountAPI(AbstractBaseModel, AddressBase):
         Returns:
             :class:`~ape.api.providers.ReceiptAPI`
         """
-
         # NOTE: Use "expected value" for Chain ID, so if it doesn't match actual, we raise
         txn.chain_id = self.provider.network.chain_id
 
         # NOTE: Allow overriding nonce, assume user understand what this does
         if txn.nonce is None:
             txn.nonce = self.nonce
-
         elif txn.nonce < self.nonce:
             raise AccountsError("Invalid nonce, will not publish.")
 
         txn_type = TransactionType(txn.type)
         if txn_type == TransactionType.STATIC and txn.gas_price is None:  # type: ignore
             txn.gas_price = self.provider.gas_price  # type: ignore
-
         elif txn_type == TransactionType.DYNAMIC:
             if txn.max_priority_fee is None:  # type: ignore
                 txn.max_priority_fee = self.provider.priority_fee  # type: ignore
@@ -119,7 +117,7 @@ class AccountAPI(AbstractBaseModel, AddressBase):
 
         if send_everything:
             if txn.max_fee is None:
-                raise TransactionError(message="Max fee must not be null.")
+                raise TransactionError(message="Max fee must not be None.")
 
             txn.value = self.balance - txn.max_fee
 
@@ -132,7 +130,6 @@ class AccountAPI(AbstractBaseModel, AddressBase):
 
         if txn.required_confirmations is None:
             txn.required_confirmations = self.provider.network.required_confirmations
-
         elif not isinstance(txn.required_confirmations, int) or txn.required_confirmations < 0:
             raise TransactionError(message="'required_confirmations' must be a positive integer.")
 
@@ -225,11 +222,9 @@ class AccountAPI(AbstractBaseModel, AddressBase):
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
         """
-
         if isinstance(data, SignableMessage):
             if signature:
                 return self.address == Account.recover_message(data, vrs=signature)
-
             else:
                 raise ValueError(
                     "Parameter 'signature' required when verifying a 'SignableMessage'."
@@ -290,7 +285,6 @@ class AccountContainerAPI(AbstractBaseModel):
         Returns:
             :class:`~ape.api.accounts.AccountAPI`
         """
-
         for account in self.accounts:
             if account.address == address:
                 return account
@@ -307,7 +301,6 @@ class AccountContainerAPI(AbstractBaseModel):
         Args:
             account (:class:`~ape.api.accounts.AccountAPI`): The account to add.
         """
-
         self._verify_account_type(account)
 
         if account.address in self.accounts:
@@ -330,7 +323,6 @@ class AccountContainerAPI(AbstractBaseModel):
         Args:
             account (:class:`~ape.accounts.AccountAPI`): The account to remove.
         """
-
         self._verify_account_type(account)
 
         if account.address not in self.accounts:
@@ -348,7 +340,6 @@ class AccountContainerAPI(AbstractBaseModel):
         Args:
             address (``AddressType``): The address of the account to delete.
         """
-
         raise NotImplementedError("Must define this method to use `container.remove(acct)`.")
 
     def __contains__(self, address: AddressType) -> bool:
@@ -364,7 +355,6 @@ class AccountContainerAPI(AbstractBaseModel):
         Returns:
             bool: ``True`` if ``ape`` manages the account with the given address.
         """
-
         try:
             self.__getitem__(address)
             return True
@@ -373,7 +363,6 @@ class AccountContainerAPI(AbstractBaseModel):
             return False
 
     def _verify_account_type(self, account):
-
         if not isinstance(account, self.account_type):
             message = (
                 f"Container '{type(account).__name__}' is an incorrect "
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -17,7 +17,6 @@ if TYPE_CHECKING:
     from ape.contracts import ContractContainer, ContractInstance
 
 
-# NOTE: AddressBase is an AbstractBaseModel already
 class AccountAPI(AbstractBaseModel, AddressBase):
     """
     An API class representing an account.
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -256,7 +256,7 @@ class AccountContainerAPI(BaseInterfaceModel):
 
     @property
     @abstractmethod
-    def accounts(self):
+    def accounts(self) -> Iterator[AccountAPI]:
         """
         Iterate over all accounts.
 
@@ -302,7 +302,7 @@ class AccountContainerAPI(BaseInterfaceModel):
         """
         self._verify_account_type(account)
 
-        if account.address in self.accounts:
+        if account.address in self.accounts:  # type: ignore
             raise AccountsError(f"Account '{account.address}' already in container.")
 
         self._verify_unused_alias(account)
@@ -324,7 +324,7 @@ class AccountContainerAPI(BaseInterfaceModel):
         """
         self._verify_account_type(account)
 
-        if account.address not in self.accounts:
+        if account.address not in self.accounts:  # type: ignore
             raise AccountsError(f"Account '{account.address}' not known.")
 
         self.__delitem__(account.address)
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -47,9 +47,8 @@ class AddressBase(AbstractBase):
             bool: comparison result
         """
 
-        return self.conversion_manager.convert(
-            self, AddressType
-        ) == self.conversion_manager.convert(other, AddressType)
+        convert = self.conversion_manager.convert
+        return convert(self, AddressType) == convert(other, AddressType)
 
     def __dir__(self) -> List[str]:
         """
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -4,7 +4,7 @@ from ape.types import AddressType
 from ape.utils import BaseInterface, abstractmethod
 
 
-class AddressBase(BaseInterface):
+class BaseAddress(BaseInterface):
     """
     A base address API class. All account-types subclass this type.
     """
@@ -18,7 +18,7 @@ class AddressBase(BaseInterface):
 
     def __eq__(self, other: object) -> bool:
         """
-        Compares :class:`~ape.api.AddressBase` / ``str`` objects by converting to ``AddressType``.
+        Compares :class:`~ape.api.BaseAddress` / ``str`` objects by converting to ``AddressType``.
 
         Returns:
             bool: comparison result
@@ -98,7 +98,7 @@ class AddressBase(BaseInterface):
         return len(self.code) > 0
 
 
-class Address(AddressBase):
+class Address(BaseAddress):
     """
     A generic blockchain address.
 
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -41,7 +41,7 @@ class AddressBase(AbstractBase):
 
     def __eq__(self, other: object) -> bool:
         """
-        Compares :class:`~ape.api.AddressBase`/``str`` objects by converting to ``AddressType``.
+        Compares :class:`~ape.api.AddressBase` / ``str`` objects by converting to ``AddressType``.
 
         Returns:
             bool: comparison result
@@ -57,7 +57,6 @@ class AddressBase(AbstractBase):
         Returns:
             List[str]: Method names that IPython uses for tab completion.
         """
-
         return [
             "address",
             "balance",
@@ -69,7 +68,6 @@ class AddressBase(AbstractBase):
         ]
 
     def __repr__(self) -> str:
-
         return f"<{self.__class__.__name__} {self.address}>"
 
     def __str__(self) -> str:
@@ -79,7 +77,6 @@ class AddressBase(AbstractBase):
         Returns:
             str: The stringified address.
         """
-
         return self.address
 
     @property
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -24,10 +24,9 @@ class AddressBase(AbstractBase):
         Returns:
             :class:`~ape.api.providers.ProviderAPI`
         """
-
         if self.network_manager.active_provider is None:
             raise AddressError(
-                f"Incorrectly implemented provider API for class {type(self).__name__}"
+                f"Incorrectly implemented provider API for class '{type(self).__name__}'."
             )
 
         return self.network_manager.active_provider
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -1,7 +1,5 @@
 from typing import List
 
-from ape.api.providers import ProviderAPI
-from ape.exceptions import AddressError
 from ape.types import AddressType
 from ape.utils import BaseInterface, abstractmethod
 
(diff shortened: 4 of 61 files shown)


evanblank3:
Still not ready for the review, but putting this up to start work on merging the changes for the few left over API refactors. @fubuloubu

src/ape_compile/_cli.py
```diff
@@ -35,27 +35,34 @@ def cli(cli_ctx, file_paths, use_cache, display_size):
     a project is loaded. You do not have to manually trigger a recompile.
     """
 
-    if not file_paths and cli_ctx.project.sources_missing:
-        contracts_dir = cli_ctx.project.config.contracts_folder
+    if not file_paths and cli_ctx.project_manager.sources_missing:
+        contracts_dir = cli_ctx.project_manager.config_manager.contracts_folder
```
unparalleled-js, line 39:
```suggestion
        contracts_dir = cli_ctx.config_manager.contracts_folder
```

src/ape/cli/options.py
```diff
@@ -27,24 +27,6 @@ def __init__(self):
         self.logger = logger
         self._project = None
```
unparalleled-js, line 28:
```suggestion
```

No longer needed

src/ape/cli/options.py
```diff
@@ -27,24 +27,6 @@ def __init__(self):
         self.logger = logger
```
unparalleled-js, line 27: Inside this `init`, we should add `self.config_manager.load()`

Basically, I have add this line somewhere in order to get `ape-solidity` compile to work...
It is a matter of figuring out the best place.

The path of execution is `ape_compile._cli.py` -> `project-manager.load_contracts()`...
My thinking is that any CLI command that is getting the cli_ctx object should expect the configs to already be loaded, so doing it here makes sense... 

Thoughts?

src/ape/utils.py
```diff
@@ -41,6 +40,18 @@
...
     from singledispatchmethod import singledispatchmethod  # type: ignore
 
+if TYPE_CHECKING:
+    from ape.managers.accounts import AccountManager
+    from ape.managers.chain import ChainManager
+    from ape.managers.compilers import CompilerManager
+    from ape.managers.config import ConfigManager
+    from ape.managers.converters import ConversionManager
+    from ape.managers.networks import NetworkManager
+    from ape.managers.project import ProjectManager, _DependencyManager
```
fubuloubu, line 50: I think `_DependencyManager` should have the underscore removed since we are using it all over like this

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
     return [path]
 
 
-class AbstractDataClassMeta(DataClassMeta, ABCMeta):
+class injected_before_use(property):
+    """
+    Injected properties are injected class variables that must be set before use
+    NOTE: do not appear in a Pydantic model's set of properties
+    """
+
```
fubuloubu, line 572: Explore (in a new PR) refactoring this class to accept a `type_str` argument, and automatically doing the `cast` by some sort of integration with mypy (maybe using generics?)

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    :class:`~ape.managers.compilers.CompilerManager` for more information.
+    Call method :meth:`~ape.managers.project.ProjectManager.load_contracts` in this class
+    to more easily compile sources.
     """
 
+    config_manager: ClassVar["ConfigManager"] = cast("ConfigManager", injected_before_use())
+    """
+    A reference to :class:`~ape.managers.config.ConfigManager`, which
+    manages project and plugin configurations.
+    """
```
fubuloubu, line 596: I think all of the docstrings should just be in `ape/__init__.py`, but not sure how the docs render for this

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
-    def __get__(self, *args):
-        raise ValueError("Value not set. Please inject this property before calling.")
+    plugin_manager: ClassVar["PluginManager"] = cast("PluginManager", injected_before_use())
+
+    project_manager: ClassVar["ProjectManager"] = cast("ProjectManager", injected_before_use())
+
+    query_manager: ClassVar["QueryManager"] = cast("QueryManager", injected_before_use())
+
+
+class AbstractBase(ManagerAccessBase, ABC):
```
fubuloubu, line 624: This is sort've why I used `API` instead of `Abstract` or `Base` everywhere

`BaseAPI` would look nicer here

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    """
+    Abstract class that has manager access.
+    """
+
+
+class AbstractBaseModel(AbstractBase, BaseModel):
```
fubuloubu, line 630: And then this becomes `BaseAPIModel`

src/ape_compile/_cli.py
fubuloubu, line 39: `cli_ctx` should have `config_manager` in it (can it inherit `ManagerAccessBase`?)

src/ape_ethereum/ecosystem.py
```diff
@@ -108,25 +82,19 @@ class StaticFeeTransaction(BaseTransaction):
     Transactions that are pre-EIP-1559 and use the ``gasPrice`` field.
     """
 
-    gas_price: int = None  # type: ignore
-    type: TransactionType = TransactionType.STATIC
+    gas_price: Optional[int] = Field(None, alias="gasPrice")
+    max_priority_fee: Optional[int] = Field(None, exclude=True)
+    type: TransactionType = Field(TransactionType.STATIC, exclude=True)
```
fubuloubu, line 87: Maybe we can use the `Literal` Pydantic trick here for indexing purposes

src/ape_ethereum/ecosystem.py
```diff
@@ -108,25 +82,19 @@ class StaticFeeTransaction(BaseTransaction):
     Transactions that are pre-EIP-1559 and use the ``gasPrice`` field.
     """
 
-    gas_price: int = None  # type: ignore
-    type: TransactionType = TransactionType.STATIC
+    gas_price: Optional[int] = Field(None, alias="gasPrice")
+    max_priority_fee: Optional[int] = Field(None, exclude=True)
```
fubuloubu, line 86: Looking at this, I realize that `TransactionAPI` might be over-specified

src/ape_ethereum/ecosystem.py
```diff
@@ -309,33 +294,46 @@ def create_transaction(self, **kwargs) -> TransactionAPI:
         Returns:
             :class:`~ape.api.providers.TransactionAPI`
         """
+
+        transaction_types = {
+            TransactionType.STATIC: StaticFeeTransaction,
+            TransactionType.DYNAMIC: DynamicFeeTransaction,
+        }
```
fubuloubu, line 301: We can probably use `.type` member more smartly here... :thinking:

fubuloubu (commented):
A couple of enhancements and/or fixes, but overall.... looking really good!!!!!

src/ape/cli/options.py
```diff
@@ -73,7 +55,9 @@ def verbosity_option(cli_logger):
 
     def decorator(f):
         def _set_level(ctx, param, value):
+
```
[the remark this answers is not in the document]
fubuloubu: Seriously, probably 500 or so + lines in this PR are from this config in your editor!

unparalleled-js:
Good news! Using the `ape-test` provider, I can run tests and scripts.
Bad news! I am running issues with the `geth` provider.

```
    return account.deploy(project.Fund, *args, **kwargs)
  File "/Users/jules/PycharmProjects/ape/src/ape/api/accounts.py", line 196, in deploy
    receipt = self.call(txn)
  File "/Users/jules/PycharmProjects/ape/src/ape/api/accounts.py", line 144, in call
    return self.provider.send_transaction(txn)
  File "/Users/jules/PycharmProjects/ape/src/ape_geth/providers.py", line 245, in send_transaction
    raise _get_vm_error(err) from err
ape.exceptions.VirtualMachineError: 1 validation error for Receipt
receiver
  none is not an allowed value (type=type_error.none.not_allowed)
```

I can help look more into this tomorrow.
I also have a PR of small changes against your branch I had to do to get the `ape-solidity` plugin working

fubuloubu:
> Good news! Using the `ape-test` provider, I can run tests and scripts. Bad news! I am running issues with the `geth` provider.

Good news / Great news!

src/ape_ethereum/ecosystem.py
```diff
@@ -170,17 +134,66 @@ def raise_for_status(self):
...
         if status:
             if isinstance(status, str) and status.isnumeric():
                 status = int(status)
 
             status = TransactionStatusEnum(status)
 
         txn_hash = data["hash"].hex() if isinstance(data["hash"], HexBytes) else data["hash"]
-        return cls(  # type: ignore
+
+        return Receipt(  # type: ignore
```
unparalleled-js, line 196: To get this to work with `ape-geth`, I had to change line 208 to `receiver=data["to"] or ""`

What was happening was that the data was `{"receiver": None}` for geth and  `{"receiver": ""}` for eth-tester. The empty str passes the Pydantic validation and the `None` does not.

src/ape/cli/options.py
NotPeopling2day: I agree, I added it and if we need to move it elsewhere that's cool too.

src/ape/api/providers.py
```diff
@@ -95,30 +76,33 @@ def total_transfer_value(self) -> int:
...
 
-    def as_dict(self) -> dict:
+    def as_dict(self, exclude: Optional[Set[str]] = None) -> dict:
         """
         Create a ``dict`` representation of the transaction.
 
         Returns:
             dict
         """
-        return as_dict(self)
```
unparalleled-js, line 113: Here is where we should also remove `from`.
`from` seems to cause issues down the line, I have to remove it here to get anything to work.

```python
        data = self.dict(by_alias=True, exclude_none=True, exclude=exclude)

        # Don't publish sender
        if "from" in data:
            del data["from"]

        return data

```

src/ape/__init__.py
```diff
@@ -3,96 +3,51 @@
...
-    request_header={
-        "User-Agent": USER_AGENT,
-    },
-    # What we are considering to be the starting project directory
-    project_folder=_Path.cwd(),
-)
-"""The active configs for the current project. See :class:`ape.managers.config.ConfigManager`."""
+config = _ManagerAccessBase.config_manager
+"""The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
+# NOTE: config is an injected property.
```
unparalleled-js, line 20: This comment is not really a comment, it is embedded into the doc str
Screen Shot 2022-02-25 at 09 43 49
n

Is this intentional? I don't think "injected property" means a whole lot to the user viewing this particular doc...

src/ape/_cli.py
```diff
@@ -20,14 +20,15 @@
 
 
 def display_config(ctx, param, value):
+
```
unparalleled-js, line 23: This is really nitty but I don't like the newline separating the function signature and the first line of the function

src/ape/api/accounts.py
```diff
@@ -118,6 +119,9 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
         # else: Assume user specified the correct amount or txn will fail and waste gas
 
         if send_everything:
+            if txn.max_fee is None:
+                raise TransactionError(message="Max fee must not be null.")
```
unparalleled-js, line 123:
```suggestion
                raise TransactionError(message="Max fee must not be None.")
```

None is more pythonic?

src/ape/api/config.py
```diff
@@ -13,82 +12,29 @@ class ConfigEnum(str, Enum):
...
-            return getattr(self, attrname)
+        if not hasattr(self, attr_name):
+            raise AttributeError(f"{self.__class__.__name__} has no attr '{attr_name}'")
 
-        return default_value
+        return getattr(self, attr_name)
 
 
-class ConfigDict(ConfigItem):
+class UnprocessedConfig(PluginConfig):
```
unparalleled-js, line 35: I find the name of the class confusing... `Unprocessed` makes it sound like the it is completely ignored?

unparalleled-js: Maybe one of these:

* `FreeformConfig`
* `GenericConfig`
* `AnyConfig`
* `UnspecializedConfig`

src/ape/api/config.py
```diff
@@ -13,82 +12,29 @@ class ConfigEnum(str, Enum):
...
     """
-    A config class that is generic and key-value based.
+    The default class used when no specialized class is used
```
unparalleled-js, line 37:
```suggestion
    The default class used when no specialized class is used.
```

src/ape/api/networks.py
```diff
@@ -34,40 +32,58 @@ class EcosystemAPI:
...
-    """The receipt class for this ecosystem."""
+    @abstractmethod
+    def serialize_transaction(self, transaction: "TransactionAPI") -> bytes:
+        """
+        Serialize a transaction to bytes.
 
-    block_class: Type["BlockAPI"]
-    """The block class for this ecosystem."""
+        Args:
+            transaction (:class:`~ape.api.TransactionAPI`): The transaction to encode.
```
unparalleled-js, line 49: You have to link the full path unfortunately

```suggestion
            transaction (:class:`~ape.api.providers.TransactionAPI`): The transaction to encode.
```

src/ape/api/networks.py
```diff
@@ -283,10 +294,11 @@ class ProviderContextManager:
 
     # NOTE: Class variable, so it will manage stack across instances of this object
     _connected_providers: List["ProviderAPI"] = []
+    network_manager: "NetworkManager"
 
-    def __init__(self, network_manager: "NetworkManager", provider: "ProviderAPI"):
-        self.network_manager = network_manager
+    def __init__(self, provider: "ProviderAPI", network_manager: "NetworkManager"):
         self.provider = provider
```
unparalleled-js, line 300: Do we still the to pass in the `Provider`? The `Provider` should already be accessible from the `NetworkManager` (such as via `active_provider`)

Maybe the flow here needs adjustments?

src/ape/api/networks.py
```diff
@@ -326,22 +338,16 @@ class NetworkAPI:
...
-
     data_folder: Path  # For caching any data that might need caching
     """The path to the ``.ape`` directory."""
 
-    request_header: str
+    request_header: dict
     """A shareable network HTTP header."""
 
-    _default_provider: str = ""
+    _default_provider: Optional[str] = ""
```
unparalleled-js, line 347: Is this supposed to be `Optional`? I would think it would default to `None` rather than the `""` if that is the case.

src/ape/api/providers.py
unparalleled-js: I am not 100% sure on this but somewhere, `from` needs to either be removed or set to its proper value, specifically for calls (not really transactions, although needs more testing)

src/ape/api/providers.py
```diff
@@ -308,100 +294,64 @@ def await_confirmations(self) -> "ReceiptAPI":
...
-    parent_hash: HexBytes
+    parent_hash: Any
     size: int
     timestamp: int
 
-    @classmethod
-    @abstractmethod
-    def decode(cls, data: Dict) -> "BlockAPI":
+    @validator("hash", "parent_hash", pre=True)
+    def validate_hexbytes(cls, value):
```
unparalleled-js, line 332: Either one of two things needs to happen here:

1. Rename `cls` to `self`, as this is not a classmethod (from what I can tell)
2. Put back the `@classmethod` decorator

src/ape/api/providers.py
```diff
@@ -308,100 +294,64 @@ def await_confirmations(self) -> "ReceiptAPI":
...
         """
-        Decode data to a :class:`~ape.api.BlockAPI`.
-
-        Args:
-            data (dict): A dictionary of data to decode.
-
-        Returns:
-            :class:`~ape.api.BlockAPI`
+        Custom validation method for HexBytes because pydantic treats these
```
unparalleled-js, line 334: Do we want doc strs for the validation methods or do just want comments?
Doc strs here will cause the user the maybe think this a method they can call.

I think we should exclude all validation methods from the documentation. You can do this by appending the method name to the `:exclude-members: __repr__, __weakref__, __metaclass__` list in the doc files, like `docs/methoddocs/api.md`

src/ape/api/providers.py
```diff
@@ -631,12 +583,9 @@ def mine(self, num_blocks: int = 1):
...
         """
 
-    def __repr__(self):
-        return f"<{self.__class__.__name__} {self.name}>"
-
     def _try_track_receipt(self, receipt: ReceiptAPI):
-        if self._chain:
-            self._chain.account_history.append(receipt)
+        if self.chain_manager:
+            self.chain_manager.account_history.append(receipt)
```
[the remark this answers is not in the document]
unparalleled-js: Yes, we have a custom append method that appends the receipt to an internal dict that is accessible by the sender... I don't think this is a bug.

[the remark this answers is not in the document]
unparalleled-js: https://github.com/ApeWorX/ape/blob/main/src/ape/managers/chain.py#L315-L336

src/ape/api/query.py
```diff
@@ -102,7 +99,7 @@ def estimate_query(self, query: QueryType) -> Optional[int]:
         query engine is not available for use or is unable to complete the query.
 
         Args:
-            query (``QueryType``): Query to estimate execution time for.
+            query (``QueryType``): Query for which execution time will be estimated.
```
unparalleled-js, line 102: `will be estimated` is passive-language. Consider:

```suggestion
            query (``QueryType``): Query to estimate.
```

src/ape/cli/choices.py
```diff
@@ -29,6 +29,7 @@ class Alias(click.Choice):
 
     # noinspection PyMissingConstructor
```
unparalleled-js, line 30:
```suggestion
```

We can delete the Pycharm directives for now, usually `type: ignore` also appeases pycharm and we can put those back as needed

src/ape/cli/options.py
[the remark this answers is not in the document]
unparalleled-js: Yes the spaces are a bit much! The diff is already huge.

src/ape/contracts/base.py
```diff
@@ -102,96 +128,125 @@ def __call__(self, *args, **kwargs) -> Any:
...
     def __repr__(self) -> str:
         abis = sorted(self.abis, key=lambda abi: len(abi.inputs or []))  # type: ignore
         return abis[-1].signature
 
+    @property
+    def provider(self) -> ProviderAPI:
+
+        if self.network_manager.active_provider is None:
+            raise AddressError(
+                f"Incorrectly implemented provider API for class {type(self).__name__}"
```
unparalleled-js, line 150:
```suggestion
                f"Incorrectly implemented provider API for class '{type(self).__name__}'."
```

src/ape/contracts/base.py
```diff
@@ -102,96 +128,125 @@ def __call__(self, *args, **kwargs) -> Any:
...
         return self.provider.network.ecosystem.encode_transaction(
             self.address, self.abi, *args, **kwargs
         )
 
+    @property
+    def provider(self) -> ProviderAPI:
+
+        if self.network_manager.active_provider is None:
+            raise AddressError(
+                f"Incorrectly implemented provider API for class {type(self).__name__}"
```
unparalleled-js, line 223: This error logic is used all over, we should probably abstract this out

src/ape/contracts/base.py
```diff
@@ -284,66 +428,38 @@ def __getattr__(self, attr_name: str) -> Any:
...
-        if num_matching_conditions == 0:
+        if attr_name not in set((*self._view_methods_, *self._mutable_methods_, *self._events_)):
             # Didn't find anything that matches
             # NOTE: `__getattr__` *must* raise `AttributeError`
             name = self._contract_type.name or self.__class__.__name__
             raise AttributeError(f"'{name}' has no attribute '{attr_name}'.")
 
-        elif num_matching_conditions > 1:
+        elif (
+            int(attr_name in self._view_methods_)
```
unparalleled-js, line 441: It'd be more readable to use `any()` here.. something like:

```python
elif any((attr_name in self._view_methods_, attr_name in self._mutable_methods_, attr_name in self._events_))
```

src/ape/logging.py
```diff
@@ -22,8 +22,11 @@ class LogLevel(Enum):
 
 
 def success(self, message, *args, **kws):
-    """This method gets injecting into python's `logging` module
-    to handle logging at this level."""
+    """
+    This method gets injecting into python's `logging` module
```
unparalleled-js, line 26:
```suggestion
    This method gets injected into python's `logging` module
```

src/ape/managers/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+# Wiring together the application
+_data_folder = _Path.home().joinpath(".ape")
+_project_folder = _Path.cwd()
+
+ManagerAccessBase.plugin_manager = PluginManager()
+
+ManagerAccessBase.dependency_manager = _DependencyManager(data_folder=_data_folder)
+
+ManagerAccessBase.config_manager = ConfigManager(
```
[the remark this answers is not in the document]
unparalleled-js: `config_manager` does use `dependency_manager`

src/ape/managers/base.py
```diff
@@ -0,0 +1,7 @@
+from ape.utils import ManagerAccessBase
+
+
+class ManagerBase(ManagerAccessBase):
+    """
+    Manager base that allows us to add other IPython integration features
```
unparalleled-js, line 6:
```suggestion
    Manager base that allows us to add other IPython integration features.
```

src/ape/managers/compilers.py
```diff
@@ -51,15 +47,15 @@ def registered_compilers(self) -> Dict[str, CompilerAPI]:
...
 
-        cache_key = self.config.PROJECT_FOLDER
+        cache_key = self.config_manager.PROJECT_FOLDER
         if cache_key in self._registered_compilers_cache:
             return self._registered_compilers_cache[cache_key]
 
         registered_compilers = {}
 
         for plugin_name, (extensions, compiler_class) in self.plugin_manager.register_compiler:
-            config = self.config.get_config(plugin_name)
```
unparalleled-js, line 61: Caveat here is that now we have to make sure to load the configs from the compiler plugins. I have done this in the `ape-solidity` and it works fine I think! But I just want to make sure that we know this now and that this is really what we want.

src/ape/managers/project.py
```diff
@@ -125,14 +134,17 @@ def file_needs_compiling(source: Path) -> bool:
         needs_compiling = list(filter(file_needs_compiling, sources))
 
         # Set the context in case compiling a dependency (or anything outside the root project).
-        with self.compilers.config.using_project(self.path, contracts_folder=self.contracts_folder):
-            self.compilers.config.PROJECT_FOLDER = self.path
-            self.compilers.config.contracts_folder = self.contracts_folder
-            compiled_contract_types = self.compilers.compile(needs_compiling)
+        with self.compiler_manager.config_manager.using_project(
```
unparalleled-js, line 137:
```suggestion
        with self.config_manager.using_project(
```

src/ape/managers/project.py
```diff
@@ -125,14 +134,17 @@ def file_needs_compiling(source: Path) -> bool:
...
+            self.path, contracts_folder=self.contracts_folder
+        ):
+            self.compiler_manager.config_manager.PROJECT_FOLDER = self.path
```
unparalleled-js, line 140:
```suggestion
```

This line and the one below happen already in the `using_project()` method, so I am not sure they are still needed here

src/ape/managers/project.py
```diff
@@ -680,11 +689,11 @@ def run_script(self, name: str, interactive: bool = False):
             return console()
 
     def _load_dependencies(self) -> Dict[str, PackageManifest]:
-        return {d.name: d.extract_manifest() for d in self.config.dependencies}
+        return {d.name: d.extract_manifest() for d in self.config_manager.dependencies}
 
     # @property
```
unparalleled-js, line 694: ugh commented out code? Can we not?

unparalleled-js: Commented out code is the antithesis of clean code.
It sits here rots, no one knows why it is here, everyone is afraid to delete it, it looks bad, it is hard to maintain...

I would strongly prefer to keep "ideas" out of the production code base, especially as we near  v1.  Keep "ideas" in GH issues

src/ape/plugins/__init__.py
```diff
@@ -134,6 +138,10 @@ def __repr__(self):
         return f"<{self.__class__.__name__}>"
 
     def __getattr__(self, attr_name: str) -> Iterator[Tuple[str, Tuple]]:
+
+        if attr_name == "__fields_set__":
```
unparalleled-js, line 142: I am confused as why this is needed because `PluginManager` is not a Pydantic model

src/ape/plugins/__init__.py
```diff
@@ -145,26 +153,33 @@ def get_plugin_name_and_hookfn(h):
...
             # NOTE: Some plugins return a tuple and some return iterators
-            if not isinstance(results, tuple) and hasattr(results, "__iter__"):
+            if not isinstance(results, Generator):
+                validated_plugin = self._validate_plugin(plugin_name, results)
+
+                if validated_plugin:
+                    yield validated_plugin
+
+            else:
                 # Only if it's an iterator, provider results as a series
```
unparalleled-js, line 165: This comment is wrong?
Because the `if` statement above handles generators, which are iterators.

Should this be:

```suggestion
                # Only if it's an tuple, provider results as a series
```

IDK, I am a little confused here...

unparalleled-js: Are we still handling tuples correctly??

src/ape/api/providers.py
```diff
@@ -95,30 +76,33 @@ def total_transfer_value(self) -> int:
...
 
     @abstractmethod
-    def encode(self) -> bytes:
+    def serialize_transaction(self) -> bytes:
         """
         Take this object and produce a hash to sign to submit a transaction
         """
 
-    def as_dict(self) -> dict:
+    def as_dict(self, exclude: Optional[Set[str]] = None) -> dict:
```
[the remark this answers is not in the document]
NotPeopling2day: Perfect, this is what I was looking for last night! I'll refactor it to match, thanks.

src/ape/api/address.py
```diff
@@ -55,14 +41,13 @@ def address(self) -> AddressType:
 
     def __eq__(self, other: object) -> bool:
         """
-        Compares :class:`~ape.api.AddressAPI`/``str`` objects by converting to ``AddressType``.
+        Compares :class:`~ape.api.AddressBase`/``str`` objects by converting to ``AddressType``.
```
unparalleled-js, line 44:
```suggestion
        Compares :class:`~ape.api.AddressBase` / ``str`` objects by converting to ``AddressType``.
```

src/ape/api/networks.py
```diff
@@ -34,40 +32,58 @@ class EcosystemAPI:
...
 
-    _default_network: str = LOCAL_NETWORK_NAME
+        Returns:
+            bytes
+        """
+
+    @abstractmethod
+    def decode_receipt(self, data: dict) -> "ReceiptAPI":
+        """
+        Convert data to :class:`~ape.api.ReceiptAPI`.
```
unparalleled-js, line 58:
```suggestion
        Convert data to :class:`~ape.api.providers.ReceiptAPI`.
```

src/ape/api/networks.py
```diff
@@ -34,40 +32,58 @@ class EcosystemAPI:
...
+
+        Args:
+            data (dict): A dictionary of Receipt properties.
+
+        Returns:
+            :class:`~ape.api.ReceiptAPI`
```
unparalleled-js, line 64:
```suggestion
            :class:`~ape.api.providers.ReceiptAPI`
```

src/ape/api/networks.py
```diff
@@ -34,40 +32,58 @@ class EcosystemAPI:
...
+        """
+
+    @abstractmethod
+    def decode_block(self, data: dict) -> "BlockAPI":
+        """
+        Decode data to a :class:`~ape.api.BlockAPI`.
```
unparalleled-js, line 70:
```suggestion
        Decode data to a :class:`~ape.api.providers.BlockAPI`.
```

src/ape/api/networks.py
```diff
@@ -34,40 +32,58 @@ class EcosystemAPI:
...
+
+        Args:
+            data (dict): A dictionary of data to decode.
+
+        Returns:
+            :class:`~ape.api.BlockAPI`
```
unparalleled-js, line 76:
```suggestion
            :class:`~ape.api.providers.BlockAPI`
```

src/ape/plugins/__init__.py
```diff
@@ -145,26 +153,33 @@ def get_plugin_name_and_hookfn(h):
             return h.plugin_name, getattr(h.plugin, attr_name)()
 
         for plugin_name, results in map(get_plugin_name_and_hookfn, hookimpls):
+
             # NOTE: Some plugins return a tuple and some return iterators
-            if not isinstance(results, tuple) and hasattr(results, "__iter__"):
+            if not isinstance(results, Generator):
```
unparalleled-js, line 158: Other comments hinted at this - but wondering if we are not handling tuples correctly in here

src/ape/plugins/config.py
```diff
@@ -14,12 +14,12 @@ class Config(PluginType):
...
     @hookspec
-    def config_class(self) -> Type[ConfigItem]:
+    def config_class(self) -> Type[PluginConfig]:
         """
-        A hook that returns a :class:`~ape.api.config.ConfigItem` parser class that can be
+        A hook that returns a :class:`~ape.api.config.PluginConfig` parser class that can be
         used to deconstruct the user config options for this plugins.
 
-        NOTE: If none are specified, all injected :class:`ape.api.config.ConfigItem`'s are empty.
+        NOTE: If none are specified, all injected :class:`ape.api.config.PluginConfig`'s are empty.
```
unparalleled-js, line 22:
```suggestion
        **NOTE**: If none are specified, all injected :class:`ape.api.config.PluginConfig`'s are empty.
```

src/ape/pytest/runners.py
```diff
@@ -1,101 +1,35 @@
...
+        self.pytest_config = pytest_config
         self._warned_for_missing_features = False
         ape.reverts = RevertsContextManager  # type: ignore
 
     @property
     def _network_choice(self) -> str:
-        # The option the user providers via --network (or the default).
-        return self.config.getoption("network")
-
-    def pytest_exception_interact(self, report, call):
```
unparalleled-js, line 37: Looks like we accidentally deleted all of the console related stuff?

- [ ] Make sure `-I` still works for `ape test`

src/ape/__init__.py
NotPeopling2day: I'll remove that, not as relevant now that we've changed some stuff around.

src/ape/types/signatures.py
```diff
@@ -1,11 +1,11 @@
 from typing import Iterator, Union
 
-from dataclassy import dataclass
 from eth_account.messages import SignableMessage
 from eth_utils import to_bytes
+from pydantic.dataclasses import dataclass
 
 
-@dataclass(frozen=True, slots=True, kwargs=True)
```
unparalleled-js, line 8: I am curious about the original intention here for the `kwargs` and `slots` specs.
From my understand, all Pydantic models have `__slots__` so we are good to remove that (However there may be a bug when using Cython?: https://github.com/samuelcolvin/pydantic/issues/3791)

But, `kwargs`, why were using that to begin with? @fubuloubu ?

src/ape/utils.py
```diff
@@ -349,6 +382,7 @@ def available_plugins(self) -> Set[str]:
         Returns:
             Set[str]: The plugin names as 'ape_plugin_name' (module-like).
```
unparalleled-js, line 383:
```suggestion
            Set[str]: The plugin names as ``'ape_plugin_name'`` (module-like).
```

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
         return list(path.rglob("*.*"))
 
     return [path]
 
 
-class AbstractDataClassMeta(DataClassMeta, ABCMeta):
+class injected_before_use(property):
+    """
+    Injected properties are injected class variables that must be set before use
+    NOTE: do not appear in a Pydantic model's set of properties
```
unparalleled-js, line 570:
```suggestion
    **NOTE**: Injected properties do not appear in a Pydantic model's set of properties.
```

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
     if path.is_dir():
         return list(path.rglob("*.*"))
 
     return [path]
 
 
-class AbstractDataClassMeta(DataClassMeta, ABCMeta):
+class injected_before_use(property):
+    """
+    Injected properties are injected class variables that must be set before use
```
unparalleled-js, line 569:
```suggestion
    Injected properties are injected class variables that must be set before use.
```

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    NOTE: do not appear in a Pydantic model's set of properties
+    """
+
+    def __get__(self, *args):
+        raise ValueError("Value not set. Please inject this property before calling.")
+
+
+class ManagerAccessBase:
```
unparalleled-js, line 577: Missing docs for this ?

Is this public? Do we expected users to imports / use this, perhaps when making plugins?

Otherwise, we should consider renaming it to `_ManagerAccessBase` and intentionally hiding it from the docs...

src/ape/utils.py
unparalleled-js: I am trying to figure out if this is public or not.

The `ManagerAccessBase` is currently not showing up in the docs for some reason.

If this _is_ public, I'd be fine with avoiding repeating all the doc strs here that are the same from `__init__` and instead just say at the class level docs that this class has all the same singletons from the root ape namespace and maybe link to those docs.

src/ape/utils.py
unparalleled-js: My question for us is this:  is "API" a marketing term for our abstract plugin classes?

We should avoid terms that are likely to change in the future. "Astract" is general enough that I don't see it ever changing. However, if we are confident in the same regarding "API", then I like that too.

My only point is to avoid the term if and only if it is ape-specific and not a general term for the idea. I know for sure "abstract" is general enough, but "API" can often refer to the actual implementation of an interface which may confuse users who are not familiar with ape-isms.

unparalleled-js: My question for us is this:  is "API" a marketing term for our abstract plugin classes?

We should avoid terms that are likely to change in the future. "Astract" is general enough that I don't see it ever changing. However, if we are confident in the same regarding "API", then I like that too.

My only point is to avoid the term if and only if it is ape-specific and not a general term for the idea. I know for sure "abstract" is general enough, but "API" can often refer to the actual implementation of an interface which may confuse users who are not familiar with ape-isms.

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+
+class AbstractBase(ManagerAccessBase, ABC):
+    """
+    Abstract class that has manager access.
+    """
+
+
+class AbstractBaseModel(AbstractBase, BaseModel):
+    """
+    Abstract class with manager access on a pydantic base model
```
unparalleled-js, line 632:
```suggestion
    An abstract base-class with manager access on a pydantic base model.
```

src/ape/__init__.py
```diff
@@ -4,95 +4,49 @@
...
-    # NOTE: For all HTTP requests we make
-    request_header={
-        "User-Agent": USER_AGENT,
-    },
-    # What we are considering to be the starting project directory
-    project_folder=_Path.cwd(),
-)
-"""The active configs for the current project. See :class:`ape.managers.config.ConfigManager`."""
+config = _ManagerAccessBase.config_manager
+"""The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
```
unparalleled-js, line 19:
```suggestion
"""
The active configs for the current project. See :class:`ape.managers.config.ConfigManager`.
```

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    """
+
+    class Config:
+        """
+        NOTE: Due to https://github.com/samuelcolvin/pydantic/issues/1241 we have
```
unparalleled-js, line 637: This `NOTE` seems out of place? It is referring some cached property yet is a doc str on the entire class?

unparalleled-js: Class doc strs should only be telling the user how to use the class and non-public classes should not have them.

Is this a public class?

src/ape/api/config.py
NotPeopling2day: I pulled this from @fubuloubu's original refactor. I'm leaning toward `GenericConfig` based on the doc string.

src/ape/utils.py
```diff
@@ -513,45 +557,106 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+            TypeError: cannot pickle '_thread.RLock' object
+        """
+
+        keep_untouched = (cached_property,)
+        arbitrary_types_allowed = True
+        underscore_attrs_are_private = True
+
+    def __dir__(self) -> List[str]:
+        """
+        NOTE: Integrates with IPython tab-completion
```
unparalleled-js, line 649:
```suggestion
        Should options in IPython tab-completion.
```

src/ape_ethereum/ecosystem.py
```diff
@@ -108,25 +82,19 @@ class StaticFeeTransaction(BaseTransaction):
...
-        raise NotImplementedError("Max fee is not settable for static-fee transactions.")
-
-    def as_dict(self):
-        data = super().as_dict()
-        if "gas_price" in data:
-            data["gasPrice"] = data.pop("gas_price")
-
-        data.pop("type")
+    def __init__(self, **data) -> None:
+        """
```
unparalleled-js, line 90: doc strs in `__init__` compete the class-level doc strs.
You probably just want this as a normal comment.

src/ape_plugins/_cli.py
```diff
@@ -42,17 +47,19 @@ def plugins_argument():
...
         elif len(value) == 1 and Path(value[0]).resolve().exists():
             # User passed in a path to a config file.
             config_path = Path(value[0]).expanduser().resolve()
+
             if config_path.name != CONFIG_FILE_NAME:
                 config_path = config_path / CONFIG_FILE_NAME
 
             config = load_config(config_path)
-            plugins = config.get("plugins") or []
+            plugins = config.get("plugins", [])
```
unparalleled-js, line 62: Hey. I would advise not making this change.

There is actually a difference here between what it was and what you are changing it too.

The default argument in `get()` only gets returns *if they key is not present in the dictionary*.

In the case where they key _is_ present in the dictionary, it always uses that, even if that key is `None`.

The way it was before, all falsey values get coerced to the empty list. So even if they key exists but its value is `None`, it would get turned into an empty list.

the change you are making breaks that safety!

src/ape_test/providers.py
```diff
@@ -10,6 +10,15 @@
...
 
 class LocalProvider(TestProviderAPI, Web3Provider):
+
+    _tester: PyEVMBackend
+    _web3: Web3
+
+    def __init__(self, **data) -> None:
+
+        super().__init__(**data)
+        self.__post_init__()
```
unparalleled-js, line 20: probably can just hoist everything in `__post_init__` to here, right?

src/ape/api/providers.py
NotPeopling2day: Fixing this to an internal `NOTE` for now, but it'll change if we change the data type to something pydantic can validate

src/ape/api/providers.py
```diff
@@ -308,100 +294,64 @@ def await_confirmations(self) -> "ReceiptAPI":
...
-
-        Args:
-            data (dict): A dictionary of data to decode.
-
-        Returns:
-            :class:`~ape.api.BlockAPI`
+        Custom validation method for HexBytes because pydantic treats these
+        values as bytes and throws an error
         """
+        assert isinstance(value, HexBytes)
```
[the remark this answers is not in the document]
NotPeopling2day: That's one way pydantic shows how to validate, since it doesn't understand what to do with a HexByte. It tries to treat it like bytes and it raises an error.

https://pydantic-docs.helpmanual.io/usage/validators/

src/ape/utils.py
fubuloubu: How about just `BaseInterface` (`AccountInterface`, `ProviderInterface`, etc.)

src/ape/api/config.py
fubuloubu: I mean it technically is "unprocessed"

But I'm okay with generic

src/ape/types/signatures.py
fubuloubu: I think `kwargs` just allows you to specify names via `v=....`

src/ape_test/providers.py
fubuloubu: Does Pydantic have `__post_init__`?

unparalleled-js: > Does Pydantic have `__post_init__`?

No, I don't believe so

src/ape/api/providers.py
NotPeopling2day: Pydantic does treat this as class level.

https://pydantic-docs.helpmanual.io/usage/validators/

src/ape/api/projects.py
```diff
@@ -180,7 +180,7 @@ def _target_manifest_cache_file(self) -> Path:
             version_id = f"v{version_id}"
 
         name = self.name
-        return self._data_folder / "packages" / name / version_id / f"{name}.json"
```
unparalleled-js, line 195: Can we delete _data_folder from line 161 too?

src/ape/contracts/base.py
```diff
@@ -32,16 +31,11 @@ def __init__(
         self.abi = abi
         self.provider = provider
 
-    @validator("deployment_bytecode")
-    def check_empty_contract(cls, value):
-
+    def __post_init__(cls, value):
```
unparalleled-js, line 34: I thought there was not a `__post_init__` though?

Can we just move this to inside the `__init__()`, (that was my suggestion before)

unparalleled-js: Wait, what is `value`?

I am confused here

unparalleled-js: I think this change may have been an accident?

src/ape/api/providers.py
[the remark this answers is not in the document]
unparalleled-js: Yes, but looks at the warning in that link you posted:
Screen Shot 2022-02-25 at 16 41 50

This is why it is just better to manually raise `AssertionError` when using normal python.

src/ape/managers/project.py
```diff
@@ -718,16 +724,19 @@ def dependency_types(self) -> Dict[str, Type[DependencyAPI]]:
...
 
     def decode_dependency(self, config_dependency_data: Dict) -> DependencyAPI:
+
         for key, dependency_cls in self.dependency_types.items():
+
             if key in config_dependency_data:
                 return dependency_cls(
                     **config_dependency_data,
-                    project_manager=self.project_manager,
                     _data_folder=self.DATA_FOLDER,
```
unparalleled-js, line 737:
```suggestion
```

I think we removed this?

src/ape/__init__.py
```diff
@@ -4,95 +4,49 @@
...
 from .managers.project import ProjectManager as Project
-from .managers.project import _DependencyManager
-from .plugins import PluginManager as _PluginManager
-from .utils import USER_AGENT
 
 # Wiring together the application
-_data_folder = _Path.home().joinpath(".ape")
 
-plugin_manager = _PluginManager()
+plugin_manager = _ManagerAccessMixin.plugin_manager
```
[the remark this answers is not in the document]
unparalleled-js: I don't think so either!

Plugins use `from ape import plugins`, which has everything they need already.
(That is not related to this module)

src/ape/api/providers.py
```diff
@@ -84,25 +84,16 @@ def total_transfer_value(self) -> int:
...
-
     def __repr__(self) -> str:
-        data = self.as_dict()  # NOTE: `as_dict` could be overridden
+        data = self.dict()
         params = ", ".join(f"{k}={v}" for k, v in data.items())
         return f"<{self.__class__.__name__} {params}>"
 
     def __str__(self) -> str:
-        data = self.as_dict()  # NOTE: `as_dict` could be overridden
+        data = self.dict()
```
unparalleled-js, line 96: Maybe we should call this `txn_dict` to not confuse with a Transaction's data property

src/ape/plugins/__init__.py
NotPeopling2day: Left over from looking at the previous pydantic refactor, I'll remove it.

src/ape/plugins/__init__.py
NotPeopling2day: I believe so, I inverted the logic here, moving the Tuple handling up and pushing down the iterator logic.

src/ape_plugins/_cli.py
NotPeopling2day: Thanks for breaking it down for me, that makes sense!

src/ape/managers/accounts.py
```diff
@@ -68,7 +68,7 @@ def aliases(self) -> Iterator[str]:
 
     def __iter__(self) -> Iterator[AccountAPI]:
         for container in self.containers.values():
-            for account in container:
+            for account in container.accounts():
```
unparalleled-js, line 71: Change this to `container.accounts` to get it to work!

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+    def __dir__(self) -> List[str]:
+        """
+        NOTE: Should integrate options in IPython tab-completion.
+        https://ipython.readthedocs.io/en/stable/config/integrating.html
+        """
+        # Filter out protected/private members
+        return [member for member in super().__dir__() if not member.startswith("_")]
+
+    def dict(self, *args, **kwargs) -> dict:
+        # NOTE: We do this to accomodate the aliases needed for EIP-2678 compatibility
```
fubuloubu, line 635: Comment doesn't need to be copied

```suggestion
```

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        if "by_alias" not in kwargs:
+            kwargs["by_alias"] = True
+
+        # EIP-2678: skip empty fields (at least by default)
```
fubuloubu, line 639: Comment doesn't need to be copied

```suggestion
```

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        if "exclude_none" not in kwargs:
+            kwargs["exclude_none"] = True
+
+        return super().dict(*args, **kwargs)
+
+    def json(self, *args, **kwargs) -> str:
+        # NOTE: When serializing to IPFS, the canonical representation must be repeatable
+
+        # EIP-2678: minified representation (at least by default)
```
fubuloubu, line 648: Comment doesn't need to be copied

```suggestion
```

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        if "separators" not in kwargs:
+            kwargs["separators"] = (",", ":")
+
+        # EIP-2678: sort keys (at least by default)
```
fubuloubu, line 652: Comment doesn't need to be copied

```suggestion
```

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        if "sort_keys" not in kwargs:
+            kwargs["sort_keys"] = True
+
+        # NOTE: We do this to accomodate the aliases needed for EIP-2678 compatibility
```
fubuloubu, line 656: Comment doesn't need to be copied

```suggestion
```

src/ape/utils.py
```diff
@@ -513,45 +557,117 @@ def get_all_files_in_directory(path: Path) -> List[Path]:
...
+        if "by_alias" not in kwargs:
+            kwargs["by_alias"] = True
+
+        # EIP-2678: skip empty fields (at least by default)
```
fubuloubu, line 660: Comment doesn't need to be copied

```suggestion
```

src/ape/managers/accounts.py
```diff
@@ -72,6 +66,13 @@ def aliases(self) -> Iterator[str]:
         for container in self.containers.values():
             yield from container.aliases
 
+    @property
+    def accounts(self) -> Iterator[AccountAPI]:
```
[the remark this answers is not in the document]
NotPeopling2day: Reverted this back to using iter for the manager

fubuloubu:
One note (for another PR?): when parsing the config file, it should probably display a warning on validation error instead of error

src/ape/api/networks.py
fubuloubu: This is the code that actually manages `active_provider`

fubuloubu: This is resolved, just leaving open for @unparalleled-js to read

src/ape_ethereum/ecosystem.py
fubuloubu: Add an issue to revisit Ecosystem Api and associated types

src/ape_ethereum/ecosystem.py
fubuloubu: Related to comment to revisiting EcosystemAPI

src/ape_ethereum/ecosystem.py
fubuloubu: Related to comment to revisiting EcosystemAPI

tests/integration/cli/test_accounts.py
```diff
@@ -150,6 +150,7 @@ def test_list(ape_cli, runner, test_keyfile):
     assert ALIAS in result.output
 
 
+@pytest.mark.skip(reason="Changes to underlying structure make mocks incorrect")
```
fubuloubu, line 153: Only skipped test left?

fubuloubu (approved):

src/ape_ethereum/ecosystem.py
NotPeopling2day: https://github.com/ApeWorX/ape/issues/528

src/ape/utils.py
fubuloubu: Note: bikeshed naming in another PR

src/ape/utils.py
fubuloubu: Note: bikeshed naming in another PR

src/ape/utils.py
fubuloubu: Explore how docs are structured in another PR

fubuloubu (dismissed):
Let's Go!!!!!

src/ape/managers/compilers.py
NotPeopling2day: Adding back and leaving a NOTE comment

fubuloubu (approved):

fubuloubu (approved):

src/ape/api/providers.py
NotPeopling2day: Had a discussion around this and since we are no longer wrapping it as custom, I think it's okay to leave it as the pydantic model's dict method.

fubuloubu (dismissed):

unparalleled-js:
We are still seeing an issue with static-fee transactions

unparalleled-js (dismissed):
IT'S DONE

src/ape_ethereum/ecosystem.py
```diff
@@ -200,7 +200,7 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
             contract_address=data.get("contractAddress"),
             sender=data["from"],
             receiver=data["to"] or "",
-            nonce=data.get("nonce"),
+            nonce=data.get("nonce", None) or None,
```
unparalleled-js, line 203: oops forgot about 0!

unparalleled-js (approved):
ALL CAPS APPROVAL

Issue #498, opened 2022-02-17
Ape hijacks pytest, not allowing pytest to be used independently or pytest-xdit to be used


Pet3ris:
### Environment information

Ape version was beta 0.1.3 and this issue surfaced before plug-in installation.

* Python Version: 3.8
* OS: macOS

I'm using poetry with ape as a dev dependency.

### What went wrong?

When I run poetry run pytest ape reports "no provider name test"
After that I switched to running `poetry run ape test`, which worked fine as a replacement.
However, recently started using `pytest-xdist` for test parallelization and when running `poetry run ape test -n auto`, the same error occurs. Had to uninstall ape to be able to test directly with pytest.

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

unparalleled-js:
Most pytest plugins are installed along side pytest. Ape, however, does hijack it. The work around is that you have to disable the ape-test plugin when running pytest in other projects (or use separate environments).

To disable the `ape-test` plugin in `pytest`, include the following command line option:  `-p no:ape_test`.
You can also use this in your `pyproject.toml` file. This is actually how we run the normal `ape` tests: https://github.com/ApeWorX/ape/blob/main/pyproject.toml#L21

- [ ] I think, for now anyway, we should more explicitly document this somewhere


We can also brainstorm alternatives.

fubuloubu:
> We can also brainstorm alternatives.

Wonder if we can dynamically inject the plugin only when `ape test` is called? Although I do like it being an alias for `pytest`

unparalleled-js:
> > We can also brainstorm alternatives.
> 
> Wonder if we can dynamically inject the plugin only when `ape test` is called? Although I do like it being an alias for `pytest`

I also want to look more into why it fails.. Maybe if we can detect that it is not an ApeProject, we can avoid doing any plugin stuff.

fubuloubu:
> > > We can also brainstorm alternatives.
> > 
> > 
> > Wonder if we can dynamically inject the plugin only when `ape test` is called? Although I do like it being an alias for `pytest`
> 
> I also want to look more into why it fails.. Maybe if we can detect that it is not an ApeProject, we can avoid doing any plugin stuff.

Hmm, maybe have `-p no:ape_test` by default w/ pytest registration, and then remove it in `ape_test/_cli.py`

Pet3ris:
Curious - why couldn't ape exist as a library/pytest plug-in as far as testing is concerned?

[...]

PR #510, opened 2022-02-24
fix: register pytest plugin outside of plugin

unparalleled-js:
### What I did

fixes: #498 

### How I did it

Register pytest plugin outside of `ape_test` resolves the race condition

### How to verify it

Now, you can do `pytest` or `ape test`, and it does not matter - both work for an ape project.
Also, you can run `pytest` in a normal project and that seesm to work now w/o having to do `-p no:test`

### Checklist


src/ape/pytest/plugin.py
```diff
@@ -4,8 +4,8 @@
 import pytest
 
 from ape import accounts, chain, networks, project
-from ape_test.fixtures import PytestApeFixtures
-from ape_test.runners import PytestApeRunner
+from ape.pytest.fixtures import PytestApeFixtures
```
unparalleled-js, line 7: I was wondering if we should make this namespace `ape.testing` so we an include other testing related stuff here besides just the pytest plugin stuff? Thoughts?

fubuloubu: Not opposed, but I think pytest specific things here are important for namespacing

I don't see supporting another framework for a while though.

fubuloubu (approved):

unparalleled-js:
I was told this actually _resolves_ the issue in #498 for our user!  🥳

PR #512, opened 2022-02-25
fix: issue where config file would unnecessarily get created

unparalleled-js:
### What I did

Issue where setting default config data in dependencies always caused config files to get created even when it was not necessary.

### Checklist


NotPeopling2day (approved):
:shipit:

Issue #369, opened 2021-12-22
General: User scripts to work w/o `main` or `cli` methods


NotPeopling2day:
### Overview

Update to make user scripts work whether or not they include `main` or `cli` methods in their scripts

### Specification

https://github.com/ApeWorX/ape/blob/29c8840d6dc02d7c37545e29b28269bf744842c9/src/ape_run/_cli.py#L15

fubuloubu:
Check line 46. The script is actually fully running, but it shows an error for not having the right method

PR #522, opened 2022-03-01
refactor: do script loading more dynamically

fubuloubu:
### What I did
Allow scripts that use click to function correctly. Also load in extra locals correctly using interactive. Basically entire scripting engine overhaul.

fixes: #369

### Checklist

### DIFF
--- a/docs/userguides/projects.md
+++ b/docs/userguides/projects.md
@@ -21,9 +21,9 @@ The project manager object is a representation of your current project. Access i
 from ape import project
 ```
 
-Your `project` contains all the "relevant" files, such as source files in the `contracts/` directory. The 
-`contracts/` directory is where compilers look for contracts to compile. File extensions found within the `contracts/` 
-directory determine which compiler plugin `ape` uses. Make sure to install the compiler plugins you need if they are 
+Your `project` contains all the "relevant" files, such as source files in the `contracts/` directory. The
+`contracts/` directory is where compilers look for contracts to compile. File extensions found within the `contracts/`
+directory determine which compiler plugin `ape` uses. Make sure to install the compiler plugins you need if they are
 missing by adding them to your `ape-config.yaml`'s `plugin` section, or manually adding via the following:
 
 ```bash
@@ -49,9 +49,9 @@ a.deploy(project.MyContract)
 
 ## Networks
 
-The default provider for the development network is the 
+The default provider for the development network is the
 [Ethereum Tester provider](https://github.com/ethereum/eth-testers). However, you can change the default provider per
-network using the `ape-config.yaml` file. 
+network using the `ape-config.yaml` file.
 
 ```yaml
 ethereum:
@@ -59,22 +59,36 @@ ethereum:
     default_provider: hardhat
 ```
 
-For specifying the network in an ad-hoc fashion, commands such as `run`, `test`, and `console` offer a `--network` 
+For specifying the network in an ad-hoc fashion, commands such as `run`, `test`, and `console` offer a `--network`
 option:
 
 ```bash
-ape run deploy --network ethereum:local:hardhat
+ape console --network ethereum:local:hardhat
 ```
 
 **NOTE**: If you are using the default ecosystem or network, you can omit them from the option:
 
 ```bash
-ape run deploy --network ::hardhat
+ape console --network ::hardhat
 ```
+
+## Scripts
+
+You can write scripts that run using the `ape run` command. The `ape run` command will register and run Python
+files defined under the `scripts/` directory that do not start with an `_` underscore. If the scripts take
+advantage of utilities from our [`ape.cli`](../methoddocs/cli.html) submodule,
+you can build a [Click](https://click.palletsprojects.com/) command line interface
+by defining a `click.Command` or `click.Group` object called `cli` in your file.
+Otherwise, if the script has a `main()` method, it will execute that method when called.
+
+Note that by default, `cli` scripts do not have
+[`ape.cli.network_option`](../methoddocs/cli.html?highlight=options#ape.cli.options.network_option)
+installed, giving you more flexibility in how you define your scripts.
+`main` scripts will always provide a network option to the call.
 
 ## Testing
 
-You can test your project using the `ape test` command. The `ape test` command comes with the core-plugin `ape-test`. 
+You can test your project using the `ape test` command. The `ape test` command comes with the core-plugin `ape-test`.
 The `ape-test` plugin extends the popular python testing framework
 [pytest](https://docs.pytest.org/en/6.2.x/contents.html).
 
@@ -91,14 +105,14 @@ def test_add():
 
 ### Fixtures
 
-Fixtures are any type of re-usable instances of something with configurable scopes. `pytest` handles passing fixtures 
-into each test method as test-time. You can define your own fixtures or use existing ones. The `ape-test` plugin comes 
+Fixtures are any type of re-usable instances of something with configurable scopes. `pytest` handles passing fixtures
+into each test method as test-time. You can define your own fixtures or use existing ones. The `ape-test` plugin comes
 with fixtures you will likely want to use:
 
 #### accounts fixture
 
 You have access to test accounts. These accounts are automatically funded, and you can use them to transact in your
-tests. Access each [test account](../methoddocs/api.html?highlight=testaccount#ape.api.accounts.TestAccountAPI) by 
+tests. Access each [test account](../methoddocs/api.html?highlight=testaccount#ape.api.accounts.TestAccountAPI) by
 index from the `accounts` fixture:
 
 ```python
@@ -163,8 +177,8 @@ Tests are generally divisible into three parts:
 2. Invocation
 3. Assertion
 
-In the example above, we created a fixture that deploys our smart-contract. This is an example of a 'setup' phase. 
-Next, we need to call a method on our contract. Let's assume there is a method called `is_owner()` that returns `True` 
+In the example above, we created a fixture that deploys our smart-contract. This is an example of a 'setup' phase.
+Next, we need to call a method on our contract. Let's assume there is a method called `is_owner()` that returns `True`
 when it is the owner of the contract making the transaction.
 
 This is an example of how that test may look:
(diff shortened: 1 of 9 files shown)


src/ape_run/_cli.py
```diff
@@ -1,33 +1,155 @@
...
+
+                ns["main"]()  # Call function
+
+            return call
+
+        else:
+            logger.warning(f"No 'main' method or 'cli' command in script: {relative_filepath}")
+
+            @click.command(short_help=f"Run '{relative_filepath}'")
+            def call():
```
unparalleled-js, line 74: this call could be a default call above and the `elif` above could call it before doing `ns["main"]()  # Call function`

unparalleled-js:
This seems pretty cool, I am going to try it out

unparalleled-js:
If the script does not exist (like `ape run asdfasdf`), the error you get is:

```
Try 'ape run -h' for help.

Error: No such command 'asdf'.
```

Can we make this somehow more custom, like

```
Try 'ape run <script-name> -h' for help or check if script exists.

Error: No such script or command 'asdf' found.
```

unparalleled-js:
I think my import statements are no longer working in scripts. e.g  a script like:

```python
from .utils import deploy


def main():
    deploy()
```

is failing with

```
  File "/Users/jules/PycharmProjects/ape/src/ape_run/_cli.py", line 62, in call
    eval(code, ns, ns)
  File "/Users/jules/ApeProjects/ape-demo-project/scripts/deploy.py", line 6, in <module>
    from .utils import deploy
KeyError: "'__name__' not in globals"
```

fubuloubu:
> I think my import statements are no longer working in scripts. e.g a script like:
> 
> ```python
> from .utils import deploy
> 
> 
> def main():
>     deploy()
> ```
> 
> is failing with
> 
> ```
>   File "/Users/jules/PycharmProjects/ape/src/ape_run/_cli.py", line 62, in call
>     eval(code, ns, ns)
>   File "/Users/jules/ApeProjects/ape-demo-project/scripts/deploy.py", line 6, in <module>
>     from .utils import deploy
> KeyError: "'__name__' not in globals"
> ```

I just added a small tweak where it ignores "private" files (e.g. `_utils.py`). Give that a try.

Will document this

unparalleled-js:
We could use some documentation around scripts, perhaps in `docs/userguides/projects`, just something like:

```
## Scripts

You can create scripts in the `scripts/` directory that either have a `main()` method or is a simple `click` CLI script.

Here is an example using the `main()` method:
...

Here is an example using the `cli()` method:

```

unparalleled-js:
> I just added a small tweak where it ignores "private" files (e.g. `_utils.py`). Give that a try.
> 
> Will document this

Cool!

I renamed it to `_utils` and tried both `from ._utils import` and `from _utils import` and I am still seeing the same error

unparalleled-js (dismissed):
Thank you for doing this

sabotagebeats:
I'm trying to use this branch but I'm getting this error: 
```
WARNING: Error loading scripts/deploy.py:
        TypeError: Arguments take exactly one parameter declaration, got 2.
INFO: Skipped loading: /home/doge/ape/simple-nft/scripts/deploy.py
```

tests/integration/cli/test_run.py
```diff
@@ -1,36 +1,27 @@
...
 
-@skip_projects(
-    ["empty-config", "no-config", "one-interface", "unregistered-contracts", "test", "geth"]
-)
 def test_run(ape_cli, runner, project):
-    for script_file in project.scripts_folder.glob("*.py"):
-        if script_file.stem == "__init__":
-            continue
+    result = runner.invoke(ape_cli, ["run"])
+    assert result.exit_code == 0, result.output
```
unparalleled-js, line 15: what is the thought behind allowing `ape run` to work without any arguments?

I would think at a script name should be required, no?

src/ape_run/_cli.py
```diff
@@ -1,33 +1,152 @@
...
+    short_help="Run scripts from the `scripts/` folder",
+)
 @click.option(
     "-i",
     "--interactive",
     is_flag=True,
     default=False,
     help="Drop into interactive console session after running",
 )
-@ape_cli_context()
```
unparalleled-js, line 15: How does the network get set on scripts now?
I usually run ape run <script> --network <network-path>.

Am I going to have to manually add the network_option to all of my scripts that use it now?

tests/integration/cli/projects/script/scripts/_utils.py
```diff
@@ -0,0 +1,2 @@
+# Note that any file prepended w/ underscore is ignored
+deploy = lambda: None  # noqa: E731
```
unparalleled-js, line 2:
```suggestion
def deploy():
    pass
```

This is not an anonymous method and there is no reason to use lambda then

tests/integration/cli/test_run.py
unparalleled-js, line 15: what is the thought behind allowing ape run to work without any arguments?

I would think a script name should be required, no?

unparalleled-js (commented):
I am wondering about something and am about to test it out

src/ape_run/_cli.py
fubuloubu: if we add it by default, then scripts connect to a network that don't need to

I think focusing more on "a set of functions for building scripts that integrate with ape" should be the goal here

tests/integration/cli/test_run.py
fubuloubu: To display help information about `ape run`, as well as display all the scripts that you can execute from the `scripts/` folder
```bash
$ ape run
  Run scripts from the ``scripts`` folder. ...

Options:
  -i, --interactive  Drop into interactive console session after running
  -h, --help         Show this message and exit.

Commands:
  click   Use a subcommand
  deploy  Run 'scripts/deploy.py:main'
  site    Run 'scripts/site.py'
```

unparalleled-js: That is neat!

src/ape_run/_cli.py
```diff
@@ -1,33 +1,152 @@
...
 
+        Returns:
+            :class:`~ape.managers.project.ProjectManager`
+        """
 
-@click.command(cls=NetworkBoundCommand, short_help="Run scripts from the `scripts` folder")
-@click.argument("scripts", nargs=-1)
+        from ape import project
+
+        project.config.load()
```
unparalleled-js, line 92:
```suggestion
        project.config_manager.load()
```

(if this is still needed)

tests/integration/cli/projects/script/scripts/_utils.py
```diff
@@ -0,0 +1,3 @@
+# Note that any file prepended w/ underscore is ignored
```
unparalleled-js, line 1: Can we add a note to the docs about this?

unparalleled-js (commented):
The error in the docs happens when I try to run a script,
(but is fixed when renaming `config` to  `config_manager)`
and I am still struggling to import my utils module.

I get:

```
ImportError: attempted relative import with no known parent package
```

I do have a `__init__.py` file so I am not sure why that is.

This is breaking change, so we should get this in before the release!

fubuloubu:
Note: scripts no longer launch networking processes by default, must explicitly connect or use CLI utilities for network bound commands

unparalleled-js:
There are a lot of places throughout the docs that say the `--network` option is available in the `ape run` command, so we need to fix that.

Also still would like docs showing users how to make scripts and differences between the `main()` method approach versus the `cli()` approach.

I have to migrate 90% of my scripts to be CLI based now, since they require the network option. This also means I have to repeat the same code that many times when before I didn't.

unparalleled-js (approved):
Please fix docs! Otherwise this is nice :)

I will busy for a while migrating 30 scripts 😅 
This is a big breaking change!

fubuloubu:
> There are a lot of places throughout the docs that say the `--network` option is available in the `ape run` command, so we need to fix that.

I think we should ensure this is the approach we'd want to have. It's easy enough for me to add, at least to non-cli backends, by just adding the class that adds the network option to the cli wrapper that the non-cli scripts execute within. Can just as easily add this, and you still have the option of avoiding network usage if you define a cli command without it (since that type will not get injected)

unparalleled-js:
I agree on the approach here for the reasons you have outlined! But we need to fix the places in the docs that still reference the old way, that is all I am saying.

docs/userguides/projects.md
```diff
@@ -49,32 +49,46 @@ a.deploy(project.MyContract)
...
 
 ```bash
-ape run deploy --network ::hardhat
+ape console --network ::hardhat
 ```
 
+## Scripts
+
+You can write scripts that run using the `ape run` command. The `ape run` command will register and run Python
+files defined under the `scripts/` directory, that do not start with an `_` underscore. If the scripts take
```
unparalleled-js, line 78:
```suggestion
files defined under the `scripts/` directory that do not start with an `_` underscore. If the scripts take
```

unparalleled-js, line 78:
```suggestion
files defined under the `scripts/` directory that do not start with an `_` underscore. If the scripts take
```

src/ape_run/_cli.py
```diff
@@ -1,33 +1,155 @@
...
+                except Exception as e:
+                    logger.error_from_exception(
+                        e, f"Exception while parsing script: {relative_filepath}"
+                    )
+                    return None  # Prevents stalling scripts
+
+            self._namespace[filepath.stem] = ns
+            return ns["cli"]
+
+            return None  # NOTE: Allow other scripts to load if loading script breaks
```
unparalleled-js, line 48: Does this actually ever get called?
Wouldn't the return statement on line 46 always either cause an exception to occur or actually return the CLI?

tests/integration/cli/projects/script/scripts/deploy.py
```diff
@@ -1,6 +1,11 @@
 from ape import networks
 
+# Make sure the re-import bug goes away
+from ._utils import deploy  # noqa: F401
+
 
 def main():
-    assert networks.active_provider.name == "test"
-    print("Script ran!")  # noqa: T001
+    with networks.parse_network_choice("::test"):
```
unparalleled-js, line 8: I don't think this part is needed anymore if `--network` is automatically passed in?
I am not sure how that would affect the test though

unparalleled-js (approved):
This is a really well-done PR!

I like having the `--network` option on `main`. I know that is a little inconsistent, but the flexibility is still there with the CLI commands and it makes sense to have CLIs be really flexible and main scripts to be classic.

src/ape_run/_cli.py
fubuloubu: Oh yeah, whoops. I found out that the `if` statement I had before was unreachable, forgot this line!

unparalleled-js (approved):
LGTM

PR #532, opened 2022-03-03
docs: ape init

Ninjagod1251:
### What I did

Added documentation to ape init to help the users understand and use the initialize command

### How I did it

added """ """ comments

### How to verify it

`ape init -h`

Please check my format and let me know if it is clear enough.
### Checklist


src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
...
-@click.option("--github")
+@click.option(
+    "--github", metavar="GH_username/repo", help="Add github_username/repository to clone project"
+)
 def cli(cli_ctx, github):
+    """
+    Ape Init allows the user to create an ape project with
+    default folders and ape-config.yaml
+
+        project_name/\n
```
unparalleled-js, line 20: I think the extra `\n` are not needed, but otherwise this is fantastic!

unparalleled-js, line 20: I think the extra \n are not needed, but otherwise this is fantastic!

src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
...
+            contracts/\n
+            tests/\n
+            scripts/\n
+            ape-config.yaml\n
+
+    ape-config.yaml is to manage project dependencies such as plugins.
+    Adding plugins is format sensitive.
```
unparalleled-js, line 27: This may be superfluous. 

How about we just link to the project docs guide instead?

https://docs.apeworx.io/ape/stable/userguides/projects.html

src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
 
 @click.command(short_help="Initalize an ape project")
 @ape_cli_context()
-@click.option("--github")
+@click.option(
+    "--github", metavar="GH_username/repo", help="Add github_username/repository to clone project"
```
unparalleled-js, line 13:
```suggestion
    "--github", metavar="github-org/repo", help="Clone a template from Github"
```

src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
...
+            contracts/\n
+            tests/\n
+            scripts/\n
+            ape-config.yaml\n
+
+    ape-config.yaml is to manage project dependencies such as plugins.
+    Adding plugins is format sensitive.
+
+    Ape init has an [OPTIONAL] argument to clone repositories
+    `ape init --github github_username/repository_name`
```
unparalleled-js, line 30:
```suggestion
    ``ape init --github github_username/repository_name``
```

unparalleled-js (commented):
This is great! I like the CLI docs, we should do this to the other commands.

Oh, don't forget to fix your PR title :)

I added some comments as well.

src/ape_init/_cli.py
fubuloubu: Yeah, so the docs needs double backticks to render properly

src/ape_init/_cli.py
fubuloubu: Docstrings via triple quotes render exactly as written, including all the newlines you have, so as Jules said no need for extra ones!

src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
 
 @click.command(short_help="Initalize an ape project")
 @ape_cli_context()
-@click.option("--github")
+@click.option(
+    "--github", metavar="GH_username/repo", help="Add github_username/repository to clone project"
+)
 def cli(cli_ctx, github):
+    """
+    Ape Init allows the user to create an ape project with
```
fubuloubu, line 17:
```suggestion
    ``ape init`` allows the user to create an ape project with
```

Although, click doesn't render the backticks by default... Wonder if there is a way to do this

src/ape_init/_cli.py
fubuloubu: Agree, please just link to docs on config

src/ape_init/_cli.py
```diff
@@ -9,8 +9,26 @@
...
+        project_name/\n
+            contracts/\n
+            tests/\n
+            scripts/\n
+            ape-config.yaml\n
+
+    ape-config.yaml is to manage project dependencies such as plugins.
+    Adding plugins is format sensitive.
+
+    Ape init has an [OPTIONAL] argument to clone repositories
```
fubuloubu, line 29:
```suggestion
    ``ape init`` has an option to clone repositories
```

fubuloubu (approved):

src/ape/utils.py
```diff
@@ -394,9 +394,11 @@ def get_repo(self, repo_path: str) -> GithubRepository:
         """
 
         if repo_path not in self._repo_cache:
-            self._repo_cache[repo_path] = self._client.get_repo(repo_path)
-
-        return self._repo_cache[repo_path]
+            try:
+                self._repo_cache[repo_path] = self._client.get_repo(repo_path)
+                return self._repo_cache[repo_path]
```
unparalleled-js, line 399: Ah, this is wrong because of the `if` statement.

To get around this, you could inver the if statement on line 396

unparalleled-js (commented):
`mypy` caught something
Also, don't forgot to catch up with upstream!

fubuloubu:
Just noticed `ape_init` isn't in here:

https://github.com/ApeWorX/ape/blob/main/src/ape/__modules__.py

Ninjagod1251:
> Just noticed `ape_init` isn't in here:
> 
> https://github.com/ApeWorX/ape/blob/main/src/ape/__modules__.py

I added it to module thank you

unparalleled-js (approved):
Nice work!

PR #533, opened 2022-03-04
docs: have user update their version of pip before installing

fubuloubu:
### What I did

fixes: #

### Checklist


unparalleled-js (approved):
`--upgrade` is more explicit to non-familiar users than `-U` but this is good!

Issue #526, opened 2022-03-01
Allow specifying the index of a Dataframe for query purposes,


fubuloubu:
### Overview
Allow specifying the index of a Dataframe for query purposes, either:
- block numbers (for block-based queries)
- nonces (for account transaction queries)
- timestamps (for both)

### Specification
TBD

### Dependencies
https://pandas.pydata.org/docs/user_guide/indexing.html

PR #535, opened 2022-03-04
feat: default query provider

johnson2427:
### What I did

Added a default block query under src/ape/api/managers

fixes: #526 

### How I did it

Perform query required a map function in order to deal with potential memory issues when performing large queries

### How to verify it

Must add tests, but this was how I tested it from the ape console:

```python
In [1]: a = accounts.test_accounts[0]
In [2]: a.transfer(a, 100)
In [3]: a.transfer(a, 100)
In [4]: a.transfer(a, 100)
In [5]: chain.blocks.query("number")
In [6]: chain.blocks.query("number", "timestamp")
```

### Checklist

### DIFF
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -17,9 +17,9 @@ def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
 
 class DefaultQueryProvider(QueryAPI):
     """
-     Default implementation of the ape.api.query.QueryAPI
-     Allows for the query of blockchain data using connected provider
-     """
+    Default implementation of the ape.api.query.QueryAPI
+    Allows for the query of blockchain data using connected provider
+    """
 
     def estimate_query(self, query: QueryType) -> Optional[int]:
         """
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -16,7 +16,26 @@ def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
 
 
 class DefaultQueryProvider(QueryAPI):
+    """
+    Default implementation of the ape.api.query.QueryAPI
+    Allows for the query of blockchain data
+
+    Args:
+        query: (``QueryType``): query to execute
+
+    returns:
+        pd.DataFrame: response of query in a pandas dataframe
+    """
     def estimate_query(self, query: QueryType) -> Optional[int]:
+        """
+        Estimates the time that the query will take as a timestamp
+
+        Args:
+            query (``QueryType``): The transaction data you want to query
+
+        Returns:
+             Optional[int]: Depends on whether the query can be completed
+        """
         if isinstance(query, BlockQuery):
             # NOTE: Very loose estimate of 100ms per block
             return (query.stop_block - query.start_block) * 100
@@ -24,14 +43,19 @@ class DefaultQueryProvider(QueryAPI):
         return None  # can't handle this query
 
     def perform_query(self, query: QueryType) -> pd.DataFrame:
-        provider = self.network_manager.active_provider
-        if not provider:
-            raise QueryEngineError("Not connected to a provider.")
+        """
+        Performs a query
 
+        Args:
+            query (``QueryType``): The specific transaction data you want to query
+
+        Returns:
+            pd.DataFrame: A pandas dataframe
+        """
         if isinstance(query, BlockQuery):
             blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
-            blocks_iter = map(partial(get_columns_from_item, query), blocks_iter)  # type: ignore
-            return pd.DataFrame(columns=query.columns, data=blocks_iter)
+            block_dicts_iter = map(partial(get_columns_from_item, query), blocks_iter)
+            return pd.DataFrame(columns=query.columns, data=block_dicts_iter)
 
         raise QueryEngineError(f"Cannot handle '{type(query)}'.")
 
@@ -58,13 +82,13 @@ class QueryManager(ManagerAccessMixin):
             dict[str, :class:`~ape.api.query.QueryAPI`]
         """
 
-        engines = {"__default__": DefaultQueryProvider()}
+        engines: Dict[str, QueryAPI] = {"__default__": DefaultQueryProvider()}
 
         for plugin_name, (engine_class,) in self.plugin_manager.query_engines:
             engine_name = clean_plugin_name(plugin_name)
             engines[engine_name] = engine_class()
 
-        return engines  # type: ignore
+        return engines
 
     def query(self, query: QueryType, engine_to_use: Optional[str] = None) -> pd.DataFrame:
         """
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -26,6 +26,7 @@ class DefaultQueryProvider(QueryAPI):
     returns:
         pd.DataFrame: response of query in a pandas dataframe
     """
+
     def estimate_query(self, query: QueryType) -> Optional[int]:
         """
         Estimates the time that the query will take as a timestamp
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -1,11 +1,40 @@
-from typing import Dict, Optional
+from eth_utils import keccak, to_hex, to_bytes
+from functools import partial
+from typing import Any, Dict, Optional
 
 import pandas as pd
+from pydantic import BaseModel
 
 from ape.api import QueryAPI, QueryType
+from ape.api.query import ContractEventQuery, BlockQuery, _BaseQuery
 from ape.exceptions import QueryEngineError
 from ape.plugins import clean_plugin_name
 from ape.utils import ManagerAccessMixin, cached_property
+
+
+def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
+    return {k: v for k, v in item.dict().items() if k in query.columns}
+
+
+class DefaultQueryProvider(QueryAPI):
+    def estimate_query(self, query: QueryType) -> Optional[int]:
+        if isinstance(query, BlockQuery):
+            # NOTE: Very loose estimate of 100ms per block
+            return (query.stop_block - query.start_block) * 100
+
+        return None  # can't handle this query
+
+    def perform_query(self, query: QueryType) -> pd.DataFrame:
+        provider = self.network_manager.active_provider
+        if not provider:
+            raise QueryEngineError("Not connected to a provider.")
+
+        if isinstance(query, BlockQuery):
+            blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
+            blocks_iter = map(partial(get_columns_from_item, query), blocks_iter)
+            return pd.DataFrame(columns=query.columns, data=blocks_iter)
+
+        raise QueryEngineError(f"Cannot handle '{type(query)}'.")
 
 
 class QueryManager(ManagerAccessMixin):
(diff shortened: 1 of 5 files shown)


johnson2427:
forgot to run pre-commit

src/ape/managers/query.py
```diff
@@ -1,13 +1,42 @@
...
 
+def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
+    return {k: v for k, v in item.dict().items() if k in query.columns}
+
+
+class DefaultQueryProvider(QueryAPI):
+    def estimate_query(self, query: QueryType) -> Optional[int]:
+        if isinstance(query, BlockQuery):
+            # NOTE: Very loose estimate of 100ms per block
+            return (query.stop_block - query.start_block) * 100
```
unparalleled-js, line 23: It might be nice to define the multipliers as constants at the top of the module... all next to each other, ready to be tweaked.

unparalleled-js, line 23: Are events going to have different multiplier values than blocks etc?
How are you choosing values?

src/ape/managers/query.py
```diff
@@ -1,13 +1,42 @@
...
+
+        return None  # can't handle this query
+
+    def perform_query(self, query: QueryType) -> pd.DataFrame:
+        provider = self.network_manager.active_provider
```
unparalleled-js, line 28: I believe the provider now accessible in the managers mixin so you don't have to go through `self.network_manager` anymore.

src/ape/managers/query.py
```diff
@@ -1,13 +1,42 @@
...
+        if not provider:
```
unparalleled-js, line 29: You don't have to raise this error anymore or do this check.. It happens at the mixin level now

src/ape/managers/query.py
```diff
@@ -1,13 +1,42 @@
...
 from ape.exceptions import QueryEngineError
 from ape.plugins import clean_plugin_name
 from ape.utils import ManagerAccessMixin, cached_property
 
 
+def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
+    return {k: v for k, v in item.dict().items() if k in query.columns}
+
+
+class DefaultQueryProvider(QueryAPI):
```
unparalleled-js, line 19: If this class needs docs, make sure to add those (and the methods too).

unparalleled-js (commented):
Just adding some comments!

src/ape/managers/query.py
fubuloubu: Just a loosely-chosen heuristic that is more-or-less accurate, so that the selection algorithm works correctly when other query plugins are added that can fulfill the query faster

src/ape/managers/query.py
johnson2427: Okay will check

src/ape/managers/query.py
johnson2427: Will remove

src/ape/managers/query.py
johnson2427: Was waiting to get a pass before adding docs

src/ape/managers/query.py
```diff
@@ -1,13 +1,41 @@
...
+
+    def perform_query(self, query: QueryType) -> pd.DataFrame:
+        provider = self.network_manager.active_provider
+        if not provider:
+            raise QueryEngineError("Not connected to a provider.")
+
+        if isinstance(query, BlockQuery):
+            blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
+            blocks_iter = map(partial(get_columns_from_item, query), blocks_iter)  # type: ignore
+            return pd.DataFrame(columns=query.columns, data=blocks_iter)
```
fubuloubu, line 34: This should fix the typing issue, because we are no longer re-defining the variable
```suggestion
            block_dicts_iter = map(partial(get_columns_from_item, query), blocks_iter)
            return pd.DataFrame(columns=query.columns, data=block_dicts_iter)
```

src/ape/managers/query.py
```diff
@@ -30,12 +58,13 @@ def engines(self) -> Dict[str, QueryAPI]:
             dict[str, :class:`~ape.api.query.QueryAPI`]
         """
 
-        engines = {}
+        engines = {"__default__": DefaultQueryProvider()}
```
fubuloubu, line 61: I think this will solve the typing issue below (e.g. the typing is more constrained here than it should be)
```suggestion
        engines: Dict[str, QueryAPI] = {"__default__": DefaultQueryProvider()}
```

src/ape/managers/query.py
johnson2427: I am going to leave this constant for now, we can pull it out and make it global in a future revision

src/ape/managers/query.py
johnson2427: removed

src/ape/managers/query.py
johnson2427: done

src/ape/managers/query.py
fubuloubu: Oh wait, I forgot that if you have docstrings on the base class, then the docs get propagated upwards.

Remove the docs on the methods, and just add one on the class at the upper level describing the unique properties of the default provider

fubuloubu (commented):
Add some tests and change the docs, and we are good to go!

src/ape/managers/query.py
```diff
@@ -1,13 +1,66 @@
...
+def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
+    return {k: v for k, v in item.dict().items() if k in query.columns}
+
+
+class DefaultQueryProvider(QueryAPI):
+    """
+    Default implementation of the ape.api.query.QueryAPI
+    Allows for the query of blockchain data
+
+    Args:
```
fubuloubu, line 23: This doc is incorrect, at class level you don't have args or returns

johnson2427: So add the docs on the QueryAPI?

src/ape/managers/query.py
```diff
@@ -1,13 +1,66 @@
...
+        query: (``QueryType``): query to execute
+
+    returns:
+        pd.DataFrame: response of query in a pandas dataframe
+    """
```
fubuloubu, line 28:
```suggestion
    """
    Default implementation of the ape.api.query.QueryAPI
    Allows for the query of blockchain data using connected provider
    """
```

src/ape/managers/query.py
fubuloubu: Sorry, see https://github.com/ApeWorX/ape/pull/535#discussion_r820127534

johnson2427:
Question, why don't we just query all data from chain.blocks.query since it's all available to begin with (and we have that map function that limits overloading memory) and then allow the user to choose what they want to see in the dataframe after they pull that query out?

johnson2427:
This test is failing because of tests that I did not touch. And they pass in my environment.

unparalleled-js:
The changes in this PR should help you with your failing tests: https://github.com/ApeWorX/ape/pull/549#

tests/functional/api/test_query.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_basic_query(test_provider):
+    a = accounts.test_accounts[0]
+    a.transfer(a, 100)
+    a.transfer(a, 100)
+    a.transfer(a, 100)
+    df = chain.blocks.query("number")
+    assert len(df) == 3
+    assert df["number"][0] == 0
```
fubuloubu, line 11: Do something more like this (but fill in the `...`)
```suggestion
    assert chain.blocks.query("number") == pd.Dataframe(...)
```

fubuloubu:
> Question, why don't we just query all data from chain.blocks.query since it's all available to begin with (and we have that map function that limits overloading memory) and then allow the user to choose what they want to see in the dataframe after they pull that query out? As a data guy, I'd like to see what data is available up front, then eliminate columns rather than pull piece by piece (I know you can pull a list of values, but if I wanted all the data, it would be nice to just run the query, then format my dataframe)

you can do both!

```py
df = chain.blocks.query("*")  # get everything
numbers = df["numbers"]
# OR
df = chain.blocks.query("numbers")
```

It might not be obvious now when allowing the query to de-select is handy, but when we add other provider types that query larger databases it will come in handy.

Also, some elements (that we still need to add to the underlying models) like `"transactions"` are very heavy, and will need some sort of "table join" to effectively query, which will be very expensive to do by default (unless asked for)

fubuloubu (dismissed):

fubuloubu (approved):

tests/functional/conftest.py
```diff
@@ -118,3 +119,9 @@ def sender(test_accounts):
 @pytest.fixture
 def receiver(test_accounts):
     return test_accounts[1]
+
+
+@pytest.fixture
+def test_provider():
```
unparalleled-js, line 125: You can use the existing fixture `eth_tester_provider`

tests/functional/api/test_query.py
```diff
@@ -0,0 +1,12 @@
+from ape import accounts, chain
+
+
+def test_basic_query(test_provider):
+    a = accounts.test_accounts[0]
+    a.transfer(a, 100)
```
unparalleled-js, line 6: Calling `test_provider.mine()` is probably a better way to advance the blockchain

src/ape/managers/query.py
```diff
@@ -1,13 +1,60 @@
...
+from pydantic import BaseModel
 
 from ape.api import QueryAPI, QueryType
+from ape.api.query import BlockQuery, _BaseQuery
 from ape.exceptions import QueryEngineError
 from ape.plugins import clean_plugin_name
 from ape.utils import ManagerAccessMixin, cached_property
 
 
+def get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
```
unparalleled-js, line 14:
```suggestion
def _get_columns_from_item(query: _BaseQuery, item: BaseModel) -> Dict[str, Any]:
```

this will ensure this method stays out of the docs

unparalleled-js (commented):
Looks good, mostly small stuff.

The main thing is the comment about `range()` and the stop block

src/ape/managers/query.py
fubuloubu: This method is non-optimal anyways, @NotPeopling2day discovered a nicer way via Pandas

tests/functional/api/test_query.py
unparalleled-js: I think you can ever just do `test_provider.mine(3)` to replace all 3 lines

tests/functional/api/test_query.py
```diff
@@ -0,0 +1,12 @@
+from ape import accounts, chain
```
fubuloubu, line 1:
```suggestion
from ape import chain
```

tests/functional/api/test_query.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def test_basic_query(test_provider):
+    a = accounts.test_accounts[0]
+    a.transfer(a, 100)
+    a.transfer(a, 100)
+    a.transfer(a, 100)
```
fubuloubu, line 8:
```suggestion
    test_provider.mine(3)
```

fubuloubu (approved):
Flake8 issue:
```
./src/ape/managers/chain.py:8:1: F401 'ape.api.query.QueryAPI' imported but unused
```

tests/functional/api/test_query.py
fubuloubu: goes with https://github.com/ApeWorX/ape/pull/535#discussion_r823279432

fubuloubu:
Update all of the above, and we can merge!

tests/functional/conftest.py
johnson2427: I will fix this

fubuloubu (approved):

src/ape/managers/query.py
johnson2427: @unparalleled-js why couldn't we remove the docs from the methods again?

Issue #534, opened 2022-03-04
Allow `ape-geth` and `ape-test` provider plugins to inject configs for other ecosystems


fubuloubu:
### Overview
We supported re-importing `ape-geth` and `ape-test` as plugins for another ecosystem (`ape-fantom`) and it retains their original location, however the configs for those two providers are specific to `ape-ethereum`. Allow 2nd/3rd party `EcosystemAPI` plugins to inject their own network config items into these 1st party provider's configs

### Specification
TBD

### Dependencies
#500
#506

PR #537, opened 2022-03-04
refactor: allow config to pass extra fields

NotPeopling2day:
### What I did

fixes: #534 

### How I did it

Allow passing of extra fields via the pydantic Config class

### Checklist

### DIFF
--- a/src/ape_geth/providers.py
+++ b/src/ape_geth/providers.py
@@ -9,6 +9,7 @@ from geth.accounts import ensure_account_exists  # type: ignore
 from geth.chain import initialize_chain  # type: ignore
 from geth.process import BaseGethProcess  # type: ignore
 from geth.wrapper import construct_test_chain_kwargs  # type: ignore
+from pydantic import Extra
 from requests.exceptions import ConnectionError
 from web3 import HTTPProvider, Web3
 from web3.exceptions import ContractLogicError as Web3ContractLogicError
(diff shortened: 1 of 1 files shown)


src/ape_geth/providers.py
```diff
@@ -118,6 +119,9 @@ class GethNetworkConfig(PluginConfig):
     # Make sure to run via `geth --dev` (or similar)
     local: dict = DEFAULT_SETTINGS.copy()
 
+    class Config:
+        extra = Extra.allow
```
unparalleled-js, line 123: maybe we need a comment saying what this is for? "Adding ecosystems that use Geth, such as geth-forks"

unparalleled-js: also (talked offline) but this is the wrong spot, needs to be in the class below.

After that, things seem to work in ape-fantom!

unparalleled-js (commented):
I did more testing!

* I think in that same URI method, we need to make return the default URI if `self.config`  does NOT have attr `self.network.ecosystem.name`.. Otherwise, we always have to specify the values in the config

* Error handling for non-dict extra fields that clearly are not network configs

fubuloubu:
I was able to use this with `ape-fantom` and get some stuff working for me, however I had one bug:
https://github.com/ApeWorX/ape/pull/537/files#diff-006c1d861ec6abe5c7a1135384957b5da1a101f56bb4fb991624988449b9d345R147

changed to `ecosystem_config.get(self.network.name)` because it is `__root__: dict`

unparalleled-js (approved):
This seems to work!
Also if you combine this with the logic in this PR: https://github.com/ApeWorX/ape/pull/544, the experience is even better

unparalleled-js (approved):
LGTM

PR #538, opened 2022-03-04
fix: adjust to honest python versions and upgrade web3.py dependency

unparalleled-js:
### What I did

We cannot support python versions 3.7 and 3.7.1; they are not part of the list of python versions the `ipfshttpclient` package supports: https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/master/pyproject.toml#L15-L23

And `ipfshttpclient` is a dependency of `web.py`. I raised an issue here https://github.com/ethereum/web3.py/issues/2378 to see if we can get it also upgraded there.

### How I did it

bump!

### How to verify it

install ape on your python 3.7 version (which is probably python 3.7.12 or something)

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,7 +15,7 @@ jobs:
               python-version: 3.8
 
         - name: Install Dependencies
-          run: pip install .[lint]
+          run: pip install --upgrade pip && pip install .[lint]
 
         - name: Run Black
           run: black --check .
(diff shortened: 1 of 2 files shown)


unparalleled-js:
All part of an effort to be able to install `starknet.py` and `ape` in the same python env:

https://github.com/software-mansion/starknet.py/pull/92

fubuloubu (approved):

PR #543, opened 2022-03-05
fix: issue where we were using a default mutable argument

unparalleled-js:
### What I did

We were using a default mutable argument, which is python anti-pattern than can lead to some highly unusual situations:
https://duckduckgo.com/?q=default+mutable+arugments+python&t=brave&ia=web

### How I did it

Default to None and coerce to empty list in first line of init

### How to verify it

with your eyeballs is fine

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -228,12 +228,12 @@ class ContractEvent(ManagerAccessMixin):
         self,
         contract: "ContractInstance",
         abis: List[EventABI],
-        cached_logs: List[ContractLog] = [],
+        cached_logs: List[ContractLog] = None,
     ) -> None:
         super().__init__()
         self.contract = contract
         self.abis = abis
-        self.cached_logs = cached_logs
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
Oh yes, this is a good one!

In general, this is okay w/ Pydantic, because it actually intelligently does a "copy-on-write" here instead of falling into the trap of mutable default args, but this class in particular is not a subclass of Pydantic's BaseModel

unparalleled-js:
> In general, this is okay w/ Pydantic, because it actually intelligently does a "copy-on-write" here instead of falling into the 

Wow, does pydantic even do this for the methods defined on the model or just for the model's keys?

fubuloubu:
> > In general, this is okay w/ Pydantic, because it actually intelligently does a "copy-on-write" here instead of falling into the 
> 
> Wow, does pydantic even do this for the methods defined on the model or just for the model's keys?

Just model fields ("keys")

Issue #525, opened 2022-03-01
feat: Add Error Message about no Test Directory Found


Ninjagod1251:
### Environment information

* `ape` and plugin versions:

Installed Plugins:
  solidity        0.1.0b5
  vyper           0.1.0b6
  alchemy         0.1.0b2
  tokens          0.1.0b4
  hardhat         0.1.0b3

```
$ ape --version:

0.1.0b5

$ ape plugins list

* Python Version: x.x.x
* OS: macOS/linux/win

### What went wrong?

When I run ape test on the wrong current working directory. I do not get the right error message to fix my problem. 

For example my project directory is nested on folder deeper and the error message is:

```
(ape) chris@DESKTOP-ID4V0R6:~/_src/template/ape-ethdenver$ ape test
=========================================================== test session starts ===========================================================
platform linux -- Python 3.8.10, pytest-6.2.5, py-1.11.0, pluggy-0.13.1
rootdir: /home/chris/_src/template/ape-ethdenver, configfile: pyproject.toml, testpaths: tests
plugins: eth-ape-0.1.0b5, web3-5.27.0
collected 0 items                                                                                                                         
ERROR: (ProviderNotConnectedError) Not connected to a network provider.
```


### How can it be fixed?

When we run ape test. We should say:

Console print:
`running ape test in (cwd)`

error message:
`ProjectError: No 'tests/' directory found. Are you in an Ape project?`

PR #545, opened 2022-03-05
fix: handle exiting pytest when provider never connects

unparalleled-js:
### What I did

fixes: #525

Before output:

```bash
collected 0 items / 1 error                                                                                                                                                                    
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
ERROR: (ProviderNotConnectedError) Not connected to a network provider
```

New output:

```bash
==================================================================================== no tests ran in 0.01s =====================================================================================
```

### How I did it

* Don't try to disconnect if connect never happens
* Remove logger error stmt from `ape_test._cli.py` as it just noise and does add anything (pytest output is what matters)

This works exactly the same using `pytest` without ape installed.

### How to verify it

go to a directory without tests and run `ape test`

See output expected outputs above for both `main` and this branch!

### Checklist

### DIFF
--- a/src/ape/pytest/runners.py
+++ b/src/ape/pytest/runners.py
@@ -19,6 +19,7 @@ class PytestApeRunner(ManagerAccessMixin):
     ):
         self.pytest_config = pytest_config
         self._warned_for_missing_features = False
+        self._provider_is_connected = False
         ape.reverts = RevertsContextManager  # type: ignore
 
     @property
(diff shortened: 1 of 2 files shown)


unparalleled-js:
I think it may be possible to run some of our "functional" tests using `ape test` since that works the same as `pytest` now.
We can consider that during the test refactor.

fubuloubu (approved):

PR #546, opened 2022-03-06
fix: make debug work for failed-to-load plugins

unparalleled-js:
### What I did

Currently, if a plugin fails to load, it says you can debug the reason why... but the `-v` argument does not load in time, so you really cannot. Debugging other stuff works but not the failed to load plugins; that stuff happens really early in the lifecycle now.

This PR addresses this bug and forces a more eager setting LogLevel of the logger.

### How I did it

At plugin load time, check `sys.argv` and set the logger then.

### How to verify it

Step 1: Break a plugin so it does not load
Step 2: Run `ape` notice it failed to load
Step 3: Run `ape -v debug` and notice you get a stacktrace
Step 4: Run `ape networks list -v debug` and notice the command still works AND you get a stacktrace at top for the failed to load plugin
Step 5: Notice `ape -v` fails the normal way and not in some weird way (careful arg indexing)

### Checklist

### DIFF
--- a/src/ape/plugins/__init__.py
+++ b/src/ape/plugins/__init__.py
@@ -1,10 +1,9 @@
 import functools
 import importlib
 import pkgutil
-import sys
 from typing import Any, Callable, Generator, Iterator, List, Optional, Tuple, Type, cast
 
-from ape.logging import LogLevel, logger
+from ape.logging import logger
 
 from .account import AccountPlugin
 from .compiler import CompilerPlugin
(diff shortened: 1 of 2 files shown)


src/ape/plugins/__init__.py
```diff
@@ -122,6 +123,15 @@ class PluginManager:
     _unimplemented_plugins: List[str] = []
 
     def __init__(self) -> None:
+        # NOTE: This is a hack to make 'ape -v debug' correctly debug failing plugins.
```
unparalleled-js, line 126: Other things I tried before this:

* Override something in the `click.Multicommand` we are using and force-load the argument there. But, based on where this all happens in the click flow, that did not seem possible without hacky measures (as far as I could tell)
* Put this same logic at the top of the main `__init__.py` script... that did not work either. Odd.

I am open to suggestions!

It is really nice being able to debug failing plugins now!

fubuloubu (approved):
One thing to try would be breaking an `EcosystemAPI` plugin, because due to how they work it is more catastrophic than others because at least one ecosystem is expected.

unparalleled-js:
> One thing to try would be breaking an `EcosystemAPI` plugin, because due to how they work it is more catastrophic than others because at least one ecosystem is expected.

Good! As I am testing this on an ecosystem plugin :)

The bug in the ecosystem was very deep in the dependencies too! So the trace is very helpful.

src/ape/plugins/__init__.py
unparalleled-js: Was able to put it in the logging module itself, so now when you request the logger, it sets

fubuloubu (approved):

Issue #468, opened 2022-02-04
Contract types from dependencies cannot be used out of the box


fubuloubu:
### What went wrong?
Project dependencies cannot access their contract types as `ContractContainer` instances (necessary to deploy them)
```py
In [1]: type(project.dependencies["erc4626"].Vault)
Out[1]: ethpm_types.contract_type.ContractType

In [2]: type(project.Portfolio)
Out[2]: ape.contracts.base.ContractContainer
```

which gives me this error:
```py
>       return sudo.deploy(project.dependencies["erc4626"].Token)
E       TypeError: 'ContractType' object is not callable
```

### How can it be fixed?

I think because project.dependencies is a dict to manifest contents, and not Projects

potential idea: have projects load from `ProjectAPI` and then convert to manifest, add class method `ApeProject.from_manifest` that expands a project from a manifest file, so that it can be maximally useful inside of your local project.

fubuloubu:
@unparalleled-js is this one solved?

PR #547, opened 2022-03-06
fix: bug preventing you from interacting with dependency contracts

unparalleled-js:
### What I did

fixes: #468 

### How I did it

* Exclude dependency contracts from `ProjectManager.__getattr__`
* Add `__getattr__` to `DependencyAPI` that returns `ContractContainer` for the dependency's manifest's contracts.
* Change the type of `ProjectManager.dependencies` to `Dict[str, DependencyAPI]`

### How to verify it

I think I can add a test using a local dependency but it would be some work involved. I may look into that more later.

Otherwise, you can do this to test:

Have a config like this:

```yaml
dependencies:
  - name: oz
    version: 4.4.2

solidity:
  import_remapping:
    - "@openzeppelin=oz/4.4.2"
```

And a `deploy_dependency.py` script like this:

```python


def main():
    account = accounts.test_accounts[0]
    contract = project.dependencies["oz"].AccessControl
    account.deploy(contract)
```

Live example of the one I used to test / develop this feature: https://github.com/unparalleled-js/ape-demo-project/blob/main/scripts/deploy_dependency.py

### Checklist


src/ape/managers/project.py
```diff
@@ -470,10 +470,9 @@ def __getattr__(self, attr_name: str) -> ContractContainer:
         """
 
         contracts = self.load_contracts()
+
         if attr_name in contracts:
             contract_type = contracts[attr_name]
-        elif attr_name in self.dependencies:
-            contract_type = self.dependencies[attr_name]  # type: ignore
```
fubuloubu, line 476: Wait, is this change removing the support for `project.MyDependency`?

docs/userguides/projects.md
```diff
@@ -47,6 +47,20 @@ a = accounts.load("metamask_0")
...
+To set up dependencies in your ``ape-config.yaml`` file, follow [this guide](https://docs.apeworx.io/ape/stable/userguides/config.html#dependencies).
+If you are setting dependencies from a remote source, they will download when you run `ape compile` or other commands that compile in the background.
+Dependencies only need to download and compile once.
+
+You can access dependency contracts off your root project manager the same as your project's contracts.
+
+```python
+from ape import project
+
+dependency = project.MyDependency
```
fubuloubu, line 61: Yeah, I think if you had dependencies it wouldn't let you compile anything with the same name as something in your dependencies because of name shadowing, come to think of it.

fubuloubu (approved):
Overall, a bit confused on the outcome of this PR, but dependencies in general are going to be hard to use if everything name-shadows. Manifests might be able to help with this.

p.s. as of Vyper 0.3.2, the compiler will allow you to import types from manifest files directly, which seems like the preferred way of handling things (if Solidity eventually implements the same feature)

src/ape/managers/project.py
unparalleled-js: The contract types are available on the project already, so this `if` statement would never get called if given the contract name of a dependency.

However - before you could do `project.OpenZeppelin` and have access to the project manifest. Do we still want to support that? We could still do the `from_manifest()` idea too, I just didn't know if it was as important knowing we can access any contact type we need.

Yes, I am confused to. I would like to know the ideal situation so we can strive for that.

docs/userguides/projects.md
unparalleled-js: Correct, at this time (even before this PR), you would hit a naming collision if any contract type in your project manifest had the same name.

unparalleled-js:
I am going to mark this as a draft for now until we all understand:

* the current design
* what (if anything) we would like changed about the current design
* what is the next step towards getting there

src/ape/managers/project.py
fubuloubu: Ah yes, so I think the original design here was thinking that `dependencies: Dict[str, ProjectManager]`, so that if you did `project.my_dependency.DependencyContractType` it would first do `project.__getattr__` to look up the dependency and return that `ProjectManager` instance, and then next do `my_dependency.__getattr__` to get the contract inside that `ProjectManager` instance.

I think we changed things around a lot here where `dependencies: Dict[str, PackageManifest]` instead, causing the issue since `PackageManifest.__getattr__` returns an `ethpm_types.ContractType` data model instead

src/ape/managers/project.py
```diff
@@ -228,6 +228,19 @@ def extract_manifest(self) -> PackageManifest:
             return self._extract_local_manifest(temp_project_path)
 
 
+class PackageManifestWrapper:
```
fubuloubu: Could this just be `Package`?

unparalleled-js: Ah, I deleted this class and went with a different approach that didn't involve adding a new class.

```python
def dependencies(self) -> Dict[str, DependencyAPI]:
```

fubuloubu: okay hahaha, will check it out

docs/userguides/projects.md
fubuloubu, line 61: It might be nicer to use snake_case in the example here, to make it clear it's not a ContractType
```suggestion
dependency = project.my_dependency
```

src/ape/api/projects.py
```diff
@@ -203,6 +206,15 @@ def cached_manifest(self) -> Optional[PackageManifest]:
         """
         return _load_manifest_from_file(self._target_manifest_cache_file)
 
+    def __getattr__(self, item: str) -> "ContractContainer":
+        manifest = self.extract_manifest()
+        if hasattr(manifest, item):
+            from ape.contracts import ContractContainer
```
fubuloubu, line 212: What would be the issues in flipping this cycle so that we don't have to import it internally here? I see `BaseAddress` and `Address` used in `ape.contracts`

Perhaps for another PR refactoring that out into `ape.types` somehow

fubuloubu (dismissed):

src/ape/api/projects.py
unparalleled-js: I was also wondering if it would make sense to have access to the partial off of the manager-base-mixin. That way, you can create contract containers anywhere.

docs/userguides/projects.md
unparalleled-js: Oh, I need to update these docs completely! I went with the `project.dependencies["my_dependency"].ContractTypeThing` approach because it seemed like you were favoring that yesterday

fubuloubu: oh yes, I forgot, yes please let's do that

fubuloubu (dismissed):

src/ape/managers/project.py
```diff
@@ -467,19 +467,16 @@ def __getattr__(self, attr_name: str) -> ContractContainer:
...
-        else:
-            # Fixes anomaly when accessing non-ContractType attributes.
-            # Returns normal attribute if exists. Raises 'AttributeError' otherwise.
-            return self.__getattribute__(attr_name)  # type: ignore
+        if attr_name in self.contracts:
+            source_id = self.contracts[attr_name].source_id
+            if source_id and not source_id.startswith(".cache"):
+                return ContractContainer(  # type: ignore
+                    contract_type=self.contracts[attr_name],
+                )
```
fubuloubu, line 475: probably should be an `else` and then raise `AttributeError` with the specific issue here (cannot load dependency contracts this way)

fubuloubu (dismissed):

docs/userguides/projects.md
```diff
@@ -76,7 +93,7 @@ ape console --network ::hardhat
 
 You can write scripts that run using the `ape run` command. The `ape run` command will register and run Python
 files defined under the `scripts/` directory that do not start with an `_` underscore. If the scripts take
-advantage of utilities from our [`ape.cli`](../methoddocs/cli.html) submodule,
+advantage of utilities from our [`ape.cli`](../methoddocs/cli.html#ape-cli) submodule,
```
unparalleled-js, line 96: For some reason, the link was not working without adding this.

src/ape/managers/project.py
```diff
@@ -451,35 +451,44 @@ def contracts(self) -> Dict[str, ContractType]:
...
+                        "Cannot access dependency contracts from root project manager. "
+                        f"Please use `project.dependencies['dependency_name'].{attr_name}`."
+                    )
 
-        return ContractContainer(  # type: ignore
-            contract_type=contract_type,
-        )
+                return ContractContainer(  # type: ignore
+                    contract_type=self.contracts[attr_name],
+                )
```
fubuloubu, line 487: should this be dedented to the right one? `source_id` is not a requirement to load the type

fubuloubu (dismissed):

src/ape/managers/config.py
```diff
@@ -168,12 +176,22 @@ def _plugin_configs(self) -> Dict[str, PluginConfig]:
...
 
+    def __str__(self) -> str:
+        """
+        The JSON-text version of the project config data.
+
+        Returns:
+            str
+        """
+
+        return self.json()
```
fubuloubu, line 193: Why is this returning json for str method?

fubuloubu (approved):

src/ape/managers/config.py
unparalleled-js: I don't know, I didn't write this method, it was just moved it near the `__repr__` in effort to clean this up a little.

The type of `json()` return is a `str` though.

I am not sure what uses this.

fubuloubu: yeah... this seems like a good candidate for optimizing out

unparalleled-js: 💨

src/ape/utils.py
```diff
@@ -609,6 +611,37 @@ def provider(self) -> "ProviderAPI":
...
+        """
+        Helper method for creating a ``ContractContainer``.
+
+        Args:
+            contract_type (``ContractType``): Type of contract for the container
+
+        Returns:
+            :class:`~ape.contracts.ContractContainer`
+        """
+        from ape.contracts.base import ContractContainer
```
unparalleled-js, line 624: We could make two cached properties for importing of `ContractContainer` and `ContractInstance`...
However, I suspect `python` does this already automatically. I have noticed that the second time you import something in a python session is a lot faster than the first.

fubuloubu (dismissed):

src/ape/utils.py
fubuloubu: yeah, imports are cached

NotPeopling2day (approved):

NotPeopling2day (dismissed):
🔥

src/ape/managers/compilers.py
```diff
@@ -88,8 +89,17 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
...
         for extension in extensions:
-            paths_to_compile = [path for path in contract_filepaths if path.suffix == extension]
+
+            # Filter out in-source cache files from dependencies.
+            paths_to_compile = [
+                path
+                for path in contract_filepaths
+                if path.suffix == extension
+                and os.path.commonpath([cache_folder])
+                != os.path.commonpath([str(path), cache_folder])
```
fubuloubu, line 101: Is there a way to do this without `os`?

src/ape/managers/project.py
```diff
@@ -385,6 +386,10 @@ def _project(self) -> ProjectAPI:
 
         return self._cached_projects[self.path.name]
 
+    @property
+    def _in_source_cache_folder(self) -> Path:
+        return self.contracts_folder / ".cache"
```
fubuloubu, line 391: Would you have to do this with any dependencies? Should basically have everything that's needed in the manifest

Might need to test with recursive dependencies eventually

fubuloubu: Also maybe a better name:
```suggestion
    def _dependency_cache_folder(self) -> Path:
        return self.contracts_folder / ".cache"
```

fubuloubu (approved):

src/ape/managers/project.py
unparalleled-js: This is a cache for compilers to put in files that they need to use for compiling dependencies. The files have to be in the same path as the `contracts_folder`.

I don't want to confuse it with the dependency cache that is in `.ape/packages`

fubuloubu: Fair!

unparalleled-js: Solidity uses the manifests that are cached in `.ape/packages` to put them in this cache

fubuloubu (approved):

PR #548, opened 2022-03-07
fix: bug where ecosystem implementations failed to load when their network configs did not have default providers

unparalleled-js:
### What I did

fixes: #

### How I did it

Convert config to dictionary first to not use `__getattr__` and cause recursive errors and lack of explanation

### How to verify it

Remove `default_provider` from an ecosystem plugin. Things should still "work".

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -69,10 +69,10 @@ class NetworkManager(BaseManager):
                         continue
 
                     network_config = ecosystem_config[network_name]
-                    default_provider = network_config.get("default_provider")
-                    if not default_provider:
+                    if "default_provider" not in network_config:
                         continue
 
+                    default_provider = network_config["default_provider"]
                     if default_provider in network.providers:
                         network.set_default_provider(default_provider)
                     else:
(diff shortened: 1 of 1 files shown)


src/ape/managers/networks.py
```diff
@@ -54,8 +55,8 @@ def ecosystems(self) -> Dict[str, EcosystemAPI]:
                 data_folder=self.config_manager.DATA_FOLDER / plugin_name,
                 request_header=self.config_manager.REQUEST_HEADER,
             )
-            ecosystem_config = self.config_manager.get_config(plugin_name)
-            default_network = ecosystem_config.default_network  # type: ignore
+            ecosystem_config = self.config_manager.get_config(plugin_name).dict()
```
NotPeopling2day, line 58: I'm super good with this, unless we want to change the way `get_config` works to return the dict?

unparalleled-js: I think in some contexts, the class is nice, like within the plugin itself. But here in the managers, we can't assume anything about the class

NotPeopling2day (approved):

PR #549, opened 2022-03-08
refactor: move integration tests that use local blockchain to functional tests

unparalleled-js:
### What I did

The PR moves the tests that start up a local EthTester blockchain from the `tests/integration` directory to the `tests/functional` directory.

**Note**: This means all tests now in the `functional` directory run while connected to the `EthTester` blockchain.

This will the resolve the issues that @johnson2427 is having in his failing test on his query PR: https://github.com/ApeWorX/ape/pull/535

The reason those tests are failing is because we have conflicting `eth-tester` connections AND these tests have to be in the same directory.

Also, I don't agree with the changes in this PR. All the tests in `ape` are integration tests besides the convert-related ones and some of the ones that use mocks. No worries though what we call the directories; they just need to be in the same one.

### How I did it

Moved file and stuff

### How to verify it

Passing tests 🤞🏻 

### Checklist


fubuloubu (approved):
these changes seem appropiate. I think that these look a lot more like unit tests than where they were previously, and everything left under `integration/cli` looks a lot more like an integration test because it shows the UX of the command line

Issue #518, opened 2022-02-28
A 'ProcessMGMTProvider` Mixin


unparalleled-js:
### Overview

Currently, any process-mgmt related providers (`ape-geth` and `ape-hardhat` and soon to be others) require copying a lot of the same logic regarding process mgmt e.g. signaling, handling process killing, waiting for RPCs, etc.

### Specification

(very rough sketch)

```python
class ProcessMGMTProvider(ProviderAPI):
    _process

   @property
   @abstractmethod
   def command() -> List[str]
        ...

   def check_connection() -> bool:
       # Make RPC call, verify process started up right
       
    def connect():
        self._process = Popen(self.command)  # Except using the fancy retry logic we are copying around
        if not self.check_connection():
            self._process = None
            raise ProviderError("Nope")
    
    def disconnect():
        self._process.kill(death_signaling_hook)  # Use the fancy logic for exiting the process
        self._process = None
```

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

unparalleled-js:
Design update:

the logic in hardhat for managing multiple providers (such as in the cases of x-dist testing) is not applicable across the board for subprocess providers as it assumes you are doing the JSON-RPC approach of communicating with it.

However, this encompasses a lot of logic could be shared.

I wonder if it is overkill to do something like this:

```python
class SubprocessProvider(ProviderAPI):
    ...


class HTTPSubprocessProvider(SubprocessProvider):
    ...
```

PR #551, opened 2022-03-08
feat: subprocess provider mixin class

unparalleled-js:
### What I did

fixes: #518 

### How I did it

Mostly moved logic from `ape-hardhat` to here; here is the matching `ape-hardhat` PR: https://github.com/ApeWorX/ape-hardhat/pull/30.

Now, plugin developers will need less code to create providers based on subprocesses. So we can probably create a `ganache` one easier now. I am using this elsewhere too, as you may know!

### How to verify it

I tested out both hardhat providers and and ran the `ape-hardhat` test suite locally.
You can do something similar! Try using both hardhat provider in a script or the console or tests. Make sure it works as before.

### Checklist


src/ape/api/providers.py
```diff
@@ -719,3 +737,188 @@ def connection_str(self) -> str:
...
+    """
+
+    PROCESS_WAIT_TIMEOUT = 15
+    process: Optional[Popen] = None
+    is_stopping: bool = False
+
+    @property
+    @abstractmethod
+    def process_name(self) -> str:
+        """The name of the process, such as ``Hardhat node``."""
```
fubuloubu, line 754: Is this note correct?

fubuloubu: Or wait, is this just used for display purposes? or calling subprocess?

fubuloubu (approved):

src/ape/api/providers.py
unparalleled-js: Yes this is just for display purposes, like the log statements `"Starting blank"`.

Perhaps there is a better way? hmmmm

The actual command comes from the abstract method `build_command()`.
Both `Hardhat` providers use this hook as needed, like the `mainnet-fork` provider includes the `--fork` and `--block-number` flags.

unparalleled-js: Oh, and the first command in `ape-hardhat` is the `npx` bin path, so I can't find a reliable way to know the display name of a process based on its command list

fubuloubu: Okay, was confused for a sec lol

PR #555, opened 2022-03-10
feat: allow network choice filtering

unparalleled-js:
### What I did

Allows user to filter the network choices available. Helpful for limiting available networks per plugin CLI commands or user scripts who only care about a select few of options.

### How I did it
Made `network_choices` a method that accepts the filter strs

### How to verify it
`network_option()` still works and you can additionally filter out stuff via `ecosystem="fantom"` to only get `fantom` network options

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -133,8 +133,12 @@ class NetworkManager(BaseManager):
 
         return self.ecosystems[attr_name]
 
-    @property
-    def network_choices(self) -> Iterator[str]:
+    def get_network_choices(
+        self,
+        ecosystem_filter: Optional[str] = None,
+        network_filter: Optional[str] = None,
+        provider_filter: Optional[str] = None,
+    ) -> Iterator[str]:
         """
         The set of all possible network choices available as a "network selection"
         e.g. ``--network [ECOSYSTEM:NETWORK:PROVIDER]``.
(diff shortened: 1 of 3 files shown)


src/ape/managers/networks.py
```diff
@@ -151,27 +155,36 @@ def network_choices(self) -> Iterator[str]:
             Iterator[str]: An iterator over all the network-choice possibilities.
         """
         for ecosystem_name, ecosystem in self.ecosystems.items():
+            if ecosystem_filter and ecosystem_name != ecosystem_filter:
+                continue
+
```
NotPeopling2day: Might be thinking about this wrong, but what if they type it incorrectly (i.e. etthereum)? Would we want a default here? Maybe an error message?

unparalleled-js: - [ ] Validate ecosystem / network / provider names

src/ape/managers/networks.py
```diff
@@ -147,31 +190,84 @@ def network_choices(self) -> Iterator[str]:
...
             Iterator[str]: An iterator over all the network-choice possibilities.
         """
-        for ecosystem_name, ecosystem in self.ecosystems.items():
-            yield ecosystem_name
-            for network_name, network in ecosystem.networks.items():
-                if ecosystem_name == self.default_ecosystem.name:
-                    yield f":{network_name}"
 
-                yield f"{ecosystem_name}:{network_name}"
+        def _validate_filter(arg: Optional[Union[List[str], str]], options: Set[str]):
```
fubuloubu, line 205: Turn this into a staticmethod or function?

fubuloubu (commented):
Seems like a lot of work to core to add a feature to `ape_networks`'s CLI

Do you think it will be useful outside of `ape_networks`?

src/ape/managers/networks.py
unparalleled-js: > Seems like a lot of work to core to add a feature to `ape_networks`'s CLI

The majority of the work in core is done to support the filtering in `ape.cli.options.network_option()`

unparalleled-js: Alternatively, we can do the filtering for `network_option` within the `ape.cli` namespace.

`ape_networks` is only touching small changes in core.

fubuloubu (approved):
Chatted off-line, these changes will support other necessary use cases in plugins being developed

fubuloubu (approved):

fubuloubu (approved):

PR #561, opened 2022-03-12
fix: avoid issues with pip by making it a core dependency

fubuloubu:
### What I did
Should hopefully take care of a few issues that arise from not having the latest version of `pip` when installing

### How I did it
Added `pip` to core dependencies

### How to verify it
Validate plugin installs

### Checklist

- [x] ~~New test cases have been added~~

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,9 +15,7 @@ jobs:
               python-version: 3.8
 
         - name: Install Dependencies
-          run: |
-            python -m pip install --upgrade pip
-            pip install .[lint]
+          run: pip install .[lint]
 
         - name: Run Black
           run: black --check .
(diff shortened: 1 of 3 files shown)


NotPeopling2day (approved):

fubuloubu:
Not working ugh
https://github.com/ApeWorX/ape-fantom/actions/runs/1973927487

unparalleled-js (approved):

unparalleled-js:
I think this solution does not meet our needs.

Reason: We need to ensure pip is up-to-date _before_ running the `setup.py`. Otherwise, I think the install process uses the `pip` that was installed from the start and does not update in time.

unparalleled-js (changes requested):
I think we should probably close this; I don't think it will work :/
It was a nice idea though!

fubuloubu:
You're right, I don't think this would work correctly.

PR #566, opened 2022-03-14
feat: console extras

mikeshultz:
### What I did

Added the ability for a user to add a Python script (`ape_console_extras.py`) to their project directory (or their global `.ape` directory) to provide often used vars and utilities upon launch of the console.

### How I did it

Using [Python's importlib's `SourceFileLoader`](https://docs.python.org/3/library/importlib.html#importlib.machinery.SourceFileLoader) it loads a module, includes it's non-private symbols in the IPython namespace. If a function named `ape_init_extras()` exists, it will execute it given the namespace symbols as kwargs created for the console so far.

### How to verify it

[I'm not sure what this section means]

### Checklist


mikeshultz:
I'm trying to craft some tests for this but am unable to craft a negative (failing) test for `ape console`.   It's difficult to get IPython embedded console to exit with a non-zero code.  Here's an example of one but it ends up terminating the `pytest` process and not the `ape console` process (I suppose `runner.invoke()` does not actually execute another process).  

    @skip_projects(["geth"])
    @pytest.mark.parametrize("item", __all__)
    def test_console_assert(ape_cli, runner, item):
        result = runner.invoke(
            ape_cli, ["console"], input="import os; os.kill(os.getpid(), 15)", catch_exceptions=True
        )
        assert result.exit_code == 137

I don't think the existing console tests can fail, either.  Any tips?

docs/userguides/console.md
```diff
@@ -0,0 +1,51 @@
...
+Then both imported util functions and `WETH_ADDRESS` will be available when you launch the console.
+
+```python
+In [1]: decode_hex(WETH_ADDRESS)
+Out[1]: b"\xc0*\xaa9\xb2#\xfe\x8d\n\x0e\\O'\xea\xd9\x08<ul\xc2"
+```
+
+### Init Function
+
+If you include a function named `init_extras`, it will be executed with the symbols from the existing namespace being provided as keyword arguments.  This allows you to alter the scripts namespace using locals already included in the Ape namespace.  If you return a `dict`, these values will be added to the console namespace.  For example, you could setup an initialized Web3.py object by using one from an existing Ape Provider.
```
fubuloubu, line 35: Maybe `ape_init_extras` for function name, sort of like pytest hooks

fubuloubu (commented):
Does this work with both `ape test -I` and `ape run -I`?

docs/userguides/console.md
```diff
@@ -0,0 +1,51 @@
...
+Then `web3` will be available to use immediately.
+
+```python
+In [1]: web3.eth.chain_id
+Out[1]: 1
+```
+
+### Global Extras
+
+You can also add an `ape_console_extras.py` file to the global ape config directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
```
unparalleled-js, line 51:
```suggestion
You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
```

^ Just to not confuse config-file related stuff

src/ape_console/_cli.py
```diff
@@ -24,6 +32,57 @@ def cli(cli_ctx, network):
...
+
+
+def load_console_extras(namespace: dict[str, Any]) -> dict[str, Any]:
+    """load and return namespace updates from ape_console_extras.py  files if
+    they exist"""
+    global_extras = config.DATA_FOLDER.joinpath(CONSOLE_EXTRAS_FILENAME)
+    project_extras = config.PROJECT_FOLDER.joinpath(CONSOLE_EXTRAS_FILENAME)
+
+    for extras_file in [global_extras, project_extras]:
+        if extras_file.is_file():
```
unparalleled-js, line 55: To save an elongated indent, you can do this:

```suggestion
        if not extras_file.is_file():
            continue
```

and then de-indent everything below.

unparalleled-js, line 55: Are we using `is_file` just to check if it exists? Maybe we should just use the `exists()` method since the intent is more obvious, and if it existed and was not a file, maybe that should be an error anyways.

src/ape_console/_cli.py
```diff
@@ -24,6 +32,57 @@ def cli(cli_ctx, network):
...
+                # from the original namespace
+                func_spec = inspect.getfullargspec(init_extras)
+                init_kwargs: Dict[str, Any] = {k: namespace.get(k) for k in func_spec.args}
+
+                # Execute functionality with existing console namespace as
+                # kwargs.
+                extras = init_extras(**init_kwargs)
+
+                # If init_extras returned a dict expect it to be new symbols
+                if type(extras) == dict:
```
unparalleled-js, line 71:
```suggestion
                if isinstance(extras, dict):
```

This will make it so subclasses of `dict` are also supported

docs/userguides/console.md
mikeshultz: If you'd like, can do

mikeshultz:
> As far as your testing question goes, what error conditions are you trying to replicate? When errors happen in the console, they just get printed there and it does not cause the console to break necessarily.

Right now, I'm trying to create *any* failing test for `ape console` CLI.  I can't get anything to fail.  I don't think the existing test can fail at all, even if the expected locals from `__all__` are missing.  I feel like I tried everything yesterday but couldn't get it to fail.

src/ape_console/_cli.py
mikeshultz: `is_file` is more explicit because if it were a directory things would get unexpectedly wonky.

src/ape_console/_cli.py
```diff
@@ -24,6 +32,57 @@ def cli(cli_ctx, network):
...
+    assert spec is not None
+
+    module = module_from_spec(spec)
+    loader.exec_module(module)
+
+    return module
+
+
+def load_console_extras(namespace: dict[str, Any]) -> dict[str, Any]:
+    """load and return namespace updates from ape_console_extras.py  files if
```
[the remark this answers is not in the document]
mikeshultz: Double tick?

docs/userguides/console.md
```diff
@@ -0,0 +1,51 @@
...
+## Namespace Extras
+
+You can also create scripts to be included in the console namespace by adding a file (`ape_console_extras.py`) to your root project directory.  All non-private symbols from this file will be included in the console namespace.  Private symbols are prefixed by an underscore (`_`).
+
+An example file might look something like this:
+
+```python
+from eth_utils import encode_hex, decode_hex
+
+WETH_ADDRESS = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
```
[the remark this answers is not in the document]
mikeshultz: I thought it was a good generic example but happy for alternatives.  I also thought maybe store a comon ABI, but that is excessively long for an example.  I'll mull it over but if you have any suggestions, lmk.

docs/userguides/console.md
```diff
@@ -0,0 +1,51 @@
...
+
+In [1]: weth = Contract(address='0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', contract_type=project.contracts['ERC20'])
+
+In [2]: weth.totalSupply()
+Out[2]: 7065522116676615294347037
+```
+
+## Namespace Extras
+
+You can also create scripts to be included in the console namespace by adding a file (`ape_console_extras.py`) to your root project directory.  All non-private symbols from this file will be included in the console namespace.  Private symbols are prefixed by an underscore (`_`).
```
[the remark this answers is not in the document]
mikeshultz: Happy to go with internal since it's more relevant, but the exclusion would also include private.

src/ape_console/_cli.py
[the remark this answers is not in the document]
unparalleled-js: I suppose this is not a public method so never mind that.

The double ticks are for sphinx

src/ape_console/_cli.py
unparalleled-js: Ya at first i thought we should just raise an error in that situation... but I changed my mind! Let's let it be.

unparalleled-js:
> Right now, I'm trying to create _any_ failing test for `ape console` CLI

The output from `IPython` is present in `result.output`.
Rather than trying to get it to raise an exception, you can verify the exception's output is present in the `result.output`.

e.g., with the command `input=my_thing_that_does_not_exist; exit()` (or whatever it is)
and then do `assert "NameError: name 'my_thing_that_does_not_exist' is not defined" in result.output`

mikeshultz:
> Does this work with both `ape test -I` and `ape run -I`?

@fubuloubu  Woops, missed this one.  I did not consider that other things might use this console.  Looks like `ape run -I` runs `console()` directly.  I'm not sure I have an opinion on this.  On one hand, I don't personally have a use case for it so I have a hard time imagining it's usefulness.  On the other, it's probably safe if these extras are included in interactive consoles as well since overwrites are prevented.  Anyone else have opinions on this?

> The output from `IPython` is present in `result.output`.
> Rather than trying to get it to raise an exception, you can verify the exception's output is present in the `result.output`.

That's a good idea, I'll see if I can get that to work.

mikeshultz:
> > > Does this work with both `ape test -I` and `ape run -I`?
> > 
> > 
> > @fubuloubu Woops, missed this one. I did not consider that other things might use this console. Looks like `ape run -I` runs `console()` directly. I'm not sure I have an opinion on this. On one hand, I don't personally have a use case for it so I have a hard time imagining it's usefulness. On the other, it's probably safe if these extras are included in interactive consoles as well since overwrites are prevented. Anyone else have opinions on this?
> 
> A lot of people tend to use this to launch into terminal sessions and debug their scripts and tests. I'd say it gets a fair amount of use.

Do you think these extras should be included in those console instances or excluded?

fubuloubu:
> > > > Does this work with both `ape test -I` and `ape run -I`?
> > > 
> > > 
> > > @fubuloubu Woops, missed this one. I did not consider that other things might use this console. Looks like `ape run -I` runs `console()` directly. I'm not sure I have an opinion on this. On one hand, I don't personally have a use case for it so I have a hard time imagining it's usefulness. On the other, it's probably safe if these extras are included in interactive consoles as well since overwrites are prevented. Anyone else have opinions on this?
> > 
> > 
> > A lot of people tend to use this to launch into terminal sessions and debug their scripts and tests. I'd say it gets a fair amount of use.
> 
> Do you think these extras should be included in those console instances or excluded?

Ah, well it would be nice for it to have the same behavior everywhere the console is made availability, but not sure I'd call that a strict requirement if it is very difficult to figure out.

mikeshultz:
> > > > > Does this work with both `ape test -I` and `ape run -I`?
> > > > 
> > > > 
> > > > @fubuloubu Woops, missed this one. I did not consider that other things might use this console. Looks like `ape run -I` runs `console()` directly. I'm not sure I have an opinion on this. On one hand, I don't personally have a use case for it so I have a hard time imagining it's usefulness. On the other, it's probably safe if these extras are included in interactive consoles as well since overwrites are prevented. Anyone else have opinions on this?
> > > 
> > > 
> > > A lot of people tend to use this to launch into terminal sessions and debug their scripts and tests. I'd say it gets a fair amount of use.
> > 
> > 
> > Do you think these extras should be included in those console instances or excluded?
> 
> Ah, well it would be nice for it to have the same behavior everywhere the console is made availability, but not sure I'd call that a strict requirement if it is very difficult to figure out.

I think as-is these extras will be included in the `ape run -I` console since it just calls `console()` straight up.  Not sure about `ape test -I` though since it doesn't seem to invoke the Ape console.  I'm unsure of the scope of work for that.  I'll take a brief pass to see what that might entail.

fubuloubu:
> > > > > > Does this work with both `ape test -I` and `ape run -I`?
> > > > > 
> > > > > 
> > > > > @fubuloubu Woops, missed this one. I did not consider that other things might use this console. Looks like `ape run -I` runs `console()` directly. I'm not sure I have an opinion on this. On one hand, I don't personally have a use case for it so I have a hard time imagining it's usefulness. On the other, it's probably safe if these extras are included in interactive consoles as well since overwrites are prevented. Anyone else have opinions on this?
> > > > 
> > > > 
> > > > A lot of people tend to use this to launch into terminal sessions and debug their scripts and tests. I'd say it gets a fair amount of use.
> > > 
> > > 
> > > Do you think these extras should be included in those console instances or excluded?
> > 
> > 
> > Ah, well it would be nice for it to have the same behavior everywhere the console is made availability, but not sure I'd call that a strict requirement if it is very difficult to figure out.
> 
> I think as-is these extras will be included in the `ape run -I` console since it just calls `console()` straight up. Not sure about `ape test -I` though since it doesn't seem to invoke the Ape console. I'm unsure of the scope of work for that. I'll take a brief pass to see what that might entail.

It uses the same setup (launching `console` here w/ any extras from the test), so if it works for `ape run -I` it should be good for `ape test -I`

mikeshultz:
> I am mostly interested in tests that make sure the namespaces get updates correctly. You can just add a new project in `tests/cli/integration/projects/` maybe named `console-scripts` or something. Testing can be a little weird, as your tests will run for every project by default but there are decorators you can use to make it only run for your new project. I am happy to assist there!

@unparalleled-js  I was thinking about just writing an `ape_console_extras.py` file into the project for each test.  That would allow me to create different files to test different things (e.g. with the ape_init_extras func, with locals, global file, etc).  

Also, since the existing test is parameterized for all projects, I assume that was important.  Do you think there's more value to creating project(s) just for the extras test versus testing all projects?

unparalleled-js:
> @unparalleled-js I was thinking about just writing an `ape_console_extras.py` file into the project for each test. That would allow me to create different files to test different things (e.g. with the ape_init_extras func, with locals, global file, etc).

I am good with this!

mikeshultz:
CI should go green now, and I think I addressed all the feedback up until now.  Take a look again when you get a chance, thanks for the helpful reviews.

docs/userguides/console.md
```diff
@@ -0,0 +1,62 @@
+# Ape Console
+
+Ape provides an [IPython](https://ipython.readthedocs.io/) interactive console with useful pre-defined locals to interact with your project.
+
+```bash
+ape console --network ethereum:mainnet
+
+In [1]: networks.active_provider.get_block("latest")
```
fubuloubu, line 8: Can you use a different call here? The structure of the block is something we are working on, it might change and we don't want to have to update the block struct here

docs/userguides/console.md
```diff
@@ -0,0 +1,62 @@
...
+_networks = None
+
+
+def latest(key="number"):
+    return getattr(_networks.active_provider.get_block("latest"), key)
+
+
+def ape_init_extras(networks):
+    """Setup _networks so util func can access it"""
+    global _networks
```
fubuloubu, line 30: Is using a global something you've been looking to do?

mikeshultz: It was intentional.  It allows the function to alter `_networks` that gets included in the namespace.  

It could be argued this is questionable style and could be structured differently to return a func in the dict from `ape_init_extras()` but I wanted to include this as an example of altering symbols within the file.

docs/userguides/console.md
mikeshultz: Do you have a recommendation?  I'm still familiarizing myself with the Ape API.

fubuloubu:
```py
chain.blocks.head.timestamp
```

docs/userguides/console.md
fubuloubu: I think it might be pretty advanced for an example. These docs are going to be referenced by a lot of newcomers, so I'd like to use simpler things as examples, like maybe importing a popular dependency like matplotlib or Decimal

mikeshultz: I am not at all familiar with matplotlib.  You do have me worried I'm writing for the wrong audience though.  I can simplify my example, but I'm not sure I can write for other types of users very well.  Simplifying this example might look like:

    from eth_utils import encode_hex, decode_hex


    def ape_init_extras(networks):
        """Setup _networks so util func can access it"""
        return {
            "latest": lambda key: getattr(networks.active_provider.get_block("latest"), key)
        }

I could create another example to parse token transfer logs and drop value of Decimal but that's probably equally not simple (`print('Hello, World')`?).  I'll have to mull it over but open to suggestions.

tests/integration/cli/test_console.py
```diff
@@ -4,13 +4,99 @@
...
 
 
+def no_console_error(result):
+    return "NameError" not in result.output and "AssertionError" not in result.output
+
+
+def destroy_ape_console_extras(project):
+    try:
+        project.config_manager.DATA_FOLDER.joinpath("ape_console_extras.py").unlink()
+    except FileNotFoundError:
```
unparalleled-js, line 14: You could jut check `is_file()` first rather than doing a whole try and catch (twice)

unparalleled-js (commented):
I am going to test this out today

unparalleled-js:
Testing I have done:

- [ ] project extras
- [ ] global extras
- [ ] project extras take precedence over global extras
- [ ] extras are available in `ape run -I <script>`
- [ ] extras are available in `ape test -I` when an exception occurs

tests/integration/cli/test_console.py
mikeshultz: Works for me

tests/integration/cli/test_console.py
```diff
@@ -4,13 +4,99 @@
...
         ape_cli, ["console", "-v", "debug"], input=f"{item}\nexit\n", catch_exceptions=False
     )
     assert result.exit_code == 0, result.output
+    assert no_console_error(result), result.output
+
+
+@skip_projects(["geth"])
+@pytest.mark.parametrize("folder", ["PROJECT_FOLDER", "DATA_FOLDER"])
+def test_console_extras(project, folder, ape_cli, runner):
+    destroy_ape_console_extras(project)
```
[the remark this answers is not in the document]
mikeshultz: Good call, adding

mikeshultz:
I don't know what I'm supposed to do about the `commit-msg` check failing.  That wasn't in the `CONTRIBUTING.md` doc and seems a bit too late now.

fubuloubu:
> I don't know what I'm supposed to do about the `commit-msg` check failing. That wasn't in the `CONTRIBUTING.md` doc and seems a bit too late now.

No worries about that, it only runs on branches, not PRs

docs/userguides/console.md
```diff
@@ -0,0 +1,56 @@
...
+Then `web3` will be available to use immediately.
+
+```python
+In [1]: web3.eth.chain_id
+Out[1]: 1
+```
+
+### Global Extras
+
+You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
```
unparalleled-js, line 56:
```suggestion
You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of what project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
```

unparalleled-js (approved):
Looks amazing! Is there anything left on this? I had one super minor docs comment

mikeshultz:
Not sure what it means that I "dismissed" your review.  First time I've ever seen that on GH.  Maybe happened when I updated from upstream/main?

But I'm done from my end, if you're all happy with it feel free to merge.  Thanks!

unparalleled-js:
> Not sure what it means that I "dismissed" your review. First time I've ever seen that on GH. Maybe happened when I updated from upstream/main?

It is a rule we have setup on this repo!

Thanks :) We will merge this in soon

unparalleled-js (approved):
LGTM

docs/userguides/console.md
```diff
@@ -0,0 +1,56 @@
...
+An example file might look something like this:
+
+```python
+from eth_utils import encode_hex, decode_hex
+
+
+def ape_init_extras(networks):
+    return {
+        "latest": lambda key: getattr(networks.active_provider.get_block("latest"), key)
+    }
```
fubuloubu, line 25: Actually, in keeping with the explanation above, this doesn't mention `ape_init_extras` yet, so do it like so
```suggestion
latest = lambda key: getattr(networks.active_provider.get_block("latest"), key)
```

fubuloubu (approved):
Thanks for this feature, I really like it!

docs/userguides/console.md
mikeshultz: Agreed

docs/userguides/console.md
```diff
@@ -19,10 +19,7 @@ An example file might look something like this:
 from eth_utils import encode_hex, decode_hex
 
 
-def ape_init_extras(networks):
-    return {
-        "latest": lambda key: getattr(networks.active_provider.get_block("latest"), key)
-    }
+latest = lambda key: getattr(networks.active_provider.get_block("latest"), key)
```
unparalleled-js, line 22: This actually causes issues with the linters, if they were to run.
anonymous methods are not meant to have name, so when you do `latest = <anon method>`, you might as well just do 

```python
def latest(key):
   return getattr(networks.active_provider.get_block("latest"), key)
```

mikeshultz: Far be it from me to argue with the linters.  Do you mind making that change?  I won't be able to until this evening.

docs/userguides/console.md
```diff
@@ -0,0 +1,53 @@
...
+
+You can also create scripts to be included in the console namespace by adding a file (`ape_console_extras.py`) to your root project directory.  All non-internal symbols from this file will be included in the console namespace.  Internal symbols are prefixed by an underscore (`_`).
+
+An example file might look something like this:
+
+```python
+from eth_utils import encode_hex, decode_hex
+
+
+latest = lambda key: getattr(networks.active_provider.get_block("latest"), key)
```
unparalleled-js, line 22:
```suggestion
def latest(key):
    return getattr(networks.active_provider.get_block("latest"), key)
```

fubuloubu (approved):

PR #575, opened 2022-03-14
fix: only run actions once

fubuloubu:


fubuloubu:
Didn't work the way I expected

PR #576, opened 2022-03-14
feat: support configuring default ecosystem

unparalleled-js:
### What I did

Allows user to config a default ecosystem.

### How I did it

* Add new item to config
* Set it as the default
* Also called `load()` before trying to get a config because of a "race" condition.

### How to verify it

* Create a project using an ecosystem that is not Ethereum
* Add this to config: `default_ecosystem: the-ecosystem` replacing `the-ecosystem` with whatever ecosystem you are using.
* Run `ape console` and make sure it uses your default ecosystem

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -4,7 +4,7 @@ import yaml
 
 from ape.api import EcosystemAPI, ProviderAPI, ProviderContextManager
 from ape.api.networks import LOCAL_NETWORK_NAME
-from ape.exceptions import ConfigError, NetworkError
+from ape.exceptions import NetworkError
 
 from .base import BaseManager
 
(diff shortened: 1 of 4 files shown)


src/ape/managers/networks.py
```diff
@@ -97,10 +98,7 @@ def ecosystems(self) -> Dict[str, EcosystemAPI]:
             ecosystem_config = self.config_manager.get_config(plugin_name).dict()
             default_network = ecosystem_config.get("default_network", LOCAL_NETWORK_NAME)
 
-            if default_network and default_network in ecosystem.networks:
```
unparalleled-js, line 100: I removed these parts because we already raise custom errors in the `set_default_()` methods

fubuloubu (approved):

Issue #579, opened 2022-03-15
Non-Ethereum addresses do not work in to_address


unparalleled-js:
### Environment information

ape v0.1.0
python 3.7.12
macOS

### What went wrong?

When using a "custom" Ecosystem plugin for a non-Ethereum network, addresses do not work because of the `ape.utils.to_address()` method.

```
  File "**/python3.7/site-packages/eth_utils/address.py", line 64, in to_normalized_address
    "Unknown format {}, attempted to normalize to {}".format(value, hex_address)
ValueError: Unknown format 1059844984421814259306102818863337139307291816605262763186519932397694697550, attempted to normalize to 0x257d9db5994c162fe30c4ac81e53956ee7df2dd6f08b9051c60b4e60357404e
```

### How can it be fixed?

1. One idea that is mehh

Catch exceptions and just return the given value in `to_address()` if it fails.

I am still thinking on this one.. It has become my primary priority.

PR #580, opened 2022-03-15
fix: handle non-length 42 addresses in to_address utility method

unparalleled-js:
### What I did

fixes: #579 

### How I did it

* Copy code from `eth_utils.to_checksum_address()` method
* Replace part that assumes the length is 42 with the length of `hex_address` that gets created.

### How to verify it

See tests!

Or, install an ecosystem plugin that does not have addresses with this length and make sure you can do stuff like `ape accounts list --all` and they show up.

### Checklist

### DIFF
--- a/src/ape/utils.py
+++ b/src/ape/utils.py
@@ -10,14 +10,15 @@ from collections import namedtuple
 from functools import lru_cache
 from io import BytesIO
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Mapping, Optional, Set, cast
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Mapping, Optional, Set, Union, cast
 
 import pygit2  # type: ignore
 import requests
 import yaml
 from eth_account import Account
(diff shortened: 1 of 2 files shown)


fubuloubu:
So, it would be helpful to understand the address encoding format for an `EcosystemAPI` plugin. Ethereum defines one, the checksum hex format, but many non-EVM chains do not use the same format. Even the Ethereum 2.0 Beacon chain uses a different format.

We probably have to be a bit more abstract about converting "raw addresses" to/from their string representations, perhaps through an `abstractmethod` on `EcosystemAPI`

unparalleled-js:
> So, it would be helpful to understand the address encoding format for an `EcosystemAPI` plugin. Ethereum defines one, the checksum hex format, but many non-EVM chains do not use the same format. Even the Ethereum 2.0 Beacon chain uses a different format.
> 
> We probably have to be a bit more abstract about converting "raw addresses" to/from their string representations, perhaps through an `abstractmethod` on `EcosystemAPI`

I will create a new branch for this and close this one for now

Issue #569, opened 2022-03-14
Add `timedelta` to `chain.mine`


fubuloubu:
Allow `timedelta` keword arg, which is mutually exclusive with `timestamp` kwarg, that sets `chain.pending_timestamp += timedelta` and then mines a block

https://github.com/ApeWorX/ape/blob/504862f334074517efdb48a4e76251f506c119b6/src/ape/managers/chain.py#L485

PR #585, opened 2022-03-17
feat: add deltatime kwarg to chain.mine

johnson2427:
### What I did
added deltatime capability to chain.ChainManager.mine

fixes: #569

### How I did it
Added the deltatime kwarg to ChainManager.mine

### How to verify it
Create test (will do following some back and forth) and test the new mining capability

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -483,7 +483,12 @@ class ChainManager(BaseManager):
         self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
 
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None, deltatime: Optional[timedelta] = None) -> None:
+    def mine(
+        self,
+        num_blocks: int = 1,
+        timestamp: Optional[int] = None,
+        deltatime: Optional[timedelta] = None,
+    ) -> None:
         if timestamp:
             self.pending_timestamp = timestamp
         if deltatime:
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -3,7 +3,6 @@ from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 import pandas as pd
 
-from ape import convert
 from ape.api import BlockAPI, ReceiptAPI
 from ape.api.address import BaseAddress
 from ape.api.query import BlockQuery
@@ -492,5 +491,5 @@ class ChainManager(BaseManager):
         if timestamp:
             self.pending_timestamp = timestamp
         elif deltatime:
-            self.pending_timestamp = convert(deltatime, int)
+            self.pending_timestamp = deltatime
         self.provider.mine(num_blocks)
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -1,4 +1,3 @@
-from datetime import timedelta
 import time
 from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
 
@@ -487,7 +486,7 @@ class ChainManager(BaseManager):
         self,
         num_blocks: int = 1,
         timestamp: Optional[int] = None,
-        deltatime: Optional[timedelta] = None,
+        deltatime: Optional[int] = None,
     ) -> None:
         if timestamp:
             self.pending_timestamp = timestamp
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -3,6 +3,7 @@ from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 import pandas as pd
 
+from ape import convert
 from ape.api import BlockAPI, ReceiptAPI
 from ape.api.address import BaseAddress
 from ape.api.query import BlockQuery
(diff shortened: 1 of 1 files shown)


johnson2427:
I probably should have left this issue. I thought it was more complicated at first, then realized we have a converter for this, so it was really much simpler than I previously thought

NotPeopling2day:
Yeah, maybe try changing the typing and let pydantic handle it for you?

johnson2427:
@NotPeopling2day can you give me an example of what you mean

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
...
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None) -> None:
+    def mine(
+        self,
+        num_blocks: int = 1,
+        timestamp: Optional[int] = None,
+        deltatime: Optional[int] = None,
+    ) -> None:
         if timestamp:
             self.pending_timestamp = timestamp
+        if deltatime:
```
unparalleled-js, line 493: are both `timstamp` and `deltatime` arguments allowed together?

johnson2427: needs to be elif, and we need to choose which overrides the other

NotPeopling2day:
> @NotPeopling2day can you give me an example of what you mean

Apologies, when I quickly looked at this I thought it was a part of the BlockAPI we were working on yesterday.

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
...
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None) -> None:
+    def mine(
+        self,
+        num_blocks: int = 1,
+        timestamp: Optional[int] = None,
+        deltatime: Optional[int] = None,
+    ) -> None:
         if timestamp:
             self.pending_timestamp = timestamp
+        elif deltatime:
```
unparalleled-js, line 493: I think it would be confusing if the user passes in both and the method succeeds but silently ignored one of the parameters.

I think we raise a `ValueError` if both params are present at the same time

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
         self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
 
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None) -> None:
+    def mine(
```
unparalleled-js, line 485: Can we get docs for this method?

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
...
+        self,
+        num_blocks: int = 1,
+        timestamp: Optional[int] = None,
+        deltatime: Optional[int] = None,
+    ) -> None:
```
unparalleled-js, line 490: Can we add a test or 2?

unparalleled-js (commented):
Hi!

* Requesting that we raise `ValueError`
* Requests docs + tests

thank you :)

src/ape/managers/chain.py
fubuloubu: first case should be `if timestamp and deltatime: raise`

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
         self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
 
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None) -> None:
+    def mine(
+        self,
+        num_blocks: int = 1,
+        timestamp: Optional[int] = None,
+        deltatime: Optional[int] = None,
```
fubuloubu, line 489: can we call it `timedelta`? Does that still read well?
```suggestion
        timedelta: Optional[int] = None,
```

src/ape/managers/chain.py
```diff
@@ -482,7 +482,14 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
         self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
 
-    def mine(self, num_blocks: int = 1, timestamp: Optional[int] = None) -> None:
```
fubuloubu, line 485: Yeah, this method should have had a docstring before, good time to add it!

src/ape/managers/chain.py
johnson2427: I didn't use timedelta because it is a class within datetime, and I don't want any confusion

src/ape/managers/chain.py
johnson2427: I figured, I'll add the docstring after I get back from the kids soccer practice

src/ape/managers/chain.py
johnson2427: I was waiting to get this to pass here, there was a connection issue so I had to run a few times before it did.

src/ape/managers/chain.py
johnson2427: I think we should leave it as deltatime, but you make the last decision on that

fubuloubu (approved):

unparalleled-js:
It would be nice to have some context on the `ValueError` as well.
It looks like it is just raising `ValueError` with no message, so users may not know what has gone away

Issue #588, opened 2022-03-18
Add tests for chain.ChainManager.mine


johnson2427:
### Overview
We merged some changes to mine without testing

### Specification
```python
import pytest


def test_mine():
    ...
```

fubuloubu:
My bad for merging #569 prematurely

unparalleled-js:
Another comment: https://github.com/ApeWorX/ape/pull/585#issuecomment-1072552562

PR #589, opened 2022-03-18
Added a test for chain.ChainManager.mine

johnson2427:
### What I did
Added a test to chain.ChainManager.mine.

** will do this as I build changes from now on **

fixes: #588

### How I did it
chain.ChainManager.mine
```python
self.pending_timestamp += deltatime
```

Test added:
```python
def test_set_pending_timestamp_with_deltatime(chain):
    start_timestamp = chain.pending_timestamp
    chain.mine(deltatime=5)
    new_timestamp = chain.pending_timestamp
    assert new_timestamp - start_timestamp - 5 <= 1
```

### How to verify it
The test should verify this

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -483,7 +483,12 @@ class ChainManager(BaseManager):
         self.provider.revert(snapshot_id)
         self.account_history.revert_to_block(self.blocks.height)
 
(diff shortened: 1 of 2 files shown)


unparalleled-js (changes requested):
Thanks for coming back and adding a test :)

Can we also get a negative test and error message adjustment!
(see other comment)

unparalleled-js (approved):
LGTM!

There is a missing positive test for just setting timestamp (and not timedelta),
but I am good with this!

Thanks!

Issue #592, opened 2022-03-19
`parent_hash` field is set incorrectly in Ethereum ecosystem


fubuloubu:
https://github.com/ApeWorX/ape/blob/202596d88a7d69f01bc6f3a34442ca0d5e40e162/src/ape_ethereum/ecosystem.py#L215

PR #593, opened 2022-03-19
fix: parent_hash field is set incorrectly

fubuloubu:
### What I did
Noticed that Ethereum blocks had the same value set for both `hash` and `parent_hash`, so I fixed it

fixes: #592


fubuloubu:
Maybe add a test showing that `block(0).hash == block(1).parent_hash`

unparalleled-js (approved):
Good catch!

Issue #595, opened 2022-03-21
Leverage query manager to complete historical block range queries


johnson2427:
### Elevator pitch:
The query provider system will intelligently find the fastest way to provide data for historical queries, often much faster than querying a provider directly. 

### Value:
Allows for querying of block data and fetching it much faster when alternative data sources are connected vs. calling the provider directly. 

### Dependencies:
None

### Design approach:
Pull the logic from the `ChainManager.blocks.range` function and replace with calls to `QueryManager` instead. Move the previous logic to `DefaultQueryProvider` so that by default it will query the provider if no faster method of providing this data is found. 

### Task list:

* [ ] Remove range functionality from chain_manager
* [ ] Put that functionality into query_manager (confirm this)

### Estimated completion date:
March 22nd, 2022

### Design review:

(Please leave a comment to sign off)

fubuloubu:
Approved! (with edits)

PR #599, opened 2022-03-22
refactor: pulled logic from ChainManager.blocks.range to DefaultQueryProvider

johnson2427:
### What I did

Pull the range longic from ChainManager.blocks.range and place that functionality in DefaultQueryProvider

fixes: #595

### How I did it
Grabbed the functionality from ChainManager.blocks.range (which was in BlockContainer) and pulled it to DefaultQueryProvider

### How to verify it
Currently, the only tests that I have ran are:
```bash
ape console
```

```python
a = accounts.test_accounts[0]
a.transfer(a, 100)
a.transfer(a, 100)
a.transfer(a, 100)
chain.blocks.query("number", "timestamp")
```

```
   number   timestamp  size
0       0  1647968057   517
1       1  1647986057   627
2       2  1647986154   627
```

I also ran pytest test_basic_query(), which passed

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -134,6 +134,57 @@ class BlockContainer(BaseManager):
 
         return self.query_manager.query(query)
 
+    def range(
+        self, start_or_stop: int, stop: Optional[int] = None, step: int = 1
+    ) -> Iterator[BlockAPI]:
+        """
+        Iterate over blocks. Works similarly to python ``range()``.
+
+        Raises:
+            :class:`~ape.exceptions.ChainError`: When ``stop`` is greater
+                than the chain length.
+            :class:`~ape.exceptions.ChainError`: When ``stop`` is less
+                than ``start_block``.
+            :class:`~ape.exceptions.ChainError`: When ``stop`` is less
+                than 0.
+            :class:`~ape.exceptions.ChainError`: When ``start`` is less
+                than 0.
+
+        Args:
+            start_or_stop (int): When given just a single value, it is the stop.
+              Otherwise, it is the start. This mimics the behavior of ``range``
+              built-in Python function.
+            stop (Optional[int]): The block number to stop before. Also the total
+              number of blocks to get. If not setting a start value, is set by
+              the first argument.
+            step (Optional[int]): The value to increment by. Defaults to ``1``.
+             number of blocks to get. Defaults to the latest block.
+
+        Returns:
+            Iterator[:class:`~ape.api.providers.BlockAPI`]
+        """
+
+        if stop is None:
+            stop = start_or_stop
+            start = 0
+        else:
+            start = start_or_stop
+
+        if stop > len(self):
+            raise ChainError(
+                f"'stop={stop}' cannot be greater than the chain length ({len(self)}). "
+                f"Use '{self.poll_blocks.__name__}()' to wait for future blocks."
+            )
+        elif stop < start:
+            raise ValueError(f"stop '{stop}' cannot be less than start '{start}'.")
+        elif stop < 0:
+            raise ValueError(f"start '{start}' cannot be negative.")
+        elif start_or_stop < 0:
+            raise ValueError(f"stop '{stop}' cannot be negative.")
+
+        for i in range(start, stop, step):
+            yield self._get_block(i)
+
     def poll_blocks(
         self,
         start: Optional[int] = None,
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -134,9 +134,7 @@ class BlockContainer(BaseManager):
 
         return self.query_manager.query(query)
 
-    def range(
-        self, start_or_stop: int, step: int = 1
-    ) -> Iterator[BlockAPI]:
+    def range(self, start_or_stop: int, step: int = 1) -> Iterator[BlockAPI]:
         return self.query_manager.get_blocks(start_or_stop, stop=None, step=step)
 
     def poll_blocks(
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -137,53 +137,7 @@ class BlockContainer(BaseManager):
     def range(
         self, start_or_stop: int, stop: Optional[int] = None, step: int = 1
     ) -> Iterator[BlockAPI]:
-        """
-        Iterate over blocks. Works similarly to python ``range()``.
-
-        Raises:
-            :class:`~ape.exceptions.ChainError`: When ``stop`` is greater
-                than the chain length.
-            :class:`~ape.exceptions.ChainError`: When ``stop`` is less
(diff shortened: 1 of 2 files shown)


src/ape/managers/query.py
```diff
@@ -38,10 +38,61 @@ def perform_query(self, query: QueryType) -> pd.DataFrame:
 
     @perform_query.register
     def perform_block_query(self, query: BlockQuery) -> pd.DataFrame:
-        blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
+        blocks_iter = self.range(query.start_block, query.stop_block)
         block_dicts_iter = map(partial(get_columns_from_item, query), blocks_iter)
         return pd.DataFrame(columns=query.columns, data=block_dicts_iter)
 
+    def range(
```
unparalleled-js, line 45: I think the name of this method has to change if we do this...

`DefaultQueryProvider.range()` does not indicate it returns blocks necessarily.

src/ape/managers/chain.py
```diff
@@ -134,57 +134,6 @@ def query(
 
         return self.query_manager.query(query)
 
-    def range(
```
unparalleled-js, line 137: It is imperative we delete this method?
Could we instead just to delegate to the query manager?

unparalleled-js (changes requested):
Tests will be broken now from this I image.
I think we should keep the method and just delegate to the query provider.
Also we need to rename the method on the query provider so it makes more sense.

src/ape/managers/chain.py
johnson2427: not sure what you mean by delegate, I can leave range, and then pull the functionality to the DefaultQueryProvider, and then rename the function

johnson2427:
I can return that range function to ChainManager

then add this:

```python
    def get_blocks(self, start_or_stop: int, stop: Optional[int] = None) -> Iterator[BlockAPI]:
        return self.chain_manager.blocks.range(start_or_stop, stop)
```

src/ape/managers/chain.py
unparalleled-js: By delegate, I just mean call it from the query provider

src/ape/managers/query.py
```diff
@@ -38,10 +38,13 @@ def perform_query(self, query: QueryType) -> pd.DataFrame:
 
     @perform_query.register
     def perform_block_query(self, query: BlockQuery) -> pd.DataFrame:
-        blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
+        blocks_iter = self.get_blocks(query.start_block, query.stop_block)
         block_dicts_iter = map(partial(get_columns_from_item, query), blocks_iter)
         return pd.DataFrame(columns=query.columns, data=block_dicts_iter)
 
+    def get_blocks(self, start_or_stop: int, stop: Optional[int] = None) -> Iterator[BlockAPI]:
+        return self.chain_manager.blocks.range(start_or_stop, stop)
```
unparalleled-js, line 46: I think this should be the other way? Do it the same way you had before, but instead of deleting the method in the `BlockContainer`, call the query method

johnson2427: lololol

johnson2427:
I need help fixing chain.BlockContainer.range
I'm not sure why I can't call the query.get_blocks function at the moment, figured other eyes would quicken this up

src/ape/managers/chain.py
fubuloubu: This was the original intent, yes. Perform a block query with all the fields selected, and then serialize it into a set of blocks and yield those.

fubuloubu:
> I need help fixing chain.BlockContainer.range
> 
> Should I move that get_blocks function from DefaultQueryProvider to QueryManager? Or import DefaultQueryProvider?

No, so the functions you are replacing under `BlockContainer` should delegated to the `QueryManager.query(query=BlockQuery(...))` method, and then take the result and serialize it back to the `BlockAPI` subclass that the current ecosystem is using

unparalleled-js (dismissed):
LGTM

src/ape/managers/query.py
```diff
@@ -114,3 +114,55 @@ def query(self, query: QueryType, engine_to_use: Optional[str] = None) -> pd.Dat
             engine.update_cache(query, result)
 
         return result
+
+    def get_blocks(
```
unparalleled-js, line 118: Actually... I would prefer this was different and that `stop_block` was the actual last block you want to get. And then the `range()` implementation on the `BlockContainer` was deal with that accordingly...

johnson2427: I believe stop_block is the last block you want to get, it defaults to the last block of the chain, but you can call for a block you want

unparalleled-js: If you look at the docs you copied in, it says `" The block number to stop before. Also the total
              number of blocks to get. If not setting a start value, is set by
              the first argument."`

This is because we wanted splicing behavior for the range method.

For this method, we want `stop_block` to be the last block you get. It does not currently work that way I don't think.

However, we still want it to work that way from the `BlockContainer.range()` perspective... So some refactoring is still needed

johnson2427: ohhh yes, I follow, I see what you mean. I think if I change that though, it will break all of these tests

johnson2427:
Just a heads up, I had to force a push in order for the checks to run. GitHub wasn't down

src/ape/managers/query.py
unparalleled-js: Make `chain_manager.blocks.range()` work exactly like it did before EXCEPT it now utilizes the `QueryManager` (which does the heavy lifting via most of the logic you copy in). The one difference is that in query_manager, make `stop_block` refer to the actual last block you want to get.

So it won't break the `chain_manager.range()` tests because we still want that to mimic the python splice behavior.

And then add new tests for this method in the `QueryManager`.

unparalleled-js (changes requested):
Just waiting on get_blocks to be inclusive of stop_block (whilst maintaining splice-like behavior in range)

src/ape/managers/chain.py
```diff
@@ -137,53 +137,7 @@ def query(
...
-        elif stop < start:
-            raise ValueError(f"stop '{stop}' cannot be less than start '{start}'.")
-        elif stop < 0:
-            raise ValueError(f"start '{start}' cannot be negative.")
-        elif start_or_stop < 0:
-            raise ValueError(f"stop '{stop}' cannot be negative.")
-
-        for i in range(start, stop, step):
-            yield self._get_block(i)
+        return self.query_manager.get_blocks(start_or_stop, stop, step)
```
fubuloubu, line 140: So, I was hoping here you could form and use the queries and call in via `query_manager.query` to conform to the API, and then dissect the result here in order to provide the iterated result decoded from the dataframe

PR #600, opened 2022-03-23
fix: ape plugins list showed core plugins when it was not supposed to

Ninjagod1251:
### What I did

fixes: # 
add strip() to the check core plugins check to rm the trailing white spaces

### How I did it

add strip() to the check core plugins check to rm the trailing white spaces

### How to verify it

`ape plugins list` should no longer have core plugins listed

### Checklist


unparalleled-js (approved):
LGTM

src/ape/utils.py
```diff
@@ -386,14 +386,15 @@ class GithubClient:
     An HTTP client for the Github API.
     """
 
-    TOKEN_KEY = "GITHUB_ACCESS_TOKEN"
+    TOKEN_KEY_OPTIONS = ("GITHUB_ACCESS_TOKEN", "GH_ACCESS_TOKEN")
```
unparalleled-js, line 389: had to do this because you cannot have env vars in github actions with the prefix `GITHUB_`

.github/workflows/test.yaml
```diff
@@ -55,6 +55,9 @@ jobs:
                 os: [ubuntu-latest, macos-latest]   # eventually add `windows-latest`
                 python-version: [3.7, 3.8, 3.9, "3.10"]
 
+        env:
+          GITHUB_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```
fubuloubu, line 59: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

fubuloubu:
```suggestion
          GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

fubuloubu (approved):

PR #601, opened 2022-03-23
chore: bump ethpm-types version

NotPeopling2day:
Bumps the ethpm-types version requirement from beta to >0.1.0, <0.2.0.


fubuloubu (changes requested):
V0.1.0 not released yet, can we do >=b2

PR #607, opened 2022-03-27
feat!: ecosystem decode and encode address

unparalleled-js:
BREAKING CHANGE: Adds a new abstract method and deletes `utils` method `to_address`.

### What I did

fixes: #579 

Allow ecosystem API to decode and encode addresses

### How I did it

new abstract method

### How to verify it

See test
Also going to make sure Starknet can do its thang. I wll move out of draft then.

### Checklist


src/ape_ethereum/ecosystem.py
```diff
@@ -31,6 +31,7 @@
 from ape.contracts._utils import LogInputABICollection
 from ape.exceptions import DecodingError, OutOfGasError, SignatureError, TransactionError
 from ape.types import AddressType, ContractLog
+from ape.utils import to_address
```
fubuloubu, line 34: Curious if we should just move that import here (and no need to rename it anymore)

Then later in v0.2.0 we would delete it from utils, maybe wrap a "deprecation warning" around it

fubuloubu (approved):
Will the conversion manager plugins need to use this?

src/ape_ethereum/ecosystem.py
unparalleled-js: There are other breaking changes in this PR (new abstract methods are being added). Could we delete it now then?

fubuloubu: > There are other breaking changes in this PR (new abstract methods are being added). Could we delete it now then?

Yes!

NotPeopling2day (approved):
LGTM

unparalleled-js:
Are we okay with 0.2.0 being the next release or should I hold off on merging breaking changes?
I have a whole starknet staging branch as well.

fubuloubu:
> Are we okay with 0.2.0 being the next release or should I hold off on merging breaking changes? I have a whole starknet staging branch as well.

Let's merge any other un-breaking changes, create a v0.1.4 release next week, and then do v0.2.0 in Nashville

src/ape_accounts/accounts.py
```diff
@@ -60,7 +59,7 @@ def keyfile(self) -> dict:
 
     @property
     def address(self) -> AddressType:
-        return to_address(self.keyfile["address"])
+        return to_checksum_address(self.keyfile["address"])
```
unparalleled-js, line 62: A bigger TODO is to make `AccountAPI` know which ecosystem an account can work.

Then, we can use a non-connected version of the ecosystem to decode the address.
Otherwise if we do `self.provider.network.ecosystem.decode_address` here, it will require a provider to run the command `ape accounts list`.

fubuloubu:
Edited the PR title and PR body to be consistent with https://www.conventionalcommits.org/en/v1.0.0/

The rules there state that `!` should be added after the commit type to draw attention to a breaking change, and then the `BREAKING CHANGE:` footer should be in the PR body.

fubuloubu (approved):

fubuloubu (approved):

PR #616, opened 2022-03-29
fix: version issue to display in installed sections

Ninjagod1251:
### What I did

Fixed the output for `ape plugins list` to have the installed section display the versions

fixes: #

### How I did it

formatted the display to include it. It had it earlier but it broke. 

### How to verify it

install plugins and check with `ape plugins list` and/or `ape plugins list --all`

### Checklist

### DIFF
--- a/tests/integration/cli/test_plugins.py
+++ b/tests/integration/cli/test_plugins.py
@@ -11,3 +11,11 @@ def test_list_excludes_core_plugins(ape_cli, runner):
     assert "console" not in result.output, "console is not supposed to be in Installed Plugins"
     assert "networks" not in result.output, "networks is not supposed to be in Installed Plugins"
     assert "geth" not in result.output, "networks is not supposed to be in Installed Plugins"
+
+
+@pytest.mark.xfail(strict=False, reason="Requires plugins installed")
+@skip_projects_except(["test"])  # Only run on single project to prevent rate-limiting in CI/CD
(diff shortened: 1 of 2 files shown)


unparalleled-js (approved):
LGTM

tests/integration/cli/test_plugins.py
```diff
@@ -11,3 +11,11 @@ def test_list_excludes_core_plugins(ape_cli, runner):
     assert "console" not in result.output, "console is not supposed to be in Installed Plugins"
     assert "networks" not in result.output, "networks is not supposed to be in Installed Plugins"
     assert "geth" not in result.output, "networks is not supposed to be in Installed Plugins"
+
+
+@pytest.mark.xfail(strict=False, reason="Requires plugins installed")
```
fubuloubu, line 16: if you want to run tests in a separate CI environment, or perhaps after another intermediate step, you can use markers to acheive this: https://docs.pytest.org/en/6.2.x/example/markers.html

so basically, do like:

```
pytest -m "no:plugins_installed" ...
ape install plugin-name
pytest -m "plugins_installed" ...
```

And then have `no:plugins_installed" be the default config for pytest

fubuloubu (approved):

tests/integration/cli/test_plugins.py
Ninjagod1251: okay ill take a look into this thank you

PR #617, opened 2022-03-29
fix: recursion error when accessing chain ID for non-connected provider

unparalleled-js:
### What I did

Issue causing network ID recursion and double definiton of chain ID / network ID

### How I did it

Call into `super()` when not  connected to get chain ID set from init (the static one) when not  connected

### How to verify it

Make sure you can check the chain ID for a non-connected provider
**NOTE**: Does not really work for non-local networks

### Checklist

### DIFF
--- a/src/ape_test/providers.py
+++ b/src/ape_test/providers.py
@@ -2,13 +2,12 @@ from eth_tester.backends import PyEVMBackend  # type: ignore
 from eth_tester.exceptions import TransactionFailed  # type: ignore
 from eth_utils.exceptions import ValidationError
 from web3 import EthereumTesterProvider, Web3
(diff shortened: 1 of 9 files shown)


NotPeopling2day (dismissed):

src/ape_test/providers.py
```diff
@@ -8,6 +8,8 @@
 from ape.types import SnapshotID
 from ape.utils import gas_estimation_error_message
 
+DEFAULT_CHAIN_ID = 61
```
unparalleled-js, line 11: I am trying to figure out where this gets set in `eth-tester`.
It isn't released yet but I think they are configuring having this get set more configurably

fubuloubu (approved):

fubuloubu (approved):

PR #618, opened 2022-03-29
refactor: leverage query manager to complete historical block range queries

johnson2427:
### What I did
Pulled functionality of the chain.blocks.range to DefaultQueryProvider

fixes: #595

### How I did it
Refactor of the range function and DefaultQueryProvider as well as some changes to api.query

### How to verify it
Built in tests for chain and query

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -20,17 +20,26 @@ class _BaseQuery(BaseModel):
     @classmethod
     @abstractmethod
     def all_fields(cls) -> List[str]:
+        """
+        Validates fields that are called during a block query.
+
+        Returns:
+            List[str]: list of columns to be returned in pandas
+                dataframes during block query.
+        """
         ...
 
     @validator("columns")
-    def check_columns(cls, data):
+    def check_columns(cls, data: str) -> Union[List[str], str]:
         all_fields = cls.all_fields()
         if len(data) == 1 and data[0] == "*":
             return all_fields
         else:
-            assert len(set(data)) == len(data), f"Duplicate fields in {data}"
+            if len(set(data)) != len(data):
+                raise ValueError(f"Duplicate fields in {data}")
             for d in data:
-                assert d in all_fields, f"Unrecognized field '{d}', must be one of {all_fields}"
+                if d not in all_fields:
+                    raise ValueError(f"Unrecognized field '{d}', must be one of {all_fields}")
         return data
 
 
@@ -65,7 +74,7 @@ class _BaseAccountQuery(BaseModel):
     stop_nonce: NonNegativeInt
 
     @root_validator(pre=True)
-    def check_start_nonce_before_stop_nonce(cls, values):
+    def check_start_nonce_before_stop_nonce(cls, values: Dict):
         if values["stop_nonce"] < values["start_nonce"]:
             raise ValueError(
                 f"stop_nonce: '{values['stop_nonce']}' cannot be less than "
@@ -118,7 +127,7 @@ class ContractMethodQuery(_BaseBlockQuery):
 
     @classmethod
     def all_fields(cls) -> List[str]:
-        # TODO: Figure out how to get the event ABI as a class property
+        # TODO: Figure out how to get the method ABI as a class property
         #   for the validator
         return [o.name for o in cls.method.outputs if o.name is not None]  # just for mypy
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -25,7 +25,7 @@ class _BaseQuery(BaseModel):
 
         Returns:
             List[str]: list of columns to be returned in pandas
-                dataframes during block query.
+            dataframes during block query.
         """
         ...
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -27,7 +27,6 @@ class _BaseQuery(BaseModel):
             List[str]: list of columns to be returned in pandas
             dataframes during block query.
         """
-        ...
 
     @validator("columns")
     def check_columns(cls, data: List[str]) -> List[str]:
@@ -69,7 +68,7 @@ class BlockQuery(_BaseBlockQuery):
         return list(BlockAPI.__fields__)
 
 
-class _BaseAccountQuery(BaseModel):
+class _BaseAccountQuery(_BaseQuery):
     start_nonce: NonNegativeInt = 0
     stop_nonce: NonNegativeInt
 
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -74,7 +74,7 @@ class _BaseAccountQuery(BaseModel):
     stop_nonce: NonNegativeInt
 
     @root_validator(pre=True)
-    def check_start_nonce_before_stop_nonce(cls, values: Dict):
+    def check_start_nonce_before_stop_nonce(cls, values: Dict) -> Dict:
         if values["stop_nonce"] < values["start_nonce"]:
             raise ValueError(
                 f"stop_nonce: '{values['stop_nonce']}' cannot be less than "
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -2,18 +2,36 @@ from typing import Any, Dict, List, Optional, Union
 
 import pandas as pd
 from ethpm_types.abi import EventABI, MethodABI
-from pydantic import BaseModel, NonNegativeInt, root_validator
+from pydantic import BaseModel, NonNegativeInt, root_validator, validator
 
-from ape._compat import Literal
 from ape.types import AddressType
 from ape.utils import BaseInterfaceModel, abstractmethod
+
+from .providers import BlockAPI
+from .transactions import TransactionAPI
 
 QueryType = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractMethodQuery"]
 
 
 class _BaseQuery(BaseModel):
-    type: str  # Used as discriminator
+
     columns: List[str]
+
+    @classmethod
+    @abstractmethod
+    def all_fields(cls) -> List[str]:
+        ...
+
+    @validator("columns")
+    def check_columns(cls, data):
+        all_fields = cls.all_fields()
+        if len(data) == 1 and data[0] == "*":
+            return all_fields
+        else:
+            assert len(set(data)) == len(data), f"Duplicate fields in {data}"
+            for d in data:
+                assert d in all_fields, f"Unrecognized field '{d}', must be one of {all_fields}"
+        return data
 
 
 class _BaseBlockQuery(_BaseQuery):
@@ -37,7 +55,9 @@ class BlockQuery(_BaseBlockQuery):
     blocks between ``start_block`` and ``stop_block``.
     """
 
-    type: Literal["blocks"] = "blocks"
+    @classmethod
+    def all_fields(cls) -> List[str]:
+        return list(BlockAPI.__fields__)
 
 
 class _BaseAccountQuery(BaseModel):
@@ -61,8 +81,11 @@ class AccountQuery(_BaseAccountQuery):
     of transactions made by ``account`` between ``start_nonce`` and ``stop_nonce``.
     """
 
-    type: Literal["accounts"] = "accounts"
     account: AddressType
+
+    @classmethod
+    def all_fields(cls) -> List[str]:
+        return list(TransactionAPI.__fields__)
 
 
 class ContractEventQuery(_BaseBlockQuery):
@@ -71,9 +94,16 @@ class ContractEventQuery(_BaseBlockQuery):
     logs emitted by ``contract`` between ``start_block`` and ``stop_block``.
     """
 
-    type: Literal["contract_events"] = "contract_events"
     contract: AddressType
     event: EventABI
+
+    @classmethod
+    def all_fields(cls) -> List[str]:
+        # TODO: Figure out how to get the event ABI as a class property
+        #   for the validator
+        return [
+            i.name for i in cls.event.inputs if i.name is not None
+        ]  # if i.name is not None just for mypy
 
 
 class ContractMethodQuery(_BaseBlockQuery):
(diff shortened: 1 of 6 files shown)


src/ape/api/query.py
```diff
@@ -2,19 +2,37 @@
...
 
 
 class _BaseQuery(BaseModel):
-    type: str  # Used as discriminator
+
     columns: List[str]
 
+    @classmethod
+    @abstractmethod
+    def all_fields(cls) -> List[str]:
```
unparalleled-js, line 22: Does this need a doc str?

src/ape/api/query.py
```diff
@@ -2,19 +2,37 @@
...
+        ...
+
+    @validator("columns")
+    def check_columns(cls, data):
+        all_fields = cls.all_fields()
+        if len(data) == 1 and data[0] == "*":
+            return all_fields
+        else:
+            assert len(set(data)) == len(data), f"Duplicate fields in {data}"
```
unparalleled-js, line 31: NOTE: This assertions get removed if running Python outside of debug mode. Are you sure that is what we want? Otherwise, just raise `AssertionError`s manually

src/ape/api/query.py
```diff
@@ -61,32 +81,47 @@ class AccountQuery(_BaseAccountQuery):
...
     """
 
-    type: Literal["contract_calls"] = "contract_calls"
     contract: AddressType
     method: MethodABI
     method_args: Dict[str, Any]
 
+    @classmethod
+    def all_fields(cls) -> List[str]:
+        # TODO: Figure out how to get the event ABI as a class property
```
unparalleled-js, line 121: Is this TODO only supposed to be on the class above (ContractEventQuery) and not here?
Or should this say `# TODO: Figure out how to get the method ABI as a class property`

src/ape/managers/chain.py
```diff
@@ -182,8 +182,11 @@ def range(
         elif start_or_stop < 0:
             raise ValueError(f"stop '{stop}' cannot be negative.")
 
-        for i in range(start, stop, step):
-            yield self._get_block(i)
+        # Note: the range stop block is a non-inclusive stop.
+        #       Where as the query method is an inclusive stop.
+        results = self.query("*", start_block=start, stop_block=stop - 1)  # type: ignore
```
unparalleled-js, line 187: Should all of this go in the `_get_block()` helper method on this class so other public methods can make use of this optimization, such as `poll_blocks()`?

src/ape/managers/chain.py
```diff
@@ -182,8 +182,11 @@ def range(
...
+        for _, row in results.iterrows():
+            yield BlockAPI.parse_obj(row.to_dict())
```
unparalleled-js, line 189: Shouldn't this be `self.provider.network.ecosystem.decode_block()`? I believe this will have abstract methods on it and not work too great otherwise

src/ape/managers/query.py
```diff
@@ -38,7 +38,14 @@ def perform_query(self, query: QueryType) -> pd.DataFrame:
 
     @perform_query.register
     def perform_block_query(self, query: BlockQuery) -> pd.DataFrame:
-        blocks_iter = self.chain_manager.blocks.range(query.start_block, query.stop_block)
+        if not self.network_manager.active_provider:
```
unparalleled-js, line 41: It is not necessary to check if `network_manager.active_provider is not None`... Just use `self.provider` which automatically does this.

unparalleled-js (commented):
Cool, nice work! I have some questions...

src/ape/api/query.py
```diff
@@ -2,19 +2,37 @@
...
 from ape.utils import BaseInterfaceModel, abstractmethod
 
+from .providers import BlockAPI
+from .transactions import TransactionAPI
+
 QueryType = Union["BlockQuery", "AccountQuery", "ContractEventQuery", "ContractMethodQuery"]
 
 
 class _BaseQuery(BaseModel):
-    type: str  # Used as discriminator
```
fubuloubu, line 15: Note, we really don't need a discriminator to deserialize with

src/ape/managers/chain.py
```diff
@@ -182,8 +182,11 @@ def range(
         elif start_or_stop < 0:
             raise ValueError(f"stop '{stop}' cannot be negative.")
 
-        for i in range(start, stop, step):
-            yield self._get_block(i)
+        # Note: the range stop block is a non-inclusive stop.
+        #       Where as the query method is an inclusive stop.
```
fubuloubu, line 186:
```suggestion
        # NOTE: the range `stop_block` parameter is a non-inclusive stop, while the
        #       `.query` method uses an inclusive stop, so we must adjust downwards.
```

src/ape/managers/query.py
```diff
@@ -38,7 +38,14 @@ def perform_query(self, query: QueryType) -> pd.DataFrame:
...
+            raise QueryEngineError("Not connected to an active network")
+        blocks_iter = map(
+            self.network_manager.active_provider.get_block,
+            # NOTE: the range stop block is a non-inclusive stop.
+            #       Where as the query method is an inclusive stop.
```
fubuloubu, line 46:
```suggestion
        # NOTE: the query `stop_block` parameter is an inclusive stop, while the
        #       `range` function uses a non-inclusive stop, so we must adjust upwards.
```

tests/functional/api/test_query.py
```diff
@@ -3,7 +3,7 @@
 
 def test_basic_query(eth_tester_provider):
     chain.mine(3)
```
fubuloubu, line 5: Note that latent failure that there is actually 4 blocks, the genesis block and the 3 mined blocks

fubuloubu (approved):

src/ape/api/query.py
```diff
@@ -2,19 +2,37 @@
...
+
     columns: List[str]
 
+    @classmethod
+    @abstractmethod
+    def all_fields(cls) -> List[str]:
+        ...
+
+    @validator("columns")
+    def check_columns(cls, data):
```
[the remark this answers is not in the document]
fubuloubu: Sure

src/ape/api/query.py
fubuloubu: Yes

src/ape/api/query.py
fubuloubu: Ah yes, raise `ValueError` instead here to match typical pydantic validators

src/ape/api/query.py
fubuloubu:
```suggestion
        # TODO: Figure out how to get the method ABI as a class property
```

src/ape/managers/chain.py
fubuloubu: `poll_blocks` uses range, so we are good:

https://github.com/ApeWorX/ape/blob/main/src/ape/managers/chain.py#L230

src/ape/managers/chain.py
fubuloubu: Yes, please do this!

src/ape/managers/query.py
fubuloubu: Ah yes, please do this

unparalleled-js (changes requested):
see comments

tests/functional/api/test_query.py
```diff
@@ -7,3 +11,28 @@ def test_basic_query(eth_tester_provider):
...
+    columns = list(df_all.columns)
+    assert [
+        "gas_data",
+        "consensus_data",
+        "hash",
+        "number",
+        "parent_hash",
+        "size",
+        "timestamp",
+    ] == columns
```
fubuloubu, line 24: Little easier to read:
```suggestion
    assert columns == [
        "gas_data",
        "consensus_data",
        "hash",
        "number",
        "parent_hash",
        "size",
        "timestamp",
    ]
```

tests/functional/api/test_query.py
```diff
@@ -7,3 +11,28 @@ def test_basic_query(eth_tester_provider):
...
+
+
+def test_block_query(eth_tester_provider):
+    chain.mine(3)
+    with pytest.raises(ValidationError) as err:
+        BlockQuery(columns=["columns"], start_block=0, stop_block=2)
+    assert "Unrecognized field 'columns" in str(err.value)
```
fubuloubu, line 31: Might be nicer to show a typo:
```suggestion
        BlockQuery(columns=["numbr"], start_block=0, stop_block=2)
    assert "Unrecognized field 'numbr'" in str(err.value)
```

tests/functional/api/test_query.py
```diff
@@ -7,3 +11,28 @@ def test_basic_query(eth_tester_provider):
...
+
+
+def test_account_query(eth_tester_provider):
+    chain.mine(3)
+    with pytest.raises(ValidationError) as err:
+        AccountQuery(columns=["colu"], start_nonce=0, stop_nonce=2)
+    assert "validation error for AccountQuery" in str(err.value)
```
fubuloubu, line 38: To match the test above
```suggestion
        AccountQuery(columns=["none"], start_nonce=0, stop_nonce=2)
    assert "Unrecognized field 'none'" in str(err.value)
```

tests/functional/api/test_query.py
```diff
@@ -7,3 +11,28 @@ def test_basic_query(eth_tester_provider):
...
+        "consensus_data",
+        "hash",
+        "number",
+        "parent_hash",
+        "size",
+        "timestamp",
+    ] == columns
+
+
+def test_block_query(eth_tester_provider):
```
fubuloubu, line 27: We could rewrite this test and the one below as follows:

```py
@pytest.mark.parametrize(
    "query_type,incorrect_column",
    [(BlockQuery, "numbr"), (AccountQuery, "none")],
)
def test_query_incorrect_names(query_type, incorrect_column):
    # Mismatching column raises
    with pytest.raises(ValidationError) as err:
        query_type(columns=[incorrect_column])  # NOTE: fails validation before any other checks
    # error message shows mismatching columns
    assert f"Unrecognized field '{incorrect_column}'" in str(err.value)

Comment (johnson2427) [reply tests/functional/api/test_query.py:27]: I think there is an issue doing this because of start and stop
BlockQuery is looking for start and stop `block`
AccountQuery is looking for start and stop `nonce`

Comment (fubuloubu) [reply tests/functional/api/test_query.py:27]: > I think there is an issue doing this because of start and stop BlockQuery is looking for start and stop `block` AccountQuery is looking for start and stop `nonce`

Ah yes, `start_*` had a default value, but `stop_*` does not in both cases.

**Comment by fubuloubu** [comment]
wild that only mac is breaking

File: tests/functional/api/test_query.py (hunk 15 above)
Comment (johnson2427) [reply tests/functional/api/test_query.py:27]: We can resolve this in a future revision

File: src/ape/managers/chain.py (hunk 5 above)
Comment (johnson2427) [reply src/ape/managers/chain.py:189]: I'm not sure what you mean by this. Should we be doing this:
```python
yield self.provider.network.ecosystem.decode_block()
```
?

src/ape/api/query.py
```diff
@@ -2,19 +2,46 @@
...
+        Validates fields that are called during a block query.
+
+        Returns:
+            List[str]: list of columns to be returned in pandas
+                dataframes during block query.
+        """
+        ...
+
+    @validator("columns")
+    def check_columns(cls, data: str) -> Union[List[str], str]:
```
fubuloubu, line 33: Shouldn't be a Union
```suggestion
    def check_columns(cls, data: List[str]) -> List[str]:
```

src/ape/api/query.py
```diff
@@ -37,15 +64,17 @@ class BlockQuery(_BaseBlockQuery):
...
+        return list(BlockAPI.__fields__)
 
 
 class _BaseAccountQuery(BaseModel):
     start_nonce: NonNegativeInt = 0
     stop_nonce: NonNegativeInt
 
     @root_validator(pre=True)
-    def check_start_nonce_before_stop_nonce(cls, values):
+    def check_start_nonce_before_stop_nonce(cls, values: Dict):
```
fubuloubu, line 77:
```suggestion
    def check_start_nonce_before_stop_nonce(cls, values: Dict) -> Dict:
```

src/ape/managers/chain.py
fubuloubu: Yes, call that method with `row.to_dict()`

tests/functional/api/test_query.py
```diff
@@ -1,9 +1,38 @@
...
+    with pytest.raises(ValidationError) as err:
+        BlockQuery(columns=["numbr"], start_block=0, stop_block=2)
+    assert "Unrecognized field 'numbr'" in str(err.value)
+
+
+def test_account_query(eth_tester_provider):
+    chain.mine(3)
+    with pytest.raises(ValidationError) as err:
+        AccountQuery(columns=["none"], start_nonce=0, stop_nonce=2)
+    assert "validation error for AccountQuery" in str(err.value)
```
fubuloubu, line 38: Shouldn't this also raise the unrecognized field name?

tests/functional/api/test_query.py
```diff
@@ -1,9 +1,38 @@
...
+        "timestamp",
+    ]
+
+
+def test_block_query(eth_tester_provider):
+    chain.mine(3)
+    with pytest.raises(ValidationError) as err:
+        BlockQuery(columns=["numbr"], start_block=0, stop_block=2)
+    assert "Unrecognized field 'numbr'" in str(err.value)
+
```
fubuloubu, line 32: Missing check of duplicate fields

tests/functional/api/test_query.py
johnson2427: The unrecognized field name is not coming through that error. We were able to make that happen when we were debugging, but for some reason in the pytest, it never returns that error

fubuloubu: That's not good, let's debug some of this today

src/ape/api/query.py
```diff
@@ -2,19 +2,46 @@
...
 
+    @classmethod
+    @abstractmethod
+    def all_fields(cls) -> List[str]:
+        """
+        Validates fields that are called during a block query.
+
+        Returns:
+            List[str]: list of columns to be returned in pandas
+                dataframes during block query.
```
unparalleled-js, line 28:
```suggestion
            dataframes during block query.
```

^ This space (for next-line on returns doc) causes weird things to happen. Is fixed when not indented.

tests/functional/api/test_query.py
```diff
@@ -1,9 +1,44 @@
...
+        "consensus_data",
+        "hash",
+        "number",
+        "parent_hash",
+        "size",
+        "timestamp",
+    ]
+
+
+def test_block_query(eth_tester_provider):
```
unparalleled-js, line 27: In general, it is better to use multiple test cases.

Tests should test one thing. Now, if one part of the tests fails, the whole test fails, and that makes it harder to pinpoint what is actually failing.

unparalleled-js (dismissed):
We need to figure out the Block API thing! Otherwise, we can't use the blocks from the query and they may be missing data depending on your ecosystem

fubuloubu (approved):

fubuloubu (approved):

Issue #574, opened 2022-03-14
Allow obtaining AttrDicts (or named tuples) from ABI Methods that return Structs or NamedTuples


fubuloubu:
### Overview
Sometimes when you are working with contracts that have more complex structs in it, it is very handy to be able to work with them in their native source code format (either as struct definitions, or named dicts, ABI specification is the same for both) 

### Specification
Be able to convert  them as the following:

```py
result = my_contract.methodReturnsStruct(...)
print(result)
# (1, "0xaBcD...", b"\x13...", False)
print(result.dict())
# {"my_int": 1, "my_address": "0xaBcD...", "my_bytes": b"\x13...", "my_bool": False}
```

Using an attrdict would be nice, because then we can just do this:
```py
assert my_contract.methodReturnsStruct(...).dict().my_bool
```

### Dependencies
No dependencies

fubuloubu:
Note: to detect this, the ABI outputs for a call will have a `name` field in each of them. If they do not, you cannot do this encoding because some fields will be unnamed

PR #628, opened 2022-04-02
feat!: add 'decode_returndata' abstract method to EcosystemAPI

unparalleled-js:
BREAKING CHANGE: Adds new abstract method.

### What I did

I noticed `decode_calldata()` was missing from the `EcoystemAPI` and it is expected to exist, so I added it.
Also fixed doc links related to `api.transactions.py`
Also added missing doc strs to `EcosystemAPI`

### How I did it

Added abstract method

### Checklist


src/ape/api/networks.py
```diff
@@ -204,21 +204,73 @@ def set_default_network(self, network_name: str):
...
+
+        Args:
+            **kwargs: Everything the transaction needs initialize.
+
+        Returns:
+            class:`~ape.api.transactions.TransactionAPI`
+        """
+
+    @abstractmethod
+    def decode_calldata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
```
fubuloubu, line 263: Is this why there's a `type: ignore` here?

https://github.com/ApeWorX/ape/blob/main/src/ape/contracts/base.py#L91

src/ape/api/networks.py
```diff
@@ -204,21 +204,73 @@ def set_default_network(self, network_name: str):
     def encode_deployment(
         self, deployment_bytecode: HexBytes, abi: ConstructorABI, *args, **kwargs
     ) -> "TransactionAPI":
-        ...
+        """
+        Create a deploy transaction.
```
fubuloubu, line 208:
```suggestion
        Create a deployment transaction in the given ecosystem. This may require connecting to other networks.
```

src/ape/api/networks.py
```diff
@@ -204,21 +204,73 @@ def set_default_network(self, network_name: str):
...
+            class:`~ape.api.transactions.TransactionAPI`
+        """
 
     @abstractmethod
     def encode_transaction(
         self, address: AddressType, abi: MethodABI, *args, **kwargs
     ) -> "TransactionAPI":
-        ...
+        """
+        Create a transaction object from a contract function call.
```
fubuloubu, line 225:
```suggestion
        Encode a transaction object from a contract function's abi and call arguments. Update the transaction arguments with the overrides in ``kwargs`` as well.
```

src/ape/api/networks.py
```diff
@@ -204,21 +204,73 @@ def set_default_network(self, network_name: str):
...
+
+        Returns:
+            class:`~ape.api.transactions.TransactionAPI`
+        """
 
     @abstractmethod
     def decode_logs(self, abi: EventABI, raw_logs: List[Dict]) -> Iterator[ContractLog]:
-        ...
+        """
+        Get logs from raw chain data.
```
fubuloubu, line 240:
```suggestion
        Decode any contract logs that match the given event ABI from the raw log data.
```

NotPeopling2day (commented):
LGTM, approved with the above changes

src/ape/api/networks.py
```diff
@@ -204,21 +204,75 @@ def set_default_network(self, network_name: str):
...
+
+        Args:
+            **kwargs: Everything the transaction needs initialize.
+
+        Returns:
+            class:`~ape.api.transactions.TransactionAPI`
+        """
+
+    @abstractmethod
+    def decode_calldata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
```
fubuloubu, line 265: Should it just return `Any`? I think enforcing the tuple return might be too strict

fubuloubu, line 265: Nitpick, but realized this should be called `decode_returndata` because that's the opcode in the vm

unparalleled-js: Since this is not necessarily VM specific, could we call it `decode_return_data()` ?

unparalleled-js: 💯

fubuloubu: sure

unparalleled-js:
NOTE: This PR is technically a breaking-change because it adds a new abstractmethod. However, since it is only adding it to `EcosystemAPI` and all out ecosystems currently use `ape-ethereum`, it is low-risk that this would affect any plugins (unless some users out there are already making Ecosystem plugins that we don't know about).

src/ape/contracts/base.py
```diff
@@ -88,17 +88,20 @@ def __call__(self, *args, **kwargs) -> Any:
...
 
+        # TODO: Handle struct output
+        if not isinstance(output, (list, tuple)):
+            return output
+
         # NOTE: Returns a tuple, so make sure to handle all the cases
-        if len(tuple_output) < 2:
-            return tuple_output[0] if len(tuple_output) == 1 else None
+        elif len(output) < 2:
+            return output[0] if len(output) == 1 else None
```
fubuloubu, line 102: This might not be the correct decision to make, depending on if the output is a single-member struct type, similar to resolving #574

unparalleled-js: I did leave the TODO in there for this... You think that issue needs to be fixed in this PR?

unparalleled-js:
I think I actually need a `decode_call_data()` for Starknet

fubuloubu:
> I think I actually need a `encode_call_data()` for Starknet

I thought we had this, but apparently we do not (sort of mixed in with "serialize transaction").

In favor of adding it though!

fubuloubu (approved):

fubuloubu (approved):

Issue #623, opened 2022-03-31
`ape compile` doesn't detect updates to dependencies of source files


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.4.dev3+ge9615911

$ ape plugins list
Installed Plugins:
  solidity    
  hardhat     
  vyper    
```

* Python Version: 3.9.9
* OS: linux

### What went wrong?

Updating a file did not trigger re-compiling vs. the cached copy

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

pandadefi:
here is an example from https://github.com/yearn/veYFI
First compile the code:
```
git checkout 657459797cdec60ee1739b9f63cb7b484569c4b3
ape compile -f
```

Second:
```
git checkout 1dd38049149a782b87308968593834b2ae2fbd2f
ape test
```

The test are falling because the contracts aren't re-compiled and missing https://github.com/yearn/veYFI/commit/1dd38049149a782b87308968593834b2ae2fbd2f#diff-730c6bf6b59b69b8d287c6a118eb5e741a229272167aca3e8a931ff2a0e14049R86

fubuloubu:
We are pretty sure we narrowed this down to an issue resolving updated dependencies of files that produced relevant types, which is how `Gauge.sol` is using `BaseGauge.sol` right now in your project

NotPeopling2day:
Depends on #704

NotPeopling2day:
Solution for solidity contracts is merged, this ticket is superseded by implementation tickets in vyper and cairo repos.

PR #629, opened 2022-04-03
fix: dependent sources compiling

NotPeopling2day:
### What I did
Updated the discord link to point to the ApeWorX channel
Updated `ethpm-types` version and pinning
Renamed `sources` to be more specific about being paths and not actual `Source` objects.
Added logic for setting source references i.e. dependent sources
Added logic to add source references to `needs_compiling`

- [ ] add `TODO` comments referencing what will change when breaking update is applied to `CompilerAPI`
- [ ] research what vyper will require
- [ ] open another issue for adding `type` setting/logic to `Source` objects

fixes: #623 

### How to verify it

Run through the example in the issue/ticket
Screen Shot 2022-04-02 at 9 43 28 PM

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Committing will now automatically run the local hooks and ensure that your commi
 
 ## Github Access Token
 
-If you are a member of ApeWorX and would like to install private plugins, 
+If you are a member of ApeWorX and would like to install private plugins,
 [create a Github access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token).
 
 Once you have your token, export it to your terminal session:
@@ -80,4 +80,4 @@ A pull request represents the start of a discussion, and doesn't necessarily nee
 If you are opening a work-in-progress pull request to verify that it passes CI tests, please consider
 [marking it as a draft](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).
 
-Join the Ethereum Python [Discord](https://discord.gg/PcEJ54yX) if you have any questions.
+Join the ApeWorX [Discord](https://discord.gg/7gshPtGd) if you have any questions.
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -99,7 +99,7 @@ class ProjectAPI(BaseInterfaceModel):
     @classmethod
     def _create_manifest(
         cls,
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -80,4 +80,4 @@ A pull request represents the start of a discussion, and doesn't necessarily nee
 If you are opening a work-in-progress pull request to verify that it passes CI tests, please consider
 [marking it as a draft](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).
 
-Join the Ethereum Python [Discord](https://discord.gg/PcEJ54yX) if you have any questions.
+Join the ApeWorX [Discord](https://discord.gg/7gshPtGd) if you have any questions.
```
fubuloubu, line 83: Does this link expire?

src/ape/api/projects.py
```diff
@@ -123,17 +123,36 @@ def _create_source_dict(
...
-                content=source.read_text(),
+                content=source_path.read_text(),
+                references=cls._create_source_reference_list(source_path=source_path),
             )
-            for source in contract_paths
+            for source_path in contract_paths
         }
 
+    @classmethod
+    def _create_source_reference_list(cls, source_path: Path) -> List[str]:
```
fubuloubu, line 139: Was picturing that this would be an API method, where all the paths are given as an input, and a Dict is returned with the references of each path to the set of references that depend on them. That would get stored in the manifest, and then referenced whenever changes are detected in the project manager to expand the set of files that need to be recompiled.

So basically, this logic should live under the solidity plugin.

Because this would be a breaking change, maybe the default would be to return a Dict with empty sets of references

src/ape/managers/project.py
```diff
@@ -93,20 +93,27 @@ def create_manifest(
 
         cached_sources = manifest.sources or {}
         cached_contract_types = manifest.contract_types or {}
-        sources = {s for s in self.sources if s in file_paths} if file_paths else set(self.sources)
+        cached_source_references = {
+            k: getattr(v, "references", []) for k, v in cached_sources.items()
```
fubuloubu, line 97: Smart for making it backwards compatible

CONTRIBUTING.md
NotPeopling2day: I don't think so, but I'll check into it.

NotPeopling2day:
Closing to split up into different tickets #703

PR #631, opened 2022-04-04
feat: allow plugins to register str converters

antazoey:
### What I did

Made it so that plugins can register converters that converts to str types.
Before, you would get a `ConversionError` saying you couldn't register this type.

**NOTE**: I am now realizing I don't actually need this feature (was a misunderstanding). However, it seems like something that should still exist?

### How I did it

Added `str` with an empty list to the pre-defined list.

### How to verify it

You can now register a converter that converts `str` in a plugin to 

### Checklist

### DIFF
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -186,6 +186,7 @@ class ConversionManager(BaseManager):
             Decimal: [],
             list: [ListTupleConverter()],
             tuple: [ListTupleConverter()],
+            str: [],
         }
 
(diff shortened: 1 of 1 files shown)


fubuloubu:
It uses the type of those keys to figure out what converters to try, not sure how this change would interoperate with addresses.

Also, what types of conversions do you picture needing to make that end up back as string types for ABI? I would think most of the time you'd just want the string to translate directly in the ABI layer.

Issue #635, opened 2022-04-06
git unauthenticated error when pulling dependencies using a github branch in ape config


storming0x:
### Environment information

* `ape` and plugin versions: 

```
$ ape --version
ape 0.1.4
$ ape plugins list
Installed Plugins:
  solidity    0.1.0
  hardhat     0.1.4
  vyper       0.1.0
```

* Python Version: 3.9.7
macOS m1: Darwin mac.local 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64

### What went wrong?

Trying to add to ape-config.yaml dependencies using branch option results in the errors below when running `ape compile`

```yaml
 - name: ds-test
    github: dapphub/ds-test
    branch: master
```

Error output:
```sh
INFO: Cloning branch 'master' from 'ds-test'.

...error log

  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

unparalleled-js:
I think I got something working that uses https instead! Would this be okay for now? See the attached PR.

Note that you have to also set the value `contracts_folder` in your dependency config because it is `src` and not the default `contracts/`.

storming0x:
> I think I got something working that uses https instead! Would this be okay for now? See the attached PR.
> 
> Note that you have to also set the value `contracts_folder` in your dependency config because it is `src` and not the default `contracts/`.

yes, i think https would solve my issue.

PR #639, opened 2022-04-06
fix: use encrypted protocol (https) for cloning repositories

unparalleled-js:
### What I did

fixes: #635

* Replace `git://` with `https://` for now. I tried to use `git@` but it does not seem to be a supported protocol in `pygit2`. A TODO is to add Auth to `pygit2`. We have it for PyGithub but cloning requires using `git` proper since it is not part of the `PyGithub`.
* Also fixed a bug with setting the contracts path in the dependency, which is needed to use the `ds-test` repo.

**NOTE**: This only happens to `branch` dependencies. `ds-test` requires branches since it does not have releases on Github. It is more ideal to use official releases however. Authentication currently works on real releases.

### How I did it

```python
url = repo.git_url.replace("git://", "https://")
```

### How to verify it

Try specifying the `ds-test` dependency like this:

```yaml
  - name: ds-test
    github: dapphub/ds-test
    branch: master
    contracts_folder: src
```

### Checklist


src/ape/managers/project.py
```diff
@@ -69,87 +69,102 @@ def create_manifest(
...
+            if not config_file.exists():
+                if self.name:
+                    config_data["name"] = self.name
+                if self.version:
+                    config_data["version"] = self.version
 
-            if path not in cached_sources:
-                return True  # New file added
+                config_data["contracts_folder"] = self.contracts_folder.name
+                if config_data:
```
fubuloubu, line 83: Won't this always be true now?

fubuloubu (approved):

PR #648, opened 2022-04-08
fix: use step parameter to block_container.range() method

unparalleled-js:
### What I did

Noticed we were no longing using the `step` parameter on `chain.blocks.range()` method.
Fixed it!

### How I did it

Honor `step`.

### How to verify it

See test :)

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -45,6 +45,7 @@ class _BaseQuery(BaseModel):
 class _BaseBlockQuery(_BaseQuery):
     start_block: NonNegativeInt = 0
     stop_block: NonNegativeInt
+    step: NonNegativeInt = 1
 
     @root_validator(pre=True)
     def check_start_block_before_stop_block(cls, values):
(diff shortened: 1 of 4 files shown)


src/ape/managers/chain.py
```diff
@@ -185,8 +185,13 @@ def range(
         # Note: the range `stop_block` is a non-inclusive stop, while the
         #       `.query` method uses an inclusive stop, so we must adjust downwards.
         results = self.query("*", start_block=start, stop_block=stop - 1)  # type: ignore
```
fubuloubu, line 187: Curious if we add step to the block query type, and then that allows queries to deselect a lot of unused information when querying larger sets of blocks (say, every thousand blocks)

unparalleled-js: Ah, yeah this probably makes sense at the query level

src/ape/api/query.py
```diff
@@ -45,6 +45,7 @@ def check_columns(cls, data: List[str]) -> List[str]:
 class _BaseBlockQuery(_BaseQuery):
     start_block: NonNegativeInt = 0
     stop_block: NonNegativeInt
+    step: NonNegativeInt = 1
```
fubuloubu, line 48: Should be Positive

fubuloubu (approved):

Issue #581, opened 2022-03-15
Interactive mode disconnects from provider when running scripts


antazoey:
### Environment information

ape v0.1.0
python 3.7.12
macOS

### What went wrong?

When using `-i` with `ape run`, it disconnects from the network before opening the console. Note: this does not happen in EthTester but it happens in all the other networks.

(As an aside, the `-i` is stuck on the left side of the script name. That maybe fixed at the same time as this, depending on what the solution entails)

### How can it be fixed?

I think we need to add a more clever way to manage the network in scripts.
I don't know yet!

unparalleled-js:
Another scripts bug:  ApeExceptions seem to not show up. The script ends without any output

unparalleled-js:
Seems like in some cases, as reported in #645, the interactive shell does not open at all

fubuloubu:
@NotPeopling2day follow up if this one is solved

sabotagebeats:
@fubuloubu I was unable to replicate this, however I also wasn't able to get `-I` to drop into interactive mode, but I think that might be a separate bug.

unparalleled-js:
I am not able to reproduce this anymore as of 0.6., but one thing to be aware of: if your `ape_console_extras.py` errors, it may crash `-I` and your provider and you may still get this error, that was happening to me.

antazoey:
OK so this is still a problem.
What happens - it is only a problem if you don't have a reference to the provider in your script.
This was also reported by users!

z80dev:
![image](https://github.com/ApeWorX/ape/assets/83730246/b2dcc67a-dfe3-40e1-8513-b05839698794)

PR #649, opened 2022-04-08
fix: issue when console would not launch when exception raised

unparalleled-js:
### What I did

fixes: #581

### How I did it

Catch the exception on invoke, check for `interactive`, launch console.
As far issue #581  goes, it seems like that is already fixed??? I am 

### Checklist

### DIFF
--- a/src/ape_run/_cli.py
+++ b/src/ape_run/_cli.py
@@ -1,3 +1,4 @@
+import traceback
 from pathlib import Path
 from runpy import run_module
 from typing import Any, Dict, Union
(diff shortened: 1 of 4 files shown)


src/ape_run/_cli.py
```diff
@@ -16,6 +18,19 @@ def __init__(self, *args, **kwargs):
...
+            return super().invoke(ctx)
+        except Exception:
+            if ctx.params["interactive"]:
+                # Print the exception trace and then launch the console
+                err_info = traceback.format_exc()
+                click.echo(err_info)
+                self._launch_console()
+            else:
+                # Don't handle error - raise exception as normal.
+                raise
```
unparalleled-js, line 32: NOTE: When you raise without arguments, the interpreter looks for the last exception raised and handled. It then acts the same as if you used raise with the most recent exception type, value and traceback.

https://stackoverflow.com/questions/18001721/raise-with-no-argument

fubuloubu (approved):

PR #674, opened 2022-04-28
chore: update mypy

fubuloubu:


src/ape/managers/query.py
```diff
@@ -22,7 +22,7 @@ class DefaultQueryProvider(QueryAPI):
     """
 
     @singledispatchmethod
-    def estimate_query(self, query: QueryType) -> Optional[int]:
+    def estimate_query(self, query: QueryType) -> Optional[int]:  # type: ignore
```
NotPeopling2day, line 25: If you add `@singledispatchmethod` to the `QueryAPI` methods you don't have to ignore these

fubuloubu: But then every implementation of `QueryAPI` would have to use `singledispatchmethod` which is unnecessarily restrictive

NotPeopling2day (approved):

unparalleled-js (approved):

PR #677, opened 2022-04-28
fix: issue with -v debug when CLI plugins failed to load

unparalleled-js:
### What I did

When using CLI plugins that fails, the `-v debug` flag would not register and I would not get a stack trace whereas it does for the non-CLI plugins. So, I figured out that it was more race conditions from click. Click actually calls the callback methods more than once. It does first to set the default value and again after it processes user args. So, it would get set to INFO, then the CLI plugin would fail to load, and then the DEBUG message would never get seen.
I fixed this.

### How I did it

Make the callback for the `verbosity_option` set from sys argv.
This makes it so plugins can still use this option (as currently it is not automatically given - will need to maybe look into `click-extensions` or whatever its called.

### How to verify it

Mess up a plugin, such as a config related error.
Like what I did is add key `hardhat:` in `ape-config.yaml` and don't put anything underneath it. It should error the plugin out.
Now, with this PR, you can use `-v debug` and can see the full stack trace, the same as it works elsewhere and should work.

### Checklist

### DIFF
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -54,17 +54,10 @@ def verbosity_option(cli_logger):
     names_str = f"{', '.join(level_names[:-1])}, or {level_names[-1]}"
 
     def decorator(f):
-        def _set_level(ctx, param, value):
-            log_level = getattr(LogLevel, value.upper(), None)
-            if log_level is None:
-                raise click.BadParameter(f"Must be one of {names_str}, not {value}.")
-
-            cli_logger.set_level(log_level.name)
-
         return click.option(
             "--verbosity",
             "-v",
-            callback=_set_level,
+            callback=lambda ctx, param, value: cli_logger._load_from_sys_argv(),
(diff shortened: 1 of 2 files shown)


src/ape/cli/options.py
```diff
@@ -54,17 +54,10 @@ def verbosity_option(cli_logger):
...
-            if log_level is None:
-                raise click.BadParameter(f"Must be one of {names_str}, not {value}.")
-
-            cli_logger.set_level(log_level.name)
-
         return click.option(
             "--verbosity",
             "-v",
-            callback=_set_level,
+            callback=lambda ctx, param, value: cli_logger._load_from_sys_argv(default=value),
```
fubuloubu, line 60: This has to be a lambda?

unparalleled-js: No?
We can make a new definition, but it is just a 1-liner method so it seems easier.
What you got against lambdas?

NotPeopling2day (approved):

Issue #573, opened 2022-03-14
ABI Methods that return arrays return tuples instead of lists


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.0

$ ape plugins list
Installed Plugins:
  vyper           0.1.0
  solidity        0.1.0
  hardhat         0.1.0
```

* Python Version: 3.9.9
* OS: linux

### What went wrong?

`getDelegated` returns `address[] memory` in Solidity:
```py
>       assert voter.getDelegated(shark) == [whale]
E       AssertionError: assert ('0xc89d42189...d628176a1e7',) == [<TestAccount...5d628176A1E7>]
```

### How can it be fixed?
Overall, return values need to be parsed better into Python types

#574

PR #684, opened 2022-04-30
fix: add handling for struct outputs

fubuloubu:
### What I did
Adds handling for methods that output struct values

fixes: #574
fixes: #573 
requires: https://github.com/ApeWorX/ethpm-types/pull/28

### How I did it
If we detect that a return value is a struct (a tuple with members that have names), then create a dataclass-based "struct definition" and return the data inside of that ad-hoc struct.

### How to verify it
Create a contract with structs in it:

```vyper
struct Test:
    a: address
    b: bytes32

struct Nested:
    t: Test

@view
@external
def foo() -> Test:
    return Test({a: msg.sender, b: block.prevhash})


@view
@external
def bar() -> Nested:
    return Nested({t: Test({a: msg.sender, b: block.prevhash})})
```

then try it:
```py
>>> r = c.foo()
>>> r.a  # use like a struct
'0x1e59ce931b4cfea3fe4b875411e280e173cb7a9c'
>>> r[0]  # use like a tuple too
'0x1e59ce931b4cfea3fe4b875411e280e173cb7a9c'
# NOTE: Output values need further processing to become real addresses
```

### Checklist

- [ ] Handle nested structs
- [ ] Consider allowing structs to have dict-like access too (e.g. `r["a"]` in example)


src/ape_ethereum/ecosystem.py
```diff
@@ -154,25 +155,56 @@ def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
...
                     value = self.decode_address(value)
+                except InsufficientDataBytes as err:
+                    raise DecodingError() from err
+
+            output_values.append(value)
+
+        if (
+            len(abi.outputs) == 1
+            and abi.outputs[0].components
+            and all(c.name != "" for c in abi.outputs[0].components)
```
NotPeopling2day, line 181: Currently failing `error: "ABIType" has no attribute "components"`

fubuloubu: requires https://github.com/ApeWorX/ethpm-types/pull/28 to merge

src/ape_ethereum/ecosystem.py
```diff
@@ -154,25 +155,56 @@ def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
...
+                        "__getitem__": lambda self, index: tuple(
+                            getattr(self, field) for field in self.__dataclass_fields__
+                        ).__getitem__(index)
+                    },
+                )
 
-                output_values.append(value)
+            struct_def = create_struct(
+                # NOTE: unnamed output structs appear as tuples with named members,
+                #       but we don't know of what struct type they are (because ABI encoding sucks)
```
unparalleled-js, line 199:
```suggestion
                #       but we don't know of what struct type they are (because of ABI encoding limitations)
```

src/ape_ethereum/ecosystem.py
```diff
@@ -154,25 +155,56 @@ def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
...
+
+            output_values.append(value)
+
+        if (
+            len(abi.outputs) == 1
+            and abi.outputs[0].components
+            and all(c.name != "" for c in abi.outputs[0].components)
+        ):
+
+            def create_struct(name: str, types: List[ABIType]):
```
unparalleled-js, line 184: Since this is only used the once directly after, we might not need this as a method

unparalleled-js (commented):
I can do some testing on this tomorrow!

src/ape_ethereum/ecosystem.py
fubuloubu: This might be a good utility function, to have a standard way of describing ABI struct defs

fubuloubu:
> I can do some testing on this tomorrow!

This PR is starting off as a proof-of-concept, feel free to make any changes necessary to clean up the code or improve it's functionality!

unparalleled-js:
One edge case I've noticed is that:

```solidity
    function getSecret() public onlyOwner view returns(uint256 secret) {
        return 123;
    }
```

has the same output ABI as a struct with a single item in it.

It makes sense to just return the single value in this case, but the user loses that (pointless) name given to it

unparalleled-js:
One edge case I've noticed is that:

```solidity
    function getSecret() public onlyOwner view returns(uint256 secret) {
        return 123;
    }
```

has the same output ABI as a struct with a single item in it.

It makes sense to just return the single value in this case, but the user loses that (pointless) name given to it

fubuloubu:
> It makes sense to just return the single value in this case, but the user loses that (pointless) name given to it

Yes, I think it does make more sense to return the single value, however in this scenario would it still be a "tuple" at the top level? I'd think not

Also as an aside, a common way that named outputs is often used in Solidity code is like the following

```solidity
function getSecret() public onlyOwner view returns(uint256 secret) {
        ...
        secret = 123; // Output is now set
        ...
    }
```

tests/functional/conftest.py
```diff
@@ -180,15 +401,46 @@ def owner(test_accounts):
...
+
+
+@pytest.fixture(params=("solidity", "vyper"))
+def contract_container(
+    request, solidity_contract_container, vyper_contract_container, networks_connected_to_tester
+):
+    return solidity_contract_container if request.param == "solidity" else vyper_contract_container
+
+
+@pytest.fixture(params=("solidity", "vyper"))
```
unparalleled-js, line 444: All tests that use this fixture use both a contract compiled with solidity as well as one with vyper and they have almost exactly the same ABI so they work for most tests.

NotPeopling2day (approved):
LGTM

unparalleled-js:
I just noticed this PR does not handle struct output.
I am wondering if `EcosystemAPI` needs change for that to work properly.. I see there is a TODO in `contracts/base.py` for handling struct output

src/ape_ethereum/ecosystem.py
```diff
@@ -88,6 +91,34 @@ class Block(BlockAPI):
     """
 
 
+def parse_output_type(output_type: str) -> Union[str, Tuple]:
```
fubuloubu, line 94: If you're unsure about a recursive function, it's usually a good indication to introduce fuzzing into the mix. Perhaps something like:

```py
from hypothesis import strategies as st

PRIMITIVE_TYPES = ["uint256", "address", ...]

@given(complex_type=st.tuples(st.one_of(PRIMITIVE_TYPES), min_length=1, max_length=20))
def test_output_parsing(complex_type):
    assert parse_output_type(",".join(complex_type)) == complex_type
```

unparalleled-js:
- [ ] Edge case that is not currently ideal: tuples with structs in them. The entire thing gets interpreted as a single tuple because of ABI strangeness .. Going to look into possible solutions

src/ape_ethereum/ecosystem.py
unparalleled-js: I added a test; it helped me find an edge case.

src/ape_ethereum/structs.py
```diff
@@ -0,0 +1,91 @@
+from dataclasses import make_dataclass
+from typing import Any, List, Optional, Union
+
+from ethpm_types.abi import ABIType, MethodABI
+
+
+class StructParser:
```
fubuloubu, line 7: Can we move this to `ape.utils`?

src/ape/utils/misc.py
```diff
@@ -0,0 +1,405 @@
...
+    """
+    Injected properties are injected class variables that must be set before use
+    **NOTE**: do not appear in a Pydantic model's set of properties.
+    """
+
+    def __get__(self, *args):
+        raise ValueError("Value not set. Please inject this property before calling.")
+
+
+class ManagerAccessMixin:
```
fubuloubu, line 247: Feel like this stuff and below should be in their own separate modules

fubuloubu:
Looks good!

unparalleled-js (approved):

PR #689, opened 2022-05-03
feat: allow ecosystem-specific kwargs in deploy

unparalleled-js:
### What I did

In `ape-starknet` to deploy to `alpha mainnet` currently, you need a token for the whitelisted contract you wish to deploy.
There are several ways to set this, but one way should be via `deploy(token="MY_TOKEN")`. However - currently, that is not liked by the ape core framework. This PR addresses this!

### How I did it

Only convert known kwargs and return the rest as-is for ecosystem plugins to handle on their own.

### How to verify it

TBD

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -17,15 +17,18 @@ if TYPE_CHECKING:
     from ape.managers.networks import NetworkManager
 
 
-def _convert_kwargs(kwargs, converter):
-    return {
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):

unparalleled-js:
@fubuloubu Sorry, just added a test!

fubuloubu (approved):

PR #692, opened 2022-05-04
fix: always return chain id from network when using live network

unparalleled-js:
### What I did

Fixes an issue when using `mainnet-fork:hardhat` and you do `vars(provider)`, it would fail with `ProviderNotConnectedError` because it as trying to get the chain ID from the mainnet upstream provider (which is not connected really).

### How I did it

The fix is to do exactly the same thing that we are doing except **in a better, more optimized order**.
The solution: Detect if using a live network and return the hardcoded chain ID without first checking if `web3` is a thing.

### How to verify it

Connect to mainnet fork:

```bash
ape console --network :mainnet-fork:hardhat
```

Run this:

```python
vars(networks.provider)
```

It should not fail anymore. Before, it did.

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -20,7 +20,7 @@ from pydantic import Field, validator
 from web3 import Web3
 
 from ape.api.config import PluginConfig
-from ape.api.networks import NetworkAPI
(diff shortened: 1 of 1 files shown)


NotPeopling2day (approved):
LGTM

Issue #671, opened 2022-04-26
Sign: [y/N]: after first signed tx, this message doesn't react on N


SwapOperator:
### Environment information

* OS: macOS
* Python Version: x.x.x
* `ape` and plugin versions: 0.2.1

```sh
$ ape --version
0.2.1

$ ape plugins list
ape-infura 0.2.0
ape-vyper 0.2.0
ape-etherscan 0.2.0
```

* Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
contracts_folder: contracts
```

### What went wrong?

Well, I use APE on testnet to generate a bunch of data for my web interface. With old version (I'm not sure what exactly version I have, maybe 1.5 or maybe 1.6) I ran with command ```ape run deploy_testing --network ethereum:rinkeby:infura``` I signed the first tx:
```
Sign:  [y/N]: y
Enter passphrase to unlock 'trader03' []: 
Leave 'trader03' unlocked? [y/N]: y
```
After that I just leave unlocked account and wait when script will done.

Well, I decided to rewrite some test from Brownie to APE and check how it will going. I installed ape-hardhat through PyCharm Python Interpreter. As I understood APE updated too automatically. 

If the script has:

```
    owner = accounts.load("trader03")
    owner.set_autosign(True)
``` 

I hasn't any problems. But if delete these strings and continue to Sign Tx with hands when the next message appear:
```
Sign: [y/N]: 
```
it will react only positive, so if I type y,Y,n,N, or tap on ENTER, the result will be always Yes.

How to reproduce:

token: https://github.com/vyperlang/vyper/blob/master/examples/tokens/ERC20.vy

deploy_testing.py:

```
from ape import project
from ape import accounts

class InitToken(object):

    def load_accs(self, type):
        acc = accounts.load("trader03")

        if type > 0:
          acc.set_autosign(True)

        return acc1

    def create_token(self, name, symbol, decimals, supply,  owner):
        token = owner.deploy(project.Token, name, symbol, decimals, supply)
        return token

    def token_approve(self, token, receiver, amount, sender):
        return token.approve(receiver, int(amount), sender=sender)

random_addr = 0xDf032Bc4B9dC2782Bb09352007D4C57B75160B15
### works fine with set_autosign
deploy = InitToken()

main_account = deploy.load_accs(1)
main_token = start_deploy.create_token("NAME", "SMBL", main_account)
[...]

PR #693, opened 2022-05-04
fix: bug preventing "no" from working in signing prompt

unparalleled-js:
### What I did

fixes: #671 

### How I did it

Removed unneeded check and added lots of tests to make sure is working.
Please let me know if there are missing test cases.

### How to verify it

1. Transact using keyfile account and leave account unlocked
2. Transact again in same session only when it asks you to sign, say "no". It should not sign! Before, it was.

### Checklist


NotPeopling2day (approved):
LGTM

PR #694, opened 2022-05-04
chore: upgrade pytest

unparalleled-js:
### What I did

Bump pytest because I keep installing Ape in an environment where pytest is in the 7 range.

### How I did it

bumps!

### How to verify it

Your tests to run / work as expected for your ape projects!!!
Remember: pytest is a core dependency.

### Checklist


NotPeopling2day (dismissed):
LGTM

fubuloubu (approved):

fubuloubu:
although probably need to update every plugin too

unparalleled-js:
> although probably need to update every plugin too

It only affects dev-installs of plugins, so it is not critical / breaking. But yeah, it would need to happen

unparalleled-js:
Let's save this for the next big release anyway and then adjust pytest dev dependency at the same time!

unparalleled-js:
I am pretty sure we can support both 6 and 7, which makes this not a breaking change

fubuloubu (approved):
As long as the wide pin causes no breaking changes, that's fine, I do worry about features we might use which exist in one major and not the other

PR #697, opened 2022-05-06
docs: update ape console with examples

Ninjagod1251:
### What I did

fixes: #

### How to verify it

We added examples to our userguide for the console.

### Checklist

### DIFF
--- a/docs/userguides/console.md
+++ b/docs/userguides/console.md
@@ -52,3 +52,128 @@ Out[1]: 1
 ### Global Extras
 
 You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of what project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
+
+### Console Variables
+
+* **accounts**
+
+provides access to account manager for loading, creating, and interfacing with accounts
+
+```bash
+$ ape accounts generate demo_account
+Add extra entropy for key generation...: 
+Create Passphrase: 
+Repeat for confirmation: 
+SUCCESS: A new account '0xB1b1A92A6f7435ba375e4cc470aED1e951e92fe5' has been added with the id 'demo_account'
+$ ape console
+
+In [1]: accounts.load("demo_account")
+Out[1]: <KeyfileAccount address=0xB1b1A92A6f7435ba375e4cc470aED1e951e92fe5 alias=demo_account>
+IN [2]: exit
+
+$ ape accounts delete demo_account
+Enter passphrase to delete 'demo_account' []: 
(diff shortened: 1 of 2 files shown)


docs/userguides/console.md
```diff
@@ -52,3 +52,128 @@ Out[1]: 1
 ### Global Extras
 
 You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of what project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
+
+### Console Variables
```
unparalleled-js, line 56: I think less is more for console namespace. We should just create a table with a small description; you can even just link to the manager Python docs for each item in the namespace.

Markdown table: https://markdown.land/markdown-table
Example doc to link to: https://docs.apeworx.io/ape/stable/methoddocs/managers.html#module-ape.managers.accounts

Ninjagod1251: The manager python doc does not have examples of what a user can do with them.

The purpose of this example is show the user how to use the console effectively. 

when @NotPeopling2day is going to do the tutorial. It would be nice to have something convenient to reference to.


We could move all the example to the managers. But it is not clear for a user to know to go to managers to look for sample uses of the console variables

unparalleled-js: I am not saying move anything, just can link to it and keep it more consolidated here

unparalleled-js: Look at the testing guide; it lists all the available fixtures but does not explain what accounts are in ape - just links to the part that does, so it is in one spot and easy to lookup and does not require keeping N number of docs about the same thing in-line with each other

fubuloubu: > Look at the testing guide; it lists all the available fixtures but does not explain what accounts are in ape - it is in one spot and easy to lookup but does keeps it focuses on the fixture-side of things rather than going into depth about what are accounts are in ape, etc

Chris and I talked a bit at length about this today, definitely recommend going the route of focusing on the right things in the right sections, and deep linking to other content where relevant

Ninjagod1251:
For reference: we are going to move this information to ape-academy and just post a link to it.

unparalleled-js:
> For reference: we are going to move this information to ape-academy and just post a link to it.

I think this guide does need an enumerated list of all the available namespace items though. Just a like a simple list:

```
accounts | Access to the account manager](link/to/account-manager)
project    | Access to the project manager](link/to/account-manager)
```

unparalleled-js:
@Ninjagod1251  This is what I was trying to get you to do: https://github.com/ApeWorX/ape/pull/770
I do love the linking to the Ape Academy as well!

I am going to close this PR as the work is being done both in #770 as well as #768

PR #698, opened 2022-05-08
feat: support contract namespaces

unparalleled-js:
### What I did

* Allows interface-based projects to use namespace-techniques to avoid duplicate contract names. This allows the re-use of `interface.json` so long as it is in a different directory.

### How I did it

* Create a bridge class `ContractNamespace` that serves up the next node (either another `ContractNamespace` or the `ContractContainer` (the leaf).
* Change `pm` first-class plugin to honor set the full `.`-based path contract names.

### How to verify it

Create a project like this:

```yaml
contracts:
  namespace0:
    - inteface.json
  namespace1:
     namespace:2
        - interface.json
```

See test for fuller example!

**Also, this will help with cairo specifically**:  https://github.com/ApeWorX/ape-cairo/pull/6

### Checklist


src/ape_pm/compiler.py
```diff
@@ -26,18 +26,20 @@ def compile(
...
             abi = json.loads(path.read_text())
 
-            source_id = (
-                str(get_relative_path(path, base_path))
-                if base_path and path.is_absolute()
-                else str(path)
-            )
+            if base_path:
+                source_id = str(get_relative_path(path.absolute(), base_path))
+                contract_name = source_id.replace(".json", "").replace("/", ".")
```
fubuloubu, line 31: Was curious if it was important that interface compilers use the "."

fubuloubu (approved):

src/ape_pm/compiler.py
unparalleled-js: Not really; I was thinking it was a nice feature and it helps us test that this feature works without relying on `ape-cairo` to do it.

But I don't necessarily need it so perhaps it is best to not change that for now.

fubuloubu (approved):

Issue #701, opened 2022-05-09
Query: Flatten BlockAPI and migrate to Transaction/Blocks Types


fubuloubu:
### Elevator pitch:
For database work, we want `TransactionAPI` and `BlockAPI` to be non-abstract types instead of api classes

### Value:
Remove them as API classes allows a cleaner refactor in prep for query engine work

### Dependencies:
N/A

### Design approach:
See Task List

### Task list:

- [x] Flatten `BlockAPI` structure (remove `GasDataAPI` and `ConsensusAPI`)
- [x] Migrate `TransactionAPI.serialize_transaction` to `EcosystemAPI`
- [ ] ~~Migrate `ReceiptAPI` and `BlockAPI` to `ape.types` and subclass `BaseModel` instead of `BaseInterfaceModel`~~

### Estimated completion date:
Weds 5/11

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

johnson2427:
I'm hung up on the last part here for the TransactionAPI

fubuloubu:
We may need to work with @unparalleled-js on this

unparalleled-js:
The third part seems good to be so long as Ecosystem plugins can still subclass the `TransactionAPI` and `BlockAPI` classes.

> I'm hung up on the last part here for the TransactionAPI

What about?

> We may need to work with @unparalleled-js on this

👋🏻

fubuloubu:
> so long as Ecosystem plugins can still subclass the `TransactionAPI` and `BlockAPI` classes.

This is actually the part we might need to work with you on changing

unparalleled-js:
> This is actually the part we might need to work with you on changing

Schedule some time on my calendar! Or else, draft a spec and send it to me and I will let you know if it will work for Starknet. (Is this about Starknet?)

johnson2427:
> The third part seems good to be so long as Ecosystem plugins can still subclass the `TransactionAPI` and `BlockAPI` classes.
> 
> > I'm hung up on the last part here for the TransactionAPI
> 
> What about?
> 
> > We may need to work with @unparalleled-js on this
> 
> 👋🏻

[...]

PR #705, opened 2022-05-10
Refactor/block api

johnson2427:
### What I did

Flattened the BlockAPI structure by removing GasDataAPI and ConsensusAPI
Migrated TranasctionAPI.serialize_transactions to EcosystemAPI

Currently working on subclassing with BaseModel instead of BaseInterfaceModel for TransactionAPI. This was completed for BlockAPI.

fixes: #701

### How I did it

See #### what I did

### How to verify it

run pytests, not complete yet though

### Checklist


src/ape_ethereum/ecosystem.py
```diff
@@ -136,7 +116,7 @@ def encode_address(cls, address: AddressType) -> RawAddress:
         return str(address)
 
     def serialize_transaction(self, transaction: TransactionAPI) -> bytes:
-        return transaction.serialize_transaction()
+        return self.serialize_transaction(transaction)
```
fubuloubu, line 119: Should be just moving this field over:

https://github.com/ApeWorX/ape/blob/3f154b239a783fc4659ba70bd489b2a580002379/src/ape_ethereum/transactions.py#L43

fubuloubu (commented):
I think a potentially complicating factor is the existence of multiple types of transactions that may need to be decoded differently to fulfil the current `TransactionAPI` format
https://github.com/ApeWorX/ape/blob/3f154b239a783fc4659ba70bd489b2a580002379/src/ape_ethereum/transactions.py#L61
https://github.com/ApeWorX/ape/blob/3f154b239a783fc4659ba70bd489b2a580002379/src/ape_ethereum/transactions.py#L78

src/ape/utils/__init__.py
```diff
@@ -29,6 +31,8 @@
     generate_dev_accounts,
 )
 
+EMPTY_BYTES32 = HexBytes("0x0000000000000000000000000000000000000000000000000000000000000000")
```
unparalleled-js, line 34: might make more sense to put this in `ape.utils.misc.py`

src/ape_ethereum/ecosystem.py
```diff
@@ -168,21 +148,19 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
...
-            gas_data=BlockGasFee.decode(data),
-            consensus_data=BlockConsensus.decode(data),
-            number=data.get("number"),
-            size=data.get("size"),
-            timestamp=data.get("timestamp"),
-            hash=data.get("hash"),
-            # TODO: when we flatten the Block structure, remove this hack.
-            parent_hash=data.get("parentHash") or data.get("parent_hash"),
-        )
+        if "gas_limit" in data:
```
unparalleled-js, line 151: Could you do something like 

```python
data["gasLimit"] = data.pop("gas_limit", data.get("gasLimit"))
```

for all these?
Example:

```python
>>> data
{'fooBar': 3}
>>> data.pop("foo_bar", data.get("fooBar"))
3
```

src/ape/api/transactions.py
```diff
@@ -54,12 +54,6 @@ def total_transfer_value(self) -> int:
 
         return self.value + self.max_fee
 
-    @abstractmethod
-    def serialize_transaction(self) -> bytes:
```
unparalleled-js, line 58: This method is called in at least two places other than blocks (providers and accounts).
Maybe move the logic to `EcosystemAPI` if you need it out of here?

unparalleled-js (commented):
Looking shnazzy

src/ape/api/providers.py
```diff
@@ -78,6 +71,15 @@ def validate_hexbytes(cls, value):
             raise ValueError(f"Hash `{value}` is not a valid Hexbyte.")
         return value
 
+    @validator("transaction_ids", each_item=True, pre=True)
+    def validate_transactions(cls, value):
+        return HexBytes(value)
```
NotPeopling2day, line 76: Just in case, I guess we should check that the value isn't already a `HexBytes` value?

```suggestion
    @validator("transaction_ids", each_item=True, pre=True)
    def validate_transactions(cls, value):
        if isinstance(value, HexBytes):
            return value
        return HexBytes(value)
```

johnson2427 (commented):
Feel pretty good about the changes here, all tests pass, blocks and transactions have been flattened. LMK what you guys think!

src/ape/api/accounts.py
```diff
@@ -209,7 +209,9 @@ def check_signature(
                     "Parameter 'signature' required when verifying a 'SignableMessage'."
                 )
         elif isinstance(data, TransactionAPI):
-            return self.address == Account.recover_transaction(data.serialize_transaction())
+            return self.address == Account.recover_transaction(
+                data.serialize_transaction()  # type: ignore
```
unparalleled-js, line 213: Is this type ignore needed? I think `recover_transaction()` should be able to accept bytes

unparalleled-js, line 213: Is this type ignore needed? I think recover_transaction() should be able to accept bytes

src/ape/api/providers.py
```diff
@@ -70,17 +71,31 @@ class BlockConsensusAPI(BaseInterfaceModel):
 
 class BlockAPI(BaseInterfaceModel):
     """
-    An abstract class representing a block and its attributes.
+    An abstract class representing block consensus-data,
```
unparalleled-js, line 74: Is this a copy-paste error? The `BlockAPI` contains more than just `consensus-data`.

src/ape/utils/__init__.py
```diff
@@ -55,6 +57,7 @@
     "GeneratedDevAccount",
     "generate_dev_accounts",
     "get_all_files_in_directory",
+    "EMPTY_BYTES32",
```
unparalleled-js, line 60: This is included twice. (line 49 and line 60)

src/ape_ethereum/ecosystem.py
```diff
@@ -252,7 +232,7 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfo]:
         return None
 
     def serialize_transaction(self, transaction: TransactionAPI) -> bytes:
-        return transaction.serialize_transaction()
+        return self.serialize_transaction(transaction)
```
unparalleled-js, line 235: This will cause an infinite loop!
We can delete this method if only needs to call the base class.
Otherwise, use `super()`

src/ape_ethereum/ecosystem.py
```diff
@@ -284,21 +264,19 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
...
+        if "gas_limit" in data:
+            data["gasLimit"] = data.pop("gas_limit")
+        if "gas_used" in data:
+            data["gasUsed"] = data.pop("gas_used")
+        if "parent_hash" in data:
+            data["parentHash"] = data.pop("parent_hash")
+        if "transaction_ids" in data:
+            data["transactions"] = data.pop("transaction_ids")
+        if "total_difficulty" in data:
+            data["totalDifficuly"] = data.pop("total_difficulty")
```
unparalleled-js, line 276:
```suggestion
            data["totalDifficulty"] = data.pop("total_difficulty")
```

unparalleled-js (changes requested):

fubuloubu (dismissed):
Looking pretty good otherwise though

src/ape/api/networks.py
```diff
@@ -2,11 +2,17 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Type
 
+from eth_account import Account as EthAccount  # type: ignore
+from eth_account._utils.legacy_transactions import (
```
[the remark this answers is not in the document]
unparalleled-js: Since we are using `==` in setup.py for the eth-account version, this is not a big deal. I think we have a ticket in the backlog for replacing this elsewhere too anyways.

unparalleled-js (dismissed):
Looks good! I am good with merging this after 0.2.8 release.

src/ape/api/providers.py
```diff
@@ -73,14 +74,26 @@ class BlockAPI(BaseInterfaceModel):
...
-    hash: Optional[Any] = None
+    hash: Optional[Any] = None  # NOTE: pending block does not have a hash
     number: Optional[int] = None
-    parent_hash: Optional[Any] = None
+    parent_hash: Any = Field(
+        EMPTY_BYTES32, alias="parentHash"
+    )  # NOTE: genesis block has no parent hash
     size: int
     timestamp: int
+    gas_limit: int = Field(alias="gasLimit")
```
unparalleled-js, line 85: I think the gas and consensus information should only be added to the models in `ape_ethereum`.

And the DBs should be ecosystem specific, right? Starknet TXNs and blocks differ a lot of Ethereum's.

You can see Starknet's blocks here: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L963-L1024

unparalleled-js: Here is a better link: https://starknet.io/docs/hello_starknet/cli.html#get-block

You can see blocks have `gas_price` but nothing else. No base fee. etc.
And transactions only have `max_fee`. Not priority fee etc.

johnson2427 (commented):
Looks good to me. Not quite sure how this affects the Cache yet, but I'll make it work!

PR #706, opened 2022-05-11
feat: handle structs within tuples and arrays in decode_returndata

unparalleled-js:
### What I did

Allows output like this to work properly:

```vyper
@view
@external
def getNestedStructWithTuple1() -> (NestedStruct1, uint256):
    return (NestedStruct1({t: MyStruct({a: msg.sender, b: block.prevhash}), foo: 1}), 1)

@view
@external
def getNestedStructWithTuple2() -> (uint256, NestedStruct2):
    return (2, NestedStruct2({foo: 2, t: MyStruct({a: msg.sender, b: block.prevhash})}))
```

### How I did it

Try to parse structs always and all the way down.

### How to verify it

See new test added
Old tests pass

### Checklist


unparalleled-js:
NOTE: Arrays of structs is not working, and it seems it may be related to `eth-abi`, will have to circle back on this: https://github.com/ethereum/eth-abi/issues/169

- [ ] try to upgrad eth-abi and figure out

src/ape/utils/abi.py
```diff
@@ -86,6 +123,7 @@ def length(struct) -> int:
         # NOTE: Should never be "_{i}", but mypy complains and we need a unique value
         [m.name or f"_{i}" for i, m in enumerate(types)],
         namespace={"__getitem__": get_item, "__eq__": is_equal, "__len__": length},
+        bases=(Struct,),
```
fubuloubu, line 126: Why use empty class as base? Just for subclass checks?

fubuloubu (approved):

src/ape/utils/abi.py
unparalleled-js: Ya, thought about putting a method but then we have to use `getattr`, and I thought the subclass check seemed nicer
but am open to alternatives.

fubuloubu: I like it, it works. Just add a comment about that use case

src/ape/utils/abi.py
```diff
@@ -12,14 +27,28 @@ def __init__(self, method_abi: MethodABI):
     def default_name(self) -> str:
         return f"{self.method_abi.name}_return"
 
-    def parse(self, output_types: List[ABIType], values: List) -> Any:
+    def parse(self, output_types: List[ABIType], values: Union[List, Tuple]) -> Any:
```
NotPeopling2day, line 30: Doc string? Not sure that it's needed since it's a util?

NotPeopling2day (approved):

unparalleled-js:
- [ ] Welp... Found another issue with structs containing arrays.

fubuloubu (approved):

PR #712, opened 2022-05-12
feat: cache contract types from deploy and explorer

unparalleled-js:
### What I did

* Caches contract types by address in each network.

* Swaps out `ape.Contract` with `chain.contracts.instance_at`

* Caches contract types from explorer when the explorer is used to fetch it.

**I believe this is a necessary step for getting an environment that I can play with traces in, as it will allow me to interact locally with a more complex transaction and do the tracing from there.**

### How I did it

* Have in-memory cache where everything is put
* Have disc cache where contracts are put in live networks

### How to verify it

See tests!

### Checklist


unparalleled-js:
Failing hypothesis test resolved by this PR: https://github.com/ApeWorX/ape/pull/711

src/ape/utils/basemodel.py
```diff
@@ -74,7 +76,36 @@ def provider(self) -> "ProviderAPI":
             raise ProviderNotConnectedError()
         return self.network_manager.active_provider
 
-    def create_contract_container(self, contract_type: "ContractType") -> "ContractContainer":
+    def create_contract_type(self, address: AddressType) -> Optional[ContractType]:
```
fubuloubu, line 79: The name being kind of weird (maybe `get_contract_type` makes more sense)

fubuloubu:
```suggestion
    def get_contract_type(self, address: Union[str, AddressType]) -> Optional[ContractType]:
```

And then add this to the body:
```py
converted_address: AddressType = conversion_manager.convert(address, AddressType)
```

src/ape/utils/basemodel.py
```diff
@@ -74,7 +76,36 @@ def provider(self) -> "ProviderAPI":
...
+        Returns:
+            ContractType
+        """
+        network = self.provider.network
+        address = self.conversion_manager.convert(address, AddressType)
+        contract_type = None
+
+        if address in network.contract_cache:
+            # Check contract cache (e.g. previously deployed/downloaded contracts)
+            contract_type = network.contract_cache[address]
```
fubuloubu, line 95: The idea was eventually to store these on disk, but this works for now

Directory structure:

```
.ape/
  [ecosystem-name]/
    [network-name]/
      contract-types/
        [address].json  # ABI-json file
```

unparalleled-js: This directory structure seems opposite? We need to be able to look up a contract type by address but this structure is for looking up the address of a contract type.

We get an address from the trace and need to be able to look it up. Also, caching the results from explorer.get_contract_type(address).

fubuloubu:
```py
json_abi = json.loads(Path(f"~/.ape/{ecosystem_name}/{network_name}/contract-types/{address}.json").read_text())
```

banteg:
a good add would be a lock per ecosystem-network-address to avoid bombarding etherscan with requests from different threads when they all miss the cache simultaneously

fubuloubu:
> a good add would be a lock per ecosystem-network-address to avoid bombarding etherscan with requests from different threads when they all miss the cache simultaneously

We don't have threads, but this would be something to consider if running Ape in a multi-threaded environment

src/ape/managers/chain.py
```diff
@@ -359,6 +363,73 @@ def revert_to_block(self, block_number: int):
...
+class ContractCache(BaseManager):
+    _local_contracts: Dict[AddressType, ContractType] = {}
+
+    @property
+    def network(self) -> NetworkAPI:
+        return self.provider.network
+
+    @property
+    def _contract_types_path(self) -> Path:
+        return self.network.ecosystem.data_folder / "contract_types"
```
fubuloubu: should be `data_folder / self.network.name / "contract_types"`

src/ape/managers/chain.py
```diff
@@ -359,6 +363,73 @@ def revert_to_block(self, block_number: int):
         }
 
 
+class ContractCache(BaseManager):
```
fubuloubu, line 366: Can it subclass `dict`?

Then will have these methods:
```py
ct = chain.contracts[addr]
ct = chain.contract.get(addr, None)
```
etc.

could maybe even have `chain.contracts[addr] = ct`

src/ape/contracts/base.py
```diff
@@ -661,15 +661,15 @@ def deploy(self, *args, **kwargs) -> ContractInstance:
...
+            contract_instance.address, contract_instance.contract_type
+        )
         return contract_instance
 
 
 def _Contract(
     address: Union[str, BaseAddress, AddressType],
-    networks: "NetworkManager",
+    chain: "ChainManager",
     conversion_manager: "ConversionManager",
```
fubuloubu, line 673: Having to pass these through is annoying, would be nice to add this as a method to `ContractCache` instead to leverage `BaseManager`

tests/functional/test_accounts.py
```diff
@@ -79,9 +79,17 @@ def test_transfer_using_type_0(sender, receiver):
...
+    contract = owner.deploy(contract_container)
+    assert contract.address
+
+    # Verify can reload same contract from cache
+    contract_from_cache = ape.Contract(contract.address)
+    assert contract_from_cache.contract_type == contract.contract_type
+    assert contract_from_cache.address == contract.address
+
+    # Clean up for next test
+    del chain.contracts._local_contracts[contract_from_cache.address]
```
fubuloubu, line 92: An `assert` raised above could bork this up

src/ape/managers/chain.py
```diff
@@ -359,6 +363,113 @@ def revert_to_block(self, block_number: int):
...
+    """
+
+    _local_contracts: Dict[AddressType, ContractType] = {}
+
+    @property
+    def _network(self) -> NetworkAPI:
+        return self.provider.network
+
+    @property
+    def _contract_types_path(self) -> Path:
```
fubuloubu, line 384: Might be clearer to rename this to `_contract_types_cache`

src/ape/managers/chain.py
```diff
@@ -359,6 +363,113 @@ def revert_to_block(self, block_number: int):
...
+            return None
+
+        try:
+            contract_type = self._network.explorer.get_contract_type(address)
+        except Exception as err:
+            logger.error(f"Unable to fetch contract type at '{address}' from explorer.\n{err}")
+            return None
+
+        # Cache contract so faster look-up next time.
+        self._cache_contract_to_disk(address, contract_type)
```
fubuloubu, line 464: I believe `ExplorerAPI.get_contract_type` has a return of `Optional[ContractType]` so should probably check it's not `None` before caching?

src/ape/managers/chain.py
```diff
@@ -359,6 +363,113 @@ def revert_to_block(self, block_number: int):
...
+    def _get_contract_type_from_disk(self, address: AddressType) -> Optional[ContractType]:
+        if not self._contract_types_path.is_dir():
+            return None
+
+        address_file = self._contract_types_path / f"{address}.json"
+        if not address_file.is_file():
+            return None
+
+        contract_type_data = json.loads(address_file.read_text())
+        return ContractType.parse_obj(contract_type_data)
```
fubuloubu, line 451: I think `BaseModel` has `.parse_raw()` and `.json()` methods to avoid having to import the json module here

fubuloubu (commented):
Looking pretty good so far

src/ape/managers/chain.py
```diff
@@ -359,6 +363,113 @@ def revert_to_block(self, block_number: int):
...
+        if is_local or self._network.name.endswith("-fork"):
+            # For fork networks, try the local cache first.
+            contract_type = self._local_contracts.get(address)
+
+        if contract_type or is_local:
+            return contract_type
+
+        return self._get_contract_type_from_disk(address) or self._get_contract_type_from_explorer(
+            address
+        )
```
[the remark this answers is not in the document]
unparalleled-js: Black ran

src/ape/managers/chain.py
unparalleled-js: hmm, yes but still have to have custom definitions for them because this is not a static list

src/ape/managers/chain.py
```diff
@@ -359,6 +361,130 @@ def revert_to_block(self, block_number: int):
...
+        Cache the given contract type. Contracts are cached in memory per session.
+        In live networks, contracts also get cached to disc at
+        ``.ape/{ecosystem_name}/{network_name}/contract_types/{address}.json``
+        for faster look-up next time.
+
+        Args:
+            address (AddressType): The on-chain address of the contract.
+            contract_type (ContractType): The contract's type.
+        """
+
```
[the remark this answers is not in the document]
unparalleled-js: tickets > TODOs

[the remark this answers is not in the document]
unparalleled-js: https://github.com/ApeWorX/ape/issues/733
https://github.com/ApeWorX/ape/issues/734

fubuloubu (approved):
Looking great for a first draft! Future PRs to come with non-immutable contracts and proxy handling

Issue #713, opened 2022-05-13
Bug with the fixture scoping feature added in 0.2.0 


de33:
### Environment information

* OS: macOS 
* Python Version: 3.8.2
* `ape` and plugin versions: ape 0.2.2.dev1, vyper 0.2.0 solidity 0.2.0

```sh
$ cat ape-config.yaml
plugins:
- name: vyper
- name: solidity


dependencies:
  - name: OpenZeppelin
    github: OpenZeppelin/openzeppelin-contracts
    version: 4.5.0

solidity:
  import_remapping:
    - "@OpenZeppelin=OpenZeppelin/4.5.0"


Tried to run `ape test` in local version of https://github.com/fubuloubu/ERC4626.git

### 

ERC4626 main % ape test
INFO: Compiling 'ERC4626.sol'.
=================================================== test session starts ===================================================
platform darwin -- Python 3.8.2, pytest-6.2.5, py-1.11.0, pluggy-0.13.1
rootdir: /Users/david/ERC4626
plugins: eth-ape-0.2.2.dev1+g145ef053, web3-5.29.0
collected 6 items

tests/test_methods.py EEEEEE                                                                                        [100%]

========================================================= ERRORS ==========================================================
________________________________________ ERROR at setup of test_asset[VyperVault] _________________________________________

.0 = <list_iterator object at 0x11a0b3490>

>   scopes = [item._fixtureinfo.name2fixturedefs[f][0].scope for f in item.fixturenames]
E   KeyError: 'request'

/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/site-packages/eth_ape-0.2.2.dev1+g145ef053-py3.8.egg/ape/pytest/runners.py:99: KeyError
_______________________________________ ERROR at setup of test_asset[SolidityVault] _______________________________________

.0 = <list_iterator object at 0x11a76eb80>

>   scopes = [item._fixtureinfo.name2fixturedefs[f][0].scope for f in item.fixturenames]
E   KeyError: 'request'

/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/site-packages/eth_ape-0.2.2.dev1+g145ef053-py3.8.egg/ape/pytest/runners.py:99: KeyError
_____________________________________ ERROR at setup of test_max_methods[VyperVault] ______________________________________

.0 = <list_iterator object at 0x11a7824f0>

[...]

PR #716, opened 2022-05-13
fix: bug preventing the use of pytest built-in fixtures in ape tests

unparalleled-js:
### What I did

fixes: #713 
Also adds docs to fixtures

### How I did it

Use session fixturemanager as point-of-reference for getting fixture info 

### How to verify it

You can no run the tests in https://github.com/fubuloubu/ERC4626

### Checklist

### DIFF
--- a/src/ape/pytest/fixtures.py
+++ b/src/ape/pytest/fixtures.py
@@ -15,14 +15,27 @@ class PytestApeFixtures(ManagerAccessMixin):
 
     @pytest.fixture(scope="session")
     def accounts(self) -> List[TestAccountAPI]:
+        """
+        The ape account test manager containing pre-funded test accounts.
+        """
+
         return self.account_manager.test_accounts
 
     @pytest.fixture(scope="session")
     def chain(self) -> ChainManager:
+        """
+        The chain manager for manipulating the blockchain in your test
+        setups and tear-downs.
+        """
+
         return self.chain_manager
 
     @pytest.fixture(scope="session")
     def project(self) -> ProjectManager:
+        """
+        The project manager for accessing contract types and dependencies.
+        """
+
         return self.project_manager
 
     def _isolation(self) -> Iterator[None]:
--- a/src/ape/pytest/fixtures.py
+++ b/src/ape/pytest/fixtures.py
@@ -10,6 +10,10 @@ from ape.utils import ManagerAccessMixin
 
 
 class PytestApeFixtures(ManagerAccessMixin):
+    # NOTE: Avoid including links, markdown, or rst in method-docs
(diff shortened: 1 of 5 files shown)


src/ape/pytest/fixtures.py
```diff
@@ -15,14 +15,27 @@ def __init__(self):
 
     @pytest.fixture(scope="session")
     def accounts(self) -> List[TestAccountAPI]:
+        """
+        The ape account test manager containing pre-funded test accounts.
```
unparalleled-js, line 19: NOTE: When you do `pytest -q --fixtures` with ape installed, you get special docs for each fixture built from these doc strs.

So I want to avoid putting links in these.

Before, it was read and bad looking because they are missing. Now, it looks more profess

tests/integration/cli/projects/test/tests/test_fixture_isolation.py
```diff
@@ -27,3 +27,7 @@ def test_isolation_first(alice, bob, chain):
 def test_isolation_second(bob, chain):
     assert chain.provider.get_block("latest").number == 1
     assert bob.balance == 1_000_001 * 10**18
+
+
+def test_can_use_built_in_fixtures(chain, capsys):
```
unparalleled-js, line 32: This is the test within in the tests that prove the bug is resolved. If you add this test on `main`, the tests in `test_test.py` fail.

src/ape/pytest/fixtures.py
fubuloubu: Can we add a comment saying that we don't want links

src/ape/pytest/runners.py
```diff
@@ -95,8 +95,13 @@ def pytest_runtest_setup(self, item):
...
 
-        # list of scopes for each fixture of the test (including `autouse` fixtures)
-        scopes = [item._fixtureinfo.name2fixturedefs[f][0].scope for f in item.fixturenames]
+        fixture_map = item.session._fixturemanager._arg2fixturedefs
+        scopes = []
+        for name, fixture in fixture_map.items():
+            if name in item.fixturenames:
+                definitions = fixture_map[name]
+                for definition in definitions:
+                    scopes.append(definition.scope)
```
fubuloubu, line 104: Might be a little faster and easier to read
```suggestion
        scopes = [
            definition.scope
            for name, definitions in fixture_map.items()
            if name in item.fixturenames
            for definition in definitions
        ]
```

fubuloubu (dismissed):

fubuloubu (approved):

fubuloubu (approved):

Issue #728, opened 2022-05-16
bug: `ape_accounts.accounts.KeyfileAccount.transfer(receiver)` tries to send entire balance


sabotagebeats:
### Environment information

* OS: linux wsl2 ubuntu
* Python Version: 3.9.9
* `ape` and plugin versions: 
`ape 0.2.4`
```
Installed Plugins:
  hardhat      0.2.1
  solidity     0.2.3
  etherscan    0.2.1
  alchemy      0.2.0
  fantom       0.2.0
  tokens       0.2.0
  vyper        0.2.1.dev1+g80c2bf9
  infura       0.2.0
  ens          0.2.0
```

```
hardhat:
  fork:
    ethereum:
      mainnet:
        upstream_provider: alchemy(apeworx)
```

### What went wrong?

`ape_accounts.accounts.KeyfileAccount.transfer(receiver)` will attempt to send entire balance and then fail for not enough balance for gas. 
`VirtualMachineError: Sender does not have enough balance to cover transaction value and gas  (has 999999999981624999999999, needs 999999999997705462516999)`
I would hate to send an entire balance accidentally because I forgot the amount argument. 
I'm not sure if this is intended to be like this or not. 

```
In [27]: accounts.test_accounts[0].transfer(accounts.test_accounts[1])
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
File ~/.pyenv/versions/apeworx/lib/python3.9/site-packages/ape_test/providers.py:82, in LocalProvider.send_transaction(self, txn)
     81 try:
---> 82     txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
     83 except ValidationError as err:

File ~/.pyenv/versions/apeworx/lib/python3.9/site-packages/web3/eth.py:831, in Eth.send_raw_transaction(self, transaction)
    830 def send_raw_transaction(self, transaction: Union[HexStr, bytes]) -> HexBytes:
--> 831     return self._send_raw_transaction(transaction)

File ~/.pyenv/versions/apeworx/lib/python3.9/site-packages/web3/module.py:57, in retrieve_blocking_method_call_fn.<locals>.caller(*args, **kwargs)
     56 result_formatters, error_formatters, null_result_formatters = response_formatters
---> 57 result = w3.manager.request_blocking(method_str,
     58                                      params,
     59                                      error_formatters,
     60                                      null_result_formatters)
     61 return apply_result_formatters(result_formatters, result)

[...]

PR #731, opened 2022-05-18
fix: transfer send_everything does not work properly

sabotagebeats:
### What I did
`transfer(receiver)` now throws error message
`transfer(receiver, send_everything=False)` now throws error message
`transfer(receiver, send_everything=True)` now works

fixes: https://github.com/ApeWorX/ape/issues/728

### How I did it
discussed at length https://github.com/ApeWorX/ape/issues/728

### How to verify it
test cases have been added 

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -95,8 +95,17 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         if send_everything:
             if txn.max_fee is None:
                 raise TransactionError(message="Max fee must not be None.")
-
-            txn.value = self.balance - txn.max_fee
+            if txn.gas_limit:
+                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
+                error_message = f"Sender does not have enough {self.balance}\
+                    to cover transaction value {txn.value} and \
+                    gas {txn.max_fee * txn.gas_limit}"
+            else:
+                txn.value = self.balance - txn.max_fee
+                error_message = f"Sender does not have enough {self.balance}\
+                    to cover transaction value {txn.value} and \
+                    gas {txn.max_fee}"
+            assert txn.value > 0, error_message
 
         txn.signature = self.sign_transaction(txn)
         if not txn.signature:
@@ -136,6 +145,15 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
 
         if value:
             txn.value = self._convert(value, int)
+
+        if not value:
+            assert (
+                "send_everything" in kwargs
+            ), "Transfer without value argument requires kwarg send_everything=True"
+            assert kwargs[
+                "send_everything"
+            ], "Transfer without value argument requires kwarg send_everything=True"
+            return self.call(txn, kwargs["send_everything"])
 
         return self.call(txn, send_everything=value is None)
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -145,9 +145,13 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
 
         if not value:
             if "send_everything" not in kwargs:
-                raise ValueError("Transfer without value argument requires kwarg send_everything=True")
-            if kwargs['send_everything'] != True:
-                raise ValueError("Transfer without value argument requires kwarg send_everything=True")
+                raise ValueError(
+                    "Transfer without value argument requires kwarg send_everything=True"
+                )
+            if not kwargs["send_everything"]:
+                raise ValueError(
+                    "Transfer without value argument requires kwarg send_everything=True"
+                )
             return self.call(txn, kwargs["send_everything"])
 
         return self.call(txn, send_everything=value is None)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -100,8 +100,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
                 txn.value = self.balance - (txn.max_fee * txn.gas_limit)
                 error_message = error_message + f"{txn.max_fee * txn.gas_limit}"
             else:
-                txn.value = self.balance - txn.max_fee
-                error_message = error_message + f"{txn.max_fee}"
+                raise TransactionError(message="The txn.gas_limit is not set.")
             assert txn.value > 0, error_message
 
         txn.signature = self.sign_transaction(txn)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -146,7 +146,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
                     "Kwarg send_everything=True requires transfer without value argument"
                 )
             txn.value = self._convert(value, int)
-            return self.call(txn, send_everything=value is None)
+            return self.call(txn)
 
         elif not kwargs.get("send_everything"):
             raise ValueError("Transfer without value argument requires kwarg send_everything=True")
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -102,7 +102,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
                 txn.value = self.balance - (txn.max_fee * txn.gas_limit)
             if txn.value <= 0:
                 error_message = error_message + f"{txn.max_fee * txn.gas_limit}"
-                raise ValueError(message=error_message)
+                raise ValueError(error_message)
 
         txn.signature = self.sign_transaction(txn)
         if not txn.signature:
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -93,18 +93,15 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         txn = self.prepare_transaction(txn)
 
         if send_everything:
+            error_message = "Sender does not have enough to cover transaction value and gas: "
             if txn.max_fee is None:
                 raise TransactionError(message="Max fee must not be None.")
             if txn.gas_limit:
                 txn.value = self.balance - (txn.max_fee * txn.gas_limit)
-                error_message = f"Sender does not have enough {self.balance}\
-                    to cover transaction value {txn.value} and \
-                    gas {txn.max_fee * txn.gas_limit}"
+                error_message = error_message + f"{txn.max_fee * txn.gas_limit}"
             else:
                 txn.value = self.balance - txn.max_fee
-                error_message = f"Sender does not have enough {self.balance}\
-                    to cover transaction value {txn.value} and \
-                    gas {txn.max_fee}"
+                error_message = error_message + f"{txn.max_fee}"
             assert txn.value > 0, error_message
 
         txn.signature = self.sign_transaction(txn)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -144,12 +144,10 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             txn.value = self._convert(value, int)
 
         if not value:
-            assert (
-                "send_everything" in kwargs
-            ), "Transfer without value argument requires kwarg send_everything=True"
-            assert kwargs[
-                "send_everything"
-            ], "Transfer without value argument requires kwarg send_everything=True"
+            if "send_everything" not in kwargs:
+                raise ValueError("Transfer without value argument requires kwarg send_everything=True")
+            if kwargs['send_everything'] != True:
+                raise ValueError("Transfer without value argument requires kwarg send_everything=True")
             return self.call(txn, kwargs["send_everything"])
 
         return self.call(txn, send_everything=value is None)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -97,13 +97,12 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
                 raise TransactionError(message="Max fee must not be None.")
             if not txn.gas_limit or txn.gas_limit is None:
                 raise TransactionError(message="The txn.gas_limit is not set.")
-            elif txn.value <= 0:
+            txn.value = self.balance - (txn.max_fee * txn.gas_limit)
+            if txn.value <= 0:
                 raise ValueError(
                     f"Sender does not have enough to cover transaction value and gas: \
                     {txn.max_fee * txn.gas_limit}"
                 )
-            else:
-                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
 
         txn.signature = self.sign_transaction(txn)
         if not txn.signature:
@@ -150,12 +149,9 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             return self.call(txn, send_everything=value is None)
 
         elif not kwargs.get("send_everything"):
-            raise ValueError(
-                "Transfer without value argument requires kwarg send_everything=True"
-            )
+            raise ValueError("Transfer without value argument requires kwarg send_everything=True")
         else:
             return self.call(txn, send_everything=True)
-
 
     def deploy(self, contract: "ContractContainer", *args, **kwargs) -> "ContractInstance":
         """
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -149,7 +149,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             txn.value = self._convert(value, int)
 
         if not value:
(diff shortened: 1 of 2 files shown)


src/ape/api/accounts.py
```diff
@@ -137,6 +143,15 @@ def transfer(
         if value:
             txn.value = self._convert(value, int)
 
+        if not value:
+            assert (
```
unparalleled-js, line 147: No need for these assert statements. It will fail naturally if this is not true.

Asserts in python should really only be used in tests.

unparalleled-js: Also, these get stripped out with the -I flag so its not reliable

src/ape/api/accounts.py
```diff
@@ -93,10 +93,16 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
         txn = self.prepare_transaction(txn)
 
         if send_everything:
+            error_message = "Sender does not have enough to cover transaction value and gas: "
             if txn.max_fee is None:
                 raise TransactionError(message="Max fee must not be None.")
-
-            txn.value = self.balance - txn.max_fee
+            if txn.gas_limit:
+                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
```
unparalleled-js, line 100: I think there is a property on txn for total cost or something

tests/functional/test_accounts.py
```diff
@@ -60,6 +60,24 @@ def test_transfer(sender, receiver):
     assert receiver.balance == expected
 
 
+def test_transfer_without_value(sender, receiver):
+    with pytest.raises(AssertionError) as err:
```
unparalleled-js, line 64: Use ValueError in these cases.
Raise ValueError when args are not usable. Factor this in to my prior comments about using asserts.

tests/functional/test_accounts.py
```diff
@@ -60,6 +60,24 @@ def test_transfer(sender, receiver):
...
+def test_transfer_without_value_send_everything_false(sender, receiver):
+    with pytest.raises(AssertionError) as err:
+        sender.transfer(receiver, send_everything=False)
+    assert str(err.value) == "Transfer without value argument requires kwarg send_everything=True"
+
+
+def test_transfer_without_value_send_everything_true(sender, receiver):
+    initial_balance = receiver.balance
+    sender.transfer(receiver, send_everything=True)
+    assert receiver.balance > initial_balance, "Receiver has same balance after transfer"
```
unparalleled-js, line 78: Can as assert on the actual expected value to make we are properly accounting for fees?

unparalleled-js (changes requested):
Cool!!

tests/functional/test_accounts.py
sabotagebeats: this was something I was having trouble trying to figure out, since the fee seems different at times and I'm not sure how that's calculated. maybe i can use txn.total_cost?

src/ape/api/accounts.py
sabotagebeats: should these be if statements then?

src/ape/api/accounts.py
sabotagebeats: I don't see that property: 
```
(Pdb) 'total_cost' in dir(txn)
False
```

src/ape/api/accounts.py
unparalleled-js: Use ValueError

sabotagebeats: if I don't have the assert statements then it doesn't process the transaction properly in the tests: 
`FAILED tests/functional/test_accounts.py::test_transfer_without_value - KeyError: 'send_everything'`
so i believe it needs to throw to provide those errors?

tests/functional/test_accounts.py
unparalleled-js: Total_transfer_value? Hmmmm

sabotagebeats: total_transfer_value does not work, if fails with the error that the gas is higher than that number.

sabotagebeats: max_fee also doesn't work for the same reason

src/ape/api/accounts.py
sabotagebeats: ok i got it to work with valueError :D

tests/functional/test_accounts.py
sabotagebeats: ok I got it to work with ValueError

src/ape/api/accounts.py
sabotagebeats: once the total cost is calculated correctly I should be able to use that as to `assert sender.balance==0` in the `send_everything=True` test

src/ape/api/accounts.py
```diff
@@ -137,6 +143,17 @@ def transfer(
         if value:
             txn.value = self._convert(value, int)
 
+        if not value:
+            if "send_everything" not in kwargs:
+                raise ValueError(
+                    "Transfer without value argument requires kwarg send_everything=True"
+                )
+            if not kwargs["send_everything"]:
```
fubuloubu, line 151: Simpler
```suggestion
            if "send_everything" not in kwargs or not kwargs["send_everything"]:
```

tests/functional/test_accounts.py
unparalleled-js: Ok! How about we assert that the sender has no more balance?

src/ape/api/accounts.py
```diff
@@ -93,10 +93,16 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
         txn = self.prepare_transaction(txn)
 
         if send_everything:
+            error_message = "Sender does not have enough to cover transaction value and gas: "
             if txn.max_fee is None:
                 raise TransactionError(message="Max fee must not be None.")
-
-            txn.value = self.balance - txn.max_fee
+            if txn.gas_limit:
```
unparalleled-js, line 99: Should this be `if txn.gas_limit is not None` ?

src/ape/api/accounts.py
```diff
@@ -93,10 +93,16 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
+                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
+                error_message = error_message + f"{txn.max_fee * txn.gas_limit}"
+            else:
+                txn.value = self.balance - txn.max_fee
+                error_message = error_message + f"{txn.max_fee}"
+            assert txn.value > 0, error_message
```
unparalleled-js, line 105: Can we use `ValueError`?

sabotagebeats: yes definitely

tests/functional/test_accounts.py
sabotagebeats: This doesn't work because gas fees are overestimated (discussed in this morning's meeting) but I did assert that the sender has less than 1 finney now.

src/ape/api/accounts.py
sabotagebeats: I made this value error :)

src/ape/api/accounts.py
sabotagebeats: i updated to this above and it is passing tests.

src/ape/api/accounts.py
```diff
@@ -93,10 +93,16 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
         txn = self.prepare_transaction(txn)
 
         if send_everything:
+            error_message = "Sender does not have enough to cover transaction value and gas: "
```
fubuloubu, line 96: Is this used anywhere other than [here](https://github.com/ApeWorX/ape/pull/731/files#diff-c308960cdf9376a4c05b2bb028a5a79e22c8a12b4c99633580062ec04ab613e2R104)?

src/ape/api/accounts.py
```diff
@@ -93,10 +93,16 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
             if txn.max_fee is None:
                 raise TransactionError(message="Max fee must not be None.")
```
fubuloubu, line 98: Can we raise a `ValueError` if both `txn.value` and `send_everything=True` is specified?

fubuloubu (approved):
This looks great!

src/ape/api/accounts.py
sabotagebeats: sure, that seems like a good idea

src/ape/api/accounts.py
sabotagebeats: no, I can refactor this

sabotagebeats: refactored

src/ape/api/accounts.py
sabotagebeats: implemented

src/ape/api/accounts.py
```diff
@@ -135,8 +142,19 @@ def transfer(
...
 
         if value:
+            if "send_everything" in kwargs and kwargs["send_everything"]:
+                raise ValueError(
+                    "Kwarg send_everything=True requires transfer without value argument"
+                )
             txn.value = self._convert(value, int)
 
+        if not value:
+            if "send_everything" not in kwargs or not kwargs["send_everything"]:
```
unparalleled-js, line 152: This could just be `if not kwargs.get("send_everything")`

unparalleled-js: `get` returns `None` which when the key is not present which is `falsy`

src/ape/api/accounts.py
```diff
@@ -95,8 +95,15 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
                 raise TransactionError(message="Max fee must not be None.")
-
-            txn.value = self.balance - txn.max_fee
+            if not txn.gas_limit or txn.gas_limit is None:
+                raise TransactionError(message="The txn.gas_limit is not set.")
+            else:
+                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
+            if txn.value <= 0:
+                raise ValueError(
+                    f"Sender does not have enough to cover transaction value and gas: \
```
unparalleled-js, line 104: It would be nice to have a test hit this error!

unparalleled-js (dismissed):
Couple small things, but this looks gooooooood to me!

src/ape/api/accounts.py
sabotagebeats: Hmm maybe trying to send from an empty account would do it

sabotagebeats: yes that works.

sabotagebeats:
> Couple small things, but this looks gooooooood to me!

I think I implemented those, and it's passing the new test condition. I think we may be good to merge?

src/ape/api/accounts.py
```diff
@@ -95,8 +95,15 @@ def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI
...
+                    {txn.max_fee * txn.gas_limit}"
+                )
```
fubuloubu, line 106: Just a little nit on style
```suggestion
            elif txn.value <= 0:
                raise ValueError(
                    f"Sender does not have enough to cover transaction value and gas: \
                    {txn.max_fee * txn.gas_limit}"
                )
            else:
                txn.value = self.balance - (txn.max_fee * txn.gas_limit)
```

src/ape/api/accounts.py
```diff
@@ -135,8 +142,19 @@ def transfer(
...
             txn.value = self._convert(value, int)
 
+        if not value:
+            if not kwargs.get("send_everything"):
+                raise ValueError(
+                    "Transfer without value argument requires kwarg send_everything=True"
+                )
+            return self.call(txn, kwargs["send_everything"])
+
         return self.call(txn, send_everything=value is None)
```
fubuloubu, line 158: A little more streamlined here
```suggestion
            txn.value = self._convert(value, int)
            return self.call(txn, send_everything=value is None)

        elif not kwargs.get("send_everything"):
            raise ValueError(
                "Transfer without value argument requires kwarg send_everything=True"
            )
        else:
            return self.call(txn, send_everything=True)

```

fubuloubu (approved):
Looking good! Some style nits and can merge

src/ape/api/accounts.py
fubuloubu: The nitpick: it's sometimes difficult to tell if you have one large if condition followed by the second that the two blocks are unrelated. It's often between to combine them into one larger `if..elif...else` statement this way if you can (or add extra spacing to delineate if you can't)

src/ape/api/accounts.py
fubuloubu: The nitpick: having nested if statements can be harder to read than trying to flatten them down to one level of `if..elif..else` statements, which makes it easier to reason about

unparalleled-js (approved):
LGTM

src/ape/api/accounts.py
sabotagebeats: I think the problem here is that the `txn.value` needs to be calculated before the `elif txn.value<= 0`

sabotagebeats: I updated this section of code so that it is passing checks. Can you take another look and see if you're satisfied?

src/ape/api/accounts.py
```diff
@@ -135,9 +141,17 @@ def transfer(
             txn.data = self._convert(data, bytes)
 
         if value:
+            if "send_everything" in kwargs and kwargs["send_everything"]:
```
unparalleled-js, line 144: this can just be 
```suggestion
            if not kwargs.get("send_everything"):
```

src/ape/api/accounts.py
```diff
@@ -135,9 +141,17 @@ def transfer(
...
+                raise ValueError(
+                    "Kwarg send_everything=True requires transfer without value argument"
+                )
             txn.value = self._convert(value, int)
+            return self.call(txn, send_everything=value is None)
```
unparalleled-js, line 149: since we know `send_everything` is False here
```suggestion
            return self.call(txn)
```

src/ape/api/accounts.py
sabotagebeats:
```
FAILED tests/functional/test_accounts.py::test_transfer - ValueError: Kwarg send_everything=True requires transfer without value argument
FAILED tests/functional/test_accounts.py::test_transfer_with_value_send_everything_true - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/functional/test_accounts.py::test_transfer_using_type_0 - ValueError: Kwarg send_everything=True requires transfer without value argument
FAILED tests/functional/test_accounts.py::test_send_transaction_with_bad_nonce - ValueError: Kwarg send_everything=True requires transfer without value argument
FAILED tests/functional/test_accounts.py::test_send_transaction_without_enough_funds - ValueError: Kwarg send_everything=True requires transfer without value argument
FAILED tests/functional/test_accounts.py::test_send_transaction_sets_defaults - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_sign_message_with_prompts - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_transfer_with_prompts - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_autosign_messages - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_autosign_transactions - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_unlock_with_passphrase_and_sign_message - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_unlock_from_prompt_and_sign_message - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_unlock_with_passphrase_and_sign_transaction - ValueError: Kwarg send_everything=True requires transfer without value argument
ERROR tests/functional/test_accounts.py::test_unlock_from_prompt_and_sign_transaction - ValueError: Kwarg send_everything=True requires transfer without value argument
```

unparalleled-js: Oh I see, I was reading the condition wrong

unparalleled-js (dismissed):
I think we can rid of the `value is None` thing but otherwise looks good

sabotagebeats:
passing checks. Final review?

unparalleled-js (approved):

Issue #727, opened 2022-05-16
Add Transaction Query support


fubuloubu:
### Elevator pitch:
It's useful to be able to query for transactions in multiple contexts (in terms of the account who made it, the block it's included in, etc.)

### Value:
Be able to implement query functionality across transactions.
This will help implement #392

### Dependencies:
N/A

### Task list:
* [x] Merge `ape.api.query._BaseAccountQuery` into `ape.api.query.AccountQuery`
* [x] Refactor `ape.api.query.AccountQuery` to `ape.api.query.AccountTransactionQuery`
* [x] Add `ape.api.query.BlockTransactionQuery` where `block_id` is the filter applied
* [x] Add support into `ape.managers.query.DefaultQueryProvider`
* [ ] Integrate into `ape_cache.query.CacheQueryProvider` #680 

### Estimated completion date:
5/20/2022

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

fubuloubu:
Modification to `BlockAPI`:

```py
class BlockAPI:
    ...
    # remove `transaction_ids`
    num_transactions: int = 0
```

Implemention `BlockTransactionQuery`:

```py
class BlockTransactionQuery(_BaseQuery):
    block_id: Hash

    @classmethod
    def all_fields(cls) -> List[str]:
        return list(TransactionAPI.__fields__)
```

Added to `BlockAPI`:
```py
    @cached_property
    def transactions(self) -> List[TransactionAPI]:
        query = BlockTransactionQuery(block_id=self.hash)
        txns_dataframe = self.query_manager.query(query)
        return # convert `txns_dataframe` into `List[TransactionAPI]`
```

How to do query by defualt:
```py
class DefaultQueryProvider:
    ...

    def perform_query(query: BlockTransactionQuery) -> pd.Dataframe:
         # ProviderAPI.get_transactions_for_block(block_id: Hash) -> Iterator[TransactionAPI]
         transactions_iter = self.provider.get_transactions_for_block(query.block_id)
         # return transactions as a dataframe
```

https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_block_transaction_count
https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_transaction_by_block

[...]

PR #732, opened 2022-05-18
refactor: add transaction query support

johnson2427:
### What I did

* Added `num_transactions` to `ape.api.providers.BlockAPI`
* Created `ape.api.query.BlockTransactionQuery`
* Added `transaction` method to `ape.api.providers.BlockAPI`
* Added `perform_block_transaction_query` to `ape.managers.queryDefaultQueryProvider`

fixes: #727 

### How I did it

Followed issue number 727

### How to verify it

run pytests

### Checklist


johnson2427:
This is very obviously raw, and not ready to be merged, I have a circular import issue here, and would like input on what we should do with `BlockAPI` having this `transaction` method added to it.  It causes a circular import.

src/ape/api/providers.py
```diff
@@ -23,6 +23,7 @@
 from ape.api.config import PluginConfig
 from ape.api.networks import LOCAL_NETWORK_NAME, NetworkAPI
 from ape.api.transactions import ReceiptAPI, TransactionAPI
+from ape.api.query import BlockTransactionQuery
```
fubuloubu, line 26: Oh, wait this is causing the circular import? Then yeah, I would just import it inside of `transactions` below. That should resolve it

src/ape/api/query.py
```diff
@@ -69,9 +69,23 @@ def all_fields(cls) -> List[str]:
...
+
+class AccountTransactionQuery(_BaseQuery):
+    """
+    A ``QueryType`` that collects properties of ``TransactionAPI`` over a range
+    of transactions made by ``account`` between ``start_nonce`` and ``stop_nonce``.
+    """
+
     start_nonce: NonNegativeInt = 0
     stop_nonce: NonNegativeInt
+    account: AddressType
```
fubuloubu, line 88: Just want reverse the ordering a bit here:
```suggestion
    account: AddressType
    start_nonce: NonNegativeInt = 0
    stop_nonce: NonNegativeInt
```

src/ape/api/query.py
```diff
@@ -69,9 +69,23 @@ def all_fields(cls) -> List[str]:
         return list(BlockAPI.__fields__)
 
 
-class _BaseAccountQuery(_BaseQuery):
+class BlockTransactionQuery(_BaseQuery):
+    block_id: Any
```
fubuloubu, line 73:
```suggestion
    """
    A ``QueryType`` that collects properties of ``TransactionAPI`` over a range of
    transactions collected inside the ``BlockAPI` object represented by ``block_id``.
    """
    block_id: Any
```

src/ape/managers/query.py
```diff
@@ -46,6 +46,11 @@ def perform_block_query(self, query: BlockQuery) -> pd.DataFrame:
         block_dicts_iter = map(partial(get_columns_from_item, query), blocks_iter)
         return pd.DataFrame(columns=query.columns, data=block_dicts_iter)
 
+    @perform_query.register
+    def perform_transaction_query(self, query: BlockTransactionQuery) -> pd.DataFrame:
+        transactions_iter = self.provider.get_transaction(query.block_id)
```
fubuloubu, line 51: This won't work, needs a new method `ProviderAPI.get_transactions_by_block(block_id: HexBytes) -> Iterator[TransactionAPI]`

fubuloubu:
Another, more meta thought: perhaps `QueryManager`/`QueryAPI` subclasses needs to work with sets of iterators returned from `.perform_query` and `.update_cache` methods, and let the downstream `.query` methods instead convert to `pd.DataFrame` objects, so we don't have all of these conversion back and forth occuring until the data is finally asked for

src/ape/managers/query.py
johnson2427: Yeah, I knew this wouldn't work, really was focusing on the rest of this breaking before I came back to this

src/ape/api/providers.py
```diff
@@ -614,6 +634,11 @@ def get_transaction(self, txn_hash: str, required_confirmations: int = 0) -> Rec
         )
         return receipt.await_confirmations()
 
+    def get_transactions_by_block(self, block_id: HexBytes) -> Iterator[TransactionAPI]:
+        block = self.provider.get_block(block_id)
```
fubuloubu, line 638: May need to pass-through `full_transactions` boolean to `.get_block`

src/ape/api/providers.py
```diff
@@ -81,6 +83,12 @@ def validate_hexbytes(cls, value):
             raise ValueError(f"Hash `{value}` is not a valid Hexbyte.")
         return value
 
+    @cached_property
+    def transactions(self) -> List[TransactionAPI]:
+        query = BlockTransactionQuery(block_id=self.hash)
+        df_txns = self.query_manager.query(query)
+        return list(map(lambda x: TransactionAPI(x), df_txns.values.tolist()))
```
fubuloubu, line 90: Functionaliity the same
```suggestion
        return list(map(TransactionAPI, df_txns.values.tolist()))
```

src/ape/api/providers.py
johnson2427: not quite sure how to do this yet, will review

fubuloubu: This is how Web3py exposes this feature. Our `ProviderAPI` implementations use it under the hood:
https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_block

fubuloubu (dismissed):

fubuloubu:
@unparalleled-js I think we can merge this?

johnson2427:
I don't have a test for this yet. I started to make the test, but it's not finished yet.

johnson2427:
Very basic test here, BlockTransactionQuery is working as expected. Trying to get this to query through the DefaultQueryProvider at the moment

src/ape/api/providers.py
unparalleled-js: From the `ProviderAPI` perspective, the boolean makes sense I think

tests/functional/api/test_query.py
```diff
@@ -24,6 +25,25 @@ def test_basic_query(eth_tester_provider):
     ]
 
 
+def test_block_transaction_query(sender, receiver, chain, eth_tester_provider):
+    eth_tester_provider.get_transactions_by_block("0")
```
unparalleled-js, line 29: Can we make assertions about the result from this call as well?

src/ape/api/query.py
```diff
@@ -66,10 +71,31 @@ class BlockQuery(_BaseBlockQuery):
 
     @classmethod
     def all_fields(cls) -> List[str]:
+        from .providers import BlockAPI
```
unparalleled-js, line 74: Would another way to resolve the circular import be to make a new module `providers/blocks.py` ?
That way, we are not doing anything with the query manager in `providers.py`. I am used to thinking of the provider as the last stop before RPC so I was thrown off by the import of query manager. But it makes sense for the `BlockAPI` so that is why I am thinking it deserves its own module, especially if it also resolves this circular import

unparalleled-js (commented):
Neat! Does it work? Some more tests and docs would be nice, I want to try this stuff out but don't really know how. If I can try it out or see more thorough tests, I will more readily approve.

src/ape/api/providers.py
```diff
@@ -81,6 +83,12 @@ def validate_hexbytes(cls, value):
             raise ValueError(f"Hash `{value}` is not a valid Hexbyte.")
         return value
 
+    @cached_property
```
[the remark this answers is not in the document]
johnson2427: by pending you mean current block? Not sure how this would affect anything

src/ape/api/query.py
johnson2427: I'd be for this

tests/functional/api/test_query.py
johnson2427: I removed this, we didn't need it

PR #740, opened 2022-05-20
fix: issue when compiling a source where references was None

unparalleled-js:
### What I did

fixes: #

### Checklist

### DIFF
--- a/src/ape/managers/project/types.py
+++ b/src/ape/managers/project/types.py
@@ -93,7 +93,7 @@ class BaseProject(ProjectAPI):
             cached_source_reference_paths = {
                 source_id: [
                     self.contracts_folder.joinpath(Path(s))
-                    for s in getattr(source, "references", [])
+                    for s in getattr(source, "references", []) or []
                 ]
                 for source_id, source in cached_sources.items()
             }


NotPeopling2day (approved):

src/ape/managers/project/types.py
```diff
@@ -93,7 +93,7 @@ def create_manifest(
             cached_source_reference_paths = {
                 source_id: [
                     self.contracts_folder.joinpath(Path(s))
-                    for s in getattr(source, "references", [])
+                    for s in getattr(source, "references", []) or []
```
fubuloubu, line 96: isn't this what the 2nd argument in `getattr` does?
https://docs.python.org/3/library/functions.html#getattr

fubuloubu: Oh I see, maybe this is clearer
```suggestion
                    for s in getattr(source, "references", None) or []
```

NotPeopling2day: It's getting the attribute for `references` but the actual value is None 
telegram-cloud-photo-size-1-4918042827766737427-y

fubuloubu: tbh I'd eventually like to clean up `ethpm-types` when we eventually break from the standard, it would make the schema way cleaner

PR #743, opened 2022-05-23
fix: mainnet fork naming

pandadefi:
### What I did
I can see that mainnet fork has two different names.
I have made changes so that it's using `mainnet_fork`

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Ape provides an IPython interactive console with useful pre-defined locals to in
 To interact with a deployed contract in a local environment, start by opening the console:
 
 ```bash
-ape console --network :mainnet-fork:hardhat
+ape console --network :mainnet_fork:hardhat
 ```
 
 Visit [Ape Console](https://docs.apeworx.io/ape/stable/commands/console.html) to learn how to use Ape Console.
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -127,7 +127,7 @@ Set default network and network providers:
 
 ```yaml
 ethereum:
-  default_network: mainnet-fork
+  default_network: mainnet_fork
   mainnet_fork:
     default_provider: hardhat
 ```
--- a/tests/functional/test_networks.py
+++ b/tests/functional/test_networks.py
@@ -22,7 +22,7 @@ def test_get_network_choices_filter_ecosystem(networks):
 
 
 def test_get_network_choices_filter_network(networks):
-    actual = {c for c in networks.get_network_choices(network_filter="mainnet-fork")}
+    actual = {c for c in networks.get_network_choices(network_filter="mainnet_fork")}
     assert actual == set()
 
 
--- a/tests/integration/cli/test_networks.py
+++ b/tests/integration/cli/test_networks.py
@@ -28,7 +28,7 @@ ecosystems:
     providers:
     - name: geth
       isDefault: true
-  - name: mainnet-fork
+  - name: mainnet_fork
     providers: []
   - name: ropsten
     providers:


unparalleled-js (changes requested):
Hmmmmm what exactly is going on? Are you getting an error somewhere?

pandadefi:
> Hmmmmm what exactly is going on? Are you getting an error somewhere?

Isn't it confusing to have the network name with `_` and `-` ?

I have tried with the following config:
```
ethereum:
  default_network: mainnet-fork
  mainnet_fork:
    default_provider: hardhat
```

and got: `ape_hardhat.exceptions.HardhatProviderError: Invalid upstream-fork URL. Can't be same as local Hardhat node.`


Then I have tried:

```
ethereum:
  default_network: mainnet_fork
  mainnet_fork:
    default_provider: hardhat
```
with the error: ERROR: (NetworkError) 'mainnet_fork' is not a valid network for ecosystem 'ethereum'.

unparalleled-js:
> and got: `ape_hardhat.exceptions.HardhatProviderError: Invalid upstream-fork URL. Can't be same as local Hardhat node.`

You are getting this error because your default `mainnet` provider is `geth` (which is the default out-of-the-box) and `geth` is configured to run on the port `8545`. Also, your hardhat node is configured to run on the port `8545`.

If you are using `geth`, you need to either change your hardhat port or the geth port, etc.
If you are not using `geth`, you need to either change your default mainnet provider OR change your upstream-fork provider.

e.g. 

```
ethereum:
  default_network: mainnet_fork
  mainnet_fork:
    default_provider: hardhat
  mainnet:
    default_provider: alchemy
```

or

```
hardhat:
  fork:
    ethereum:
      mainnet:
        upstream_provider: alchemy
```

unparalleled-js:
> Isn't it confusing to have the network name with `_` and `-` ?

I think it is only `mainnet-fork`.
We convert it one spot for the sake of using `__getattr__` e.g. since you can't have hyphens in python

unparalleled-js:
to make things less confusing, I am thinking we should change the default port for hardhat when using `fork` mode to account for the likelihood of geth running on 8545

pandadefi:
> hardhat:
>   port: 8555

With this changes, I am still having some issues it seems like the hardhat started by ape isn't running on `8555` since I am getting a new error 
```
ape.exceptions.RPCTimeoutError: Timed out waiting for successful RPC connection to the 'Hardhat node' process (20 seconds).
```

unparalleled-js:
> With this changes, I am still having some issues it seems like the hardhat started by ape isn't running on `8555` since I am getting a new error
> 
> ```
> ape.exceptions.RPCTimeoutError: Timed out waiting for successful RPC connection to the 'Hardhat node' process (20 seconds).
> ```

I get this error when `geth` is not running. Geth is for sure running?

One thing you can do is run the hardhat command `npx hardhat node --hostname 127.0.0.1 --port 8555 --fork <your-geth-url>`. If that fails, it will tell you more info.

pandadefi:
> One thing you can do is run the hardhat command `npx hardhat node --hostname 127.0.0.1 --port 8555 --fork <your-geth-url>`. If that fails, it will tell you more info.

I am confused I see a message `INFO: Starting 'Hardhat node' process.` so I am expecting not to have to start hardhat.

fubuloubu:
> So it seems hardhat wasn't starting because I was missing.
> 
> ```
>   mainnet:
>     default_provider: infura
> ```

This seems like a bug

unparalleled-js:
> I am confused I see a message `INFO: Starting 'Hardhat node' process.` so I am expecting not to have to start hardhat.

Right, you typically don't have to manually start Hardhat but it is a good way to debug problems.
And yes, you have tell Hardhat what network you want to use for getting fork data, either Infura, Alchemy, or Geth. You can do this by either setting the default provider for that network or by setting the upstream provider (I have a comment above on this PR explaining this). It is also mentioned in the `Hardhat` plugin readme.

fubuloubu:
> > This seems like a bug
> 
> How so?

I think in that hardhat is launching, geth is not launched, but it seems to make it look like a hardhat failure, not a geth one

unparalleled-js:
> I think in that hardhat is launching, geth is not launched, but it seems to make it look like a hardhat failure, not a geth one

We can look into an improved error, but this is a hardhat issue, not geth. Hardhat node fails to launch because you gave it a bad fork URL.

fubuloubu:
> > I think in that hardhat is launching, geth is not launched, but it seems to make it look like a hardhat failure, not a geth one
> 
> We can look into an improved error, but this is a hardhat issue, not geth. Hardhat node fails to launch because you gave it a bad fork URL.

Ah gotcha. Could it raise a better error about the selected `upstream_provider` having a configuration issue?

unparalleled-js:
> Ah gotcha. Could it raise a better error about the selected `upstream_provider` having a configuration issue?

That would be easy to do if we change it to require explicit declaration of the upstream provider rather than falling back to the default Mainnet provider. Otherwise, the challenge is knowing whether the using is actually trying to connect to that provider or if they did not configure it. I think for this reason, I always prefer setting `upstream_provider` even if I already changed my default Mainnet provider.

unparalleled-js:
See pros and cons here: https://github.com/ApeWorX/ape-hardhat/issues/62

Issue #700, opened 2022-05-09
Testing: Display Trace


fubuloubu:
### Elevator pitch:
Print a Tree

### Value:
It is often very nice to be able to print out a trace of the call tree made during a transaction for debugging purposes, or displaying stack traces when a revert occurs. Here are some examples in various frameworks:

Brownie:
https://eth-brownie.readthedocs.io/en/stable/core-transactions.html?highlight=trace#call-traces
(added in https://github.com/eth-brownie/brownie/pull/679)

Foundry:
ipfs://bafybeihapxqz43vadpqicxulwihrj6pqnvrxavf7hruk6xk45es57ds3hi/forge/traces.html

Hardhat:
https://hardhat.org/hardhat-network/#solidity-stack-traces

### Dependencies:
- Using https://github.com/ApeWorX/evm-trace
- https://github.com/ApeWorX/ape/tree/feat/tracing @NotPeopling2day's branch

### Design approach:
Add `ReceiptAPI.show_trace()` (TBD naming) to print the raw tree, then make the tree pretty

### Task list:
- [ ] Print raw tree (showing addresses and calldata of unknown contract types)
- [ ] "Lift" tree by finding contract types we know (either from Etherscan or our internal deployments) and ABI decoding
- [ ] Pretty-print "lifted" call tree

### Estimated completion date:
Friday 5/13

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

unparalleled-js:
Blocked by this: https://github.com/ApeWorX/ape/issues/374

Need behind-the-scenes caching of local deployments.

unparalleled-js:
Working on Parity support

PR #746, opened 2022-05-23
feat: show_trace() method on ReceiptAPI

unparalleled-js:
### What I did

fixes: #700 

### How I did it

Override the display node title and use `rich` console to output.
Use colors.
Fetch contract information from cache or explorer.

### How to verify it

Try viewing some traces of your favorite transactions and let me know what looks weird.
Also just let me know your thoughts on the output in general.

Output attachment: 
Screen Shot 2022-06-09 at 00 29 27

Notes about decisions made:

* Address args don't get quotes
* Try decoding bytes32 if they look like strings and give quotes
* Strings get quotes
* Convert structs to string like `(key=value)`
* Only give color to values, not type names
* When there is only a single argument, exclude type name

### Checklist


unparalleled-js:
Requires https://github.com/ApeWorX/evm-trace/pull/5 release

src/ape/api/transactions.py
```diff
@@ -243,3 +274,219 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+            return [self._decode_value(v) for v in value]
+
+        elif isinstance(value, Struct):
+            return {k: v for k, v in value.items()}
+
+        return decoded_value
+
+
+@dataclass()
+class _MethodTraceSignature:
```
fubuloubu, line 398: This might be a good util method?

src/ape/api/transactions.py
```diff
@@ -243,3 +274,219 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+                value += ","
+
+            value += "\n"
+
+        value += _SPACING * depth
+        value += "]"
+        return value
+
+
+def _get_method_called(selector: str, contract_type: ContractType) -> Optional[MethodABI]:
```
fubuloubu, line 485: This might be a good util method?

src/ape/api/transactions.py
unparalleled-js: Perhaps! I tend to prefer moving things to `utils` on an as-needs basis. In the future, if something needs this functionality, we may need to adjust it slightly before making it public.

src/ape/api/transactions.py
```diff
@@ -243,3 +274,219 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+        elif isinstance(self.return_value, dict):
+            return _dict_to_str(self.return_value, _RETURN_VALUE_TRACE_COLOR)
+
+        elif isinstance(self.return_value, (list, tuple)):
+            return f"[{_RETURN_VALUE_TRACE_COLOR}]{_list_to_str(self.return_value)}[/]"
+
+        return f"[{_RETURN_VALUE_TRACE_COLOR}]{self.return_value}"
+
+
+def _dict_to_str(dictionary: Dict, color: str) -> str:
```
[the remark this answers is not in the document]
unparalleled-js: I suppose we could have a full `utils.traces.py` module?

setup.py
```diff
@@ -113,11 +112,12 @@
         "rich>=10.14,<11",
         "tqdm>=4.62.3,<5.0",
         "typing-extensions ; python_version<'3.8'",
-        "web3[tester]>=5.29.0,<6.0",
-        "eth_abi==2.1.1",
-        "eth-utils==1.10.0",
-        "eth-rlp==0.2.1",
         "python-dateutil>=2.8.2,<3.0",
+        "web3[tester]>=6.0.0b2,<7.0",
```
unparalleled-js, line 116: I didn't mean to push these changes yet.. May undo. I was seeing if it help wit the `<?>` but it hasn't.

I really want to upgrade web3.py to the 6.0 range but its still in beta so may wait.

- [ ] Either upgrade to released 6.0.0 or undo these changes

unparalleled-js:
- [ ] Add method flag to `show_trace()` that shows **all** the trace's information
- [ ] Also set flag when user does `--verbosity debug`

unparalleled-js:
- [ ] Handle delegatecall / forwards, might require https://github.com/ApeWorX/ape/pull/750

src/ape/api/transactions.py
banteg: here is a better solution for your consideration, you can do `contract_type.mutable_methods[calldata]`
https://github.com/ApeWorX/ethpm-types/pull/34

src/ape/api/transactions.py
```diff
@@ -1,20 +1,38 @@
...
 if TYPE_CHECKING:
     from ape.contracts import ContractEvent
 
 
+_METHOD_NAME_TRACE_COLOR = "bright_green"
+_ARGUMENT_VALUE_TRACE_COLOR = "bright_magenta"
+_RETURN_VALUE_TRACE_COLOR = "bright_blue"
+_FAILURE_TRACE_COLOR = "bright_red"
+_WRAP_THRESHOLD = 50
+_SPACING = "    "
```
banteg, line 33: consider two spaces

src/ape/managers/chain.py
```diff
@@ -439,6 +440,12 @@ def get(
             return default
 
         contract_type = self._get_contract_type_from_disk(address)
+        if not contract_type:
+            contract_type = self._get_contract_type_from_explorer(address)
```
banteg, line 444: duplicate code

src/ape/api/providers.py
```diff
@@ -36,8 +36,14 @@
 )
```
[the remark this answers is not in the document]
unparalleled-js: Once I update with main, it should go away

src/ape/api/transactions.py
```diff
@@ -168,6 +188,14 @@ def ran_out_of_gas(self) -> bool:
             same amount of gas as the given ``gas_limit``.
         """
 
+    @property
+    def trace(self) -> Iterator[TraceFrame]:
+        """
+        The trace of the transaction, if available from your provider.
+        """
+
+        return self.provider.get_transaction_trace(txn_hash=self.txn_hash)
```
unparalleled-js, line 197: We should find a way to cache this request, as it is really slow. Thoughts?

src/ape/api/transactions.py
unparalleled-js: Done!

src/ape/api/transactions.py
unparalleled-js: Thanks! This was implemented

docs/userguides/transactions.md
```diff
@@ -121,3 +121,63 @@ In your `ape-config.yaml` file, add the following:
...
+receipt = contract.methodWithoutArguments()
+receipt.show_trace()
+```
+
+The trace might look something like:
+
+```bash
+Call trace for '0xd15e59e4cca4aaf3ac57be8b71d35bb53549ec3ec1833da183d3fa5f4f609b83'
+Sender=0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C
+ContractA.methodWithoutArguments() -> 0x00..7a9c [411973 gas]
```
fubuloubu, line 143: How big is the return value? Curious what heuristic makes it truncate

docs/userguides/transactions.md
```diff
@@ -121,3 +121,63 @@ In your `ape-config.yaml` file, add the following:
...
+contract = project.Contract.deploy(sender=owner)
+receipt = contract.methodWithoutArguments()
+receipt.show_trace()
+```
+
+The trace might look something like:
+
+```bash
+Call trace for '0xd15e59e4cca4aaf3ac57be8b71d35bb53549ec3ec1833da183d3fa5f4f609b83'
+Sender=0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C
```
fubuloubu, line 142: Oh, this is clever!

wonder if you can call this `tx.origin` to match with the Solidity/Vyper environment variable exposing the origin account

docs/userguides/transactions.md
```diff
@@ -121,3 +121,63 @@ In your `ape-config.yaml` file, add the following:
...
+ContractA.methodWithoutArguments() -> 0x00..7a9c [411973 gas]
+├── ContractB.methodB1(foo="bar", bar=42) [401601 gas]
+│   ├── ContractC.getSomeList() -> [
+│   │     3425311345134513461345134534531452345,
+│   │     111344445534535353,
+│   │     993453434534534534534977788884443333
+│   │   ] [369688 gas]
+│   └── ContractC.methodC1(foo="bar", baz=36, ban=ContractA) [363455 gas]
```
fubuloubu, line 150: As we discussed before, gas needs some work to be more meaningful. Will probably figure this out better w/ gas reporting work. (just making a note here)

src/ape/api/transactions.py
fubuloubu: [`itertools.tee`](https://docs.python.org/3/library/itertools.html#itertools.tee) is really nice as it let's you make copies of iterators for parallel usage. Perhaps suggest doing this for now (always keep a `tee` copy around in case we ask again), but eventually might need to get into the `QueryAPI` and caching it to cache db for long-term usage (with appropiate pruning).

unparalleled-js:
NOTE: I have a bunch of tests staged in ape-hardhat but will wait to push the branch until this is released.
For now, you can use this project I made to demo traces: https://github.com/unparalleled-js/ape-trace-project

```bash
ape run show 0x123..
```

src/ape/managers/chain.py
```diff
@@ -422,6 +422,7 @@ def get(
 
         Args:
             address (AddressType): The address of the contract.
+            default
```
fubuloubu, line 425: is there a default description missing here?

fubuloubu (dismissed):
Overall, looking super solid here. Great work!

docs/userguides/transactions.md
unparalleled-js: It is currently always being truncated but we can definitely make it only happen via some heuristic.

- [ ] conditional bytes truncation

docs/userguides/transactions.md
unparalleled-js: Done!

fubuloubu (dismissed):

src/ape/utils/misc.py
```diff
@@ -230,6 +231,26 @@ def inner(*args, **kwargs):
     return inner
 
 
+class cached_iterator(property):
```
unparalleled-js, line 234: @fubuloubu I used your `itertools.tee()` recommendation in combo with the UX of `@cached_property` so now you can do

```python
@cached_iterator
def trace() -> Iterator[TraceFrame]
   ..
```

and it only makes the RPC once.
I have tests on the hardhat plugin for this that I will push up tomorrow.

fubuloubu: You van probably unit test this impl directly and then we can merge this PR nowish

src/ape/utils/misc.py
```diff
@@ -230,6 +231,26 @@ def inner(*args, **kwargs):
...
+    """
+    A cached iterator decorator. Use it as you would ``@cached_property``.
+    The intent is to prevent repeated fetches of data.
+    """
+
+    cache: Optional[Tuple[Iterator[Any], ...]] = None
```
fubuloubu, line 240: Here this acts as a class variable, which probably isn't what you want it to do. Just move this line to `__init__` or use `hasattr` in `__get__`

banteg:
We should use [parity-style tracing](https://openethereum.github.io/JSONRPC-trace-module) when available. It's faster, it returns almost the exact calltree we need that only remains to be decorated and the format is orders of magnitude more concise. I'll demonstrate:
```python
In [16]: %time a = chain.provider.web3.manager.request_blocking('debug_traceTransaction', [txhash, {"enableMemory": True, "enableReturnData": True}])
CPU times: user 3.92 s, sys: 121 ms, total: 4.04 s
Wall time: 6.47 s

In [17]: len(pickle.dumps(a))
Out[17]: 100800152

In [18]: %time b = chain.provider.web3.manager.request_blocking('trace_transaction', [txhash])
CPU times: user 7.12 ms, sys: 3.62 ms, total: 10.7 ms
Wall time: 433 ms

In [19]: len(pickle.dumps(b))
Out[19]: 110596
```
As you can see, it's 15x faster and 900x smaller. Also investigate a scope flag in Geth which seems to allow getting much smaller traces https://github.com/ethereum/go-ethereum/pull/21832#issuecomment-1067757859

src/ape/utils/misc.py
```diff
@@ -231,6 +232,22 @@ def inner(*args, **kwargs):
...
+    """
+
+    def __init__(self, fget=None, fset=None, fdel=None, doc=None):
+        self.cache: Optional[Tuple[Iterator[Any], ...]] = None
+        super().__init__(fget=fget, fset=fset, fdel=fdel, doc=doc)
+
+    def __get__(self, obj, objtype=None):
+        iterator = self.cache[1] if self.cache else self.fget(obj)
+        self.cache = tee(iterator)
+        return self.cache[0]
```
fubuloubu, line 248:
```suggestion
        iterator, self.cache = tee(self.cache if self.cache else self.fget(obj))
        return iterator
```

src/ape/utils/misc.py
```diff
@@ -231,6 +232,22 @@ def inner(*args, **kwargs):
...
 
 
+class cached_iterator(property):
+    """
+    A cached iterator decorator. Use it as you would ``@cached_property``.
+    The intent is to prevent repeated fetches of data.
+    """
+
+    def __init__(self, fget=None, fset=None, fdel=None, doc=None):
+        self.cache: Optional[Tuple[Iterator[Any], ...]] = None
```
fubuloubu, line 242: To match with below
```suggestion
        self.cache: Optional[Iterator] = None
```

unparalleled-js:
- [ ] Include `value` in decorated tree when `> 0` (suggested by @banteg )

src/ape/api/transactions.py
```diff
@@ -259,3 +282,312 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+        """
+        Display the complete sequence of contracts and methods called during
+        the transaction.
+
+        Args:
+            verbose (bool): Set to ``True`` to include more information.
+            file (IO[str]): The file to send output to. Defaults to stdout.
+        """
+        tree_factory = CallTraceParser(self, verbose=verbose)
+        call_tree = self.provider.get_call_tree(self.txn_hash)
```
fubuloubu, line 296: What happens if this fails?

fubuloubu (dismissed):

fubuloubu (approved):

src/ape/api/transactions.py
fubuloubu: Ping on this

src/ape/api/transactions.py
```diff
@@ -259,3 +294,339 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+            ):
+                suffix = default_message
+            else:
+                decoded_result = decode_abi(("string",), call_tree.returndata[4:])
+                if len(decoded_result) == 1:
+                    suffix = f'reverted with message: "{decoded_result[0]}"'
+                else:
+                    suffix = default_message
+
+            console.print(f"🚫 [bold red]{suffix}[/]")
```
fubuloubu, line 326: Does this emoji render?

src/ape/api/transactions.py
fubuloubu: ping on this

src/ape/api/transactions.py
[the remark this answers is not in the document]
fubuloubu: ping on this

src/ape/api/transactions.py
```diff
@@ -243,3 +274,219 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+
+        if do_wrap:
+            arguments_str += "\n"
+
+        index += 1
+
+    return f"{arguments_str})"
+
+
+def _list_to_str(ls: Union[List, Tuple], depth: int = 0) -> str:
```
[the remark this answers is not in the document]
fubuloubu: ping on this

fubuloubu (approved):

src/ape/api/transactions.py
unparalleled-js: I suppose it may depend on your Terminal, but it has for me everywhere.
Do we still want it?

src/ape/api/transactions.py
unparalleled-js: you get:

```
ERROR: (APINotImplementedError) Attempted to call method 'ProviderAPI.get_call_tree', method not supported.
```

src/ape/api/transactions.py
fubuloubu: We don't really have emojis anywhere else, I think the color is fine

fubuloubu (dismissed):

src/ape/api/transactions.py
unparalleled-js: Providers implement `get_call_tree()` however they can!

`providers` that don't support the `trace_transaction` use the full trace in their implementation of `get_call_tree()`
An example of that is `ape-hardhat`!
https://github.com/ApeWorX/ape-hardhat/pull/71/files#r898447679

Providers that support both implement the full trace still via `get_transaction_trace()` but still use `get_call_tree()` to be the parity trace.
An example of this provider is Alchemy: https://github.com/ApeWorX/ape-alchemy/pull/15/files

unparalleled-js: Providers implement `get_call_tree()` however they can!

`providers` that don't support the `trace_transaction` use the full trace in their implementation of `get_call_tree()`
An example of that is `ape-hardhat`!
https://github.com/ApeWorX/ape-hardhat/pull/71/files#r898447679

Providers that support both implement the full trace still via `get_transaction_trace()` but still use `get_call_tree()` to be the parity trace.
An example of this provider is Alchemy: https://github.com/ApeWorX/ape-alchemy/pull/15/files

fubuloubu (approved):

fubuloubu (approved):

Issue #734, opened 2022-05-18
handle proxy contracts in contract types cache


unparalleled-js:
### Overview

If a contract is a proxy, how can we detect + handle this best in the contracts cache logic from 

https://github.com/ApeWorX/ape/pull/712#pullrequestreview-976042673

### Specification

Describe the syntax and semantics of how you would like to see this feature implemented. The more detailed the better!

Remember, your feature is much more likely to be included if it does not involve any breaking changes.

### Dependencies

https://github.com/ApeWorX/ape/pull/712#pullrequestreview-976042673

PR #750, opened 2022-05-25
feat: detect minimal proxies

banteg:
### What I did
Automatically infer ContractType for minimal proxies

fixes: #734

### How I did it
Match known bytecodes, storage slots and views, then request the target's ContractType.

API changes:
- EcosystemAPI: add `get_proxy_info`
- ProviderAPI: add `get_storage_at`

### How to verify it
- [EIP-1167](https://eips.ethereum.org/EIPS/eip-1167)
- [Vyper](https://github.com/vyperlang/vyper/blob/0b3f3b329e59cf90bbff8a883b49f1166eb716a2/vyper/functions/functions.py#L1420)
- [0xSplits Clones](https://github.com/0xSplits/splits-contracts/blob/main/contracts/libraries/Clones.sol#L32)
- [EIP-1967](https://eips.ethereum.org/EIPS/eip-1967)
- [EIP-1822](https://eips.ethereum.org/EIPS/eip-1822)
- [OpenZeppelin UpgradeableProxy](https://github.com/OpenZeppelin/openzeppelin-labs/blob/master/initializer_contracts/contracts/UpgradeabilityProxy.sol)
- [Gnosis Safe](https://github.com/safe-global/safe-contracts/blob/c36bcab46578a442862d043e12a83fec41143dec/contracts/proxies/GnosisSafeProxy.sol#L29-L34)
- [EIP-1967 Beacon Proxy](https://ethtx.info/0xbcf2cd0fac6278f89dcbcfc61c0b85e1f25cce1f40c5fbf48abbe2b93732281e/)
- [EIP-987 Delegate Proxy](https://eips.ethereum.org/EIPS/eip-897)

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -370,7 +370,8 @@ class EcosystemAPI(BaseInterfaceModel):
             address (str): The address of the contract.
 
         Returns:
-            ProxyInfoAPI or None if the contract does not use any known proxy pattern.
+            Optional[:class:`~ape.api.networks.ProxyInfoAPI`]: Returns ``None`` if the contract 
+            does not use any known proxy pattern.
         """
         return None
 
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -357,7 +357,7 @@ class EcosystemAPI(BaseInterfaceModel):
 
         return data
 
-    def proxy_info(self, address: AddressType) -> Any:
+    def get_proxy_info(self, address: AddressType) -> Any:
         return None
 
 
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -357,6 +357,9 @@ class EcosystemAPI(BaseInterfaceModel):
 
         return data
 
+    def proxy_info(self, address: AddressType) -> Any:
+        return None
+
 
 class ProviderContextManager:
     """
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -23,7 +23,12 @@ LOCAL_NETWORK_NAME = "local"
 
 
 class ProxyInfoAPI(BaseModel):
+    """
+    Information about a proxy contract.
+    """
+
     target: AddressType
+    """The address of the implementation contract."""
 
 
 class EcosystemAPI(BaseInterfaceModel):
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Type
 
 from ethpm_types.abi import ConstructorABI, EventABI, MethodABI
 from hexbytes import HexBytes
+from pydantic import BaseModel
 
 from ape.exceptions import NetworkError, NetworkNotFoundError
 from ape.types import AddressType, ContractLog, RawAddress
@@ -19,6 +20,10 @@ if TYPE_CHECKING:
 
 
 LOCAL_NETWORK_NAME = "local"
+
+
+class ProxyInfoAPI(BaseModel):
+    target: AddressType
 
 
 class EcosystemAPI(BaseInterfaceModel):
@@ -357,7 +362,7 @@ class EcosystemAPI(BaseInterfaceModel):
 
         return data
 
-    def get_proxy_info(self, address: AddressType) -> Any:
+    def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
         """
         Information about a proxy contract such as proxy type and implementation address.
 
@@ -365,7 +370,7 @@ class EcosystemAPI(BaseInterfaceModel):
             address (str): The address of the contract.
 
         Returns:
-            Any: ProxyInfo or None if the contract does not use any known proxy pattern.
+            ProxyInfoAPI or None if the contract does not use any known proxy pattern.
         """
         return None
 
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -358,6 +358,15 @@ class EcosystemAPI(BaseInterfaceModel):
         return data
 
     def get_proxy_info(self, address: AddressType) -> Any:
+        """
+        Information about a proxy contract such as proxy type and implementation address.
+
+        Args:
+            address (str): The address of the contract.
+
+        Returns:
+            Any: ProxyInfo or None if the contract does not use any known proxy pattern.
+        """
         return None
 
 
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -375,7 +375,7 @@ class EcosystemAPI(BaseInterfaceModel):
             address (str): The address of the contract.
 
         Returns:
-            Optional[:class:`~ape.api.networks.ProxyInfoAPI`]: Returns ``None`` if the contract 
+            Optional[:class:`~ape.api.networks.ProxyInfoAPI`]: Returns ``None`` if the contract
             does not use any known proxy pattern.
         """
         return None
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -168,6 +168,7 @@ class ProviderAPI(BaseInterfaceModel):
 
         Args:
             address (str): The address of the contract.
+            slot (int): Storage slot to read the value of.
 
         Returns:
             bytes: The value of the storage slot.
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -172,7 +172,6 @@ class ProviderAPI(BaseInterfaceModel):
         Returns:
             bytes: The value of the storage slot.
         """
-        raise NotImplementedError("get_storage_at is not implemented by this provider")
 
     @abstractmethod
     def get_nonce(self, address: str) -> int:
@@ -604,6 +603,9 @@ class Web3Provider(ProviderAPI, ABC):
 
     def get_code(self, address: str) -> bytes:
         return self.web3.eth.get_code(address)  # type: ignore
+
+    def get_storage_at(self, address: str, slot: int) -> bytes:
+        return self.web3.eth.get_storage_at(address, slot)  # type: ignore
 
     def send_call(self, txn: TransactionAPI) -> bytes:
         return self.web3.eth.call(txn.dict())
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -161,7 +161,17 @@ class ProviderAPI(BaseInterfaceModel):
             bytes: The contract byte-code.
         """
 
+    @raises_not_implemented
     def get_storage_at(self, address: str, slot: int) -> bytes:
+        """
+        Gets the raw value of a storage slot of a contract.
+
+        Args:
+            address (str): The address of the contract.
+
+        Returns:
+            bytes: The value of the storage slot.
+        """
         raise NotImplementedError("get_storage_at is not implemented by this provider")
 
     @abstractmethod
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -161,6 +161,9 @@ class ProviderAPI(BaseInterfaceModel):
             bytes: The contract byte-code.
         """
 
+    def get_storage_at(self, address: str, slot: int) -> bytes:
+        raise NotImplementedError("get_storage_at is not implemented by this provider")
+
     @abstractmethod
     def get_nonce(self, address: str) -> int:
         """
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -458,18 +458,25 @@ class ContractCache(BaseManager):
     def resolve_proxy(self, address: AddressType) -> Optional[AddressType]:
         code = self.provider.get_code(address).hex()[2:]
         patterns = [
-            r"363d3d373d3d3d363d73(.{40})5af43d82803e903d91602b57fd5bf3",  # eip-1167 minimal proxy contract
-            r"366000600037611000600036600073(.{40})5af4602c57600080fd5b6110006000f3",  # vyper <0.2.9 create_forwarder_to
-            r"36603057343d52307f830d2d700a97af574b186c80d40429385d24241565b08a7c559ba283a964d9b160203da23d3df35b3d3d3d3d363d3d37363d73(.{40})5af43d3d93803e605b57fd5bf3",  # 0xsplits clones
+            # eip-1167 minimal proxy contract
+            r"363d3d373d3d3d363d73(.{40})5af43d82803e903d91602b57fd5bf3",
+            # vyper <0.2.9 create_forwarder_to
+            r"366000600037611000600036600073(.{40})5af4602c57600080fd5b6110006000f3",
+            # 0xsplits clones
+            r"36603057343d52307f830d2d700a97af574b186c80d40429385d24241565b08a7c559ba283a964d9"
+            + r"b160203da23d3df35b3d3d3d3d363d3d37363d73(.{40})5af43d3d93803e605b57fd5bf3",
         ]
         for pattern in patterns:
             if match := re.match(pattern, code):
                 return self.conversion_manager.convert(match.group(1), AddressType)
 
         slots = [
-            int(self.provider.web3.keccak(text="eip1967.proxy.implementation").hex(), 16) - 1,  # eip-1967 standard proxy storage slots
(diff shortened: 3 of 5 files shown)


banteg:
need to drop python 3.7 to use walrus operator

banteg:
converted to draft, will also add EIP-1967 and EIP-1822

src/ape/managers/chain.py
```diff
@@ -450,6 +455,19 @@ def get(
 
         return contract_type or default
 
+    def resolve_proxy(self, address: AddressType) -> Optional[AddressType]:
+        code = self.provider.get_code(address).hex()[2:]
+        patterns = [
+            r"363d3d373d3d3d363d73(.{40})5af43d82803e903d91602b57fd5bf3",  # eip-1167 minimal proxy contract
+            r"366000600037611000600036600073(.{40})5af4602c57600080fd5b6110006000f3",  # vyper <0.2.9 create_forwarder_to
+            r"36603057343d52307f830d2d700a97af574b186c80d40429385d24241565b08a7c559ba283a964d9b160203da23d3df35b3d3d3d3d363d3d37363d73(.{40})5af43d3d93803e605b57fd5bf3",  # 0xsplits clones
```
Ninjagod1251, line 463:
```suggestion
            r"363d3d373d3d3d363d73(.{40})5af43d82803e903d91602b57fd5bf3",
            # eip-1167 minimal proxy contract
            r"366000600037611000600036600073(.{40})5af4602c57600080fd5b6110006000f3",
            # vyper <0.2.9 create_forwarder_to
            r"36603057343d52307f830d2d700a97af574b186c80d40429385d24241565b08a7c559ba2"
            + r"83a964d9b160203da23d3df35b3d3d3d3d363d3d37363d73(.{40})5af43d3d93803e605b57fd5bf3",
            # 0xsplits clones
```

Ninjagod1251: To make flake8 happy. :)

banteg:
possible optimizations:
- cache get_code, being mindful of selfdestruct and create2, these should probably be cached with a limited ttl
- add proxy contract type and cache it (but for how long? cache invalidation problem)
- add batch jsonrpc request machinery, so we can get code and all the needed slots in one request
- (high effort) scan all recently deployed contracts to determine the most likely order different proxy types

src/ape/managers/chain.py
```diff
@@ -450,6 +458,66 @@ def get(
 
         return contract_type or default
 
+    def resolve_proxy(self, address: AddressType) -> Optional[AddressType]:
```
unparalleled-js, line 461: We need to probably move this to `ape-ethereum` and the `EcosystemAPI`

`decode_proxy(address)`.

This way, plugins like `ape-starknet` can decode their own proxy standards.

src/ape/managers/chain.py
```diff
@@ -450,6 +458,66 @@ def get(
...
+            # eip-1822 universal upgradeable proxy standard (uups)
+            self.provider.web3.keccak(text="PROXIABLE"),
+        ]
+        for slot in slots:
+            storage = self.provider.web3.eth.get_storage_at(address, slot)
+            if sum(storage) != 0:
+                return self.conversion_manager.convert(storage[-20:].hex(), AddressType)
+
+        # eip-1967 beacon proxy
+        slot = int(self.provider.web3.keccak(text="eip1967.proxy.beacon").hex(), 16) - 1
```
unparalleled-js, line 491: We can add tests for these! Let me know if you need help with that.
The way I would approach is add pre-compiled proxies that point to the addresses of other contracts found in `tests/functional/data/contract_types` and interact with the proxies and show that its the same. If you added the proxies to the parametrized `tests.functional.conftests.contract_instance` fixture, the tests that interact with the contracts with also run against the proxies.

src/ape/managers/chain.py
```diff
@@ -450,6 +458,66 @@ def get(
...
+        storage = self.provider.web3.eth.get_storage_at(address, slot)
```
unparalleled-js, line 492: This `web3.eth` specific call is a big reason why we need to move this to `ape-Ethereum`, but also maybe we should add a `ProviderAPI` method `get_storage_at()` that is `@raise_not_implemented` by default.

src/ape/managers/chain.py
```diff
@@ -450,6 +458,66 @@ def get(
...
+            abi = MethodABI(
+                type="function",
+                name="implementation",
+                stateMutability="view",
+                outputs=[ABIType(type="address")],
+            )
+            beacon = self.conversion_manager.convert(storage[-20:].hex(), AddressType)
+            return ContractCall(abi, beacon)()
+
+        # gnosis safe proxy
```
unparalleled-js, line 503: Enumerating all the proxies in a doc would be nice! There are so many. Even if it happens in the doc str of the method, that is fine. Else, we may need `contracts.md` userguide (in `docs/userguides`). If you go that route, you don't have to thoroughly create it, just initialize it, add the enumeration, and we will fill it in later with non-relevant details to this PR.

unparalleled-js (commented):
Wow, you are amazing @banteg 

The only real blocker for me that prevents me from merging this is that it isn't multi-chain friendly. See comment about moving logic to `ape-ethereum` and adding an API method in `EcosystemAPI`! Thoughts @fubuloubu ?

Adding tests would help me sleep at night but I understand that it may be a beast to take on, but I am happy to assist.

src/ape/managers/chain.py
```diff
@@ -441,6 +444,11 @@ def get(
         contract_type = self._get_contract_type_from_disk(address)
 
         if not contract_type:
+            # Contract could be a minimal proxy
+            target = self.resolve_proxy(address)
+            if target:
+                return self.get(target)
```
unparalleled-js, line 450: Just thought of this: Maybe this could happen later, but we should also be caching these as well and we may need to adjust our caching schema to include the proxy type.

src/ape/managers/chain.py
unparalleled-js: But directly using `web3` on the provider assumes the provider is a `Web3Provider` which is does not necessarily have to be.

src/ape/managers/chain.py
banteg: added `proxy_info` with some metadata which we can cache. i think we also need to add ttl or user-initiated cache invalidation.

src/ape/managers/chain.py
banteg: added `ProxyType` enum

src/ape/managers/chain.py
```diff
@@ -441,6 +441,12 @@ def get(
         contract_type = self._get_contract_type_from_disk(address)
 
         if not contract_type:
+            # Contract could be a minimal proxy
+            if hasattr(self.provider.network.ecosystem, "proxy_info"):
+                proxy = self.provider.network.ecosystem.proxy_info(address)
```
fubuloubu, line 446: Should definitely add this method to `EcosystemAPI` abstract base class with a default implementation that returns empty list

src/ape_ethereum/ecosystem.py
```diff
@@ -42,6 +46,23 @@
...
+    Beacon = 2  # eip-1967 beacon proxy
+    UUPS = 3  # # eip-1822 universal upgradeable proxy standard
+    Vyper = 4  # vyper <0.2.9 create_forwarder_to
+    Clones = 5  # 0xsplits clones
+    GnosisSafe = 6
+    OpenZeppelin = 7  # openzeppelin upgradeability proxy
+
+
+@dataclass
+class ProxyInfo:
```
fubuloubu, line 61: This might be better off living alongside `EcosystemAPI` as a standard type

src/ape/managers/chain.py
unparalleled-js, line 446: Can we name to `get_proxy_info()`?
`proxy_info` sounds more like a property name than a method

src/ape_ethereum/ecosystem.py
unparalleled-js: That would make the `ProxyType` enum kind of complicated though?
What if an ecosystem defines new types?

src/ape/managers/chain.py
unparalleled-js: Yes, then you don't have to use `hasattr()`, the method will just always exists.

fubuloubu: Exactly!

src/ape_ethereum/ecosystem.py
fubuloubu: Oh, yeah fair

I think instead of `dataclass` it should maybe use `pydantic.BaseModel` then so it can serialize

sslivkoff:
this PR is an excellent reference on commonly-used proxy patterns

is there a reason why EIP-897 is not included? I've encountered many proxy contracts that just have an `implementation()` function. I don't think that would be detected by the current version of `proxy_info()`

banteg:
> is there a reason why EIP-897 is not included? I've encountered many proxy contracts that just have an `implementation()` function. I don't think that would be detected by the current version of `proxy_info()`

my omission, the goal is to include as many patterns as possible, so the user doesn't have to worry about how to instantiate contracts

banteg:
noticed this when adding eip-897, an implementation is also detected as proxy, although pointing to a zero address implementation.

for now i added a check for zero address, but maybe we should have an option for `ContractCache.get` to ignore proxies, so it can't end up in an endless loop.

are there any examples of proxies nested more than one level?

src/ape_ethereum/ecosystem.py
```diff
@@ -147,6 +168,89 @@ def decode_address(cls, raw_address: RawAddress) -> AddressType:
...
+
+        keccak = self.provider.web3.keccak
+        slots = {
+            ProxyType.Standard: int(keccak(text="eip1967.proxy.implementation").hex(), 16) - 1,
+            ProxyType.Beacon: int(keccak(text="eip1967.proxy.beacon").hex(), 16) - 1,
+            ProxyType.OpenZeppelin: keccak(text="org.zeppelinos.proxy.implementation"),
+            ProxyType.UUPS: keccak(text="PROXIABLE"),
+        }
+        for type, slot in slots.items():
+            storage = self.provider.web3.eth.get_storage_at(address, slot)
```
banteg, line 195: mypy warns web3 is not guaranteed to be the provider, maybe we should add getting to a provider api

src/ape/managers/chain.py
banteg: i would appreciate help with the tests, thank you

src/ape/api/networks.py
```diff
@@ -357,6 +357,9 @@ def get_network_data(self, network_name: str) -> Dict:
 
         return data
 
+    def get_proxy_info(self, address: AddressType) -> Any:
```
unparalleled-js, line 360: typically, we include docs for public methods in the API classes.

src/ape/api/providers.py
```diff
@@ -161,6 +161,9 @@ def get_code(self, address: str) -> bytes:
             bytes: The contract byte-code.
         """
 
+    def get_storage_at(self, address: str, slot: int) -> bytes:
```
unparalleled-js, line 164: For consistency's sake, we have a `@raises_not_implemented` util decorator

unparalleled-js (approved):
This is looking good!
Docs and tests would be nice, but we can always add those ourselves.

Thank you for your contribution :)

src/ape/api/networks.py
```diff
@@ -357,6 +357,18 @@ def get_network_data(self, network_name: str) -> Dict:
...
+        """
+        Information about a proxy contract such as proxy type and implementation address.
+
+        Args:
+            address (str): The address of the contract.
+
+        Returns:
+            Any: ProxyInfo or None if the contract does not use any known proxy pattern.
```
fubuloubu, line 368: Can return type instead be `Optional[ProxyInfo]`?

src/ape/api/providers.py
```diff
@@ -161,6 +161,19 @@ def get_code(self, address: str) -> bytes:
...
+        """
+        Gets the raw value of a storage slot of a contract.
+
+        Args:
+            address (str): The address of the contract.
+
+        Returns:
+            bytes: The value of the storage slot.
+        """
+        raise NotImplementedError("get_storage_at is not implemented by this provider")
```
fubuloubu, line 175: No need for this with `raises_not_implemented` decorator!

fubuloubu: Also, I think Web3Provider can implement this one by default maybe, which geth also inherits

src/ape_ethereum/ecosystem.py
```diff
@@ -42,6 +46,23 @@
...
+    Vyper = 4  # vyper <0.2.9 create_forwarder_to
+    Clones = 5  # 0xsplits clones
+    GnosisSafe = 6
+    OpenZeppelin = 7  # openzeppelin upgradeability proxy
+    Delegate = 8  # eip-897 delegate proxy
+
+
+class ProxyInfo(BaseModel):
+    type: ProxyType
+    target: AddressType
```
fubuloubu, line 63: Maybe a way to do this in alignment with `get_proxy_info` in `EcosystemAPI` is to make a `ProxyInfoAPI` object with only `.target` in it, and then here you would subclass and add the type. Really, the target it just what you want when you use it downstream.

src/ape_ethereum/ecosystem.py
```diff
@@ -147,6 +168,89 @@ def decode_address(cls, raw_address: RawAddress) -> AddressType:
     def encode_address(cls, address: AddressType) -> RawAddress:
         return str(address)
 
+    def is_proxy(self, address: AddressType) -> bool:
+        return self.get_proxy_info(address) is not None
```
fubuloubu, line 172: Maybe we don't need this afterall w/ `Optional`, can just do `.get_proxy_info(address).target or address`

src/ape/api/networks.py
banteg: It can, but ProxyInfo can vary by ecosystem

src/ape_ethereum/ecosystem.py
banteg: .target will fail if it returns None

src/ape/api/networks.py
banteg: Added ProxyInfoAPI

fubuloubu (dismissed):
Looking solid!

src/ape/api/networks.py
```diff
@@ -21,6 +22,10 @@
 LOCAL_NETWORK_NAME = "local"
 
 
+class ProxyInfoAPI(BaseModel):
```
unparalleled-js, line 25: right now, `ProxyInfo` is not showing up in the docs. But if you a class doc to it, it does, something like

```python
    """
    Information about a proxy contract.
    """
```

And to get the `target` field to show up, you also have to add a doc, something like:

```
    """The address to the implementation contract."""
```

src/ape/api/networks.py
```diff
@@ -357,6 +362,18 @@ def get_network_data(self, network_name: str) -> Dict:
...
 
+    def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
+        """
+        Information about a proxy contract such as proxy type and implementation address.
+
+        Args:
+            address (str): The address of the contract.
+
+        Returns:
+            ProxyInfoAPI or None if the contract does not use any known proxy pattern.
```
unparalleled-js, line 373:
```suggestion
            Optional[:class:`~ape.api.networks.ProxyInfoAPI`]: Returns ``None`` if the contract 
            does not use any known proxy pattern.
```

This would link to the class so the user can see that it has an address field.
Only works

src/ape/api/providers.py
```diff
@@ -161,6 +161,18 @@ def get_code(self, address: str) -> bytes:
             bytes: The contract byte-code.
         """
 
+    @raises_not_implemented
+    def get_storage_at(self, address: str, slot: int) -> bytes:
+        """
+        Gets the raw value of a storage slot of a contract.
+
+        Args:
+            address (str): The address of the contract.
```
unparalleled-js, line 170: `slot` missing from args doc

unparalleled-js (approved):
Some small issues with the docs, but I am happy to address those later.

banteg:
I think that would be all from me here. Huge thanks to Jules and Doggie for your help and guidance in understanding the architecture.

unparalleled-js (approved):
LGTM

fubuloubu (approved):
Looks really slick!

PR #757, opened 2022-05-28
feat!: support decoding multiple ABIs at the same time, including ds-note library logs

banteg:
### What I did
- if no abi is provided to `Receipt.decode_logs`, it would fetch all contract types that have emitted logs and decode all events. undecoded logs are returned as is.
- add decoding of [ds-note](https://github.com/dapphub/ds-note) which is commonly found in MakerDAO contracts

### How I did it
added ds-note decoding to ecosystem api and added a fallback to it in the receipt api. this is very cheap, it can deduce whether it's a note from byte padding in the topic before fetching any contract types. then it finds a function which matches the selector and decodes the data based on the function abi.

### How to verify it
```
list(chain.provider.get_transaction('0x460cb2476a7a0e8e1933d935a1c386851d5c007e957dbf9bd2e32a8494b38e9d').decode_logs())
```

See also my similar pull request [in brownie](https://github.com/eth-brownie/brownie/pull/985).

### Checklist


src/ape/api/transactions.py
```diff
@@ -200,6 +203,56 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+        except StopIteration:
+            return None
+
+        # in versions of ds-note the data field uses either (uint256,bytes) or (bytes) encoding
+        # instead of guessing, assume the payload starts right after the selector
+        data = decode_hex(log["data"])
+        input_types = [i.canonical_type for i in abi.inputs]
+        values = decode_abi(input_types, data[data.index(selector) + 4 :])  # noqa: E203
+
+        return ContractLog(  # type: ignore
```
banteg, line 247: I think `ContractLog` misses the `address` field, it would be useful for dealing with logs emitted from multiple contracts

src/ape/api/transactions.py
```diff
@@ -200,6 +203,56 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+                if selector == keccak(text=func.selector)[:4]
+            )
+        except StopIteration:
+            return None
+
+        # in versions of ds-note the data field uses either (uint256,bytes) or (bytes) encoding
+        # instead of guessing, assume the payload starts right after the selector
+        data = decode_hex(log["data"])
+        input_types = [i.canonical_type for i in abi.inputs]
+        values = decode_abi(input_types, data[data.index(selector) + 4 :])  # noqa: E203
```
banteg, line 245: Black clashes with flake8 [over slices](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices)

src/ape/api/transactions.py
```diff
@@ -200,6 +203,56 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+        contracts = {log["address"] for log in self.logs}
+        contract_types = {addr: self.chain_manager.contracts.get(addr) for addr in contracts}
+        decoded = []
+        for log in self.logs:
+            note = self.decode_ds_note(contract_types[log["address"]], log)
+            if note:
+                decoded.append(note)
+            else:
+                try:
+                    event_abi = next(
```
banteg, line 217: This could be simplified if we add choice by selector to ethpm_types. I see it like this:
```python
ContractType.events["Transfer"]
ContractType.events[selector]
ContractType.events[keccak(selector)]
```
I'll make a pull request.

src/ape/api/transactions.py
```diff
@@ -200,6 +203,56 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
 
         yield from self.provider.network.ecosystem.decode_logs(abi, self.logs)
 
+    @property
+    def events(self):
+        contracts = {log["address"] for log in self.logs}
+        contract_types = {addr: self.chain_manager.contracts.get(addr) for addr in contracts}
+        decoded = []
+        for log in self.logs:
+            note = self.decode_ds_note(contract_types[log["address"]], log)
```
banteg, line 212: To get this working with decode_logs, we need to make it more generic, and pass it the full ContractType.

src/ape/api/transactions.py
unparalleled-js: one workaround for this is to set the index as a variable first

src/ape/api/transactions.py
```diff
@@ -200,6 +203,56 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
 
         yield from self.provider.network.ecosystem.decode_logs(abi, self.logs)
 
+    @property
+    def events(self):
```
unparalleled-js, line 207: Should we call these `logs` instead of events? I am used to thinking of the event as the definition in the contract and the log as the data in python

fubuloubu: We already have `.logs` property, so this makes sense. It's basically decoding all the logs according to the event definitions in the contract cache

unparalleled-js: I am guessing this should actually be in `ape_ethereum.transactions.Receipt`

banteg:
ready for review

src/ape_ethereum/ecosystem.py
```diff
@@ -490,3 +490,37 @@ def decode_value(t, v) -> Any:
                 block_hash=log["blockHash"],
                 block_number=log["blockNumber"],
             )  # type: ignore
+
+    def decode_ds_note(self, log: dict) -> Optional[ContractLog]:
```
fubuloubu, line 494: Is this just a useful helper?

fubuloubu (approved):

src/ape_ethereum/ecosystem.py
banteg: yes, it's used to decode [custom logs](https://github.com/dapphub/ds-note) which encode the function call and args. these are common in makerdao contracts.

src/ape/api/transactions.py
```diff
@@ -231,10 +233,23 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+                abi = abi.abi
 
-        yield from self.provider.network.ecosystem.decode_logs(abi, self.logs)
+            yield from self.provider.network.ecosystem.decode_logs(abi, self.logs)
+        else:
+            # if abi is not provided, decode all events
+            for log in self.logs:
+                contract_type = self.chain_manager.contracts.get(log["address"])
+                try:
+                    event_abi = contract_type.events[log["topics"][0]]  # type: ignore
```
unparalleled-js, line 246: `contract_type` could be `None` here which would cause `AttributeError`
and if we fail to find a contract type, we should maybe raise an error

src/ape_ethereum/ecosystem.py
unparalleled-js: We could call it `decode_dappsys_note()` to give more context (optional)

src/ape/api/transactions.py
unparalleled-js, line 246: any alternative to using `# type ignore` here?

maybe:

```
selector = str(log["topics"][0])
event_abi = contract_type.events[selector]
```

unparalleled-js (approved):
My only real concern is the potential `AttributeError`, but looks good!

src/ape/api/transactions.py
banteg: i think we could, but the type should be bytes or HexBytes

src/ape_ethereum/ecosystem.py
banteg: this library is more known as ds-note

src/ape/api/transactions.py
unparalleled-js: Ah, I meant this line here: `contract_type = self.chain_manager.contracts.get(log["address"])`

`contract_type` is `Optional[ContractType]` so it may not have the `.events` attribute if it's `None`. I am surprised `mypy` is not picking it up though, so maybe I am missing something.

But in any case, I am happy to fix anything remaining and get this through :)

src/ape/api/transactions.py
```diff
@@ -231,10 +233,27 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+                if not contract_type:
+                    logger.warning(f"Failed to locate contract at '{log['address']}'.")
+                    continue
+
+                try:
+                    event_abi = contract_type.events[log["topics"][0]]
+                    yield from self.provider.network.ecosystem.decode_logs(event_abi, [log])
+                except (StopIteration, KeyError, DecodingError):
+                    try:
+                        yield self.provider.network.ecosystem.decode_ds_note(log)  # type: ignore
```
unparalleled-js, line 254: Realizing this is an Ethereum specific feature here...

I am actually wondering if this should be in the `EcosystemAPI.decode_logs` method (in ape_Ethereum) and kept as an implementation detail

unparalleled-js:
I am taking a deep dive into this feature and adjusting some things. I want find a better way to detect DS notes.
In the process of doing so, I converted the ds note repo to an project and am tests there that will use this feature: https://github.com/unparalleled-js/ds-note

tests/functional/test_receipt.py
```diff
@@ -2,12 +2,58 @@
...
+
+
+def test_decode_logs_specify_abi_as_event(invoke_receipt, vyper_contract_instance):
+    abi = vyper_contract_instance.NumberChange
+    logs = [log for log in invoke_receipt.decode_logs(abi=abi)]
+    assert len(logs) == 1
+    assert logs[0].newNum == 1
+
+
+def test_decode_logs_with_ds_notes(ds_note_test_contract, owner):
```
unparalleled-js, line 36: @banteg We have ds note tests now!
NOTE: I having trouble with events from payable methods though... Not sure why! `contract.test_4` triggers a payable method.
These tests are taken from the library itself

src/ape/api/transactions.py
banteg: yes, i think it should. it's better if we abstract it into the ecosystem's `decode_logs`

src/ape_ethereum/transactions.py
```diff
@@ -156,6 +159,54 @@ def raise_for_status(self):
...
+    ) -> Iterator[ContractLog]:
+        if not abi:
+            # Check for DS-Note library logs.
+            for log in self.logs:
+                try:
+                    yield self._decode_ds_note(log)
+                except DecodingError:
+                    continue
+
+        return super().decode_logs(abi)
```
fubuloubu, line 173: So this will add two loops worth of processing to this method, is there a better way of folding this in, or a heuristic that can be used to reduce the amount of processing? Picturing a transaction w/ several thousand logs it might be slow-ish

fubuloubu (approved):
Needs working CI, one comment for consideration but non-blocking

src/ape/api/transactions.py
```diff
@@ -231,10 +233,52 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
+                    logger.warning(f"Failed to locate contract at '{address}'.")
+                    continue
+
+                log_topics = log.get("topics", [])
+                if not log_topics:
+                    raise ValueError("Missing 'topics' in log data")
+
+                selector = log["topics"][0]
+                if selector not in contract_type.events:
+                    # Likely a library log
```
unparalleled-js, line 258: @fubuloubu I did this to address your feedback
So now instead of looping over `self.logs` twice, it will only attempt to decode library logs if it does not locate the selector in the contact type (the heuristic).

The default impl of `def decode_library_log()` returns `None`, so it's not a breaking change.
Otherwise, ecosystems can override it to attempt to process lib logs.

unparalleled-js:
@banteg Lots has changed since you first opened the PR! Let us know your thoughts on those changes before this lands, if you can :)

And as always, thank you for your contribution and invaluable insights

banteg:
the changes look good

unparalleled-js:
Bump! Can we get these changes in for staging purposes against https://github.com/ApeWorX/ape/pulls

fubuloubu (approved):

fubuloubu (dismissed):

src/ape/api/transactions.py
```diff
@@ -233,14 +236,52 @@ def decode_logs(
...
+        if abi:
+            if not isinstance(abi, (list, tuple)):
+                abi = [abi]
 
-        for event_abi in abi:
-            if not isinstance(event_abi, EventABI):
-                event_abi = event_abi.abi
+            for event_abi in abi:
+                if not isinstance(event_abi, EventABI):
+                    event_abi = event_abi.abi
```
banteg, line 245: this will do nothing

src/ape/api/transactions.py
```diff
@@ -233,14 +236,52 @@ def decode_logs(
...
 
-            yield from self.provider.network.ecosystem.decode_logs(event_abi, self.logs)
+                yield from self.provider.network.ecosystem.decode_logs(event_abi, self.logs)
+        else:
+            # If ABI is not provided, decode all events
+            logs_map: Dict[str, Dict[str, Tuple[EventABI, List[Dict]]]] = {}
+            for log in self.logs:
+                address = log["address"]
+                contract_type = self.chain_manager.contracts.get(address)
```
banteg, line 253: i think this is slow, we better do this:
1. extract all addresses from the logs first
2. get all contracts in parallel
3. build a map of `address → selector → abi`
4. decode logs using the above map

src/ape/managers/chain.py
```diff
@@ -504,6 +505,44 @@ def __getitem__(self, address: AddressType) -> ContractType:
 
         return contract_type
 
+    def get_all(
```
fubuloubu, line 508: Not a huge fan of this name

fubuloubu (approved):
Would it be possible to move the updates to the contract cache to a separate PR?

src/ape/managers/chain.py
unparalleled-js: Ok, give me an alternative

unparalleled-js:
> Would it be possible to move the updates to the contract cache to a separate PR?

Yes, it will have to merge before this one then

src/ape/managers/chain.py
unparalleled-js: Or at least tell me why you don't like it

banteg:
found another issue, logs might come in a wrong order, i'll fix so this assumption holds
```python
from operator import attrgetter
tx = '0x460cb2476a7a0e8e1933d935a1c386851d5c007e957dbf9bd2e32a8494b38e9d'
logs = list(chain.provider.get_transaction(tx).decode_logs())
assert logs == sorted(logs, key=attrgetter('block_number', 'log_index'))
```

unparalleled-js:
- [ ] Merge this one first: https://github.com/ApeWorX/ape/pull/897

src/ape/managers/chain.py
fubuloubu: Sorry, it wasn't very descriptive

src/ape/api/transactions.py
```diff
@@ -233,14 +237,34 @@ def decode_logs(
...
+            yield from self.provider.network.ecosystem.decode_logs(event_abis, self.logs)
 
-            yield from self.provider.network.ecosystem.decode_logs(event_abi, self.logs)
+        else:
+            # If ABI is not provided, decode all events
+            addresses = {x["address"] for x in self.logs}
+            contract_types = self.chain_manager.contracts.get_multiple(addresses)
+            # address → selector → abi
+            selectors = {
+                address: {keccak(text=abi.selector): abi for abi in contract.events}
```
fubuloubu, line 253: `keccak` makes it ethereum-specific, can this fall through into an `EcosystemAPI` function (maybe `decode_all_logs(...)`)

fubuloubu (changes requested):

src/ape/api/networks.py
```diff
@@ -299,18 +299,33 @@ def encode_transaction(
         """
 
     @abstractmethod
-    def decode_logs(self, abi: EventABI, raw_logs: List[Dict]) -> Iterator[ContractLog]:
+    def decode_logs(
+        self, events: Union[EventABI, List[EventABI]], logs: List[Dict]
+    ) -> Iterator["ContractLog"]:
```
fubuloubu, line 304: Wonder if this might be somewhat better UX:
```suggestion
    def decode_logs(
        self, raw_logs: Iterable[Dict], *event_abis: List[EventABI],
    ) -> Iterator["ContractLog"]:
```

Used like this:
```py
txn.decode_logs(project.MyContract.EventA, project.MyContract.EventB, ...)
```
or even like this:
```py
txn.decode_logs(*project.MyContract.events)

File: src/ape_ethereum/ecosystem.py [hunk 17]
```diff
@@ -442,13 +443,54 @@ def decode_logs(
...
-                block_number=to_int(log["blockNumber"]),
-                block_hash=log["blockHash"],
-                log_index=to_int(log["logIndex"]),
-            )
+            )  # type: ignore
+
+    def decode_library_log(self, log: Dict) -> Optional[ContractLog]:
+        return self._decode_ds_note(log)
+
+    def _decode_ds_note(self, log: Dict) -> Optional[ContractLog]:
```
fubuloubu, line 459: Could this function live under `ape_ethereum.transactions.Receipt` instead, and be used directly instead of the indirection involved w/ adding `EcosystemAPI.decode_library_log`. It doesn't seem to be used by other ecosystems, so I think it would be more at home reducing the overall complexity of the API layer a bit.

src/ape/api/networks.py
unparalleled-js: Your suggested change shows `raw_logs` as the first parameter but the use-cases show the events as coming first. I am guessing the raw data would have to come first in this case since it is `*args` for the event types?

So it would be like:

```python
log_data = get_log_data()  # Pretend
decoded_logs = txn.decode_logs(log_data, project.MyContract.EventA, project.MyContract.EventB)
```

src/ape_ethereum/ecosystem.py
unparalleled-js: Would that limit the use-case of decoding DS Notes outside the context of having a receipt?
Like what if I got the data from someplace else and I want to decode it, I could use provider method alone by having it this way.

fubuloubu: You can still use the method, it's just private. I don't have a problem with it being a public method either, just don't think it should be an API method

src/ape/api/networks.py
unparalleled-js: So this won't do a ton for UX then...

Inside receipt API implementations we do it like this:


```
self.network.ecosystem.decode_logs(log_filter.events, logs)
```

and you suggestion would make it so we have to do

```
self.network.ecosystem.decode_logs(logs, *log_filter.events)
```

it seems like it add unnecessary work to the Python code and an extra character `*`

unparalleled-js:
[Doing QA with @dtdang and wearing that hat right now]

Tried to follow the `How to verify it section` but doing this:

`ape console`

`list(chain.provider.get_transaction('0x460cb2476a7a0e8e1933d935a1c386851d5c007e957dbf9bd2e32a8494b38e9d').decode_logs())`

and it hangs for 2 minutes before timing out.

I realize now that I am supposed to use mainnet. However - I think the default timeout should be smaller for local networks. Thoughts?

- [ ] Make timeout smaller for local networks

src/ape/api/networks.py
fubuloubu: Yes, but the set of events you can give it can be just one event (no need to wrap in a list), several events of your choosing, or all of the events applied with the `*` args syntax. It feels much more flexible

fubuloubu:
> I think the default timeout should be smaller for local networks. Thoughts?

Yes, I think this makes sense

unparalleled-js:
> Yes, I think this makes sense

Done! https://github.com/ApeWorX/ape/pull/911/files
I made it a separate PR as it was a bigger refactor than expected.

src/ape/api/networks.py
```diff
@@ -305,13 +305,13 @@ def encode_transaction(
         """
 
     @abstractmethod
-    def decode_logs(self, abi: EventABI, raw_logs: List[Dict]) -> Iterator[ContractLog]:
+    def decode_logs(self, logs: List[Dict], *events: EventABI) -> Iterator["ContractLog"]:
```
banteg, line 308: the variadic argument feels a bit weird here

src/ape_ethereum/ecosystem.py
```diff
@@ -468,13 +464,12 @@ def decode_logs(
...
             event_arguments = abi.decode(topics, log["data"])
-
             yield ContractLog(
-                name=abi.event_name,
+                block_hash=log["blockHash"],
+                block_number=to_int(log["blockNumber"]),
                 contract_address=self.decode_address(log["address"]),
                 event_arguments=event_arguments,
+                log_index=log["logIndex"],
+                name=abi.event_name,
```
banteg, line 473: should we say `event_name` to be more clear?

src/ape_ethereum/ecosystem.py
```diff
@@ -468,13 +464,12 @@ def decode_logs(
...
 
             event_arguments = abi.decode(topics, log["data"])
-
             yield ContractLog(
-                name=abi.event_name,
+                block_hash=log["blockHash"],
+                block_number=to_int(log["blockNumber"]),
                 contract_address=self.decode_address(log["address"]),
                 event_arguments=event_arguments,
+                log_index=log["logIndex"],
```
banteg, line 472: should we add a `transaction_index` field?

src/ape/managers/chain.py
unparalleled-js: `chain.contracts.get_multiple(addresses)` seems just as descriptive as `chain.contracts[address]` or `chain.contracts.get(address)`

fubuloubu: Sure, that works

src/ape/api/networks.py
unparalleled-js: @fubuloubu wanted it and I was hesitant at first, but I kind of like how it feels by not having to wrap single ABIs in a list first.

But I can go either way still, let's reach a consensus

banteg:
wen merge

unparalleled-js:
- [ ] Currently experiencing weird hanging behavior,, Will check box once resolved

src/ape/api/networks.py
fubuloubu: It's a breaking change anyways, live life on the edge 🤘

src/ape/types/__init__.py
```diff
@@ -168,6 +169,28 @@ class ContractLog(BaseModel):
...
     """The index of the log on the transaction."""
 
+    transaction_index: Optional[int] = None
+    """
+    The index of the transaction's position when the log was created.
+    Is `None` when from the pending block.
+    """
+
+    class Config:
+        alias_generator = to_snake_case
```
fubuloubu, line 179: What's this do?

src/ape/utils/misc.py
```diff
@@ -251,6 +251,10 @@ def to_int(value) -> int:
     raise ValueError(f"cannot convert {repr(value)} to int")
 
 
+def to_snake_case(string: str) -> str:
```
fubuloubu, line 254: There's a library with a bunch of well-tested string conversion functions like this

fubuloubu (commented):
What's with the use of snake casing everywhere?

src/ape/utils/misc.py
unparalleled-js: EthTester raw responses contain snake case but turns out we can handle that elsewhere

fubuloubu (approved):

Issue #758, opened 2022-05-29
idempotence issue with contract calls


banteg:
### Environment information

* OS: macOS
* Python Version: 3.9.10
* `ape` and plugin versions: development

### What went wrong?

```python
from ape import *

with networks.ethereum.mainnet.use_provider('geth'):
    multicall = Contract("0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696")
    calls = [
        ["0xf66Cd2f8755a21d3c8683a10269F795c0532Dd58", b"\x95\xd8\x9bA"],
        ["0xf66Cd2f8755a21d3c8683a10269F795c0532Dd58", b"1<\xe5g"],
    ]
    for i in range(3):
        print(multicall.tryAggregate.call(False, calls))
```

One might expect the same call would return the same value here if called twice, but for some reason on the second call output decoding errors out.
```python
(returnData(success=True, returnData=HexBytes('0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000007436f726544414f00000000000000000000000000000000000000000000000000')), returnData(success=True, returnData=HexBytes('0x0000000000000000000000000000000000000000000000000000000000000012')))
...
eth_abi.exceptions.NonEmptyPaddingBytes: Boolean must be either 0x0 or 0x1.  Got: b'\x02'
```

banteg:
There must be a hint somewhere there
```python
from ape import *

def main():
    multicall = Contract("0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696")
    calls = [
        ["0xf66Cd2f8755a21d3c8683a10269F795c0532Dd58", b"1<\xe5g"],
    ]
    for i in range(3):
        print(multicall.tryAggregate.call(False, calls))
```
Outputs
```python
returnData(success=True, returnData=HexBytes('0x0000000000000000000000000000000000000000000000000000000000000012'))
returnData(success=True, returnData=HexBytes('0x0000000000000000000000000000000000000000000000000000000000000001'))
returnData(success=True, returnData=HexBytes('0x0000000000000000000000000000000000000000000000000000000000000001'))
```

banteg:
The problem occurs here. The left pane shows how the `output_types` value changes between the two runs.

<img width="1465" alt="Screen Shot 2022-05-29 at 23 05 56" src="https://user-images.githubusercontent.com/4562643/170887418-7835d006-2e04-4719-9094-659031e36cbf.png">

[...]

PR #759, opened 2022-05-29
fix: abi idempotence

banteg:
### What I did
Fixed an issue where ABI were modified in place.

fixes: #758

### How I did it
Copy ABI before modification.

### How to verify it
See linked issue.

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #762, opened 2022-05-31
fix: array struct idempotence issue

unparalleled-js:
### What I did

fixes: #758

or to be used as a reference for https://github.com/ApeWorX/ape/pull/759

I will look deeper tomorrow to better access and work with @banteg 

### Checklist


src/ape/utils/abi.py
```diff
@@ -90,7 +90,7 @@ def parse(self, output_types: List[ABIType], values: Union[List, Tuple]) -> Any:
             return values
 
         elif has_tuple_array_return:
-            output_type = output_types[0]
+            output_type = output_types[0].copy(deep=True)
```
unparalleled-js, line 93: We can also disable mutation from ABIType in `ethpm-types` to help prevent this from happening again. then, we will need to use object initialization or parsing here

unparalleled-js: OK, switched to the object initialization strategy!

banteg (approved):

Issue #751, opened 2022-05-26
Multiple event with the same name in different contracts error with `as already been declared`


pandadefi:
### Environment information

* OS: macOS
* Python Version: x.x.x
* `ape` and plugin versions: 0.2.7

### What went wrong?

After an update from 0.2.2 to 0.2.7 I am getting the following error when trying to run `ape test` on https://github.com/yearn/veYfi/

```
E                   ape.exceptions.ContractError: Multiple events with the same ABI defined in 'Gauge'.
```

unparalleled-js:
I did some digging on this. The `Gauge` contract has two events in its ABI named `Initialize`. They can be described as:

```python
EventABI(type='event', name='Initialized', inputs=[EventABIType(name='stakingToken', type='address', components=None, internalType='address', indexed=True), EventABIType(name='rewardToken', type='address', components=None, internalType='address', indexed=True), EventABIType(name='owner', type='address', components=None, internalType='address', indexed=True), EventABIType(name='rewardManager', type='address', components=None, internalType='address', indexed=False), EventABIType(name='ve', type='address', components=None, internalType='address', indexed=False), EventABIType(name='veYfiRewardPool', type='address', components=None, internalType='address', indexed=False)], anonymous=False)

EventABI(type='event', name='Initialized', inputs=[EventABIType(name='version', type='uint8', components=None, internalType='uint8', indexed=False)], anonymous=False)
```

Only the first one is actually defined in the contract. I have not yet found where the other one is coming from.
I have figured out a solution (check signature instead of name pre-error-ing). However, I don't know how to replicate this issue in the tests yet.

unparalleled-js:
Ope found it: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/proxy/utils/Initializable.sol#L72

PR #771, opened 2022-06-02
docs: improve error message when multiple events with same name defined

unparalleled-js:
### What I did

fixes: #751 

### How I did it

Well, it does not actually resolve their error but gives more detail on how to fix it.
It seems problematic to have two events with the same name so I am guessing it is probably best to raise an error.

An alternative is to rename the events to `Initialized_0` and `Initialized_1` and so on (plus a WARNING), but that seems not ideal from the project's perspective.

Any other ideas? Is  `veYfi` able to rename `Gauge.Initialized` event?

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -503,7 +503,8 @@ class ContractInstance(BaseAddress):
         for abi in self.contract_type.events:
             if abi.name in events:
                 raise ContractError(
-                    f"Multiple events with the same ABI defined in '{self.contract_type.name}'."
+                    f"Multiple events named '{abi.name}' in '{self.contract_type.name}'.\n"
+                    f"Could one be coming from a base class?"
                 )
 
             events[abi.name] = abi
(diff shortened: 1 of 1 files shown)


src/ape/contracts/base.py
```diff
@@ -503,7 +503,8 @@ def _events_(self) -> Dict[str, ContractEvent]:
         for abi in self.contract_type.events:
             if abi.name in events:
                 raise ContractError(
-                    f"Multiple events with the same ABI defined in '{self.contract_type.name}'."
+                    f"Multiple events named '{abi.name}' in '{self.contract_type.name}'.\n"
+                    f"Could one be coming from a base class?"
```
fubuloubu, line 507:
```suggestion
                    f"Could one be coming from an imported file?"
```

fubuloubu (dismissed):

unparalleled-js:
I am really trying to think if there is a better solution to do.. Like are we able to detect if the event is coming from an outside source?

PR #776, opened 2022-06-02
refactor: add transaction query support

johnson2427:
### What I did

* Added `num_transactions` to `ape.api.providers.BlockAPI`
* Created `ape.api.query.BlockTransactionQuery`
* Added `transaction` method to `ape.api.providers.BlockAPI`
* Added `perform_block_transaction_query` to `ape.managers.query.DefaultQueryProvider`

fixes: #727 

### How I did it

Followed issue number 727

### How to verify it

run pytests

### Checklist


src/ape/api/providers.py
```diff
@@ -277,6 +285,18 @@ def get_transaction(self, txn_hash: str) -> ReceiptAPI:
             The receipt of the transaction with the given hash.
         """
 
+    @abstractmethod
+    def get_transactions_by_block(self, block_id: HexBytes) -> Iterator[TransactionAPI]:
```
unparalleled-js, line 289: `block_id` sounds like it can also be a block num. Perhaps we should just call this `block_hash` ?

unparalleled-js: Or could we make this type `ape.types.BlockID` instead? And that way we can support block numbers

johnson2427: I think I'd rather do block_id. And currently, we don't have the ability to use a hash for this query.

unparalleled-js: https://github.com/ApeWorX/ape/blob/main/src/ape/types/__init__.py#L21

johnson2427: I should have been more clear. Currently, our validation for `BlockTransactionQuery` fails because of the hash. We don't have a method that includes all types for that Model.

unparalleled-js: If it's called `block_id`, I am going to assume the type is `BlockID`.

Also, here the type is `HexBytes`, so I am assuming it is the hash? But you are saying the hash is not working?

I don't think the at the API level that we should be concerned with query-related data models. If we are able to use the `query_manager`, that is great, but otherwise we can just use the provider or something. Or figure out how to adjust the query mechanism to handle searching by block_num or block_hash.

unparalleled-js: Actually seeing as how it it always just hashes here when you are getting the full transaction, I see the argument for only supporting the hashes.

But the arg name should be `txn_hash` in that case instead of `block_id` ? To as not confuse with `BlockID` ?

johnson2427: This isn't looking for a txn_hash, this is looking for a group of transactions from a block with a block hash.

unparalleled-js: `bloch_hash` then..

I am saying that if it is called `block_id`, I am going to assume the type is `BlockID`. I am just hoping you rename the parameter name...............

johnson2427: @unparalleled-js I'm just informing you to make sure we're on the same page in the hope that something clicks for you and you have a simple solution. Otherwise, this requires me to build a new method, and a whole different level of design. I'm trying to solve this without adding a new method that does nearly the exact same thing as two other methods that already exist. I'm not sitting here trying to correct you, I don't care for showing you up

unparalleled-js: OK, yeah I guess I am just extremely lost. I don't get why we can't rename the parameter. But it is ok! I will trust you guys and your constraints. No worries :)

johnson2427: No, I changed the name of the parameter. I'm not arguing against that. I have renamed the parameter, just haven't pushed the new changes, because I have to map out how to get this to validate

fubuloubu (approved):

johnson2427:
I had to rebase before we merged

fubuloubu (approved):

PR #783, opened 2022-06-07
fix: bug where \r characters in contracts causes endless compiling

unparalleled-js:
### What I did and How I did it

Fixes https://github.com/ApeWorX/ape-cairo/issues/14

For some reason, in that user's project (the one linked in the ticket), the source files bytes are different than its text encoding as utf8. The source content parameter is text, and that is what the checksum has to be based on. So the only solution I could think of was to first read the text as utf8, then encode, and then compare checksums. Now things work.

### How to verify it

When you compile https://github.com/supsam89/game_guilds/tree/main/contracts multiple times,
it recognizes that the content has not changed.

### Checklist


unparalleled-js:
- [ ] try adding test

unparalleled-js:
Ok, I found the strange characters that are showing up:

```python
>>> bsplit[0]
'%lang starknet\r'
>>> tsplit[1]
''
>>> tsplit[0]
'%lang starknet'
>>> 
```

the bytes version contains return characters for some reason!

tests/integration/cli/test_compile.py
```diff
@@ -68,14 +76,60 @@ def test_compile(ape_cli, runner, project, clean_cache):
...
+    assert "Compiling 'Interface.json'" in result.output
+
+    # Verify that the next time, it does not need to recompile (no changes)
+    result = runner.invoke(ape_cli, ["compile"], catch_exceptions=False)
+    assert result.exit_code == 0, result.output
+    assert "Compiling 'Interface.json'" not in result.output
+
+
+@skip_projects_except(["multiple-interfaces"])
+def test_compile_when_source_contains_return_characters(ape_cli, runner, project, clean_cache):
```
unparalleled-js, line 103: haha.

This is the test that tests the bugfix

src/ape/api/projects.py
```diff
@@ -137,7 +137,7 @@ def _create_source_dict(
                     hash=compute_checksum(source_path.read_bytes()),
                 ),
                 urls=[],
-                content=source_path.read_text(),
+                content=source_path.read_text("utf8"),
```
NotPeopling2day, line 140: I'm guessing we will want to do a similar treatment to the solidity plugin? 

https://github.com/ApeWorX/ape-solidity/blob/3dee61309d10b796da6027e9abf4828763fef197/ape_solidity/compiler.py#L29

NotPeopling2day (approved):

src/ape/api/projects.py
fubuloubu: I believe this is at a layer above the compilers?

PR #785, opened 2022-06-07
fix: correct type of block ID in abstractmethod

unparalleled-js:
### What I did

The type of `block_id` does not seem to match the implementing behavior

### How I did it
changed it!

### How to verify it
Tell me I am wrong if I am

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -290,7 +290,8 @@ class ProviderAPI(BaseInterfaceModel):
             The receipt of the transaction with the given hash.
         """
 
-    @abstractmethod
+    # TODO: After 0.2.8 release, make this @abstractmethod
+    @raises_not_implemented
     def get_transactions_by_block(self, block_id: BlockID) -> Iterator[TransactionAPI]:
         """
         Get the information about a set of transactions from a block.
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -291,12 +291,12 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def get_transactions_by_block(self, block_id: HexBytes) -> Iterator[TransactionAPI]:
+    def get_transactions_by_block(self, block_id: BlockID) -> Iterator[TransactionAPI]:
         """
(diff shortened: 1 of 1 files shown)


src/ape/api/providers.py
```diff
@@ -291,12 +291,12 @@ def get_transaction(self, txn_hash: str) -> ReceiptAPI:
         """
 
     @abstractmethod
```
unparalleled-js, line 293: If we make this `@raises_not_implemented()`, it will prevent it from being a breaking change on the next release

src/ape/api/providers.py
```diff
@@ -291,12 +291,12 @@ def get_transaction(self, txn_hash: str) -> ReceiptAPI:
...
-    def get_transactions_by_block(self, block_id: HexBytes) -> Iterator[TransactionAPI]:
+    def get_transactions_by_block(self, block_id: BlockID) -> Iterator[TransactionAPI]:
         """
         Get the information about a set of transactions from a block.
 
         Args:
-            block_id (HexBytes): The hash of a block.
+            block_id (:class:`~ape.types.BlockID`): The ID of the block.
```
johnson2427, line 299: This gives us more control, absolutely should have been done this way

johnson2427 (dismissed):
I'd like to see @fubuloubu's answer, but I agree with the changes

src/ape/api/providers.py
unparalleled-js: Ya I saw it was done in the implementation method just not in the abstract. Implementations can always choose not to support a certain ID type too.

src/ape/api/providers.py
fubuloubu: Should eventually be `@abstractmethod` though, just add a TODO to change this after 0.2.8 release

johnson2427 (approved):
I think you were just handling a rebase, so I'm giving you the approval again, in the event that you'd like to merge

unparalleled-js:
@johnson2427 The PR gets auto-dismissed when new changes occur after Approval

johnson2427 (approved):
Ah, I see, that note from GitHub was just a little misleading.

PR #800, opened 2022-06-11
feat!: add access list transaction type, add transaction api txn_hash, fix block transaction query

banteg:
### What I did
- correctly instantiate transactions coming from `BlockTransactionQuery`, previously they were missing `data` and `signature`
- add `txn_hash` property to `TransactionAPI`
- add `AccessListTransaction` type and optional access list for `DynamicFeeTransaction`
- fix a bug with whitespace bytes being stripped when parsing txn data

### How to verify it
`chain.blocks[height].transactions` should yield transactions with correct calculated `txn_hash`

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -1,6 +1,7 @@
 import time
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
 
+from ethpm_types import HexBytes
 from ethpm_types.abi import EventABI
 from pydantic.fields import Field
 from tqdm import tqdm  # type: ignore
@@ -77,7 +78,7 @@ class TransactionAPI(BaseInterfaceModel):
         return f"{self.__class__.__name__}:\n  {params}"
 
     @abstractmethod
-    def txn_hash(self):
+    def txn_hash(self) -> HexBytes:
         """
         The hash of the transaction.
         """
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -1,8 +1,6 @@
 import time
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
 
-from eth_utils import keccak
-from ethpm_types import HexBytes
 from ethpm_types.abi import EventABI
 from pydantic.fields import Field
 from tqdm import tqdm  # type: ignore
@@ -78,9 +76,11 @@ class TransactionAPI(BaseInterfaceModel):
         params = "\n  ".join(f"{k}: {v}" for k, v in data.items())
         return f"{self.__class__.__name__}:\n  {params}"
 
-    @property
+    @abstractmethod
     def txn_hash(self):
-        return HexBytes(keccak(self.serialize_transaction()))
+        """
+        The hash of the transaction.
+        """
 
 
 class ConfirmationsProgressBar:
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -1,6 +1,8 @@
 import time
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
 
+from eth_utils import keccak
+from ethpm_types import HexBytes
 from ethpm_types.abi import EventABI
 from pydantic.fields import Field
 from tqdm import tqdm  # type: ignore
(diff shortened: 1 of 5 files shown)


banteg:
also caught an interesting off-by-one problem
```pytest
import rich
from hexbytes import HexBytes


def test_wtf(networks, chain):
    with networks.ethereum.mainnet.use_provider("geth"):
        for height in range(chain.blocks.height, chain.blocks.height - 100, -1):
            print(height)
            web3_transactions = chain.provider.web3.eth.get_block(height, True).transactions
            ape_transactions = chain.blocks[height].transactions
            for a, b in zip(web3_transactions, ape_transactions):
                rich.print(f"[red]{a.hash.hex()}")
                rich.print(f"web3={a}")
                rich.print(f"ape={b}")
                assert HexBytes(a.input) == HexBytes(b.data), "data mismatch"
                rich.print(f"[red]{b.txn_hash.hex()}")
                assert a.hash == b.txn_hash, "hash mismatch"
```

image

src/ape/utils/basemodel.py
```diff
@@ -127,7 +127,6 @@ class Config:
         keep_untouched = (cached_property, singledispatchmethod)
         arbitrary_types_allowed = True
         underscore_attrs_are_private = True
-        anystr_strip_whitespace = True
```
banteg, line 130: this has caused data to get malformed on load if it started or ended with bytes `[9, 10, 11, 12, 13, 32]`, as in `['\t', '\n', '\x0b', '\x0c', '\r', ' ']`

src/ape_ethereum/transactions.py
```diff
@@ -84,6 +90,25 @@ class DynamicFeeTransaction(BaseTransaction):
     max_priority_fee: Optional[int] = Field(None, alias="maxPriorityFeePerGas")
     max_fee: Optional[int] = Field(None, alias="maxFeePerGas")
     type: Union[int, str, bytes] = Field(TransactionType.DYNAMIC.value)
+    access_list: List[AccessList] = Field(default_factory=list, alias="accessList")
```
banteg, line 93: EIP-1559 transactions can have access list too, [see the spec](https://eips.ethereum.org/EIPS/eip-1559).

src/ape/api/transactions.py
```diff
@@ -76,6 +78,10 @@ def __str__(self) -> str:
         params = "\n  ".join(f"{k}: {v}" for k, v in data.items())
         return f"{self.__class__.__name__}:\n  {params}"
 
+    @property
+    def txn_hash(self):
+        return HexBytes(keccak(self.serialize_transaction()))
```
fubuloubu, line 83: hmm, could this be an API method? then have EcosystemAPI fill it out

src/ape/utils/basemodel.py
fubuloubu: No idea why this was there....

Looks like @unparalleled-js might have added in #684?

unparalleled-js: I refactored in that PR and split up Utils.py.

I don't think I added it initially.

src/ape/api/transactions.py
```diff
@@ -76,6 +77,12 @@ def __str__(self) -> str:
         params = "\n  ".join(f"{k}: {v}" for k, v in data.items())
         return f"{self.__class__.__name__}:\n  {params}"
 
+    @abstractmethod
+    def txn_hash(self) -> HexBytes:
```
unparalleled-js, line 81: one note is that in `ReceiptAPI`, the `txn_hash` property is of type `str` (for DB reasons?)
I think it should be fine as `HexBytes` here but just FYI

unparalleled-js:
NOTE: Breaking change since `txn_hash` is abstract
I think we should do `0.3.0` next anyway?

src/ape/api/transactions.py
fubuloubu: It should be changed to `HexBytes` in `ReceiptAPI` too, @johnson2427 and I were just talking about this inconsistency

fubuloubu (approved):

src/ape/api/transactions.py
unparalleled-js: HexBytes works with me. It is an int in Starknet but it can be translated.

PR #810, opened 2022-06-16
refactor: breaking changes PR

unparalleled-js:
### What I did

* Renamed `start` and `stop` to `start_block` and `stop_block` respectively in `poll_logs` to lessen confusion between that in the `range()` methods (which use the slice approach for `start` and `stop`).
* make the `-i` flag capitalized like `-I` in the `ape run` CLI command so that it matches its equivalent in the `ape test` command, for consistency's sake since the flag does the same thing.

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -441,7 +441,9 @@ class ContractEvent(ManagerAccessMixin):
         ) - required_confirmations
 
         for new_block in self.chain_manager.blocks.poll_blocks(
-            start=start_block, stop=stop_block, required_confirmations=required_confirmations
+            start_block=start_block,
+            stop_block=stop_block,
+            required_confirmations=required_confirmations,
         ):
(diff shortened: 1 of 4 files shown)


fubuloubu (dismissed):

johnson2427 (approved):
On my local machine, I have a test failing but it looks like it's an issue with an assert.

```    
def test_block_range_negative_start(chain_at_block_5):
        with pytest.raises(ValueError) as err:
            _ = [b for b in chain_at_block_5.blocks.range(-1, 3, step=2)]
    
>       assert str(err.value) == "start '-1' cannot be negative."
E       assert "stop '3' cannot be negative." == "start '-1' c... be negative."
E         - start '-1' cannot be negative.
E         ?   ^^^  ^^
E         + stop '3' cannot be negative.
E         ?   ^^  ^
```

PR #814, opened 2022-06-18
refactor: don't disconnect from ProviderAPI when using context manager

fubuloubu:
### What I did
When switching to another network using the network switcher, currently ape will disconnect and reconnect so that only one provider is connected at a time. This seems like an over-optimization, since it is unlikely (or in some cases impossible) that a dissimilar provider would interfere with another. Therefore, it seems prudent to remove this restriction, which helps improve some other things as well.

potentially related:
- #581
- https://github.com/ApeWorX/ape-ens/issues/13
- https://github.com/ApeWorX/ape-ens/pull/9

### How I did it
Removed the disconnect when connecting to another provider, and the symmetric connect when disconnecting from that provider. The context manager will still connect and disconnect to the provider it is targeting for the next level in the stack.

### How to verify it
Try a bunch of iterations of `with networks.<use_provider>(...)`

### Checklist
- [ ] ~~New test cases have been added~~ No test cases required
- [ ] ~~Documentation has been updated~~ No documentation updates required 
- [ ] Needs lots of testing with all provider plugins and across many different network switches


unparalleled-js:
This does not seem to fix #581 (just tried it out!)

Commands that use `NetworkBoundCommand` exit the connected network context at the end of the command, and the `-I` flag in scripts starts the console after the command is done executing _unless_ an exception occurs.

fubuloubu:
> This does not seem to fix #581 (just tried it out!)

Ah, was hoping it might, but still needed if it doesn't solve that particular issue

unparalleled-js (approved):

PR #815, opened 2022-06-18
chore: fix .md and small changes to docs 

Ninjagod1251:
### What I did

I added a better link to the .md file. It works now.
Added small MacOS notes to make the quickstart a bit more understanding
Same thing with contribute guide.

fixes: #

### Checklist


Ninjagod1251:
I would like to merge this PR on the last day before NYC so we can more add small changes if we need to.

CONTRIBUTING.md
```diff
@@ -16,6 +16,9 @@ python setup.py install
 
 # install the developer dependencies (-e is interactive mode)
 pip install -e .[dev]
+
+# MacOS install the developer dependencies (-e is interactive mode)
+pip install -e .'[dev]'
```
unparalleled-js, line 21: I believe the quotes work on both, so perhaps we should just include the example with quotes since it should work for everyone? that is what we do elsewhere. Rather than list this twice.

README.md
```diff
@@ -52,13 +52,24 @@ require one for your use case.
 * [virtualenv](https://pypi.org/project/virtualenv/)
 * [venv](https://docs.python.org/3/library/venv.html)
 
+NOTE: MacOS sometimes has issues with installing since it comes with python2.7 by default.
```
unparalleled-js, line 55:
```suggestion
**NOTE**: For MacOS users, we advise installing in a virtual environment to avoid interfering with OS-level site packages.
```

It does not matter the Python version, we should tell people not to install Ape outside a virtual env unless they are using some simpler linux box

README.md
```diff
@@ -52,13 +52,24 @@ require one for your use case.
...
+
+### via `homebrew`
```
unparalleled-js, line 57: What??? We support Homebrew?
Can you provide an example command for this?

```bash
brew install eth-ape
```

or what is it? I did not know this.

README.md
```diff
@@ -52,13 +52,24 @@ require one for your use case.
...
+
+[brew](https://brew.sh/)
+If you are installing ape on a MacOS. Most users find it easier to install it on brew. It is the same process as pip.
```
unparalleled-js, line 60: This first sentence is not a real sentence. It just ends! Maybe you meant to use a comma with the next sentence?

README.md
```diff
@@ -52,13 +52,24 @@ require one for your use case.
...
+* (MacOS Alternative Option) Install via homebrew [brew](https://formulae.brew.sh/formula/virtualenv)
+
 Create your virtual environment folder
 
 ```bash
 python3 -m venv /path/to/new/environment
 source <venv_folder>/bin/activate
 ```
 
+NOTE: the path to venv is not a real path.
```
unparalleled-js, line 71: I think that is obvious, probably don't need to mention

README.md
fubuloubu: I don't think that's what was meant, I think he meant you can install `virtualenv` from homebrew

unparalleled-js: Ah, it is a sub-header among other sub-headers related to installing ape. Maybe we should put it above near the other mention of `virtual env` and just say `**NOTE**: You can also use homebrew(link) to install virtual env.`

Ninjagod1251: Can we support homebrew in the future?

README.md
Ninjagod1251: WELL I THOUGHT IT WAS. BUT IVE BEEN SENT SCREENSHOTS WITH PEOPLE NOT GETTING IT
😆

CONTRIBUTING.md
```diff
@@ -15,6 +15,8 @@ source venv/bin/activate
 python setup.py install
 
 # install the developer dependencies (-e is interactive mode)
+
+```bash
```
fubuloubu, line 19: Maybe you'd want each step here to be it's own code block?

fubuloubu: Also, do we even use this file?

unparalleled-js (approved):

PR #823, opened 2022-06-22
feat: add set_balance test provider API

unparalleled-js:
### What I did

Adds `set_balance()` API to `TestProviderAPI` class that raises not implemented by default.
Hardhat and Foundry both use this!

### Checklist


src/ape/api/providers.py
```diff
@@ -557,9 +561,15 @@ def mine(self, num_blocks: int = 1):
...
-    def test_config(self) -> PluginConfig:
-        return self.config_manager.get_config("test")
+    @raises_not_implemented
+    def set_balance(self, address: AddressType, amount: int):
+        """
+        Change the balance of an account.
+
+        Args:
+            address (AddressType): An address on the network.
+            amount (int): The balance to set in the address.
```
fubuloubu, line 571: Can this API support conversions? Or do you think that's better off upstream (e.g. `ChainManager`)?

fubuloubu (approved):

src/ape/api/providers.py
unparalleled-js: Yeah, I think that is good point because that is how I made the Hardhat one work so you can do stuff like

```python
provider.set_balance(me, "1000 ETH")
```

fubuloubu: For sure, or it can also be:
```py
chain.set_balance(me, "1000 ETH")
```

unparalleled-js:
We can better out-of-the-box support if this feat request is honored:

https://github.com/ethereum/eth-tester/issues/236

Also, I will adjust + add more tests to `ape-hardhat` and `ape-foundry` where the API is actually implemented.

src/ape/api/providers.py
fubuloubu: > does it make more sense in chain then? chain is like the high-level provider sometimes

Yeah, if it's there then you can do the conversion step and leave `ProviderAPI` to only accept integer value

src/ape/managers/chain.py
```diff
@@ -759,3 +759,16 @@ def mine(
         elif deltatime:
             self.pending_timestamp += deltatime
         self.provider.mine(num_blocks)
+
+    def set_balance(self, account: Union[BaseAddress, AddressType], amount: Union[int, str]):
+        if isinstance(account, BaseAddress):
+            account = account.address  # type: ignore
```
fubuloubu, line 765: Doesn't `self.conversion_manager.convert(account, AddressType)` work here?

src/ape/managers/chain.py
```diff
@@ -759,3 +759,16 @@ def mine(
...
+
+        if isinstance(amount, str) and len(str(amount).split(" ")) > 1:
+            # Support values like "1000 ETH".
+            amount = self.conversion_manager.convert(amount, int)
+        elif isinstance(amount, str):
+            # Support hex strings
+            amount = int(amount, 16)
```
fubuloubu, line 772: Isn't there a hex converter to`int`?

fubuloubu (approved):

tests/functional/test_chain.py
```diff
@@ -185,3 +185,8 @@ def test_contract_caches_default_contract_type_when_used(solidity_contract_insta
     # Ensure we don't need the contract type when creating it the second time.
     contract = ape.Contract(address)
     assert isinstance(contract, ContractInstance)
+
+
+def test_set_balance(chain, test_accounts):
+    with pytest.raises(APINotImplementedError):
```
fubuloubu, line 191: Wait, does `eth-tester` not support this method?

unparalleled-js: It didnt seem like it but i created an issue for them

src/ape/api/providers.py
```diff
@@ -379,6 +379,16 @@ def mine(self, num_blocks: int = 1):
             NotImplementedError: Unless overridden.
         """
 
+    @raises_not_implemented
+    def set_balance(self, address: AddressType, amount: int):
```
NotPeopling2day, line 383: Thank you for discussing this choice and future alternatives to making these test methods, like setting the balance or timestamp, only available on local/test chains.

NotPeopling2day (approved):
LGTM

PR #824, opened 2022-06-22
feat: added pandas back to chain

johnson2427:
### What I did

Pandas was removed from `DefaultQueryProvider`, and in turn pulled the dataframe functionality out of Ape all together.

Pandas was brought back into the fold in the chain manager

fixes: #

### How I did it

Instead of return the map from the `DefaultQueryProvider`, in `chain.blocks.query`, we are now pulling the map apart and injecting it into a dataframe.

### How to verify it

Run the tests for `test_query.py` and `test_chain.py`

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -69,9 +69,9 @@ class BlockQuery(_BaseBlockQuery):
 
     @classmethod
     def all_fields(cls) -> List[str]:
-        from .providers import BlockAPI
+        from ape_ethereum.ecosystem import Block
 
-        return list(BlockAPI.__fields__)
+        return list(Block.__fields__)
 
 
 class BlockTransactionQuery(_BaseQuery):
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -1,12 +1,10 @@
 from typing import Any, Dict, Iterator, List, Optional, Union
 
 from ethpm_types.abi import EventABI, MethodABI
-from pydantic import BaseModel, NonNegativeInt, PositiveInt, root_validator, validator
+from pydantic import BaseModel, NonNegativeInt, PositiveInt, root_validator
 
 from ape.types import AddressType
 from ape.utils import BaseInterfaceModel, abstractmethod
-
-from .transactions import TransactionAPI
 
 QueryType = Union[
     "BlockQuery",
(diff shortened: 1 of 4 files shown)


src/ape/api/query.py
```diff
@@ -69,9 +69,9 @@ class BlockQuery(_BaseBlockQuery):
 
     @classmethod
     def all_fields(cls) -> List[str]:
-        from .providers import BlockAPI
+        from ape_ethereum.ecosystem import Block
 
-        return list(BlockAPI.__fields__)
+        return list(Block.__fields__)
```
fubuloubu, line 74: Keep this the way it was, we should open up a ticket for how to do it better. In general, should not be pulling in a 1st party plugin into ape core.

src/ape/managers/chain.py
```diff
@@ -77,7 +78,7 @@ def __len__(self) -> int:
 
         return self.height + 1
 
-    def __iter__(self) -> Iterator[BlockAPI]:
+    def __iter__(self) -> Iterator[pd.Series]:
```
fubuloubu, line 82: Keep this the way it was

src/ape/managers/chain.py
```diff
@@ -136,12 +137,14 @@ def query(
             step=step,
             engine_to_use=engine_to_use,
         )
-
-        return self.query_manager.query(query)
+        vals = self.query_manager.query(query)
+        vals = map(lambda val: val.dict(by_alias=False), vals)
+        df = pd.DataFrame(columns=query.columns, data=vals)
+        return df
```
fubuloubu: Eventually should turn the lambda into a utility function
```suggestion
        vals = map(lambda val: val.dict(by_alias=False), self.query_manager.query(query))
        return pd.DataFrame(columns=query.columns, data=vals)
```

src/ape/managers/chain.py
```diff
@@ -136,12 +137,14 @@ def query(
...
 
     def range(
         self, start_or_stop: int, stop: Optional[int] = None, step: int = 1
-    ) -> Iterator[BlockAPI]:
+    ) -> Iterator[pd.Series]:
```
fubuloubu, line 148: Revert the way it was

src/ape/managers/chain.py
```diff
@@ -166,7 +169,7 @@ def range(
              number of blocks to get. Defaults to the latest block.
 
         Returns:
-            Iterator[:class:`~ape.api.providers.BlockAPI`]
+            Iterator[:class:`pd.Series`]
```
fubuloubu: Revert the way it was

src/ape/managers/chain.py
```diff
@@ -136,12 +137,14 @@ def query(
             step=step,
             engine_to_use=engine_to_use,
         )
-
-        return self.query_manager.query(query)
+        vals = self.query_manager.query(query)
```
fubuloubu, line 141: Should probably flop this back to using `self.range(...)` and then converting to dataframe

src/ape/managers/chain.py
```diff
@@ -190,15 +193,15 @@ def range(
         # Note: the range `stop_block` is a non-inclusive stop, while the
         #       `.query` method uses an inclusive stop, so we must adjust downwards.
         results = self.query("*", start_block=start, stop_block=stop - 1, step=step)  # type: ignore
```
fubuloubu, line 196: Pull in all of the logic to call `self.query_manager.query(query)` and yield from that instead

Then update `self.query` to call this function and convert to a dataframe instead

src/ape/managers/chain.py
```diff
@@ -190,15 +193,15 @@ def range(
...
+        for idx, row in results.iterrows():
+            yield row  # type: ignore
 
     def poll_blocks(
         self,
         start_block: Optional[int] = None,
         stop_block: Optional[int] = None,
         required_confirmations: Optional[int] = None,
-    ) -> Iterator[BlockAPI]:
+    ) -> Iterator[pd.Series]:
```
fubuloubu: Keep this the way it was

src/ape/api/query.py
fubuloubu: Side note: starting to question whether `columns` is even necessary with the changes to how `QueryAPI` functions

There's no longer really any need to filter out columns anymore with how the iterator works

fubuloubu (dismissed):

fubuloubu (commented):
Looking for approval from @NotPeopling2day and @johnson2427 before merging, I believe this resolves all of the issues

src/ape/api/query.py
fubuloubu: Note: I went with the refactor as described above

NotPeopling2day (approved):
LGTM

johnson2427:
LGTM

NotPeopling2day (approved):

Issue #821, opened 2022-06-21
Make default transaction type configurable one ape-ethereum


pandadefi:
### Overview

The optimism plugin required to copy and modify some of the Ethereum methods in order to change the default transaction types since optimism doesn't support EIP1559.

See: https://github.com/ApeWorX/ape-optimism

### Specification

Adding a variable to `class Ethereum(EcosystemAPI): ` that corresponds to the default transaction type that can be modified by the class inheriting from Ethereum.

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #825, opened 2022-06-23
feat: make default transaction type a function that can be overridden

pandadefi:
### What I did

I have made the default transaction type a function in order for classes inheriting from `Ethereum` to be able to set the default transaction type. This is a change I would like to introduce to simplify integration with non EIP1559 compliant chains.
I have made the integration with optimism this changes will allow me to simplify the ape-optimism codebase and reduce the amount of code that got copied from the parent class reducing the chances of changes breaking the plugin.

https://github.com/ApeWorX/ape-optimism/blob/main/ape_optimism/ecosystem.py#L33

fixes: #821

### Checklist


banteg:
if you specify `gas_price`, ape will guess the transaction type i think

fubuloubu:
I would say two things:
1. The real solution for `ape-optimism` is to make it so only the transaction types that it supports are valid choices. This is clearly not made easy by the code in core.
2. A config variable might be a better solution to specify the default type, which you could override for whatver reason

banteg:
maybe create transaction could also make it possible to create access list type transactions when they specify both `gas_price` and `access_list`. it's not critical, since eip-1559 type is better in every regard, but maybe it's something we should support for completeness.

banteg (approved):

fubuloubu (approved):

PR #829, opened 2022-06-24
feat: allow relative indexing block numbers in query

fubuloubu:
### What I did
Allow relative indexing for start and stop block numbers in `chain.blocks.query`

```py
# Query the last 50 blocks
chain.blocks.query("*", stark_block=-50)

# Query the lat 20 blocks older than 30 behind head
chain.blocks.query("*", start_block=-50, stop_block=-30)
```

### How I did it
Make relative to absolute by adding `len(chain.blocks)`

### How to verify it
Try it out!

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -121,8 +121,14 @@ class BlockContainer(BaseManager):
             pd.DataFrame
         """
 
+        if start_block < 0:
+            start_block = len(self) + start_block
+
         if stop_block is None:
             stop_block = self.height
+
+        elif stop_block < 0:
+            stop_block = len(self) + stop_block
 
         elif stop_block > len(self):
             raise ChainError(
--- a/tests/functional/api/test_query.py
+++ b/tests/functional/api/test_query.py
@@ -31,6 +31,12 @@ def test_basic_query(eth_tester_provider):
     ]
 
 
+def test_relative_block_query(eth_tester_provider):
+    chain.mine(10)
+    df = chain.blocks.query("*", start_block=-8, stop_block=-2)
+    assert len(df) == 7
+
+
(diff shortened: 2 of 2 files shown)


src/ape/managers/chain.py
```diff
@@ -121,9 +121,15 @@ def query(
             pd.DataFrame
         """
 
+        if start_block < 0:
```
unparalleled-js, line 124: maybe add a test or two?

might be unrelated, but also wondering about negative step

fubuloubu: Handling negative steps would be a bit of work, I'll skip that for now.

Adding some tests

fubuloubu: see [1bc374d](https://github.com/ApeWorX/ape/pull/829/commits/1bc374dc1d4afac799123948ea1f8c26008eec7a)

tests/functional/api/test_query.py
```diff
@@ -31,6 +31,12 @@ def test_basic_query(eth_tester_provider):
     ]
 
 
+def test_relative_block_query(eth_tester_provider):
+    chain.mine(10)
+    df = chain.blocks.query("*", start_block=-8, stop_block=-2)
+    assert len(df) == 7
```
unparalleled-js, line 37: might be nice to check we have the right block numbers, at least at the start and stop blocks

unparalleled-js (approved):

tests/functional/api/test_query.py
fubuloubu: Thanks for this, think I identified an off-by-one bug

fubuloubu: nvm, false alarm. `-1` means "last block"

unparalleled-js (approved):

Issue #803, opened 2022-06-13
Odd conversion with stringified numbers


unparalleled-js:
### Environment information

* OS: macOS
* Python Version: 3.7
* `ape` and plugin versions:

### What went wrong?

Launch python

```bash
python
```

import `convert` and use it on a stringified number:

```python
>>> from ape import convert
>>> convert("235", int)
-54737251200
>>> convert("2353", int)
12100406400
>>> convert("23533", int)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/*/PycharmProjects/ape/src/ape/managers/converters.py", line 251, in convert
    raise ConversionError(f"No conversion registered to handle '{value}'.")
ape.exceptions.ConversionError: No conversion registered to handle '23533'.
```

The last line is what I would expect all of them to do but for some reason the first couple trigger registered converters and become some unusual strings.

**NOTE**: I uninstall all plugins to verify that this is coming from ape core.

### How can it be fixed?

* Figure out the registered converter triggering
* See why and if can make it not

PR #832, opened 2022-06-27
fix: prevent TimestampConverter from accepting single integer strings

helloibis:
### What I did

Update the conditional in `TimestampConverter.is_convertible()` to
return false if the string has no whitespace or has more than two
pieces (i.e. whitespace separated tokens)

Also added a test against bad timestamp strings

fixes: #803 

### How I did it

`dateutil.parse()` will actually accept a single integer string, and the output
depends on what number range the integer falls under. Numbers like `6` affect
the date or month, while `2011` affects the year. So if a number is in a valid range
`parse()` will pass meaning the `TimestampConverter` gets used.

### How to verify it

Simple integers without units or date formatting - as described 
in #803 - should now raise `ConversionError`:

```python
ape.convert("123", int)
ape.convert("123456", int)
ape.convert("0", int)
```

### Checklist

### DIFF
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -139,7 +139,7 @@ class TimestampConverter(ConverterAPI):
         if not isinstance(value, (str, datetime, timedelta)):
             return False
         if isinstance(value, str):
-            if " " not in value and len(value.split(" ")) > 2:
+            if " " not in value or len(value.split(" ")) > 2:
                 return False
             else:
                 try:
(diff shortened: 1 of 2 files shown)


helloibis:
@NotPeopling2day @unparalleled-js Hello, tagging for review :)

Haven't been added to the Github org yet but I saw this ticket when talking
with Blake this morning and figured I could take a stab at it to get started

unparalleled-js (approved):
Heck ya, thank you

PR #835, opened 2022-06-28
feat: add networks fixture automatically to pytest integration

unparalleled-js:
### What I did

When testing multi-chain projects, we have to change the networks mid-test.
It is a good use-case for having the network manager as an auto-fixture.

### How I did it

Same as other fixtures.
Also notice the docs for usage and use-case example

### How to verify it

use the `networks` fixture in an Ape project's test or pytest fixture.
it should exist automatically when ape is installed the pytest plugin is being used.

### Checklist


fubuloubu (approved):

unparalleled-js:
I don't understand why the tests are failing. It is like it cannot find any of the changes made to the pytest integration from this PR.

The tests all pass locally for me, and I replicated the environment.

fubuloubu (dismissed):

.github/workflows/test.yaml
```diff
@@ -60,7 +60,8 @@ jobs:
 
         steps:
         - uses: actions/checkout@v2
-          fetch-depth: 0
+          with:
+            fetch-depth: 0
```
unparalleled-js, line 64: I am not 100% sure why this fixes it.
What this does is cause the tags to be fetched, so this means the version is correct when it gets installed.

🤷🏻‍♀️ 

Probably should be doing this anyway everywhere!

fubuloubu (approved):

PR #837, opened 2022-06-29
fix: issue with caching local proxies

unparalleled-js:
### What I did

Noticed when you deploy local proxies, the caching behavior was not the same as when you fetched them from the explorer.
This is because we always were caching deployed contracts without first handling it as a proxy.

### How I did it

Detect if deploying a proxy. If so, cache it as a proxy instead of a contract type.

### How to verify it

When you deploy a proxy like

```
pragma solidity ^0.8.4;

contract Proxy {

    address public implementation;
    uint256 public proxyType = 2;

    constructor(address impl) {
        implementation = impl;
    }

    function setImplementation(address implementation_) public {
        implementation = implementation_;
    }
}
```

it will cache the proxy info.

### Checklist


src/ape/managers/chain.py
```diff
@@ -516,12 +516,56 @@ def cache_deployment(self, contract_instance: ContractInstance):
...
+
+        Returns:
+            Optional[:class:`~ape.api.networks.ProxyInfoAPI`]
+        """
+
+        proxy_info = self._local_proxies.get(address) or self._get_proxy_info_from_disk(address)
+        if not proxy_info:
+            proxy_info = self.provider.network.ecosystem.get_proxy_info(address)
+            if proxy_info and self._is_live_network:
+                self._cache_proxy_info_to_disk(address, proxy_info)
```
fubuloubu, line 565: Not sure I like this, basically if it can't find proxy info it will fetch and cache it to disk? I think only `cache_proxy_info` should be allowed to do this

unparalleled-js: was just making it work the same way we made `get()` work for the `contract_types` in the same class.. but yeah I agree, I don't like the side-effect of this

helloibis (approved):

helloibis:
Looks like there's type check and test fail issues but otherwise LGTM

src/ape/managers/chain.py
```diff
@@ -516,12 +516,36 @@ def cache_deployment(self, contract_instance: ContractInstance):
...
+        deployed proxies. When you deploy a proxy locally, it will also call this method.
+
+        Args:
+            address (AddressType): The address of the proxy contract.
+            proxy_info (:class:`~ape.api.networks.ProxyInfo`): The proxy info class
+              to cache.
+        """
+
+        if self.get_proxy_info(address) and self._is_live_network:
+            return
```
johnson2427, line 542: why not cache if we have the proxy info and this is a live network?

src/ape_ethereum/ecosystem.py
```diff
@@ -161,7 +162,9 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfo]:
                 target = self.conversion_manager.convert(match.group(1), AddressType)
                 return ProxyInfo(type=type, target=target)
 
-        str_to_slot = lambda text: int(keccak(text=text).hex(), 16)  # noqa: E731
```
johnson2427, line 164: Is this because it returns a pointer here that you turn this into a method? Just curious if you were running into issues

tests/functional/conftest.py
```diff
@@ -23,10 +23,10 @@
 from ape.types import AddressType, ContractLog
 
 
-def _get_raw_contract(name: str) -> Dict:
+def _get_raw_contract(name: str) -> str:
     here = Path(__file__).parent
     contracts_dir = here / "data" / "contracts" / "ethereum" / "local"
-    return json.loads((contracts_dir / f"{name}.json").read_text())
+    return (contracts_dir / f"{name}.json").read_text()
```
johnson2427, line 29: I don't hate this, but I'm just curious why. Is it easier to deal with strings and the parse_obj here?

johnson2427 (commented):
Just a couple questions, I have no requested changes though. Just some highlights from what I saw

src/ape_ethereum/ecosystem.py
unparalleled-js: Lambdas should only be used as anonymous functions. When you assign it to a variable, it isn' really anonymous so you might as well use a `def` statement. There is some more info here: https://www.flake8rules.com/rules/E731.html

The `#noqa` statement allowed it to go uncaught.

src/ape/managers/chain.py
unparalleled-js: Because we should cache the target contract type and not the proxys, but I just realized I forgot that part!

- [ ] cache target ct

tests/functional/conftest.py
unparalleled-js: It just saves a step I guess.

Before we were:

1. Reading the file,
2. Parsing it into a dict 
3. and then parsing that into a `ContractType`.

Now, we:

1. Read the file 
2. and parse it into a `ContractType`

tests/functional/test_accounts.py
```diff
@@ -147,6 +148,23 @@ def test_deploy_and_not_publish(mocker, owner, contract_container, dummy_live_ne
...
+    proxy = owner.deploy(proxy_contract_container, target)
+    assert proxy.address in chain.contracts._local_contract_types
+    assert proxy.address in chain.contracts._local_proxies
+
+    actual = chain.contracts._local_proxies[proxy.address]
+    assert actual.target == target
+    assert actual.type == ProxyType.Delegate
+
+    # Show we get the implementation contract type using the proxy address
+    implementation = chain.contracts.instance_at(proxy.address)
```
unparalleled-js, line 164: @johnson2427 I added this part of the test to further demonstrate the bugfix here.
So now, we pass the locally deployed proxy contract address to `chain.instance_at()` (same as `ape.Contract()`) and we are given back the implementation contract type.

tests/functional/conftest.py
johnson2427: Works for me!

tests/functional/test_accounts.py
johnson2427: This is great work here. well done

johnson2427 (approved):
LGTM

PR #841, opened 2022-06-29
feat: show SubprocessProvider process logs in debug output

unparalleled-js:
### What I did

Leverage process logs in `SubprocessProvider`.

So when we do `-v debug` in a script or console, we see the output from the subprocess in our normal ape logs

### How I did it

daemon queue thingy ma-bobs. Idk I just copy code on the internet.

### How to verify it

use hardhat, starknet, or foundry with `-v debug` and watch for expected HTTP request logs

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -41,11 +41,13 @@ from ape.types import AddressType, BlockID, ContractLog, SnapshotID
 from ape.utils import (
     EMPTY_BYTES32,
     BaseInterfaceModel,
+    JoinableQueue,
     LogInputABICollection,
     abstractmethod,
     cached_property,
     gas_estimation_error_message,
     raises_not_implemented,
+    spawn,
 )
 
 
@@ -860,6 +862,9 @@ class SubprocessProvider(ProviderAPI):
     process: Optional[Popen] = None
     is_stopping: bool = False
 
+    _stdout_queue: Optional[JoinableQueue] = None
+    _stderr_queue: Optional[JoinableQueue] = None
+
     @property
     @abstractmethod
     def process_name(self) -> str:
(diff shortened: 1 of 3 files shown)


src/ape/api/providers.py
```diff
@@ -931,6 +942,28 @@ def start(self, timeout: int = 20):
                     time.sleep(0.1)
                     _timeout.check()
 
+    def produce_stdout_queue(self):
```
unparalleled-js, line 945: @johnson2427 let me know your thoughts on this madness! It seems to work.

unparalleled-js:
- [ ] Perhaps always dump the output into files even not using `-v debug`

src/ape/api/providers.py
```diff
@@ -931,6 +942,28 @@ def start(self, timeout: int = 20):
...
+            time.sleep(0)
+
+    def produce_stderr_queue(self):
+        for line in iter(self.process.stderr.readline, b""):
+            self._stderr_queue.put(line)
+            time.sleep(0)
+
+    def consume_stdout_queue(self):
+        for line in self._stdout_queue:
+            logger.debug(line.strip())
```
unparalleled-js, line 957: - [ ] decode output

src/ape/utils/process.py
```diff
@@ -0,0 +1,49 @@
...
+            yield item
+
+    def join(self, timeout=None):
+        with SubprocessTimeoutError(timeout) as _timeout:
+            while not self.empty():
+                time.sleep(0)
+                _timeout.check()
+
+
+def spawn(target, *args, **kwargs):
```
johnson2427, line 37: As long as we're aware that this is still running on the same processor, just being dropped off the main thread, I'm okay with this. I tend to avoid multi-threading in Python, and use multi-process. And considering this is more of a CPU bound issue, maybe multiprocessing is the move here? I'll thread if it's I/O bound

src/ape/utils/process.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class JoinableQueue(queue.Queue):
+    """
+    A queue that can be joined, useful for multi-processing.
+    Borrowed from the ``py-geth`` library.
+    """
+
+    def __iter__(self):
+        while True:
```
johnson2427, line 15: Need to think of edge cases that could cause infinite loop here, not saying there is, but just in case. This is just a preliminary review so far, taking note

johnson2427 (commented):
Still need to reproduce on my end to get these errors to fire off

src/ape/api/providers.py
johnson2427: how much of a headache do you think it would be to make a test for this? Normally I wouldn't think it would be, but the required ties to hardhat/starknet/foundry make me wonder if it would be a giant headache

johnson2427: If we normally do the hardhat test in these scenarios, let's keep it that way until we come up with a better way to do it

unparalleled-js:
this bad now

PR #847, opened 2022-07-01
fix: issue where subprocess provider would crash in the background

unparalleled-js:
### What I did

Fixes issue preventing foundry from working, preventing from Hardhat from working well, and currently preventing Starknet tests from passing.

Instead of sending output back to parent process, send it to a file and then everything is fine.

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -1,13 +1,15 @@
 import atexit
 import ctypes
+import logging
 import platform
 import shutil
 import sys
 import time
 from abc import ABC
+from logging import FileHandler, Formatter, Logger, getLogger
 from pathlib import Path
 from signal import SIGINT, SIGTERM, signal
-from subprocess import Popen
+from subprocess import PIPE, Popen
 from typing import IO, Any, Dict, Iterator, List, Optional, Union
 
 from eth_abi.abi import encode_single
@@ -41,11 +43,13 @@ from ape.types import AddressType, BlockID, ContractLog, SnapshotID
 from ape.utils import (
     EMPTY_BYTES32,
     BaseInterfaceModel,
+    JoinableQueue,
     LogInputABICollection,
     abstractmethod,
     cached_property,
     gas_estimation_error_message,
     raises_not_implemented,
+    spawn,
 )
 
 
@@ -862,6 +866,9 @@ class SubprocessProvider(ProviderAPI):
     process_stdout: Optional[IO] = None
     process_stderr: Optional[IO] = None
 
+    _stdout_queue: Optional[JoinableQueue] = None
+    _stderr_queue: Optional[JoinableQueue] = None
+
     @property
     @abstractmethod
     def process_name(self) -> str:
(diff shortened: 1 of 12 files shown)


src/ape/api/providers.py
```diff
@@ -921,7 +935,22 @@ def start(self, timeout: int = 20):
         else:
             logger.info(f"Starting '{self.process_name}' process.")
             pre_exec_fn = _linux_set_death_signal if platform.uname().system == "Linux" else None
-            self.process = _popen(*self.build_command(), preexec_fn=pre_exec_fn)
+
+            self.base_logs_path.mkdir(parents=True, exist_ok=True)
+            if self.stdout_logs_path.is_file():
+                self.stdout_logs_path.unlink()
+            if self.stderr_logs_path.is_file():
+                self.stderr_logs_path.unlink()
```
ca11ab1e, line 943:
```suggestion
            self.stdout_logs_path.unlink(missing_ok=True)
            self.stderr_logs_path.unlink(missing_ok=True)
```

src/ape/api/providers.py
```diff
@@ -921,7 +935,22 @@ def start(self, timeout: int = 20):
...
+            self.stdout_logs_path.touch()
+            self.stderr_logs_path.touch()
+            self.process_stdout = open(str(self.stdout_logs_path))
+            self.process_stderr = open(str(self.stderr_logs_path))
```
ca11ab1e, line 947:
```suggestion
            self.process_stdout = self.stdout_logs_path.open()
            self.process_stderr = self.stderr_logs_path.open()
```

src/ape/api/providers.py
unparalleled-js: > 

we can't actually do this yet because of python 3.7

ca11ab1e: Ah yes. That's a shame :)

johnson2427 (approved):
I'm approving the changes as they are. It does solve a collision issue with PIPE, but I'm not sure this solves our issue with how slow .cairo files are in ape-starknet. If we approve this, we can focus on that same slow test. I do think this will prevent that infinite loop we were seeing in ape-starknet

src/ape/api/providers.py
```diff
@@ -860,6 +864,9 @@ class SubprocessProvider(ProviderAPI):
     process: Optional[Popen] = None
     is_stopping: bool = False
 
+    stdout_queue: Optional[JoinableQueue] = None
```
unparalleled-js, line 867: At first tried naming these `_stdout_queue` etc.. but then Hardhat complains about the base classes conflictling.
Lol!

unparalleled-js:
- [ ] hmm, what about multiple provider runnings of the same type, won't they overwrite each other's logs?

src/ape/api/providers.py
```diff
@@ -931,6 +979,31 @@ def start(self, timeout: int = 20):
...
+            self._stderr_queue.put(line)
+            time.sleep(0)
+
+    def consume_stdout_queue(self):
+        for line in self._stdout_queue:
+            output = line.decode("utf8").strip()
+            logger.debug(output)
+            self._stdout_logger.info(output)
+            self._stdout_queue.task_done()
+            time.sleep(0)
```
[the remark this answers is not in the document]
johnson2427: time.sleep blocks the main thread. Could be the source of our slowing down of tests

unparalleled-js:
> * [x]  hmm, what about multiple provider runnings of the same type, won't they overwrite each other's logs?
> 
> Edit: decided it makes more sense to handle this as part of the `x-dist` ticket

johnson2427 (approved):
Starknet Tests passing

Issue #826, opened 2022-06-23
pickling problem with dask


banteg:
### What went wrong?

I'm trying to do some parallel processing with `dask`, which seems a good fit for ape since it mostly consists of drop-in replacements for pandas and other things we already support.

```python
from dask.distributed import Client
client = Client(processes=False)
future = client.submit(chain.provider.get_block, 'latest')
PicklingError: Can't pickle <cyfunction str_validator at 0x107e56ba0>: it's not the same object as pydantic.validators.str_validator
```
### How can it be fixed?

Idk, but a similar problem is [described here](https://stackoverflow.com/questions/62702040/dask-picklingerror-function-not-found) and seems related to `cloudpickle` and `pydantic`.

banteg:
Found another problem with `ContractLog` and `pickle`:
```python
log = ContractLog(name='TestLog', event_arguments={'a': 1}, transaction_hash=1, block_hash=2, block_number=3, index=4)
pickle.loads(pickle.dumps(log))
RecursionError: maximum recursion depth exceeded
```

PR #852, opened 2022-07-02
refactor!: contract event querying revamp

unparalleled-js:
### What I did

fixes: #826
fixes: #868 

Fixes:

* Issue where we didn't know what the contract address was for a given log
* Issue where couldn't re-create the class from the raw data.

### How I did it

* Add `contract_address` property to class
* Use `BaseModel` instead of dataclass

### Checklist


ca11ab1e (dismissed):

unparalleled-js:
I guess this could be considered a breaking change ... as it will require ecosystem plugins (e.g. starknet) to add another property to the initialization of its logs.

hmmm... may have to wait until 0.4

unparalleled-js:
TODOs:

- [ ] Use BaseModel instead of dataclass for `ContractLog`
- [ ] Rename `index` to `log_index`
- [ ] Re-think `__getattr__` (perhaps just avoid since naming conflicts
- [ ] Allow multiple ABIs passed to `ReceiptAPI.decode_logs()`
- [ ] _get_logs_in_block_range made more efficient to fetch topics in mass instead of 1 by 1
- [ ] Support OR topics? (mutliple OR topics)?
- [ ] The autodecoding work (is on a separate PR)

src/ape/types/__init__.py
```diff
@@ -75,11 +80,20 @@ def __getattr__(self, item: str) -> Any:
             item (str): The name of the property.
         """
 
+        try:
```
unparalleled-js, line 83: TBD - do we just rid of `__getattr__` due to potential naming conflicts + the myriad of issues we see when using `__getattr__` and serialization?

- [ ] Decide and / or take action

src/ape/types/__init__.py
```diff
@@ -75,11 +80,20 @@ def __getattr__(self, item: str) -> Any:
...
+            pass
+
         if item not in self.event_arguments:
             raise AttributeError(f"{self.__class__.__name__} has no attribute '{item}'.")
 
         return self.event_arguments[item]
 
+    def get(self, item: str, default: Optional[Any]) -> Any:
+        return self.event_arguments.get(item, default)
+
```
unparalleled-js, line 96: - [ ] Add __contains__ to check if input type name part of Event?

src/ape/types/__init__.py
banteg: i think replacing it with `__getitem__` would be fine

src/ape/types/__init__.py
```diff
@@ -75,11 +80,20 @@ def __getattr__(self, item: str) -> Any:
...
+            return normal_attribute
+        except AttributeError:
+            pass
+
         if item not in self.event_arguments:
             raise AttributeError(f"{self.__class__.__name__} has no attribute '{item}'.")
 
         return self.event_arguments[item]
 
+    def get(self, item: str, default: Optional[Any]) -> Any:
```
banteg, line 94: why not `__getitem__`?

src/ape/types/__init__.py
fubuloubu: In general, our guideline should be to never use `__getattr__` internally when other methods would do.

It's a great UX feature to be able to use it, but it should be possible to reference via other means as well in order to prevent infinite loops and other hard to debug situations

src/ape/types/__init__.py
unparalleled-js: `__getitem__` is also defined above! (line 72)

I like having both. Or `__contains__` or just some way of handling checking for keys

src/ape/types/__init__.py
unparalleled-js: OK - going to leave it and just not call it anywhere! 🤝

src/ape/api/providers.py
```diff
@@ -316,29 +323,30 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
-            abi (``EventABI``): The event of interest's ABI.
-            start_block (Optional[int]): Get events that occurred
-              in blocks after the block with this ID.
-            stop_block (Optional[int]): Get events that occurred
-              in blocks before the block with this ID.
-            block_page_size (Optional[int]): Use this parameter to adjust
-              request block range sizes.
-            event_parameters (Optional[Dict]): Filter by event parameter values.
+            addresses (Union[AddressType, List[AddressType]): Contract addresses.
+            abis (Union[EventABI, List[EventABI]]): Event ABIs.
```
unparalleled-js, line 338: - [ ] Refactor so that its a `topic_query: Dict[EventABI, Dict]` ! To better support OR-ing

src/ape/api/transactions.py
```diff
@@ -231,10 +233,14 @@ def decode_logs(self, abi: Union[EventABI, "ContractEvent"]) -> Iterator[Contrac
...
-            abi = abi.abi
+        if not isinstance(abi, (list, tuple)):
+            abi = [abi]
 
-        yield from self.provider.network.ecosystem.decode_logs(abi, self.logs)
+        for event_abi in abi:
+            if not isinstance(event_abi, EventABI):
+                event_abi = event_abi.abi
+
+            yield from self.provider.network.ecosystem.decode_logs(event_abi, self.logs)
```
banteg, line 243: i think the ordering is not preserved here

src/ape/api/providers.py
```diff
@@ -707,47 +706,42 @@ def get_transactions_by_block(self, block_id: BlockID) -> Iterator:
...
     def get_contract_logs(
         self,
-        address: Union[AddressType, List[AddressType]],
-        abi: Union[List[EventABI], EventABI],
-        start_block: Optional[int] = None,
-        stop_block: Optional[int] = None,
+        log_filter: LogFilter,
         block_page_size: Optional[int] = None,
-        event_parameters: Optional[Dict] = None,
     ) -> Iterator[ContractLog]:
```
banteg, line 711: could we also add reasonable parallelism to `get_contract_logs`? it's what people do anyway, might as well be the default. something like `concurrent.futures.ThreadPoolExecutor` with 4 workers would do.

src/ape/contracts/base.py
```diff
@@ -369,6 +374,7 @@ def range(
 
         start_block = None
         stop_block = None
+        addresses = [self.contract.address, *(extra_addresses or [])]
```
banteg, line 377: a bit cleaner way to write this
```suggestion
        addresses = [self.contract.address] + (extra_addresses or [])
```

src/ape/contracts/base.py
```diff
@@ -378,19 +384,19 @@ def range(
             stop_block = stop - 1
 
         stop_block = min(stop_block, self.chain_manager.blocks.height)
-        addresses = (
-            [self.contract.address, *extra_addresses]
-            if extra_addresses
-            else [self.contract.address]
+
+        search_topic_filters = (
+            [TopicFilter.parse_obj({"event": self.abi, "search_values": search_topics})]
```
banteg, line 389:
```suggestion
            [TopicFilter(event=self.abi, search_values=search_topics)]
```

src/ape/types/__init__.py
```diff
@@ -40,15 +45,136 @@
...
 
-@dataclass
-class ContractLog:
+class TopicFilter(BaseModel):
+    event: EventABI
+    search_values: Dict[str, Optional[Union[Any, List[Any]]]] = {}
+
+    @property
+    def event_signature_hash(self) -> str:
+        return add_0x_prefix(HexStr(keccak(text=self.event.selector).hex()))
```
banteg, line 54:
```suggestion
        return encode_hex(keccak(text=self.event.selector))
```

src/ape/types/__init__.py
```diff
@@ -40,15 +45,136 @@
...
+            abi_type = input_types.get(key)
+
+            if not abi_type or value is None:
+                return None
+
+            elif isinstance(value, (list, tuple)):
+                return [encode_topic_value(key, v) for v in value]
+
+            elif is_dynamic_sized_type(abi_type):
+                return to_hex(keccak(encode_single_packed(str(abi_type), value)))
```
banteg, line 78: `to_hex` main purpose seems encoding numeric values for jsonrpc. when dealing with bytes, it just delegates to `encode_hex`, so we could just it instead
```suggestion
                return encode_hex(keccak(encode_single_packed(str(abi_type), value)))
```

src/ape/types/__init__.py
```diff
@@ -40,15 +45,136 @@
...
+
+            else:
+                return to_hex(encode_single(abi_type, value))  # type: ignore
```
banteg, line 81:
```suggestion
                return encode_hex(encode_single(abi_type, value))  # type: ignore
```

banteg (approved):

banteg:
some design changes:
- since we discovered that you [can't OR](https://docs.alchemy.com/alchemy/apis/ethereum/eth-newfilter) multiple filters, i removed `TopicFilter` and moved the encoding portion to `LogFilter`
- it can be instantiated to store `events` which are then looked up and used for decoding
- currently i only have design with one event per filter, but you can manually instantiate it with multiple topic0s
- added a `from_event` class method which could help to make an instance and encode `search_topics` on creation
- also added better address type conversions

banteg:
simplified fetching contract logs and added concurrency while maintaining order

some quick benchmark results based on fetching logs of one contract on mainnet:
- 30795 logs in 67.6s (448.87 logs/s) baseline
- 30795 logs in 49.11s (627.12 logs/s) optimize
- 30795 logs in 25.49s (1,208.01 logs/s) threading

banteg:
looked more into optimizing log fetching. i ended up working with raw data bypassing web3 transformers, as they usually take more than the request itself. i also made a benchmark to select the optimal block page size and concurrency (50_000 and 8 respectively), which we could use as a default for geth.

get_logs_bench

banteg:
another thing i noticed when profiling/optimizing is `eth-hash` (and `eth-utils` by extension) uses the slower of two backends for keccak even if `pysha3` is installed. this can be solved by setting `export ETH_HASH_BACKEND="pysha3"`, which i think could be set by ape automatically. this is no joke, pysha3 is over 10x faster.

fubuloubu (dismissed):
looking good!

src/ape/contracts/base.py
```diff
@@ -292,19 +297,15 @@ def __getitem_int(self, index: int) -> ContractLog:
...
+        logs = self.provider.get_contract_logs(self.log_filter)
         try:
-            return logs_slice[index]
-        except IndexError as err:
+            if index == 0:
+                return next(logs)
+            elif index > 0:
+                return next(islice(logs, index, index + 1))
+            else:
+                return list(logs)[index]
```
banteg, line 307: @BobTheBuidler: you can optimize this by fetching the logs in reverse

imo this optimization has questionable value since "getting the last log" is the only place where it would be used

banteg:
tests passing, the fails are unrelated

banteg:
benchmarked all combinations of threading and page size for erigon and geth. ngl geth looks horrible, ergion performs 21.4x better.

i suggest we default to these parameters
- 8 threads, 40,000 blocks for erigon
- 16 threads, 5,000 blocks for geth

bench

banteg:
good to merge from my side

fubuloubu (approved):

Issue #853, opened 2022-07-03
Leverage the mypy configuration file


ca11ab1e:
### Overview

There are lots of `type: ignore` comments in imports.

Would you be interested in a PR to create the mypy configuration file to skip that [specific check](https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_missing_imports)?

### Specification

Here is the minimal `mypy.ini` file (I will open a PR to update pyproject.toml instead of using mypy.ini):

```ini
[mypy]
ignore_missing_imports = true
```

And a complete example (my preference):

```ini
[mypy]
# Ensure we know what we do
warn_redundant_casts = true
warn_unused_ignores = true
warn_unused_configs = true
# warn_unreachable = true  (to check)

# Imports management
ignore_missing_imports = true
follow_imports = normal
```

### Dependencies

None.

fubuloubu:
Ideally we'd like to find a way to get rid of the ignored if possible, the type hints have found numerous latent bugs for us. There are some scenarios that are difficult to work through though so we take it on a case by case basis

PR #855, opened 2022-07-03
feat: leverage mypy configuration

ca11ab1e:
### What I did

fixes: #853.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,11 @@ requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0"]
 
 [tool.mypy]
 exclude = "build/"
+follow_imports = "normal"
+ignore_missing_imports = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
 
 [tool.setuptools_scm]
 write_to = "src/ape/version.py"
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 from pathlib import Path
 from typing import Dict
 
-from setuptools import find_packages, setup  # type: ignore
+from setuptools import find_packages, setup
 
 here = Path(__file__).parent.absolute()
 packages_data: Dict = {}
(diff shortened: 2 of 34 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 from pathlib import Path
 from typing import Dict
 
-from setuptools import find_packages, setup  # type: ignore
```
fubuloubu, line 5: Are we not using `find_packages`?

src/ape/api/projects.py
```diff
@@ -131,8 +131,8 @@ def _create_source_dict(
         source_dict: Dict[str, Source] = {}
         for source_path in contract_filepaths:
             key = str(get_relative_path(source_path, base_path))
-            source_dict[key] = Source(  # type: ignore
```
fubuloubu, line 134: Some of these ignores were here from before we started using pydantic, so they might not be necessary anymore

pyproject.toml
```diff
@@ -3,6 +3,11 @@ requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0"]
 
 [tool.mypy]
 exclude = "build/"
+follow_imports = "normal"
+ignore_missing_imports = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
```
fubuloubu, line 10: See comment in the issue, but I don't want to activate any of these flags

ca11ab1e: Are we aligned on the fact that those flags are actually improving type checks?

For instance, if mypy tells us that a `type: ignore` comment is unused (thanks to the `warn_unused_ignores` flag), then we should remove it because mypy doesn't take t into account at all. So it's just useless code.

I will close both PR, and issue, if you confirm you do not want such flags :)

fubuloubu: I'll admit not reading that closely, in in favor of warning more strictly, but I don't want it to ignore unused imports. The others looks fine, but it might help to have a comment about what they do above each

unparalleled-js: I have had issues with `warn_unused_ignores` in the past when I tried enabling it.
The issue with `mypy` is that its results are not always reproducible. Meaning, in CI/CD, it may fail, but locally, it may be considered an "unused type ignore", which is frustrating. Because of this, I am in favor of occasionally running adhoc with the flag for `--warn-unused-ignores` and opening a PR that way.

unparalleled-js:
Closing as this PR is outdated and there has not been any progress in a while.
Re-open if there is more to come or we can land on a smaller subset of agreed changes!

PR #859, opened 2022-07-07
fix: issue where account would remain unlocked when disabling auto-sign

unparalleled-js:
### What I did

* Only unlock when enabling
* Only warn when enabling
* Only re-lock if disabling

**NOTE**: I think we consider keeping `unlock()` and `set_autosign()` separated. It would be a breaking change or I would have done it now. The reason is because it is weird if you unlock an account, turn autosign on, sign something, disable autosign, and now your account is suddenly re-locked. It is a strange side-effect in that scenario, but generally rare for the use-case if just enabling autosign for automated testnet deploy scripts etc.

### How I did it

if statements everywhere!

### How to verify it

see tests

### Checklist


fubuloubu (approved):
Seems acceptable

PR #860, opened 2022-07-07
fix: regressions with poll_logs

unparalleled-js:
### What I did

* fixes issue where block confirmations were off by 1
* fixes issue where default `stop_block` value causes issues (should default to never ending).
* automated tests for the poll_logs feature for the first time!
* fixes issue where mistakenly claims block re-orged when it didn't necessarily
* refactor: separate event related tests into their own module

### How I did it

* `>=`
* remove `stop_block` setting in `poll_logs` method
* use a fun threading thing (in the tests - not prod recommended)
* set boolean back to false _after_ reorg stage

### How to verify it
@banteg 

### Checklist


tests/functional/test_contract_event.py
```diff
@@ -0,0 +1,231 @@
...
+):
+    bad_stop_block = chain_at_block_5.blocks.height
+
+    with pytest.raises(ValueError) as err:
+        _ = [x for x in vyper_contract_instance.NumberChange.poll_logs(stop_block=bad_stop_block)]
+
+    assert str(err.value) == "'stop' argument must be in the future."
+
+
+def test_poll_logs(vyper_contract_instance, eth_tester_provider, owner, poll_daemon):
```
unparalleled-js, line 215: all other tests in this file were simply moved over **Except** for this one... this is the new test!

src/ape/managers/chain.py
```diff
@@ -279,6 +279,7 @@ def poll_blocks(
                 has_yielded = True
                 latest_confirmed_block_number = confirmable_block_number
 
+            has_yielded = False
```
fubuloubu, line 282: What's this for?

tests/functional/conftest.py
```diff
@@ -210,3 +212,42 @@ def dependency_config(temp_config):
...
+            # WARN: Will wait indefinitely for more events
+            self._handler(next(self._poller))
+            time.sleep(1)
+
+    def stop(self):
+        self.join()
+
+
+@pytest.fixture
+def poll_daemon():
```
fubuloubu, line 252:
```suggestion
def PollDaemon():
```

Even though it's a bit against python styling, this more accurately matches what the fixture represents

tests/functional/conftest.py
```diff
@@ -210,3 +212,42 @@ def dependency_config(temp_config):
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.stop()
+
+    def run(self):
+        while True:
+            if self._do_stop():
+                return
+
+            # WARN: Will wait indefinitely for more events
```
fubuloubu, line 243: Is there a way to set a timeout in `__init__` and respect it here?

tests/functional/test_contract_event.py
```diff
@@ -0,0 +1,281 @@
...
+    logs = Queue(maxsize=3)
+    poller = vyper_contract_instance.NumberChange.poll_logs()
+
+    with poll_daemon("logs", poller, logs.put, lambda: logs.full()):
```
fubuloubu, line 239: Hmm, this is pretty clean. Can this work for `poll_blocks` too?

src/ape/managers/chain.py
unparalleled-js: I just renamed it to `has_yielded_before_reorg` to help clarify.
Basically tracking if we yield blocks so that if a re-org happens, we can warn the user that we gave them some blocks that are not longer canonical

tests/functional/conftest.py
unparalleled-js: This was a good suggestion. I just added a `new_block_timeout` kwarg to both polling methods so that if a new block has not appeared in X number of seconds (float), raise an exception.

* The default for local in 10 seconds, just so tests don't hang on accident
* And the default for live networks is 3 * the `block_time` network setting because that seems okay for now

tests/functional/test_contract_event.py
unparalleled-js: yes, I just add tests for `poll_block` to prove it!

src/ape/contracts/base.py
```diff
@@ -440,6 +441,9 @@ def poll_logs(
             required_confirmations (Optional[int]): The amount of confirmations to wait
               before yielding the block. The more confirmations, the less likely a reorg will occur.
               Defaults to the network's configured required confirmations.
+            new_block_timeout (Optional[int]): The amount of time to wait for a new block before
+              quitting. Defaults to 10 seconds for local networks or ``3 * block_time`` for live
```
fubuloubu, line 445: Remove "Defaults" because it doesn't default here

fubuloubu (dismissed):

src/ape/contracts/base.py
unparalleled-js: it is the default behavior though

banteg: `block_time * 3` sounds insufficient, i can demonstrate:

```python
>>> block_times = chain.blocks.query('timestamp', start_block=chain.blocks.height - 10000, stop_block=chain.blocks.height).diff()
>>> block_times.describe()
          timestamp
count  10000.000000
mean      13.285700
std       12.749123
min        1.000000
25%        4.000000
50%        9.000000
75%       18.000000
max      121.000000
>>> desc = block_times.describe()['timestamp']
>>> desc['max'] / desc['mean']
9.107536674770618
```
as you can see, a worst case is 9.1x the average block time and 13.4x of the median block time

fubuloubu (dismissed):

unparalleled-js:
OK - some intermittent race condition here still ugh

fubuloubu (approved):

Issue #838, opened 2022-06-29
add supports for block_identifier and state_override for eth_call


banteg:
### Elevator pitch:
Add support for calls at any past block height and for state overrides for the context of the call.

### Value:
Historical calls are useful anyone who would like to learn what a certain view method has returned in the past.

State overrides are useful for patching contract storage and code, which opens up room for making private view methods public, and batch reading storage even if the original contract didn't have these features exposed.

### Dependencies:
Both features could be optional and are supported in web3.py already.

### Design approach:
web3.py and brownie already implement these features and they are supported by most clients. 

web3.py [calls the arguments](https://github.com/ethereum/web3.py/blob/51b6105c69dbcbe01be552c579c6fc9d66c07188/web3/eth.py#L267-L269) `block_identifier` and `state_override`, brownie calls them `block_identifier` and `override`.

Maybe we can use something more concise like `height` and `override`, up to you.

### Task list:
- [ ] Agree on names
- [ ] Add two optional arguments to `ProviderAPI.send_call`.
- [ ] Add two optional arguments to `ContractCallHandler.__call__` which follow the same convention.
- [ ] Same goes for `ContractCall.__init__`.

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

PR #865, opened 2022-07-09
feat: support request kwargs in web3 provider API methods

unparalleled-js:
### What I did

fixes: #838 

### How I did it

Add optional params to txn
Check if they are present during call and use them.

### How to verify it

See test
Need help with `state_override` test

### Checklist


banteg:
nice, i think it would be useful to add `block_id` to all other methods which support `block_identifier`
- [x] call
- [ ] estimate_gas
- [ ] get_storage_at
- [ ] get_transaction_count (nonce)
- [ ] get_balance (balance)
- [ ] get_code

banteg:
for state override we can borrow this for typing https://github.com/ethereum/web3.py/blob/b2cc97561d0cf89ad352e4849b13d71f51106f62/web3/types.py#L231-L244

src/ape_ethereum/transactions.py
```diff
@@ -50,6 +51,9 @@ class AccessList(BaseModel):
 
 
 class BaseTransaction(TransactionAPI):
+    block_id: Optional[BlockID] = None
```
fubuloubu, line 54: Been thinking it might be good to have this in the base model for cross-referencing (except always filled out?)

src/ape_ethereum/transactions.py
```diff
@@ -50,6 +51,9 @@ class AccessList(BaseModel):
...
+    state_override: Optional[Dict] = None
```
fubuloubu, line 55: Is this needed in the model?

fubuloubu (dismissed):

src/ape_ethereum/transactions.py
unparalleled-js: We need it for calls too, for checking historical values

src/ape_ethereum/transactions.py
unparalleled-js: If we want to add validation to the object instead of using a dict, adding it the model is a good way to do this. Else, I don't really know where to put the types.

banteg (approved):

banteg (dismissed):

src/ape/api/providers.py
```diff
@@ -615,10 +615,10 @@ def update_settings(self, new_settings: dict):
         self.provider_settings.update(new_settings)
         self.connect()
 
-    def estimate_gas_cost(self, txn: TransactionAPI) -> int:
+    def estimate_gas_cost(self, txn: TransactionAPI, *args, **kwargs) -> int:
```
fubuloubu, line 618: Should probably just be kwargs?

src/ape/api/providers.py
```diff
@@ -656,17 +656,17 @@ def get_block(self, block_id: BlockID) -> BlockAPI:
         block_data = dict(self.web3.eth.get_block(block_id))
         return self.network.ecosystem.decode_block(block_data)
 
-    def get_nonce(self, address: str) -> int:
-        return self.web3.eth.get_transaction_count(address)  # type: ignore
+    def get_nonce(self, address: str, *args, **kwargs) -> int:
```
fubuloubu, line 660: Ditto

fubuloubu (dismissed):

tests/functional/test_contract_instance.py
```diff
@@ -29,6 +29,17 @@ def test_init_specify_contract_type(
...
 
 
+def test_call_using_block_identifier(
+    vyper_contract_instance, owner, chain, networks_connected_to_tester
+):
+    contract = vyper_contract_instance
+    contract.setNumber(1, sender=owner)
+    height = chain.blocks.height
+    contract.setNumber(33, sender=owner)
+    actual = contract.myNumber(block_identifier=height)
```
fubuloubu, line 39: This is very nice! One of the query types I eventually wanted to add was a historical query of view methods over a number of blocks

src/ape/api/transactions.py
```diff
@@ -52,10 +52,8 @@ def total_transfer_value(self) -> int:
         Useful for determining if an account balance can afford
         to submit the transaction.
         """
-        if self.max_fee is None:
-            raise TransactionError(message="Max fee must not be null.")
-
-        return self.value + self.max_fee
+        max_fee = self.max_fee or 0
+        return self.value + max_fee
```
fubuloubu, line 56: This should be `self.max_fee` or `self.max_priority_fee + self.provider.base_fee`

Although, I think this might be too much detail for `TransactionAPI` and it should probably leave this open as an API method to fill in based on the structure of the transaction (whether it used base_fee) A new issue?

banteg (dismissed):

src/ape/api/transactions.py
unparalleled-js: I can undo these changes, I think I meant for this to be somewhere else

fubuloubu (approved):

Issue #792, opened 2022-06-08
Compiler Information field not stored in package manifest


sabotagebeats:
### Elevator pitch:
Add compiler settings and version information to manifest for publishing use cases

### Value:
Publishing a manifest, publishing files from a project (e.g. Etherscan verification)

### Dependencies:
Contract manifest requires compiler to be filled in to be more useful and for contract verification https://github.com/ApeWorX/ape/issues/81

### Design approach:
EIP-2678 defines an array of [`CompilerInformation`](https://eips.ethereum.org/EIPS/eip-2678#the-compiler-information-object) objects that are referenced by `Source` objects within the same manifest. These objects are used to determine which compiler types, versions, and settings that should be used to compile the (sub-)set of sources for a given manifest.

Using EIP-2678 (or adapting for our needs via #794), add support for obtaining the information for this field in `ProjectManager` and `CompilerAPI` so that we have it for later use. The current structure may be inadequate for our purposes, requiring a breaking change to `ethpm-types` that should be targeted for v0.3.0 if possible.

### Task list:
* [ ] Identify compilation process for projects
* [ ] Obtain compiler version and setting information for each (sub-)set of files to be compiled
* [ ] Leverage this information for calling each compiler (in series) to obtain the various results

### Estimated completion date:
TBD

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

fubuloubu:
p.s. next time use `Work Item` issue template for this kind of thing

sabotagebeats:
dependency for vyper: https://github.com/ApeWorX/ape/issues/850

sabotagebeats:
in progress: blocked by issue #850

unparalleled-js:
> in progress: blocked by issue #850

See latest comment! Not necessarily blocked.

Also - let's get Solidity out the door since it is in-progress and then return to vyper

[status update from retro]

sabotagebeats:
In progress. Needs refactor to place data correctly.

sabotagebeats:
merged. requires tests to be merged.

Issue #81, opened 2021-07-04
General: Add support for smart contract verification to ExplorerAPI


fubuloubu:
### Overview

When you deploy a smart contract to a public network, you often want to verify that contract's source code through one (or more) blockchain explorer(s) source code verification API (if it exists).

### Specification

Add a kwarg `verify=None` to `AccountAPI.deploy(ContractType, ..., verify=True)` that will publish the corresponding smart contract code to any loaded `ExplorerAPI` plugin

Also prototype a way to verify source code "after the fact", maybe via `ContractInstance` or `ReceiptAPI` of the deployment txn.

### Dependencies

Have to be able to link back to the `Source`(s) (or just use the package manifest directly) from a `ContractType` of the deployed contract.

unparalleled-js:
there is also an issue for this found in `ape-etherscan`: https://github.com/ApeWorX/ape-etherscan/issues/22
Now, they are linked :)

sabotagebeats:
dependency #792 #530

sabotagebeats:
related: https://github.com/ApeWorX/ape/issues/794

sabotagebeats:
@NotPeopling2day I believe this is the information you were requesting above 👍

sabotagebeats:
dependency: https://github.com/ApeWorX/ape/issues/850

sabotagebeats:
status: blocked by https://github.com/ApeWorX/ape/issues/948 and https://github.com/ApeWorX/ape/issues/530

PR #871, opened 2022-07-11
feat: compiler data in manifest

sabotagebeats:
### What I did
retrieve compiler data and insert in manifest

fixes: #792 

### How I did it
grab compiler data from cls

need compiler data to unblock #81  

### How to verify it
add testing (I need to do this) 

### Checklist


sabotagebeats:
I think this may needs manifest to be stored once compiler data is added.

src/ape/api/projects.py
```diff
@@ -115,6 +115,40 @@ def _create_manifest(
...
+                        {
+                            "name": cls.compiler_manager.registered_compilers[suffix].name,
+                            "version": cls.compiler_manager.registered_compilers[
+                                suffix
+                            ].get_versions(source_paths),
+                            "settings": None,
+                            "contractTypes": contract_types[contract_type],
+                        }
+                    )
+            i += 1
```
fubuloubu, line 150: I don't think this will work here because it's a classmethod, I would try to move it to where it actually compiles the files

sabotagebeats: it seems to mostly be working, I think I have to fix the iterator. I'll look to see how I can move it to where compilation is occurring if that seems like it would work better.

fubuloubu: If you use one of the managers in a classmethod I would suspect it may run into weird race conditions down the line

NotPeopling2day: I think it's fine to be here, goes along with the source dict logic, haven't seen any race conditions with it. Here's maybe a place to start exploring.

```python
    @classmethod
    def _create_manifest(
        cls,
        source_paths: List[Path],
        contracts_path: Path,
        contract_types: Dict[str, ContractType],
        name: Optional[str] = None,
        version: Optional[str] = None,
        initial_manifest: Optional[PackageManifest] = None,
    ) -> PackageManifest:
        manifest = initial_manifest or PackageManifest()

        if name:
            manifest.name = PackageName(name.lower())

        if version:
            manifest.version = version

        manifest.compilers = cls._create_compilers_list(source_paths, contract_types)
        manifest.sources = cls._create_source_dict(source_paths, contracts_path)
        manifest.contract_types = contract_types

        return manifest

    @classmethod
    def _create_compilers_list(
        cls,
        source_paths: List[Path],
        contract_types: Dict[str, ContractType],
    ):
        compiler_list: List[Compiler] = []

        for key, compiler in cls.compiler_manager.registered_compilers.items():
            compiler_list.append(
                Compiler(
                    name=compiler.name,
                    # TODO: Figure out if this should be str
                    version=str(compiler.get_versions(source_paths)),
                    # TODO: Find compiler settings
                    settings={},
                    # TODO: Figure out what file type/compiler for each contract_typeT
                    # TODO: You'll need to filter these to match the compiler used
                    # NOTE: Consider adding a compiler member to ContractType class?
                    # NOTE: Check data type, ethpm-types says List[str] not sure what this data should be example passes name of contract
                    contractTypes=[str(x) for x in contract_types.keys()],
                )
            )

        return compiler_list
```

src/ape/api/projects.py
```diff
@@ -144,6 +174,10 @@ def _create_source_dict(
 
         return source_dict  # {source_id: Source}
 
+    # @classmethod
```
unparalleled-js, line 177: - [ ] remove commented out code

unparalleled-js, line 177: - [ ] remove commented out code

src/ape/api/projects.py
```diff
@@ -115,6 +115,36 @@ def _create_manifest(
...
+        manifest.compilers = []
+
+        i = 0
+        for contract_type in contract_types:
+            suffix = source_paths[i].suffix
+            if cls.compiler_manager.registered_compilers[suffix].name:
+                if suffix == ".sol":
+                    compiler_version = [
+                        x
+                        for x in cls.compiler_manager.registered_compilers[suffix].get_version_map(
```
unparalleled-js, line 128: It might scale better if we make `get_version_map` a `@raise_note_implemented` method in the `CompilerAPI` and then first attempt to call it and if it raises, then resort to `get_versions()`

src/ape/api/projects.py
```diff
@@ -115,6 +115,36 @@ def _create_manifest(
...
+                    compiler_version = (
+                        cls.compiler_manager.registered_compilers[suffix].get_versions(
+                            source_paths
+                        ),
+                    )
+                manifest.compilers.append(
+                    {
+                        "name": cls.compiler_manager.registered_compilers[suffix].name,
+                        "version": compiler_version,
+                        "settings": None,
```
unparalleled-js, line 142: hmm, what is settings? How does we plan to populate in this future.

src/ape/api/projects.py
```diff
@@ -115,6 +115,36 @@ def _create_manifest(
 
         manifest.sources = cls._create_source_dict(source_paths, contracts_path)
         manifest.contract_types = contract_types
+
+        manifest.compilers = []
+
+        i = 0
+        for contract_type in contract_types:
```
unparalleled-js, line 122:
```suggestion
        for contract_name in contract_types:
```

src/ape/api/projects.py
unparalleled-js: Edit: does not apply if you do Evan's approach! So ignore this

src/ape/api/projects.py
```diff
@@ -127,6 +127,11 @@ def _create_compilers_list(
     ):
         compiler_list: List[Compiler] = []
         for key, compiler in cls.compiler_manager.registered_compilers.items():
+            contractTypes = [
+                str(x)
+                for x in contract_types.keys()
+                if key.strip(".") in contract_types[x].source_id
+            ]
```
sabotagebeats, line 134: this seems to be working!

unparalleled-js: What would happen if I had a contract named `sol.vy` ?

sabotagebeats: oh i see that would not work. I could split it by `.` and use the first index though!

sabotagebeats: this should be fixed in newest commit, and fixed the variable name to not be camel case

src/ape/api/projects.py
```diff
@@ -113,10 +113,42 @@ def _create_manifest(
...
+            compiler_list.append(
+                Compiler(
+                    name=compiler.name,
+                    # TODO: Figure out if this should be str
+                    version=str(compiler.get_versions(source_paths)),
+                    # TODO: Find compiler settings
+                    settings={},
+                    # TODO: Figure out what file type/compiler for each contract_typeT
+                    # TODO: You'll need to filter these to match the compiler used
```
sabotagebeats, line 143: I believe these are sorted in the newest commit @NotPeopling2day

src/ape/api/projects.py
```diff
@@ -113,10 +113,42 @@ def _create_manifest(
...
+            filtered_contract_types = [
+                str(x)
+                for x in contract_types.keys()
+                if key.strip(".") in contract_types[x].source_id.split(".")[-1]
+            ]
+            compiler_list.append(
+                Compiler(
+                    name=compiler.name,
+                    # TODO: Figure out if this should be str
+                    version=str(compiler.get_versions(source_paths)),
```
unparalleled-js, line 139: We should be looping through the versions used in the project and attaching individual instances of `Compiler` per version (which are str then instead of a stringified set like here)

I fear I misled you! @NotPeopling2day suggested this approach but I think it would be inaccurate to include versions of compilers not used in the active project.

unparalleled-js: ah nevermind, we use the sources are the key so this is good.

but we just need to loop over the versions then

unparalleled-js: NOTE: Take care to only include contract types that used that specific version though.. I am not sure if are we able to do this using `get_versions()` because it only returns a `set` whereas the method in `ape-solidity` `get_version_map()` returns a dictionary that maps version to contract types.

We may just want to add `get_version_map` to the API...

tests/functional/test_project.py
```diff
@@ -46,6 +46,7 @@ def test_extract_manifest(dependency_config, project_manager):
     # NOTE: Only setting dependency_config to ensure existence of project.
     manifest = project_manager.extract_manifest()
     assert type(manifest) == PackageManifest
+    assert type(manifest.compilers) == list
```
unparalleled-js, line 49: can we add an assert, like make sure the list has a single item in it with an ape-pm noticeable property

src/ape/api/projects.py
sabotagebeats: yes you're right! 
```
[Compiler(name='solidity', version="{'0.8.15', '0.6.12'}", settings={}, contractTypes=['SafeMath', 'Token', 'Payee', 'GenerativeNFT']),
 Compiler(name='vyper', version="{'0.2.16'}", settings={}, contractTypes=['VyperToken']),
 Compiler(name='ethpm', version='set()', settings={}, contractTypes=[])]
 ```
 this object doesn't assign the correct versions when there's more than one version of solidity file being compiled simultaneously! 😅

File: tests/functional/test_project.py (hunk 9 above)
Comment (sabotagebeats) [reply tests/functional/test_project.py:49]: perfect thank you 👍

File: src/ape/api/projects.py (hunk 1 above)
Comment (NotPeopling2day) [reply src/ape/api/projects.py]: Oh yeah, forgot that the ethpm will be in there

**Review by unparalleled-js — APPROVED** [review]
LGTM

File: src/ape/api/projects.py [hunk 10]
```diff
@@ -113,10 +114,55 @@ def _create_manifest(
...
+        contracts_folder = cls.config_manager.contracts_folder
+        for ext, compiler in cls.compiler_manager.registered_compilers.items():
+            sources = [x for x in source_paths if x.suffix == ext]
+            if not sources:
+                continue
+
+            try:
+                version_map = compiler.get_version_map(sources, contracts_folder)
+            except APINotImplementedError:
+                versions = list(compiler.get_versions(sources))
```
unparalleled-js, line 139: I just thought of a way we can handle this:

```python
for src in sources:
    versions = compiler.get_versions([src])
    if not versions:
        continue
    
    version = max(versions)
```

unparalleled-js: loop through the sources and this way we can handle situations where there happen to be more than one version albeit likely less accurate than an implemented `get_version_map()` API method

unparalleled-js:
@sabotagebeats is adding tests in solidity and vyper

sabotagebeats:
> @sabotagebeats is adding tests in solidity and vyper

https://github.com/ApeWorX/ape-vyper/pull/53
https://github.com/ApeWorX/ape-solidity/pull/61

src/ape/api/projects.py
unparalleled-js: realized this is not possible due to issues such as when the source does not list a pragma and the version is determined another way.

the only way to handle multi-version is to implement `get_version_map()`

src/ape/api/projects.py
```diff
@@ -113,10 +114,55 @@ def _create_manifest(
...
+                if len(versions) == 0:
+                    # Some compilers like json compiler doesn't use versioning
+                    versions = [Version("0.1.0")]
```
unparalleled-js, line 142: @sabotagebeats Talked with @NotPeopling2day 
We think we should skip the JSON compiler for now.


```suggestion
                    continue
```

unparalleled-js: and adjust the comment above to say we are skipping compilers that don't use versions since they are unlikely to be part of the published manifest

NotPeopling2day (approved):
LGTM

tests/functional/test_project.py
fubuloubu: Not sure this is a really good test

fubuloubu (approved):

PR #877, opened 2022-07-13
fix: get_virtual_machine_error() message extract

helloibis:
### What I did

Correct the check of the exception message so that `None` and `""` are handled by the conditional. 

In its current state the logic of `str(err_data.get("message"))` means if `message` isn't in the dictionary, the message will become `"None"`.

### How to verify it

Converting an error with a dictionary but no message means that the message becomes `"None"` rather than the contents of the originating exception:

```python
with ape.reverts():
    raise ape.project.provider.get_virtual_machine_error(Exception({"testing": "stuff"}))

# E       AssertionError: Transaction did not revert.
# E       However, an exception of type <class 'ape.exceptions.VirtualMachineError'> occurred: None
```

With the change:

```python
with ape.reverts():
    raise ape.project.provider.get_virtual_machine_error(Exception({"testing": "stuff"}))

# E       AssertionError: Transaction did not revert.
# E       However, an exception of type <class 'ape.exceptions.VirtualMachineError'> occurred: {'testing': 'stuff'}
```

### Checklist


helloibis:
Tests may fail due to this being a fork, I haven't set up the secrets/environment yet if needed

But I also ran into some issues around `EMFILE: Too many open files` when trying to run tests locally, I made a post in the engineering channel regarding it

fubuloubu (approved):

fubuloubu:
> Tests may fail due to this being a fork, I haven't set up the secrets/environment yet if needed

When you open a PR, the tests run in the context of our actions here

helloibis:
@fubuloubu when you create PRs via forks, [Github actually doesn't provide secrets from either the upstream env or the fork](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow)

If any tests/setup/etc utilize those environment variables or secrets then they end up failing, we encountered it recently on `ape-hardhat` where my PR removing the changelog couldn't pass tests https://github.com/ApeWorX/ape-hardhat/pull/72

Looks like `ape` doesn't use them so we're all good - Jules and I talked about maybe skipping tests or providing fallback values for repos that do rely on secrets

Issue #881, opened 2022-07-15
Integration tests open too many file descriptors


helloibis:
### Environment information

* OS: macOS 12.4 (arm64)
* Python Version: 3.10.5
* `ape` and plugin versions:

```sh
$ ape --version
0.3.5.dev5+gdf844d11

$ ape plugins list
Installed Plugins:
  tokens      0.3.0
```

### What went wrong?

Running the integration tests for the `ape` core opens too many file descriptors without closing them, on my MacBookPro18,1 the default is 256:

```shell
> ulimit -a
-t: cpu time (seconds)              unlimited
-f: file size (blocks)              unlimited
-d: data seg size (kbytes)          unlimited
-s: stack size (kbytes)             8176
-c: core file size (blocks)         0
-v: address space (kbytes)          unlimited
-l: locked-in-memory size (kbytes)  unlimited
-u: processes                       2666
-n: file descriptors                256
```

The error is shown in this example (sorry for the length):

```shell
venv > pytest tests/integration
======================================================= test session starts ========================================================
platform darwin -- Python 3.10.5, pytest-6.2.5, py-1.11.0, pluggy-0.13.1
rootdir: /Users/ibis/Documents/ape, configfile: pyproject.toml
plugins: web3-5.29.2, xdist-2.5.0, forked-1.4.0, mock-3.8.1, hypothesis-6.48.2, cov-3.0.0
collected 675 items                                                                                                                

tests/integration/cli/test_accounts.py .................................................................................     [ 21%]
tests/integration/cli/test_compile.py .......................                                                                [ 27%]
tests/integration/cli/test_console.py .............................................................The history saving thread hit an unexpected error (OperationalError('unable to open database file')).History will not be written to the database.
FFFFFFFFFFFEEFFEEEEFFFFFF [ 49%]
FFEEFFEEEEFFFFFFFFEEFFEEEEFFFFFFFFEEFFEEEEFFFFFFFFEE                                                                         [ 62%]
tests/integration/cli/test_misc.py EEEEEEEEEEEEEEEE.F...FFFEEEEEEEEEEEEEEEE.F...FFF.F...FFF.F...FFF.F...FFF                  [ 81%]
tests/integration/cli/test_networks.py EFEEFFFFEFEEFFFFEEFEEFFFFEFEEFFFF                                                     [ 90%]
tests/integration/cli/test_plugins.py xxxxx                                                                                  [ 91%]
[...]

PR #882, opened 2022-07-15
fix: disable console history in tests

helloibis:
### What I did

Prevents IPython from spawning history collection threads when running the test suite

fixes: #881 

### How I did it

Add an auto-used fixture to patch an `APE_TESTING` environment variable in while testing. The IPython embedded instance launches with history disabled if this environment variable is set.

### How to verify it

If the tests weren't failing from "too many open files", you can change the soft limit on your machine with this.

```shell
$ ulimit -n  # check current value so you can restore it later
...
$ ulimit -Sn 256  # set value to what was default in #881's repro steps
$ pytest tests/integration
```

This should produce a failure without these changes 

### Checklist

### DIFF
--- a/src/ape_console/_cli.py
+++ b/src/ape_console/_cli.py
@@ -1,11 +1,10 @@
 import faulthandler
 import inspect
-import io
 import logging
 import sys
 from importlib.machinery import SourceFileLoader
 from importlib.util import module_from_spec, spec_from_loader
-from os import getcwd
+from os import environ, getcwd
 from types import ModuleType
 from typing import Any, Dict
 
@@ -110,11 +109,8 @@ def console(project=None, verbose=None, extra_locals=None):
             project_path=project.path,
         )
 
-        try:
+        if not environ.get("APE_TESTING"):
             faulthandler.enable()  # NOTE: In case we segfault
-        except io.UnsupportedOperation:
-            # Likely running in tests
-            pass
 
     namespace = {component: getattr(ape, component) for component in ape.__all__}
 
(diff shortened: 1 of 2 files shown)


src/ape_console/_cli.py
```diff
@@ -110,11 +109,8 @@ def console(project=None, verbose=None, extra_locals=None):
             project_path=project.path,
         )
 
-        try:
+        if not environ.get("APE_TESTING"):
             faulthandler.enable()  # NOTE: In case we segfault
-        except io.UnsupportedOperation:
-            # Likely running in tests
-            pass
```
helloibis, line 117: This seemed to only arise from running the test suite, so I opted to change up enabling the fault handler only when we aren't testing (i.e. have the test env set)

src/ape_console/_cli.py
```diff
@@ -127,4 +123,11 @@ def console(project=None, verbose=None, extra_locals=None):
     if console_extras:
         namespace.update(console_extras)
 
-    IPython.embed(colors="Neutral", banner1=banner, user_ns=namespace)
+    from traitlets.config import Config
+
+    config = Config()
```
helloibis, line 128: Is it safe to assume an empty config when `not environ.get("APE_TESTING")` or should I have two separate paths and not provide the `config` kwarg at all if the env var isn't set?

tests/conftest.py
```diff
@@ -12,6 +12,11 @@
 ape.config.PROJECT_FOLDER = Path(mkdtemp()).resolve()
 
 
+@pytest.fixture(autouse=True)
+def setenviron(monkeypatch):
+    monkeypatch.setenv("APE_TESTING", "1")
```
fubuloubu, line 17: Add a comment about why

fubuloubu (dismissed):
Very good contribution!

tests/conftest.py
helloibis: Updated in 85a3b03

fubuloubu (approved):

PR #885, opened 2022-07-18
feat: stream traces in geth provider

banteg:
### what was wrong
ape has waited for a full request to arrive, which made it impossible to fetch certain large traces. now we can process them as stream of frames as they come it and massively reduce memory requirements. this feature doesn't require any api changes since we already return an iterator.

### what i did
- add `GethProvider.stream_request` which iteratively decodes a response using `ijson` and enables processing of huge multi-gigabyte traces
- use it to fetch traces

### other considerations
i also evaluated using streaming for logs, but the results were worse this way.

erigon-streaming

### checklist

### DIFF
--- a/src/ape_geth/providers.py
+++ b/src/ape_geth/providers.py
@@ -275,7 +275,7 @@ class GethProvider(Web3Provider, UpstreamProvider):
         results = ijson.sendable_list()
         coro = ijson.items_coro(results, iter_path)
 
-        resp = requests.post(self.connection_str, json=payload, stream=True)
+        resp = requests.post(self.uri, json=payload, stream=True)
         resp.raise_for_status()
 
         for chunk in resp.iter_content(chunk_size=2**17):


src/ape/api/providers.py
```diff
@@ -704,6 +706,19 @@ def get_transactions_by_block(self, block_id: BlockID) -> Iterator:
         for transaction in block.get("transactions"):  # type: ignore
             yield self.network.ecosystem.create_transaction(**transaction)  # type: ignore
 
+    def stream_request(self, method, params, iter_path="result.item"):
+        payload = {"jsonrpc": "2.0", "id": 1, "method": method, "params": params}
+        results = ijson.sendable_list()
+        coro = ijson.items_coro(results, iter_path)
+
+        resp = requests.post(self.connection_str, json=payload, stream=True)
```
unparalleled-js, line 714: There is a couple issues with this I think.

1.) This assumes that the `Web3Provider` is HTTP based.
2.) `connection_str` is only part of `UpstreamProvider`.

We can get rid of UpstreamProvider and make `connection_str` part of the root ProviderAPI class. This would be a breaking change but I am ok with it

To address the first point, we can verify that connection_str starts with http:// or https:// first or else raise an `APINotImplementedError` saying that subclass needs to implement its own stream_request

banteg: i see what you mean, i initially put it into `GethProvider` which explicitly [uses HTTPProvider](https://github.com/ApeWorX/ape/blob/74ec1f42481f0a8d940e3c085ac13fbf0de57b4a/src/ape_geth/providers.py#L176). maybe i should move it back?

src/ape_geth/providers.py
```diff
@@ -268,9 +270,21 @@ def disconnect(self):
         self._web3 = None  # type: ignore
         self._client_version = None
 
+    def stream_request(self, method, params, iter_path="result.item"):
+        payload = {"jsonrpc": "2.0", "id": 1, "method": method, "params": params}
+        results = ijson.sendable_list()
+        coro = ijson.items_coro(results, iter_path)
+
+        resp = requests.post(self.connection_str, json=payload, stream=True)
```
unparalleled-js, line 278: Optional / nit: I would have used `self.uri` instead of `self.connection_str` because the connection_str is the abstract version of the URI and the URI is specific to the provider, as is method.

unparalleled-js (approved):
nice optimization

PR #886, opened 2022-07-18
fix: bug where geth provider would not use URI from provider_settings dict

unparalleled-js:
### What I did

Fixes issue where user could not connect to adhoc geth URI via `provider_settings`.

### How I did it

Check for `uri` in 

### Checklist


NotPeopling2day (approved):
LGTM

Issue #887, opened 2022-07-19
Add a fee estimation helper


ca11ab1e:
### Overview

Add a helper to estimate gas cost:

```python
# For mutable methods (currently supported)
estimated_fee = contract.methodToCall.estimate_fee(1, sender=accounts.load("me"))

# For immutable methods (not currently supported)
estimated_fee = contract.immutableMethodToCall.estimate_fee(1, sender=accounts.load("me"))
```

### Specification

The current API to get fee estimation is as follow:

```python
txn = contract_a.methodToCall.as_transaction(1, sender=accounts.load("me"))
estimated_fees = provider.estimate_gas_cost(txn)
```

Note: it doesn't support immutable methods.

For mutable methods, the helper would look like:

```python
class ContractTransactionHandler:
    def estimate_fee(self, *args, **kwargs) -> int:
        txn = self.as_transaction(*args, **kwargs)
        return self.provider.client.estimate_gas_cost(txn)
```

And for immutable methods:

```python
class ContractCallHandler:
    def as_transaction(self: ContractCallHandler, *args, **kwargs):
        args = self._convert_tuple(args)
        selected_abi = _select_method_abi(self.abis, args)

        contract_transaction = ContractTransaction(  # type: ignore
            abi=selected_abi,
            address=self.contract.address,
        )

        transaction = contract_transaction.serialize_transaction(*args, **kwargs)
        self.provider.prepare_transaction(transaction)
        return transaction

    def estimate_fee(self, *args, **kwargs) -> int:
        txn = self.as_transaction(*args, **kwargs)
        return self.provider.client.estimate_gas_cost(txn)
```

This is a rought implementation, and a refactoring may be suitable.

### Dependencies

None.

fubuloubu:
I approve of this design. Output would be in native token units? (e.g. Wei for ethereum)

ca11ab1e:
> I approve of this design. Output would be in native token units? (e.g. Wei for ethereum)

Exactly.

PR #888, opened 2022-07-19
feat: estimate fee support for calls

unparalleled-js:
### What I did

fixes: #887 

### How I did it

* Added `estimate_fee`  helper method to `ContractTransactionHandler`
* Added ability to make transactions from calls via `as_transaction` method in `ContractCallHandler`
* Added `estimate_fee`  helper method to `ContractCallHandler`
* Adjusted 1 test and added another test.

### How to verify it

You can now estiamate fees on calls

### Checklist


ca11ab1e (approved):

ca11ab1e:
Maybe updating https://github.com/ApeWorX/ape/blob/1c2735399608ad5b76c68d562d291eb75c71aaf9/docs/userguides/transactions.md#estimate-fees would be great too.

src/ape/contracts/base.py
```diff
@@ -126,6 +126,47 @@ def __call__(self, *args, **kwargs) -> Any:
             address=self.contract.address,
         )(*args, **kwargs)
 
+    def as_transaction(self, *args, **kwargs):
+        """
+        Convert the Call to a transaction. This is useful for checking coverage
+        or estimating fees.
```
fubuloubu, line 132: Also useful for gas costs

src/ape/contracts/base.py
```diff
@@ -204,6 +245,22 @@ def as_transaction(self, *args, **kwargs) -> TransactionAPI:
...
+        Get the estimated fee (according to the provider) for the
+        contract method-invocation transaction.
+
+        Args:
+            *args: The contract method invocation arguments.
+            **kwargs: Transaction kwargs, such as value or
+              sender.
+
+        Returns:
+            int: The estimated fee to invoke the transaction.
```
fubuloubu, line 259: Add more info about the units for this value

fubuloubu (dismissed):
This is a nice API

src/ape/contracts/base.py
```diff
@@ -126,6 +126,48 @@ def __call__(self, *args, **kwargs) -> Any:
...
+        Args:
+            *args: The contract method invocation arguments.
+            **kwargs: Transaction kwargs, such as value or
+              sender.
+
+        Returns:
+            int: The estimated cost of gas to execute the transaction
+            reported in the fee-currency's smallest unit, e.g. Wei.
+        """
+        txn = self.as_transaction(*args, **kwargs)
```
ca11ab1e, line 168: In my hot-patch, I added a check right after that line:

```python
if txn.max_fee:
    return txn.max_fee
```

It may not be correct, but it was required because of those lines in `ape-starknet`: https://github.com/ApeWorX/ape-starknet/blob/6a9b090a89c71a2c1517e9502dad252452705da9/ape_starknet/provider.py#L317-L318

So after `.as_transaction()` was called, calling again `.estimate_gas_cost()` would end on a HTTP 500 error. Maybe that should be dealt at the `ape-starknet` level, I'm just rising the question :)

src/ape/contracts/base.py
```diff
@@ -204,6 +246,23 @@ def as_transaction(self, *args, **kwargs) -> TransactionAPI:
...
```
ca11ab1e, line 263: In my hot-patch, I added a check right after that line:

```python
if txn.max_fee:
    return txn.max_fee
```

It may not be correct, but it was required because of those lines in `ape-starknet`: https://github.com/ApeWorX/ape-starknet/blob/6a9b090a89c71a2c1517e9502dad252452705da9/ape_starknet/provider.py#L317-L318

So after `.as_transaction()` was called, calling again `.estimate_gas_cost()` would end on a HTTP 500 error. Maybe that should be dealt at the `ape-starknet` level, I'm just rising the question :)

src/ape/contracts/base.py
```diff
@@ -126,6 +126,48 @@ def __call__(self, *args, **kwargs) -> Any:
...
+        return txn.max_fee if txn.max_fee else self.provider.estimate_gas_cost(txn)
```
ca11ab1e, line 169: Hmm after more thinking, we should not do that check here. Let's return `self.provider.estimate_gas_cost(txn)` and let plugins handling specific cases.

ca11ab1e (dismissed):

src/ape/contracts/base.py
unparalleled-js: reverted!

ca11ab1e (approved):

docs/userguides/transactions.md
```diff
@@ -215,14 +215,15 @@ receipt = networks.provider.get_transaction(txn_hash)
...
-## Estimate Fees
-
-To estimate the fees on a transaction without sending it, use the `as_transaction()` method to get a reference to a transaction API object.
-Then, use the `ProviderAPI.estimate_gas_cost()` method with the transaction as the argument.
+## Estimate Gas Cost
 
+To estimate the gas cost on a transaction or call without sending it, use the `estimate_gas_cost()` method from the contract's transaction / call handler:
 (Assume I have a contract instance named `contract_a` that has a method named `methodToCall`)
 
 ```bash
```
fubuloubu, line 223: should this be `py`?

fubuloubu (approved):

src/ape/contracts/base.py
```diff
@@ -126,6 +126,47 @@ def __call__(self, *args, **kwargs) -> Any:
...
+        Args:
+            *args: The contract method invocation arguments.
+            **kwargs: Transaction kwargs, such as value or
+              sender.
+
+        Returns:
+            :class:`~ape.api.transactions.TransactionAPI`
+        """
+        handler = ContractTransactionHandler(self.contract, self.abis)
+        return handler.as_transaction(*args, **kwargs)
```
ca11ab1e, line 143: Why not using the property?
```suggestion
        return self.transact.as_transaction(*args, **kwargs)
```

unparalleled-js: It reads funny but ya that works. `transact` is technically a property but it is a handler object so you can call it

ca11ab1e (approved):

fubuloubu (dismissed):

ca11ab1e (approved):

fubuloubu (approved):

PR #897, opened 2022-07-20
feat: get multiple contracts method in contract cache

unparalleled-js:
### What I did

Ability to fetch multiple contract types in parallel

### How I did it

ThreadPoolExecutor and adding a method to the `chain.contracts`

### How to verify it

tests

### Checklist


src/ape/managers/chain.py
```diff
@@ -504,6 +505,42 @@ def __getitem__(self, address: AddressType) -> ContractType:
 
         return contract_type
 
+    def get_all(
```
unparalleled-js, line 508: I like `get_all` as a name, it seems to make sense considering we have `get()`.
But am open to alternatives! Just let me know or let me know what issues you have with this one.

banteg: `get_multiple` or `get_many` as an alternative

unparalleled-js:
@fubuloubu Would like to know what you are thinking on this, as it will help me make other decisions!

src/ape/managers/chain.py
```diff
@@ -504,6 +505,44 @@ def __getitem__(self, address: AddressType) -> ContractType:
 
         return contract_type
 
+    def get_multiple(
```
unparalleled-js, line 508: Could combine this with `get()` but that makes the `default=` argument kind of strange so a separate method made more sense.

fubuloubu: Let's just go with it for now.

fubuloubu (approved):

Issue #777, opened 2022-06-03
`Contract` does not raise a warning if it can't find a contract type


fubuloubu:
### Environment information

* OS: linux 
* Python Version: 3.9.9
* `ape` and plugin versions:

```sh
$ ape --version
0.2.7

$ ape plugins list
Installed Plugins:
  hardhat      0.2.2
  solidity     0.2.3
  tokens       0.2.0
  infura       0.2.0
  etherscan    0.2.1
  alchemy      0.2.0
  foundry      0.2.0
  vyper        0.2.0
  ens          0.2.0
```

### What went wrong?

There should be a warning raised if it's not possible to find a contract type for a given address using `Contract`. Here are some scenarios:
1. No `ExplorerAPI` plugins (for the given network)
2. Explorer fails to find a contract type
3. If there is no code at that address or something else

```py
# ape console --network :mainnet
In [1]: c = Contract("0x4332f8a38f14bd3d8d1553af27d7c7ac6c27278d")
# Should raise a warning here

In [2]: c
Out[2]: <Address 0x4332f8a38f14bd3d8d1553af27d7c7ac6c27278d>
```

### How can it be fixed?

It used to do this, so might be a regression

PR #898, opened 2022-07-20
refactor!: move contract instance and container helper utilities to chain.contracts

unparalleled-js:
### What I did

fixes: #777 
this was a breaking change we wanted to do since `create_contract` is a confusing name.

### How I did it

removed method entirely and use methods in `chain.contracts`

### How to verify it

name good or nah?

### Checklist


src/ape/utils/basemodel.py
```diff
@@ -89,7 +89,7 @@ def create_contract_container(self, contract_type: ContractType) -> "ContractCon
 
         return ContractContainer(contract_type=contract_type)
 
-    def create_contract(
+    def get_contract_instance(
```
unparalleled-js, line 92: actually we can get rid of this method all together

src/ape/api/accounts.py
```diff
@@ -173,16 +173,16 @@ def deploy(self, contract: "ContractContainer", *args, **kwargs) -> "ContractIns
...
-        contract_instance = self.create_contract(
-            address=receipt.contract_address,  # type: ignore
-            contract_type=contract.contract_type,
+        logger.success(f"Contract '{contract_name}' deployed to: {styled_address}")
+        contract_instance = self.chain_manager.contracts.instance_at(
+            address, contract.contract_type
         )
-        self.chain_manager.contracts[contract_instance.address] = contract_instance.contract_type
-        return contract_instance
+        # NOTE: Type ignore because we know it is a contract instance at this point
```
fubuloubu, line 183: Could throw a `assert contract_instance` in here to satisfy mypy

fubuloubu (approved):

src/ape/api/accounts.py
unparalleled-js: mypy is mad because the type is BaseAddress and not ContractInstance

Normally I would check the type to satisfy mypy (or assert the type), but importing from the `ape.contracts` namespace is challenging (circular imports)

fubuloubu: how about `assert not isinstance(contract_instance, AddressType)`?

fubuloubu (approved):

src/ape/api/accounts.py
unparalleled-js: It would be `BaseAddress`, but let me try

unparalleled-js: Oh no that won't work because `ContractInstance` is a subclass of `BaseAddress`

unparalleled-js: TBH I kind of wish `instance_at()` always returned a `ContractInstance`. I don't remember why we made it return `Address` when it does not find the contract

fubuloubu: > TBH I kind of wish `instance_at()` always returned a `ContractInstance`. I don't remember why we made it return `Address` when it does not find the contract

I don't either. Let's change that? Or have it return `Optional[ContractInstance]` maybe

fubuloubu: Or maybe `ape.Contract` is more of the API to access the contract cache?

unparalleled-js: ape.Contract calls instance_at() at under-the-hood and manager-access mixin impls use chain_manager.instance_at directly since they can't really import from the top-level namespace.

unparalleled-js: See all the changes in the latest commit! Making this change makes things way  easier and still fixes #777, as now things will fail loudly and cause alert.

fubuloubu (dismissed):

src/ape/managers/chain.py
```diff
@@ -618,16 +618,27 @@ def get(
 
         return contract_type
 
+    def get_container(self, contract_type: ContractType) -> ContractContainer:
```
fubuloubu, line 621: Is this necessary?

fubuloubu (approved):

src/ape/managers/chain.py
unparalleled-js: Yes, it pretty much is. It has to do with importing `ContractContainer`.

We have always had this method, it was in the `BaseModel` before but I think makes more sense here next to other contract-class factory methods.

The reason we need this is because there are internal parts of the code base that need to create `ContractContainer` objects. We are unable to import `ContractContainer` because of cyclical imports. Thus, (many months ago during the Pydantic refactor), we added methods to the `BaseModel` `create_contract()` and `create_contract_container()`. Recently, you pointed out that the names of these methods are misleading and I agree so that is why I created this PR (since it is a breaking change in prep for 0.4).

So either we have to do local imports of `ContractContainer` whenever that is needed or we can have this method which makes things easier.

fubuloubu: Ah okay, makes sense

fubuloubu (approved):

Issue #75, opened 2021-06-21
General: Add EIP 712 Support


ShadeUndertree:
### Elevator pitch:

We already support EIP-712 in the hardware plugins, or at least have a plan to.
We need to also do this for 1st class account plugins, such as `ape-accounts` and `ape-test`.

### Value:

Able to interact in the ecosystem better.

### Dependencies:

https://eth-account.readthedocs.io/en/stable/eth_account.html?highlight=eip-712#eth_account.messages.encode_structured_data

### Design approach:

See `ape-ledger` plugin as an example.

* Step 1: detect message type
* Step 2: If structures, encode it to EIP-191
* Step 3: sign the message (what the two plugins are already doing)

^ Different for non 1st-class account plugin that are based on other libraries. This only applies to account plugin that use `eth-account` (ones that have access to the pkey).

### Task list:

* [ ] Update `ape-accounts` to support signing EIP712 messages
* [ ] Update `ape-test` to support signing EIP712 messages

`ape-ledger` already supports this.
And we have https://github.com/ApeWorX/ape-trezor/issues/8 in `ape-trezor` that can probably be done now (we were waiting on the next release of their lib last I checked).

### Estimated completion date:

TBD.

### Design review:

This is Jules... Signing off!

unparalleled-js:
Upon further investigation, as the `trezorlib` method is finally available to us, it seems we should have a separate API method for this.

Reason:

The trezor lib requires the `dict()` version as it processes and sends over bits in chunks and needs access to the data in other ways.

So using the `eip712.signable_message` property, like we can do in `ape-ledger`, does not work so nicely.
If we had a more generic API method, we can better support EIP-712 messages in `ape-trezor`.

PR #900, opened 2022-07-21
test: ensure support for eip712

unparalleled-js:
### What I did

Adds test to ensure we support EIP-712 messages (in process of debugging something).
fixes: #75 

### How I did it

* eip712 package dependency

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Callable, Iterator, List, Optional, Type, Union
 
 import click
+from eip712.messages import SignableMessage as EIP712SignableMessage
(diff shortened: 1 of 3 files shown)


src/ape/api/accounts.py
```diff
@@ -201,7 +202,7 @@ def check_signature(
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
         """
-        if isinstance(data, SignableMessage):
+        if isinstance(data, (SignableMessage, EIP712SignableMessage)):
```
fubuloubu, line 205: Hmm, I actually like this. I don't think it would be too badly in conflict if this was supported in core (because it has paths for handling other types of messages)

fubuloubu (approved):

PR #903, opened 2022-07-22
refactor!: change order of parameters in get_user_selected_account

unparalleled-js:
### What I did

A breaking change I have wanted to do for a while is allow the prompt to be the first argument in `get_user_selected_account()`. This feels a lot more natural!

### How I did it

Change order of parameters
Add some tests

### Checklist


src/ape/cli/choices.py
```diff
@@ -27,6 +27,7 @@ class Alias(click.Choice):
 
     name = "alias"
 
+    # noinspection PyMissingConstructor
```
fubuloubu, line 30: What do these do?

fubuloubu (approved):

src/ape/cli/choices.py
unparalleled-js: ah they just make Pycharm not complain. I could get rid of these if you want and just turn off inspection locally but it might also be nice for other Pycharm users

unparalleled-js: turns out Pycharm also supports ` # noqa` expressions to do the same thing, so I did that since it is more IDE-generic

src/ape/cli/choices.py
```diff
@@ -27,7 +27,7 @@ class Alias(click.Choice):
 
     name = "alias"
 
-    def __init__(self, account_type: Optional[Type[AccountAPI]] = None):
+    def __init__(self, account_type: Optional[Type[AccountAPI]] = None):  # noqa
```
fubuloubu, line 30: Usually better to do a specific issue w/ noqa

fubuloubu (dismissed):

src/ape/cli/choices.py
unparalleled-js: I removed it because I couldn't figure it out but just FYI Pycharm yells at you from these lines.

fubuloubu (approved):

PR #906, opened 2022-07-22
Async web3 provider

BobTheBuidler:
### What I did
Added AsyncProviderAPI, AsyncWeb3Provider, AsyncGethProvider classes

### How I did it
I essentially copied the existing providers and just made everything async. There are also a few instances where I had to add async versions of functions in various parts of the code.

### How to verify it
Try the various methods you would expect to work from a Geth provider. (eth_call, eth_getLogs, ...)

### Checklist

### DIFF
--- a/src/ape/api/__init__.py
+++ b/src/ape/api/__init__.py
@@ -13,6 +13,7 @@ from .explorers import ExplorerAPI
 from .networks import EcosystemAPI, NetworkAPI, ProviderContextManager, create_network_type
 from .projects import DependencyAPI, ProjectAPI
 from .providers import (
+    AsyncWeb3Provider,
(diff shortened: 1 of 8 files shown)


johnson2427:
@fubuloubu I think this is a huge addition

fubuloubu:
> @fubuloubu I think this is a huge addition

Yes, it's 863 lines to be exact

In all seriousness, I love the submission, but we have to think about how it fits into the overall library before eopening this rabbit hole

antazoey:
@BobTheBuidler going to close, as this is like a year old and, full of conflicts, and different than what we want.
Let's set up a meeting to discuss. Post in Telegram or something.

PR #908, opened 2022-07-23
fix: hardcode chain ID for now in EthTester provider

unparalleled-js:
### What I did

As we wait for a web3.py fix, we can hardcode the chain ID in eth tester.

### How I did it
always return 131277322940537 in `chain_id` for eth tester

### Checklist

### DIFF
--- a/src/ape_test/providers.py
+++ b/src/ape_test/providers.py
@@ -3,7 +3,6 @@ from eth_tester.exceptions import TransactionFailed  # type: ignore
 from eth_utils.exceptions import ValidationError
 from web3 import EthereumTesterProvider, Web3
 from web3.middleware import simple_cache_middleware
-from web3.providers.eth_tester.defaults import API_ENDPOINTS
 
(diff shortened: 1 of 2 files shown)


src/ape_test/providers.py
```diff
@@ -45,16 +43,7 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
...
-        if self.cached_chain_id is not None:
-            return self.cached_chain_id
-        elif hasattr(self.web3, "eth"):
-            chain_id = self.web3.eth.chain_id
-        else:
-            default_value = API_ENDPOINTS["eth"]["chainId"]()
-            chain_id = int(default_value, 16)
-
-        self.cached_chain_id = chain_id
-        return chain_id
```
fubuloubu, line 57: Maybe comment out and/or add TODO note on the bug?

fubuloubu (approved):

unparalleled-js:
Ah looks like this solution won't work. We are going to have to upgrade web3.py somehow or wait for 5.0 patch from them

Issue #879, opened 2022-07-14
integration tests broken in non-ci environment


banteg:
### What went wrong?

if you have plugins installed, integration tests break since they expect a clean installation. some functional tests also only work in ci environment.

<img width="1510" alt="Screen Shot 2022-07-15 at 03 02 56" src="https://user-images.githubusercontent.com/4562643/179115238-dee73201-1ebe-430b-b008-0f7bd0c8f7dc.png">

### How can it be fixed?

alter the tests to be less sensitive to the environment they are running in.

fubuloubu:
Can probably do a query of certain things through pip to see what issues to ignore and what not, or do broader checks instead. I think asserting the display of certain elements might be overkill

unparalleled-js:
You have to run tests in an environment that does not have other plugins installed.

banteg:
another inconsistency between pre-commit and ci

<img width="1049" alt="Screen Shot 2022-07-19 at 02 23 51" src="https://user-images.githubusercontent.com/4562643/179627344-9ff17d3d-3251-408e-9e36-cf831ac557f5.png">

unparalleled-js:
> I think asserting the display of certain elements might be overkill

I think it is a good integration test for making sure a network related plugin is installed correctly.

fubuloubu:
> > I think asserting the display of certain elements might be overkill
> 
> I think it is a good integration test for making sure a network related plugin is installed correctly.

Yes, but you can do it like `assert "some text" in result.output` instead

PR #920, opened 2022-07-27
test: make tests run better locally

unparalleled-js:
### What I did

fixes: #879 

### How I did it

Make tests less strict about their assertions
Make plugin tests uninstall if they installed

### Checklist

### DIFF
--- a/tests/functional/test_networks.py
+++ b/tests/functional/test_networks.py
@@ -3,16 +3,82 @@ import pytest
 from ape.exceptions import NetworkError
 
 
+@pytest.fixture
+def network_with_no_providers(ethereum):
+    network = ethereum.get_network("rinkeby-fork")
+    default_provider = network.default_provider
+    providers = network.__dict__["providers"]
+
+    if default_provider or providers:
+        # Handle user running tests with forked-network plugins installed
+        network._default_provider = None
+        network.__dict__["providers"] = []
+
+    yield network
+
+    if default_provider or providers:
+        network._default_provider = default_provider
+        network.__dict__["providers"] = providers
+
+
 def test_get_network_choices_filter_ecosystem(networks):
     actual = {c for c in networks.get_network_choices(ecosystem_filter="ethereum")}
-    expected = {c for c in networks.get_network_choices()}
-    assert len(actual) == 27
-    assert actual == expected
+    all_choices = {
+        "::geth",
+        "::test",
+        ":goerli",
+        ":goerli-fork",
+        ":goerli:geth",
+        ":kovan",
+        ":kovan-fork",
+        ":kovan:geth",
+        ":local",
+        ":mainnet",
+        ":mainnet-fork",
+        ":mainnet:geth",
+        ":rinkeby",
+        ":rinkeby-fork",
+        ":rinkeby:geth",
+        ":ropsten",
+        ":ropsten-fork",
+        ":ropsten:geth",
+        "ethereum",
+        "ethereum:goerli",
+        "ethereum:goerli-fork",
+        "ethereum:goerli:geth",
+        "ethereum:kovan",
+        "ethereum:kovan-fork",
+        "ethereum:kovan:geth",
+        "ethereum:local",
+        "ethereum:local:geth",
+        "ethereum:local:test",
+        "ethereum:mainnet",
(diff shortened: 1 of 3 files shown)


tests/functional/test_networks.py
```diff
@@ -21,14 +73,14 @@ def test_get_network_choices_filter_provider(networks):
...
-def test_get_provider_when_no_default(networks):
-    ethereum = networks.get_ecosystem("ethereum")
-    network = ethereum.get_network("rinkeby-fork")
+def test_get_provider_when_no_default(network_with_no_providers):
     with pytest.raises(NetworkError) as err:
         # Not provider installed out-of-the-box for rinkeby-fork network
-        network.get_provider()
+        provider = network_with_no_providers.get_provider()
+        if provider:
+            assert False, f"Provider should be None but got '{provider.name}'"
```
helloibis, line 81: Should this be `assert not provider, f"Provider should..."`?

tests/integration/cli/test_networks.py
```diff
@@ -93,29 +93,40 @@ def assert_rich_text(actual: str, expected: str):
...
+    ]
+    actual_lines = [
+        x.replace("└", "").replace("├", "").replace("│", "").strip()
+        for x in actual.strip().split("\n")
+    ]
 
-    actual = "\n".join(new_lines)
-    assert actual == expected
+    for expected_line in expected_lines:
+        assert expected_line in actual_lines
```
helloibis, line 110: Is this function intended to be an exact match or more like an `x in y` match? 

If exact, you'll need to check the length and order of the two lists, since something like this could currently pass:

```python
expected = """
One
Two
Three
"""

actual = """
Four
Two
One
Three
One
"""
```

tests/integration/cli/test_plugins.py
```diff
@@ -25,8 +25,17 @@ def wrapper(f):
 
 @pytest.fixture
 def installed_plugin(runner, ape_cli):
-    # TODO: Figure out how to install plugin prior to test **that actually works**.
-    runner.invoke(ape_cli, ["plugins", "install", TEST_PLUGIN_NAME])
+    # TODO: Figure out how to correctly modify site packages mid-test run.
+    #  The installed plugin does not show in `list` command but does at end of tests.
+    is_already_installed = TEST_PLUGIN_NAME in runner.invoke(ape_cli, ["plugins", "list"]).output
+
```
helloibis, line 31: I haven't used this myself yet but I did find this when looking into changing the installation path for packages:

https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE

helloibis (commented):
Couple questions and one comment, but overall is looking good 👍

tests/integration/cli/test_networks.py
unparalleled-js: The length may be different, so it is more like `x in y`.
It is different when the test runner has additional plugins installed.
And the order is not super important.

What is important is this:

* Comment out a a setting in `ape_geth.provides.GethNetworkConfig` and then run the test. It show now fail! So we are ensuring the provider is registered to specific networks.
* Change the `default_provider` in `ape_ethereum.ecosystem.NetworkConfig` from `geth` to `test`. The tests should no fail! So we are ensuring the correct default provider is listed

There are more examples that this integration test still ensures, but that is just a couple.

tests/integration/cli/test_plugins.py
unparalleled-js: so we can point that at tempdir and that is much safer, because then we never adjusting user's actual site packages

unparalleled-js: it seems to not work when running mid python session already

unparalleled-js: TBH this fixture isn't working anyway, so I made it assume it is installed already for now.

tests/integration/cli/test_networks.py
helloibis: Gotcha, so we basically just care that an expected string (like a network name) shows up at least once in the actual output

helloibis (approved):

NotPeopling2day (dismissed):

helloibis (approved):

fubuloubu (approved):

Issue #922, opened 2022-07-28
Add query to DefaultQueryProvider for contract events


johnson2427:
### Overview

I want to add the ability to query contract events to the `DefaultQueryProvider` to be able to add the ability to cache these queries in the future.

### Specification

Add a `query` method to `base.ContractEvent` and force `range()` to go to the `query_manager` for contract events queries.

### Dependencies

No new dependencies required

PR #923, opened 2022-07-28
feat: added contract event query to default query provider

johnson2427:
### What I did

Added `LogFilter` as a `QueryType`. Then added contract events to the `DefaultQueryProvider`. Added `query` function to `base.ContractEvent` and forced the query for `range()` to go through the `query_manager`

fixes: #922 

### How I did it

Explained above

### How to verify it

`pytest -k test_transaction_contract_event_query`

### Checklist


johnson2427:
This all passes locally, not sure why this fails here

src/ape/api/query.py
```diff
@@ -12,6 +12,7 @@
     "AccountTransactionQuery",
     "ContractEventQuery",
     "ContractMethodQuery",
+    "LogFilter",  # type: ignore # noqa: F821
```
unparalleled-js, line 15: What does this do?
Since it is not defined anywhere, does this actually work?

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def query(
+        self,
+        start_block: int,
+        stop_block: Optional[int] = None,
+        search_topics: Optional[Dict[str, Any]] = None,
+        extra_addresses: Optional[List] = None,
```
fubuloubu, line 389: Hmm, was this in Banteg's version? Not sure I like this, would rather just compose multiple `.query` outputs

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
...
+              built-in Python function.
+            stop_block (Optional[int]): The block number to stop before. Also the total
+              number of blocks to get. If not setting a start value, is set by
+              the first argument.
+            search_topics (Optional[Dict]): Search topics, such as indexed event inputs,
+              to query by. Defaults to getting all events.
+            extra_addresses (Optional[List[``AddressType``]]): Additional contract
+              addresses containing the same event type. Defaults to only looking at
+              the contract instance where this event is defined.
+            engine_to_use (Optional[str]): query engine to use, bypasses query
```
unparalleled-js, line 415: Can we document the default behavior here when you don't specify which query engine to use?

src/ape/api/query.py
fubuloubu, line 15: I don't see this anywhere....

src/ape/api/query.py
```diff
@@ -12,6 +12,7 @@
     "AccountTransactionQuery",
     "ContractEventQuery",
```
fubuloubu, line 13: Was thinking this is the query type to use for querying events through the query manager

NOTE: We can make changes to this type

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
...
+        """
+
+        if start_block < 0:
+            start_block = len(self) + start_block
+
+        if stop_block is None:
+            stop_block = self.chain_manager.blocks.height
+
+        elif stop_block < 0:
+            stop_block = len(self) + stop_block
```
unparalleled-js, line 429: I think `len(self)` is supposed to be `len(self.chain_manager.block)` or something?

`__len__` in this class is defined as:

```python
    def __len__(self):
        logs = self.provider.get_contract_logs(self.log_filter)
        return sum(1 for _ in logs)
```

This is not related to stop block much!

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
...
+
+        elif stop_block > len(self):
```
unparalleled-js, line 431: same here re: len(self)

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
...
+            raise AttributeError(
+                f"'stop={stop_block}' cannot be greater than the "
+                f"chain length ({self.chain_manager.blocks.height})."
+            )
+
+        addresses = [self.contract.address] + (extra_addresses or [])
```
unparalleled-js, line 437: we should remove duplicates, e.g. `addresses = list(set([self.contract.address] + (extra_addresses or [])))`

src/ape/contracts/base.py
```diff
@@ -381,13 +381,78 @@ def __len__(self):
...
+        yield from cast(Iterator[ContractEvent], contract_events)
+
     def range(
         self,
         start_or_stop: int,
         stop: Optional[int] = None,
         search_topics: Optional[Dict[str, Any]] = None,
         extra_addresses: Optional[List] = None,
-    ) -> Iterator[ContractLog]:
+        engine_to_use: Optional[str] = None,
```
unparalleled-js, line 454: `engine_to_use` is missing from the argument's doc strs

src/ape/managers/query.py
```diff
@@ -28,6 +29,11 @@ def estimate_block_transaction_query(self, query: BlockTransactionQuery) -> int:
 
         return 100
 
+    @estimate_query.register
+    def estimate_contract_events_query(self, query: LogFilter) -> int:
+        # NOTE: Very loos estimate of 100ms per block for this query.
```
unparalleled-js, line 34:
```suggestion
        # NOTE: Very loose estimate of 100ms per block for this query.
```

src/ape/managers/query.py
```diff
@@ -28,6 +29,11 @@ def estimate_block_transaction_query(self, query: BlockTransactionQuery) -> int:
...
+        return (query.stop_block - query.start_block) * 100  # type: ignore
```
unparalleled-js, line 35: Would it make more sense to estimate per log instead of per block?

src/ape_console/_cli.py
```diff
@@ -123,7 +123,7 @@ def console(project=None, verbose=None, extra_locals=None):
     if console_extras:
         namespace.update(console_extras)
 
-    from traitlets.config import Config
+    from traitlets.config import Config  # type: ignore
```
unparalleled-js, line 126:
```suggestion
    from traitlets.config.loader import Config
```

I don't think the `type: ignore` is needed as this library publishes typing. However, I think we need to import this class from the correct spot; that is perhaps why Mypy is complaining.

tests/functional/test_query.py
```diff
@@ -47,6 +50,17 @@ def test_block_transaction_query(eth_tester_provider, sender, receiver):
     assert query[0].chain_id == 61
 
 
+def test_transaction_contract_event_query(contract_instance, owner, eth_tester_provider):
+    contract_instance.fooAndBar(sender=owner)
+    time.sleep(0.1)
+    event = list(contract_instance.FooHappened.query(start_block=-1))
```
unparalleled-js, line 56:
```suggestion
    events = list(contract_instance.FooHappened.query(start_block=-1))
```

since it is a list

tests/functional/test_query.py
```diff
@@ -47,6 +50,17 @@ def test_block_transaction_query(eth_tester_provider, sender, receiver):
...
+    assert isinstance(event[0], ContractLog)
+    assert "FooHappened" in event[0].dict().values()
+    event = list(contract_instance.FooHappened.range(start_or_stop=3))
```
unparalleled-js, line 59: same here re: events over event

unparalleled-js (commented):
I think I found some issues.

But this is really great!

src/ape/api/query.py
fubuloubu: I think it might have been something that banteg defined, and it should be using `ContractEventQuery` instead

unparalleled-js: Do we need this line of code though?

fubuloubu: I think not, I think it was copy-paste from Banteg's POC and we need to re-align to using `ContractEventQuery` instead

src/ape/managers/query.py
```diff
@@ -45,6 +51,21 @@ def perform_block_query(self, query: BlockQuery) -> Iterator:
     def perform_block_transaction_query(self, query: BlockTransactionQuery) -> Iterator:
         return self.provider.get_transactions_by_block(query.block_id)
 
+    @perform_query.register
+    def perform_contract_events_query(self, query: ContractEventQuery):
+        addresses = query.contract
+        if not isinstance(addresses, list):
+            addresses = [query.contract]  # type: ignore
```
johnson2427, line 58: mypy issue here as well, not sure how to remedy this at the moment

johnson2427 (commented):
Just a little push from the team should get this thing to merge

src/ape/contracts/base.py
```diff
@@ -381,13 +383,54 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def query(
+        self,
+        *columns: List[str],
+        block: int = -1,
```
fubuloubu, line 389: Should be `start_block` and `stop_block` because this queries over blocks

src/ape/contracts/base.py
```diff
@@ -381,13 +383,54 @@ def __len__(self):
...
+        start_block = stop_block - 1
+
+        contract_event_query = ContractEventQuery(
+            columns=columns,
+            contract=self.contract.address,
+            event=self.abi,
+            start_block=start_block,
+            stop_block=stop_block,
+        )
+        contract_events = list(
```
fubuloubu, line 420: If you don't convert this to a list, this will be faster when converting to a data frame

src/ape/contracts/base.py
```diff
@@ -381,13 +383,54 @@ def __len__(self):
...
+            self.query_manager.query(contract_event_query, engine_to_use=engine_to_use)
+        )
+        return pd.DataFrame(
+            columns=contract_events[0].dict().keys(), data=[val.dict() for val in contract_events]
```
fubuloubu, line 424: `columns` can be set to `contract_event_query.columns`

src/ape/contracts/base.py
```diff
@@ -381,13 +383,54 @@ def __len__(self):
...
+        )
+
     def range(
         self,
         start_or_stop: int,
         stop: Optional[int] = None,
         search_topics: Optional[Dict[str, Any]] = None,
         extra_addresses: Optional[List] = None,
-    ) -> Iterator[ContractLog]:
```
fubuloubu, line 390: Why remove this type hint?

fubuloubu (commented):
Looking pretty good once these are addressed

src/ape/contracts/base.py
```diff
@@ -419,15 +480,16 @@ def range(
...
+            columns=list(ContractLog.__fields__.keys()),
+            contract=addresses,
             event=self.abi,
             search_topics=search_topics,
-            addresses=addresses,
             start_block=start_block,
             stop_block=stop_block,
         )
-        yield from self.provider.get_contract_logs(log_filter)
+        yield from self.query_manager.query(contract_event_query)  # type: ignore
```
johnson2427, line 492: I removed the type hint above because this complains about the QueryAPI

fubuloubu:
@johnson2427 you merged in [a694c2c](https://github.com/ApeWorX/ape/pull/923/commits/a694c2c53205bfcb7d75a05a2545666cf3b8e786) by accident

src/ape/contracts/base.py
```diff
@@ -381,6 +383,65 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def query(
+        self,
+        *columns: List[str],
+        start_block: int = 0,
+        stop_block: int = None,
```
fubuloubu, line 390:
```suggestion
        stop_block: Optional[int] = None,
```

fubuloubu (approved):
This is solid! Let's go!

fubuloubu (approved):

PR #930, opened 2022-07-31
feat: added optional waiting for tx to finish

PatrickAlphaC:
### What I did
I added a field that you could optionally populate to wait for a tx.

Now, you could be able to do all the following:

```python
receipt = <some_tx_here>

receipt.await_confirmations(0)
receipt.await_confirmations(confirmations=0)
receipt.await_confirmations(0)
```

### How I did it
Basically, just modified the `await_confirmations` to take an optional input parameter. 

### How to verify it
```python
tx = some_contract.some_function_call(sender=account)
tx.await_confirmations(5)
```

### Checklist

### Notes

I'm not sure the best way to test confirmations... I've tested on some different chains that the numbers do work, but I'm not sure how to spin up a new local environment in the tests here... Looking for any help/thoughts.

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -238,7 +238,7 @@ class ReceiptAPI(BaseInterfaceModel):
         :class:`~api.providers.TransactionStatusEnum`.
         """
 
-    def await_confirmations(self) -> "ReceiptAPI":
+    def await_confirmations(self, confirmations: Optional[int] = None) -> "ReceiptAPI":
         """
         Wait for a transaction to be considered confirmed.
 
@@ -251,7 +251,7 @@ class ReceiptAPI(BaseInterfaceModel):
         except TransactionError:
             # Skip waiting for confirmations when the transaction has failed.
             return self
-
+        breakpoint()
         # Wait for nonce from provider to increment.
         sender_nonce = self.provider.get_nonce(self.sender)
         iterations_timeout = 20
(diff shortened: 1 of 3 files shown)


src/ape/api/transactions.py
```diff
@@ -251,7 +251,7 @@ def await_confirmations(self) -> "ReceiptAPI":
         except TransactionError:
             # Skip waiting for confirmations when the transaction has failed.
             return self
-
+        breakpoint()
```
fubuloubu, line 254: Oopsie

src/ape/api/transactions.py
```diff
@@ -238,7 +238,7 @@ def raise_for_status(self):
         :class:`~api.providers.TransactionStatusEnum`.
         """
 
-    def await_confirmations(self) -> "ReceiptAPI":
+    def await_confirmations(self, confirmations: Optional[int] = None) -> "ReceiptAPI":
```
fubuloubu, line 241: Maybe calling it `wait_for_confirmation(...) ` would be nicer?

fubuloubu (commented):
I like the overall idea, I think it needs some UX work though. Currently, all txns are awaited, but I agree that should be more configurable

src/ape/api/transactions.py
PatrickAlphaC:
```suggestion
```

src/ape/api/transactions.py
PatrickAlphaC: I'm down. Maybe it makes sense to make it it's own func? 

It is a little awkward right now... You'll get two progress bars if you do something like this:

```
receipt = <some_tx_here>

receipt.await_confirmations(5)
```

Cuz it calls this func the first time you call the tx, and then you call it again with `receipt.await_confirmations(5)`

PatrickAlphaC: Are you saying to have two functions? Or refactor this functions name.

Also... I like the name `wait` better, since it's less words. Maybe have that be an optional name for the func?

fubuloubu: yeah, I think it's awkward right now because it needs to be refactored into the original not waiting on confirmations (no progress bar) and the `wait_for_confirmation(num_confirmations=default)` function displaying it instead

PatrickAlphaC: Ahhh yeah that makes sense. I might not have bandwidth to make these fixes :/

unparalleled-js:
I know there is a lot of convo here, but we have `timeout=0` and `required_confirmations=x` and plenty of other tickets around improving receipts.

Wondering if we could close this PR.

antazoey:
Closing as is stale and sounds like we are looking at a bigger refactor

PR #933, opened 2022-08-02
fix: issue preventing provider context manager from switching in test fixtures

unparalleled-js:
### What I did

Fixes issue where the provider context as not properly handling multiple providers.

### How I did it

* Don't disconnect providers until very end of test run
* Manage a stack of provider IDs to keep track of the order of provider to pop / push... However! If the provider was previously added, it is already connected and remains so, only the stack ID changes.

### How to verify it

See these tests: https://github.com/unparalleled-js/ape-demo-starknet/tree/main/tests

### Checklist

### DIFF
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -225,24 +225,25 @@ The easiest way to achieve this is to use the `networks` provider context-manage
 
 ```python
 # Switch to Fantom mid test
-from ape import networks
-
-
-with networks.fantom.local.use_provider("test") as provider:
-    print(f"Using provider {provider.name}")
-
-""" or """
-
-with networks.parse_network_choice("fantom:local:test") as provider:
-    print(f"Using provider {provider.name}")
+def test_my_fantom_test(networks):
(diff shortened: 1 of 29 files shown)


docs/userguides/testing.md
```diff
@@ -217,3 +217,45 @@ def test_account_balance(project, owner, receiver, nft):
...
+```
+
+Use fixtures to assist as well:
+
+```python
+import pytest
+
+
+@pytest.fixture
+def stark_contract(networks, project):
```
fubuloubu, line 247: what happens if you use a provider with two different network contexts in a test?

fubuloubu (approved):
One deeper question, but otherwise looks good

docs/userguides/testing.md
unparalleled-js: as long as the chain ID is different, it will establish a new connection and maintain that connection throughout the session.

fubuloubu (dismissed):

ca11ab1e (approved):

fubuloubu (approved):

PR #940, opened 2022-08-05
refactor: change deployments_map contract type entries to be list of dicts instead of list of addresses

unparalleled-js:
### What I did

Changes the `deployments_map.json` entries to be a list of dictionaries rather than a list of addresses.
This will allow us to add more data to deployments, such as the `txn_hash`, which my be needed.

### How I did it

* When getting deployments, if it notices you are using the old schema, it will migrate it then
* I had to do some refactoring to achieve this nicely

### How to verify it

Case 0:
Deploy a contract to a testnet using main branch
Then, deploy a contract here. Make sure it worked and uses the new schema

Case 1:
Deploy a contract to a testnet using main branch
Call `get_deployments` here. Make sure it worked and uses the new schema.

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -438,13 +438,20 @@ class ContractCache(BaseManager):
         return self.provider.network
 
     @property
+    def _ecosystem_name(self) -> str:
+        return self._network.ecosystem.name
+
+    @property
     def _is_live_network(self) -> bool:
         return self._network.name != LOCAL_NETWORK_NAME and not self._network.name.endswith("-fork")
 
     @property
+    def _data_network_name(self) -> str:
+        return self._network.name.replace("-fork", "")
+
+    @property
     def _contract_types_cache(self) -> Path:
-        network_name = self._network.name.replace("-fork", "")
-        return self._network.ecosystem.data_folder / network_name / "contract_types"
+        return self._network.ecosystem.data_folder / self._data_network_name / "contract_types"
 
     @property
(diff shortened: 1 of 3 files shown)


src/ape/managers/chain.py
```diff
@@ -693,35 +696,45 @@ def get_deployments(self, contract_container: ContractContainer) -> List[Contrac
...
+            else self._load_deployments_cache()
+        )
+        ecosystem_map = deployments_map.get(self._ecosystem_name, {})
+        network_map = ecosystem_map.get(self._data_network_name, {})
+        deployments = network_map.get(contract_name)
+        if not deployments:
+            return []
+
+        if isinstance(deployments[0], str):
+            # Handle previous schema of list of addresses instead of list of dicts.
```
fubuloubu, line 716: Probably add a TODO to deprecate this in like 0.6.x or something

fubuloubu (dismissed):

src/ape/managers/chain.py
unparalleled-js: I also added it to my personal log of breaking changes I want (categorized by planned versions)

unparalleled-js: resolved by https://github.com/ApeWorX/ape/pull/940/commits/901e8cf77ee5ea1f3496cb1d82ce473c97b8c6bd

unparalleled-js:
The failing build makes little sense to me at this time!

fubuloubu (approved):

PR #943, opened 2022-08-05
fix: issue where web3.py used the wrong chain ID for eth-tester

unparalleled-js:
### What I did

@Ninjagod1251 experienced a signature error in a vyper contract a while back that we have ascribed to web3.py having the wrong chain ID. This got fixed in the web3.py 6.0.0 range (WHICH WE ARE FINALLY ABLE TO UPGRADE TO!!!!!) so the fix was to upgrade all eth-prefixed dependencies. Yay!

### How I did it

Upgrade Ethereum dependecies

### How to verify it
everything still works.
the chain ID for eth-tester should now be 131277322940537

### Checklist


setup.py
```diff
@@ -93,34 +93,45 @@
...
-        "eth-account==0.5.7",
-        "ijson>=3.1.4",
+        # ** Dependencies maintained by Ethereum Foundation **
+        "eth-abi>=3.0.1,<4",
+        "eth-account>=0.6.1,<0.7",
+        "eth-rlp>=0.3.0,<0.4",
+        "eth-utils>=2.0.0,<3",
+        "hexbytes>=0.2.2,<1",
+        "py-geth>=3.8.0,<4",
+        "web3[tester]>=6.0.0b4,<7",
```
unparalleled-js, line 130: fix to the web is because of this upgrade

setup.py
```diff
@@ -93,34 +93,45 @@
...
+        "packaging>=20.9,<21",
+        "pandas>=1.3.0,<2",
+        "pluggy>=0.13.1,<1",
+        "pydantic>=1.9.0,<2",
+        "PyGithub>=1.54,<2",
+        "pygit2>=1.7.2,<2",
+        "pytest>=6.0,<8",
+        "python-dateutil>=2.8.2,<3",
+        "pyyaml>=6.0,<7",
+        "requests>=2.25.1,<3",
```
fubuloubu, line 118: How many of these are required dependencies?

fubuloubu (dismissed):

setup.py
unparalleled-js: AFAIK, all of them in this `install_requires` list are required dependencies.
Note: I did not add any dependencies in this PR but I reorganized them such that the Ethereum foundation and Apeworx dependencies are grouped. The remaining dependencies are alphabetized so it is easier to lookup them up (as the list is quite long).

fubuloubu: Maybe... create a ticket to review dependencies lol

unparalleled-js: I will do it now instead:

Packages removes:

* duplicate `hexbytes` entry
* `importlib-metadata` - not sure if this is needed, but it is part of the template. Will put back if you say it's needed or the build fails.
* `eth-rlp`: didn't see it being used anywhere so removed it

Packages kept and reason for their usage:

* `backports.cached_property`: used wherever `@cached_property` is used
* `click`: The CLI framework we use
* `ijson`: Used by `ape-geth` for streaming logs
* `IPython`: The lib that powers `ape console`
* `importlib-metadata`: Required until we no longer support 3.7
* `packaging`: Used for parsing versions in the package manifest
* `pandas:` Used for querying
* `pluggy`: Used for plugin system
* `pydantic`: Used for API classes
* `PyGithub`: Used to talk to Github for downloading dependencies or installing plugins
* `pygit2`: Used as a `git` wrapper for installing dependencies
* `pyyaml`: Used to parse config file
* `requests`: Used in `ape-geth` for request streaming as well as checking off exceptions
* `rich`: Used for displaying trees in the terminal
* `singledispatchmethod`: Used for overloading methods with varying arg types
* `tqdm`: Used for CLI progress bars
* `typing-extensions`: Used to assist installing the tool correctly, can remove after python 3.7.
* `eth-abi`: Used for ABI parsing
* `eth-account`: Used for accounts in ape, like verifying signatures
* `eth-utils`: tools for working with Ethereum, used everywhere
* `hexbytes`: class for working with bytes data types in Ethereum
* `py-geth`: used for the ephemeral local geth process
* `web3[tester]`: Used as Ethereum client
* `eip712`: Uses a type when checking the signature for compatibility purposes with the library
* `ethpm-types`: Used for contract types and package manifest stuff
* `evm-trace`: Used for transaction tracing

NotPeopling2day (approved):
LGTM

setup.py
fubuloubu: perfect! thank you!

src/ape/api/transactions.py
```diff
@@ -30,6 +30,7 @@ class TransactionAPI(BaseInterfaceModel):
     receiver: Optional[str] = Field(None, alias="to")
     sender: Optional[str] = Field(None, alias="from")
     gas_limit: Optional[int] = Field(None, alias="gas")
+    gas_price: Optional[int] = Field(None, alias="gasPrice")
```
fubuloubu, line 33: This is a somewhat significant change here. I think it wasn't originally here because not all transaction dicts have this field, so it was open to subclasses to define extra elements they have and compute some sort of "total cost" property (that I don't see around)

fubuloubu (commented):
Besides this one comment, I approve

NotPeopling2day (approved):
LGTM

PR #954, opened 2022-08-08
test: update test_parse_type() regex

helloibis:
### What I did
The previous regex was allowing some invalid inputs such as `(,)` to get through.

### How I did it
Updated the regular expression to match expected Vyper decls like:
- type identifiers
- array declarations
- tuples of type/array
- tuples of type/array/tuple/etc

I split the various expected sub-patterns to their own variables so they're easier to read and can be commented on separately

### How to verify it

__NOTE:__ I used non-capturing groups (?:) on the regex site just to make the highlight colors a little less chaotic 👀 the groups in the Python code don't need to be non-capturing so I omitted them there

The regular expression (i.e. the value of `TYPE_DECL`) and some sample input are here:

https://regex101.com/r/dVmiYA/1

Screenshot for posterity, the non-highlighted text below shows invalid input that we don't want to use anyway:

image

### Checklist

### DIFF
--- a/tests/functional/utils/test_abi.py
+++ b/tests/functional/utils/test_abi.py
@@ -10,16 +10,16 @@ IDENT = r"(\w+(\[\])?)"
 # e.g. (int,) or (int, int) or (int, int,) etc.
 TUPLE = rf"(\({IDENT},(\s*{IDENT},?)*\))"
 
-# An optionally nested tuple.
+# Adding optional nesting to the tuple pattern.
 # e.g. (int,) or (int, int) or ((int,), int) etc.
-NESTED_TUPLE = rf"(\(({IDENT}|{TUPLE}),(\s*{IDENT}|{TUPLE},?)*\))"
+TUPLE = rf"(\(({IDENT}|{TUPLE}),(\s*({IDENT}|{TUPLE}),?)*\))"
 
 # The full type declaration pattern which matches:
 # - type identifier
 # - array declaration
 # - tuples of type/array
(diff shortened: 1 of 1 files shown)


unparalleled-js (approved):
Looks great!

tests/functional/utils/test_abi.py
```diff
@@ -3,18 +3,46 @@
...
+    # Examples matching strings from above regex:
+    #   - int
+    #   - int[]
+    #   - uint256
+    #   - (int,)
+    #   - (int[],)
+    #   - (int, int,)
+    #   - (int, int)
+    #   - (int, int, int,)
+    #   - (int, int, int,)
```
NotPeopling2day, line 37: Are these duplicates, or maybe missed removing the comma?

helloibis:
Updated 54a8bc14 just because we don't need a separate `NESTED_TUPLE` pattern, just need to redefine `TUPLE` and nest its existing value inside of it. That helps reduce the total length of the regex which should give hypothesis a little bit of a break lol

@NotPeopling2day I fixed the docstring in 1b16442c

I also updated the description to reflect the consolidated regex

tests/functional/utils/test_abi.py
helloibis: Updated in 1b16442c

unparalleled-js (dismissed):

NotPeopling2day (dismissed):
LGTM

NotPeopling2day (approved):

Issue #251, opened 2021-11-19
Support dev revert strings


unparalleled-js:
### Overview

"this feature requires transaction tracing in order to see what line of code the failure occurred on. could also just show the failed line of code directly with the assert, but the dev revert string is useful for coverage testing"

### Specification

```python
with ape.reverts(dev_message="This is a dev message"):
   ...
```

### Dependencies

Look at how `brownie` did it as an example!

helloibis:
@unparalleled-js @fubuloubu @NotPeopling2day is `dev_message` intended to be mutually exclusive with the existing `expected_message`?

Also Bryant had mentioned allowing regex in the kwargs for the message, I assume that also applies to the existing `expected_message`? If so I can update this ticket to include that requirement for both

fubuloubu:
> @unparalleled-js @fubuloubu @NotPeopling2day is `dev_message` intended to be mutually exclusive with the existing `expected_message`?

Yes, mutually exclusive checks

> Also Bryant had mentioned allowing regex in the kwargs for the message, I assume that also applies to the existing `expected_message`? If so I can update this ticket to include that requirement for both

No, please make another feature ticket for the regex support. Yes it should work for both.

helloibis:
Status update: once https://github.com/ApeWorX/ape/issues/937 is completed this will be able to be finished as well

helloibis:
Completed via https://github.com/ApeWorX/ape/pull/1125

PR #957, opened 2022-08-09
feat!: allow transaction raise on fail

helloibis:
### What I did

Begin adding in parameters to facilitate explicitly raising on transaction failure. We'll need transactions that fail to deliberately raise an exception for `RevertsContextManager`. That functionality is a dependency for the tracing ticket #251.

### How I did it

Adding `raise_on_fail: bool` parameter to the following:
- `ReceiptAPI.await_confirmations()` (default `False`)
- `ProviderAPI.get_receipt()` (default `False`)
- `ProviderAPI.send_transaction()` (default `True`)
- `AccountAPI.call()` (default `True`)
- `ImpersonatedAccount.call()` (default `True`)
- `Web3Provider.send_transaction()` (default `True`)
- `ContractTransaction.__call__()` (default `True`)
- `LocalProvider.send_transaction()` (default `True`)

### How to verify it

Added a test against flipping this parameter when calling the contract function to ensure that the transaction causes an exception rather than just returning a failed receipt

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -77,7 +77,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         self,
         txn: TransactionAPI,
         send_everything: bool = False,
-        raise_on_fail: bool = False,
+        raise_on_fail: bool = True,
     ) -> ReceiptAPI:
         """
         Make a transaction call.
@@ -437,7 +437,7 @@ class ImpersonatedAccount(AccountAPI):
         self,
         txn: TransactionAPI,
         send_everything: bool = False,
-        raise_on_fail: bool = False,
+        raise_on_fail: bool = True,
     ) -> ReceiptAPI:
         txn = self.prepare_transaction(txn)
         return self.provider.send_transaction(txn, raise_on_fail=raise_on_fail)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -73,19 +73,27 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
           :class:`~ape.types.signatures.TransactionSignature` (optional): The signed transaction.
         """
 
-    def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
+    def call(
+        self,
+        txn: TransactionAPI,
+        send_everything: bool = False,
+        raise_on_fail: bool = False,
+    ) -> ReceiptAPI:
         """
         Make a transaction call.
 
         Raises:
             :class:`~ape.exceptions.AccountsError`: When the nonce is invalid or the sender does
               not have enough funds.
-            :class:`~ape.exceptions.TransactionError`: When the required confirmations are negative.
+            :class:`~ape.exceptions.TransactionError`: When the required confirmations
+              are negative or on transaction failure if ``raise_on_fail`` is ``True``.
             :class:`~ape.exceptions.SignatureError`: When the user does not sign the transaction.
 
         Args:
             txn (:class:`~ape.api.transactions.TransactionAPI`): An invoke-transaction.
             send_everything (bool): ``True`` will send the difference from balance and fee.
+            raise_on_fail (bool): ``True`` will cause failed transactions to raise
+              `~ape.exceptions.TransactionError`.
 
         Returns:
             :class:`~ape.api.transactions.ReceiptAPI`
@@ -109,7 +117,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         if not txn.signature:
             raise SignatureError("The transaction was not signed.")
 
-        return self.provider.send_transaction(txn)
+        return self.provider.send_transaction(txn, raise_on_fail=raise_on_fail)
 
     @cached_property
     def _convert(self) -> Callable:
@@ -425,6 +433,11 @@ class ImpersonatedAccount(AccountAPI):
     def sign_transaction(self, txn: TransactionAPI) -> Optional[TransactionSignature]:
         return None
 
-    def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
+    def call(
+        self,
+        txn: TransactionAPI,
+        send_everything: bool = False,
+        raise_on_fail: bool = False,
+    ) -> ReceiptAPI:
         txn = self.prepare_transaction(txn)
-        return self.provider.send_transaction(txn)
+        return self.provider.send_transaction(txn, raise_on_fail=raise_on_fail)
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -331,7 +331,7 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = False) -> ReceiptAPI:
+    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = True) -> ReceiptAPI:
         """
         Send a transaction to the network.
 
@@ -840,7 +840,7 @@ class Web3Provider(ProviderAPI, ABC):
         txn_hash: str,
         required_confirmations: int = 0,
         timeout: Optional[int] = None,
-        raise_on_fail: bool = False,
+        raise_on_fail: bool = True,
     ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
@@ -933,7 +933,7 @@ class Web3Provider(ProviderAPI, ABC):
 
         return self._make_request("eth_getLogs", [filter_params])
 
-    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = False) -> ReceiptAPI:
+    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = True) -> ReceiptAPI:
         try:
             txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
         except ValueError as err:
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -830,7 +830,11 @@ class Web3Provider(ProviderAPI, ABC):
             raise self.get_virtual_machine_error(err) from err
 
     def get_receipt(
-        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None
+        self,
+        txn_hash: str,
+        required_confirmations: int = 0,
+        timeout: Optional[int] = None,
+        raise_on_fail: bool = True,
     ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
@@ -874,7 +878,8 @@ class Web3Provider(ProviderAPI, ABC):
                 **receipt_data,
             }
         )
-        return receipt.await_confirmations()
+
+        return receipt.await_confirmations(raise_on_fail=raise_on_fail)
 
     def get_transactions_by_block(self, block_id: BlockID) -> Iterator:
         if isinstance(block_id, str):
@@ -934,10 +939,18 @@ class Web3Provider(ProviderAPI, ABC):
             else self.network.required_confirmations
         )
 
-        receipt = self.get_receipt(txn_hash.hex(), required_confirmations=required_confirmations)
-        receipt.raise_for_status()
+        receipt = self.get_receipt(
+            txn_hash.hex(),
+            required_confirmations=required_confirmations,
+            raise_on_fail=txn._raise_on_fail,
+        )
+
+        if txn._raise_on_fail:
+            receipt.raise_for_status()
+
         logger.info(f"Confirmed {receipt.txn_hash} (total fees paid = {receipt.total_fees_paid})")
         self._try_track_receipt(receipt)
+
         return receipt
 
     def _make_request(self, endpoint: str, parameters: List) -> Any:
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -331,12 +331,18 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
+    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = False) -> ReceiptAPI:
         """
         Send a transaction to the network.
 
+        Raises:
+            :class:`~ape.exceptions.TransactionError`: When the transaction fails
+              if ``raise_on_fail`` is ``True``.
+
         Args:
             txn (:class:`~ape.api.transactions.TransactionAPI`): The transaction to send.
+            raise_on_fail (bool): ``True`` will cause failed transactions to raise
+              `~ape.exceptions.TransactionError`.
 
         Returns:
             :class:`~ape.api.transactions.ReceiptAPI`
@@ -834,7 +840,7 @@ class Web3Provider(ProviderAPI, ABC):
         txn_hash: str,
         required_confirmations: int = 0,
         timeout: Optional[int] = None,
-        raise_on_fail: bool = True,
+        raise_on_fail: bool = False,
     ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
@@ -927,7 +933,7 @@ class Web3Provider(ProviderAPI, ABC):
 
         return self._make_request("eth_getLogs", [filter_params])
 
-    def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
+    def send_transaction(self, txn: TransactionAPI, raise_on_fail: bool = False) -> ReceiptAPI:
         try:
             txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
         except ValueError as err:
@@ -942,10 +948,10 @@ class Web3Provider(ProviderAPI, ABC):
         receipt = self.get_receipt(
             txn_hash.hex(),
             required_confirmations=required_confirmations,
-            raise_on_fail=txn._raise_on_fail,
+            raise_on_fail=raise_on_fail,
         )
 
-        if txn._raise_on_fail:
+        if raise_on_fail:
             receipt.raise_for_status()
 
         logger.info(f"Confirmed {receipt.txn_hash} (total fees paid = {receipt.total_fees_paid})")
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -342,7 +342,7 @@ class ProviderAPI(BaseInterfaceModel):
         Args:
             txn (:class:`~ape.api.transactions.TransactionAPI`): The transaction to send.
             raise_on_fail (bool): ``True`` will cause failed transactions to raise
-              `~ape.exceptions.TransactionError`.
+              `~ape.exceptions.TransactionError`.  Defaults to ``True``.
 
         Returns:
             :class:`~ape.api.transactions.ReceiptAPI`
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -840,7 +840,7 @@ class Web3Provider(ProviderAPI, ABC):
         txn_hash: str,
         required_confirmations: int = 0,
         timeout: Optional[int] = None,
-        raise_on_fail: bool = True,
+        raise_on_fail: bool = False,
     ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
@@ -851,6 +851,8 @@ class Web3Provider(ProviderAPI, ABC):
               to wait before returning the receipt.
             timeout (Optional[int]): The amount of time to wait for a receipt
               before timing out.
+            raise_on_fail (bool): Whether an exception should be raised if the
+              transaction failed.
 
         Raises:
             :class:`~ape.exceptions.TransactionNotFoundError`: Likely the exception raised
@@ -948,7 +950,6 @@ class Web3Provider(ProviderAPI, ABC):
         receipt = self.get_receipt(
             txn_hash.hex(),
             required_confirmations=required_confirmations,
-            raise_on_fail=raise_on_fail,
         )
 
         if raise_on_fail:
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -852,7 +852,7 @@ class Web3Provider(ProviderAPI, ABC):
             timeout (Optional[int]): The amount of time to wait for a receipt
               before timing out.
             raise_on_fail (bool): Whether an exception should be raised if the
-              transaction failed.
+              transaction failed. Defaults to ``False``.
 
         Raises:
             :class:`~ape.exceptions.TransactionNotFoundError`: Likely the exception raised
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -41,9 +41,6 @@ class TransactionAPI(BaseInterfaceModel):
     # If left as None, will get set to the network's default required confirmations.
     required_confirmations: Optional[int] = Field(None, exclude=True)
 
-    # If true, will cause a TransactionError to be raised if the transaction fails.
-    _raise_on_fail: bool = False
-
     signature: Optional[TransactionSignature] = Field(exclude=True)
 
     class Config:
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -254,8 +254,8 @@ class ReceiptAPI(BaseInterfaceModel):
         Wait for a transaction to be considered confirmed.
 
         Args:
-            raise_on_fail (bool): If true, causes a TransactionError to be
-                                  raised if the transaction has failed.
+            raise_on_fail (bool): If ``True``, causes a :class:`~ape.exceptions.TransactionError`
+              to be raised if the transaction has failed.
 
         Returns:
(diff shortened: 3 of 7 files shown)


helloibis:
cc @NotPeopling2day @fubuloubu @unparalleled-js 

Hey all, this is just a draft because I have a few questions:

### Test Environment Txn Failures

I think previously we talked about having a transaction `raise_on_fail` flag's default change based on whether we're in a test suite or not (not just pytest, but `ape test` for projects). What are everyone's thoughts on that?

### Transaction Field / Pydantic Issues

Currently I'm running issues with how to actually define this `TransactionAPI.raise_on_fail` field I'm adding.

#### `Field(..., exclude=True)`

If I go about it the normal Pydantic way:

```python
class TransactionAPI(...):
	...

	raise_on_fail = Field(False, exclude=True)
```

Then I actually get the field when serializing transactions:

```python
>>> txn = ...  # let's say I have a transaction, like in the updated test
>>> "raise_on_fail" in txn.dict()
True
```

If I change the definition to `Field(None, exclude=True)` then I don't run into the issue with serialization but something about that still seems off.

Additionally, there are functions like `src/ape/contracts/base.py::_convert_kwargs()` which use `TransactionAPI.__fields__`. 
Cases like that would need to be updated to exclude this one-off field specifically, and I think I was still having issues after that but would need to try those changes again.

#### Private field

Prefixing the field with an underscore "hides" it from Pydantic entirely, but then there's the inverse issue of functions like `as_transaction()` or `serialize_transaction()` not working with attributes that aren't Pydantic fields.

What are everyone's thoughts? I didn't want to go start changing up `serialize_transaction()` calls since there seem to be several and I'm not sure if that's really the right approach anyway.

Thanks!

tests/functional/test_receipt.py
```diff
@@ -116,12 +116,22 @@ def test_decode_logs_unspecified_abi_gets_all_logs(owner, contract_instance):
 def test_get_failed_receipt(owner, vyper_contract_instance, eth_tester_provider):
     # Setting to '5' always fails.
     transaction = vyper_contract_instance.setNumber.as_transaction(
-        5, sender=owner, gas_limit=100000
+        5, sender=owner, gas_limit=100000, _raise_on_fail=False
```
helloibis, line 119: Specifically here is where the field doesn't work if it's not Pydantic. Not that it necessarily has to be, but it seems like it means updating all the functions that create a serialization from `**kwargs`?

src/ape/api/transactions.py
```diff
@@ -40,6 +40,9 @@ class TransactionAPI(BaseInterfaceModel):
     # If left as None, will get set to the network's default required confirmations.
     required_confirmations: Optional[int] = Field(None, exclude=True)
 
+    # If true, will cause a TransactionError to be raised if the transaction fails.
+    _raise_on_fail: bool = False
```
helloibis, line 44: This value will need to remain `False` (or `None`) by default for this not to be a breaking change, but we did have a bit of discussion last week around potentially making it `True` during tests. Is that viable or would that cause its own issues?

fubuloubu: I think making it true during tests might make it work the same as people expect

fubuloubu (dismissed):

src/ape/api/transactions.py
unparalleled-js: hmm, it must already raise on tests somewhere or else `ape.reverts()` wouldn't be working

src/ape/api/transactions.py
```diff
@@ -238,19 +241,19 @@ def raise_for_status(self):
         :class:`~api.providers.TransactionStatusEnum`.
         """
 
-    def await_confirmations(self) -> "ReceiptAPI":
+    def await_confirmations(self, raise_on_fail: bool = False) -> "ReceiptAPI":
         """
         Wait for a transaction to be considered confirmed.
 
+        Args:
+            raise_on_fail (bool): If true, causes a TransactionError to be
```
unparalleled-js, line 249:
```suggestion
            raise_on_fail (bool): If ``True``, causes a `:class:~ape.exceptions.TransactionError` to be
```

src/ape/api/providers.py
```diff
@@ -716,7 +716,11 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:
             raise self.get_virtual_machine_error(err) from err
 
     def get_transaction(
-        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None
+        self,
+        txn_hash: str,
+        required_confirmations: int = 0,
+        timeout: Optional[int] = None,
+        raise_on_fail: bool = True,
```
unparalleled-js, line 723: Should it be `False` here so we can get failed receipts easily and then have it get set to `True` during contract invocations / making transactions?

tests/functional/test_receipt.py
```diff
@@ -116,12 +116,22 @@ def test_decode_logs_unspecified_abi_gets_all_logs(owner, contract_instance):
...
     assert receipt.failed
 
 
+def test_get_failed_receipt_raise(owner, vyper_contract_instance, eth_tester_provider):
+    # Setting to '5' always fails.
+    transaction = vyper_contract_instance.setNumber.as_transaction(
+        5, sender=owner, gas_limit=100000, _raise_on_fail=True
+    )
+
+    with pytest.raises(TransactionError):
```
unparalleled-js, line 131: it would be ideal to have syntax like this work:

```python
with pytest.raises(TransactionError):
    vyper_contract_instance.setNumber(5, sender=owner, gas_limit=100000, raise_on_fail=True)
```

unparalleled-js (commented):
Are we able to avoid adding it to the model if use the same approach that we did for the `block_identifier` and `state_overrides` kwarg handling?

You can find that in `ape.api.providers.Web3Provider` in methods like `estimate_gas_cost()`

helloibis:
> Are we able to avoid adding it to the model if use the same approach that we did for the `block_identifier` and `state_overrides` kwarg handling?

@unparalleled-js yeah I think that would be viable and honestly be a little better in "where" the functionality lives (since really this flag doesn't _need_ to be on a transaction, it just needs to be specified when dealing with transactions)

Is there a preferred way to document those kind of kwargs? I see `block_identifier` coming up in the code but not docstrings or comments

unparalleled-js:
> Is there a preferred way to document those kind of kwargs? I see `block_identifier` coming up in the code but not docstrings or comments

For yours, we may want to consider adding to 0.5 API changes since it seems common enough across ecosystems to want to do this.

For the other kwargs you mentioned, it got me thinking and this what I landed on so far:
https://github.com/ApeWorX/ape/pull/966
and we can do something similar in the meanwhile to avoid having this be a breaking change.

I am open to suggestions though, on that PR.

helloibis:
> Are we able to avoid adding it to the model if use the same approach that we did for the `block_identifier` and `state_overrides` kwarg handling?
> 
> You can find that in `ape.api.providers.Web3Provider` in methods like `estimate_gas_cost()`

Yeah, I think so - I gave it a stab in e2c08e01

In the tests `test_get_failed_receipt()` / `test_get_failed_receipt_raise()` when I make the call it's using `AccountAPI.call()`. From there then that calls down to `ProviderAPI.send_transaction()`

Since those functions only take the `txn` they'd need additional kwargs added if we didn't append the `raise_on_fail` field to `TransactionAPI`. I updated the call signatures to:

```python
AccountAPI.call(self, txn: TransactionAPI, send_everything: bool, raise_on_fail: bool)
ProviderAPI.send_transaction(self, txn: TransactionAPI, raise_on_fail: bool)
```

Those two aforementioned tests now show the usage:

```python
txn = ...

# non-raising call examples
owner.call(txn, raise_on_fail=False)
provider.send_transaction(txn, raise_on_fail=False)

# raising call examples
owner.call(txn, raise_on_fail=True)
provider.send_transaction(txn, raise_on_fail=True)
```

cc @unparalleled-js @NotPeopling2day

src/ape/api/transactions.py
helloibis: Good catch! updated in deeccd8c

tests/functional/test_receipt.py
helloibis: Updated `ContractTransaction.__call__()` in 0e4d4ef3 to achieve this

cc @NotPeopling2day

helloibis:
Rebased off main, some of those commit hashes I just linked might be "old" now haha

helloibis:
With the kwarg being distributed across a few places, is there a preferred method of how we should handle its default value? I think that perhaps it could be a three-way value and we check inside the logic of the different callsites, for example:

```python
def send_transaction(self, ..., raise_on_fail: Optional[bool] = None):
    if raise_on_fail is None and os.environ.get("APE_TESTING", False):
        raise_on_fail = True
```

cc @unparalleled-js @NotPeopling2day

fubuloubu:
Not seeing a way for a contract to pass this parameter
```py
>>> foo.bar(False, sender=dev, raise_on_fail=True)
...

ContractLogicError: Transaction failed.
```

helloibis:
> Not seeing a way for a contract method to pass this parameter
> 
> ```python
> >>> foo.bar(False, sender=dev, raise_on_fail=True)
> ...
> 
> ContractLogicError: Transaction failed.
> ```

@fubuloubu that should be covered by https://github.com/ApeWorX/ape/pull/957/commits/a471f77dda2ffa86271f01d262edcfa69b6b9de7 unless I'm missing something

The test examples define `setNumber()` in the Vyper contract and then can pass `raise_on_fail` when calling it:

```python
with pytest.raises(TransactionError):
    vyper_contract_instance.setNumber(5, sender=owner, gas_limit=100000, raise_on_fail=True)
```

Not passing `raise_on_fail` (or passing `False`) will return a failed receipt instead

unparalleled-js:
I tried using this with the `ape-hardhat` plugin to run tests but I get errors `TypeError: send_transaction() got an unexpected keyword argument 'raise_on_fail'` so I am 🤔 

It did work with `local:test` and `local:geth` though.

helloibis:
> I tried using this with the `ape-hardhat` plugin to run tests but I get errors `TypeError: send_transaction() got an unexpected keyword argument 'raise_on_fail'` so I am 🤔
> 
> It did work with `local:test` and `local:geth` though.

Yeah I think we'll need corresponding PRs for the different provider plugins since those functions never took `**kwargs`, working on updating those now

fubuloubu:
> > Not seeing a way for a contract method to pass this parameter
> > ```python
> > >>> foo.bar(False, sender=dev, raise_on_fail=True)
> > ...
> > 
> > ContractLogicError: Transaction failed.
> > ```
> 
> @fubuloubu that should be covered by [a471f77](https://github.com/ApeWorX/ape/commit/a471f77dda2ffa86271f01d262edcfa69b6b9de7) unless I'm missing something

> Not passing `raise_on_fail` (or passing `False`) will return a failed receipt instead

meant to share when `raise_on_fail` was `False`, because it didn't seem to affect it's behavior

fubuloubu:
overall, I think this might need some good QA over the different circumstances that raise errors

NotPeopling2day:
@helloibis let's setup a call this afternoon and work through some of it

helloibis:
> meant to share when `raise_on_fail` was `False`, because it didn't seem to affect it's behavior

@fubuloubu is there a short example of what you mean? In `test_get_failed_receipt()` I added the `raise_on_fail` flag and flipping it between `True`/`False` causes it to either raise or not raise when the transaction reverts on `assert num != 5`

fubuloubu (approved):

fubuloubu:
will have to do a bit of QA on this

helloibis:
PR is updated and ready for review, I'm getting the corresponding provider plugin PRs ready and adding tests for them

## Current State

Just added 6f799560 to set the default value back to `True`. On our call Friday Bryant and I discussed the behavior of transactions matching the following:

```python
# By default, gas estimation runs (and can raise) and reverted transactions raise an exception (even though they're committed to chain)
>>> receipt = contract.foo(...)
TransactionError

# Setting a gas limit can bypass the estimation, but a reverted transaction will still raise
>>> receipt = contract.foo(..., gas_limit=100000)
TransactionError

# Using `raise_on_fail=False` turns off the behavior. It currently doesn't affect gas estimation but failed receipts do not raise
>>> receipt = contract.foo(..., raise_on_fail=False)
TransactionError    # is gas estimation fails
<ReceiptAPI ...>  # if gas estimation passes
```

## Future State

Later on it should have the following behavior:

```python
# Using `raise_on_fail=False` in the future will also bypass gas estimation. So you can 
>>> receipt = contract.foo(..., raise_on_fail=False)
<ReceiptAPI ...>   # no gas estimation run, always returns receipt

# Inside ape.reverts(), gas estimation always passes but failed receipts still raise
>>> with ape.reverts():
        receipt = contract.foo(...)
TransactionError
```

Because that implementation is different from the kwargs I'm adding, I think it'll be best suited as a separate PR

src/ape/api/providers.py
helloibis: @unparalleled-js talked with Bryant on Friday and we settled on some slightly different behavior, I set the value to true in 6f79956 and added an explanation at https://github.com/ApeWorX/ape/pull/957#issuecomment-1215565267

fubuloubu:
> ## Future State
> Later on it should have the following behavior:
> 
> ```python
...

> # Inside ape.reverts(), gas estimation always passes but failed receipts still raise
> >>> with ape.reverts():
>         receipt = contract.foo(...)
> TransactionError
> ```

Wanted to shout out one difference here, I think in this scenario it should also have the same behavior of bypassing gas estimation (as if `raise_on_fail=False` was added) and yes also raise up the transaction error, as if this were happening behind the scenes:

```py
receipt = contract.foo(..., raise_on_fail=False)
receipt.raise_for_status()
```

src/ape/api/accounts.py
```diff
@@ -73,19 +73,27 @@ def sign_transaction(self, txn: TransactionAPI) -> Optional[TransactionSignature
           :class:`~ape.types.signatures.TransactionSignature` (optional): The signed transaction.
         """
 
-    def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
+    def call(
+        self,
+        txn: TransactionAPI,
+        send_everything: bool = False,
+        raise_on_fail: bool = True,
```
unparalleled-js, line 80: By changing the API method, it will technically be a breaking change.
I was going to see if we could hold off on changing the actual API until we are ready for 0.5... Are API changes required to get the behavior we need? Is there a short-term solution we could implement?

tests/functional/test_receipt.py
```diff
@@ -115,13 +115,16 @@ def test_decode_logs_unspecified_abi_gets_all_logs(owner, contract_instance):
...
-    )
-    receipt = owner.call(transaction)
+    receipt = vyper_contract_instance.setNumber(5, sender=owner, gas_limit=100000)
     assert receipt.failed
 
 
+def test_get_failed_receipt_raise(owner, vyper_contract_instance, eth_tester_provider):
+    # Setting to '5' always fails.
+    with pytest.raises(TransactionError):
+        vyper_contract_instance.setNumber(5, sender=owner, gas_limit=100000, raise_on_fail=True)
```
unparalleled-js, line 125: Where is the receipt here?

unparalleled-js:
I am worried about this (from an ape-console environment):

```python
receipt = contract.foo(..., raise_on_fail=False)
```

^ this is not working form unless I also provider the `gas_limit`. I am worried this will confuse users. I am trying to find out why this. Are not able to allow the `raise_on_fail` to work for gas estimation?

tests/functional/test_receipt.py
helloibis: @unparalleled-js this raises, so it doesn't return

src/ape/api/accounts.py
helloibis: @unparalleled-js The method itself isn't the breaking change is it? Since it provides a default existing calls to it shouldn't be a problem - but the breaking change is the new behavior.

When I spoke with @fubuloubu the way it was described to me was that it should be raising exceptions by default, which is far different from current behavior (can currently raise on gas estimation issues, but doesn't raise if the transaction goes through and ends up reverting)

The way that this wouldn't be breaking is if the default was `raise_on_fail=False`, but only in its current state. When `raise_on_fail` is implemented to account for gas estimation then the default being `False` will change the current behavior as well 👀

helloibis:
> I am worried about this (from an ape-console environment):
> 
> ```python
> receipt = contract.foo(..., raise_on_fail=False)
> ```
> 
> ^ this is not working unless I also provide the `gas_limit`. I am worried this will confuse users. Are we not able to allow the `raise_on_fail` to work for gas estimation?

@unparalleled-js the issue I'm running into is that if we completely bypass gas estimation, the `txn.gas_limit` remains `None` which causes a missing field error later on when signing the transaction (via `eth_account.EthAccount`, which is requiring the field)

Even if I were to allow gas estimation to run but `pass` on any exceptions, we still end up with `txn.gas_limit is None`. Is this something that we can get around or would bypassing gas estimation only work if the user happened to supply their own `gas_limit`?

cc @NotPeopling2day

NotPeopling2day:
> > I am worried about this (from an ape-console environment):
> > ```python
> > receipt = contract.foo(..., raise_on_fail=False)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ^ this is not working unless I also provide the `gas_limit`. I am worried this will confuse users. Are we not able to allow the `raise_on_fail` to work for gas estimation?
> 
> @unparalleled-js the issue I'm running into is that if we completely bypass gas estimation, the `txn.gas_limit` remains `None` which causes a missing field error later on when signing the transaction (via `eth_account.EthAccount`, which is requiring the field)
> 
> Even if I were to allow gas estimation to run but `pass` on any exceptions, we still end up with `txn.gas_limit is None`. Is this something that we can get around or would bypassing gas estimation only work if the user happened to supply their own `gas_limit`?
> 
> Ideally `raise_on_fail=False` **will** bypass gas estimation, but I'm blocked on that part by what I just described
> 
> cc @NotPeopling2day

I might be missing some nuance here, but I think it's that you would supply a `gas_limit` value during the bypass. I believe Brownie uses an environment configuration to grab a "max" gas based off of the last block in the chain. 

https://github.com/eth-brownie/brownie/blob/4ae5f527ea86eb95766fe225a0f67620ffd36022/brownie/network/account.py#L415

unparalleled-js:
> @unparalleled-js the issue I'm running into is that if we completely bypass gas estimation, the `txn.gas_limit` remains `None` which causes a missing field error later on when signing the transaction (via `eth_account.EthAccount`, which is requiring the field)

Thanks for explaining this, that totally makes sense and now I understand the problem!
I think that @NotPeopling2day is on to something here- we may need a way for `EcosystemAPI` to define what its max fee value is. Perhaps a new API property?

tests/functional/test_receipt.py
unparalleled-js: Ah ok I think I was confused by the test name, `test_get_failed_receipt_raise` .

Can we call it something like `test_contract_method_raise_on_fail()`

src/ape/api/accounts.py
unparalleled-js: Oh, cool! I think you're right. I think it only makes `mypy` mad which is nbd.

fubuloubu:
> > @unparalleled-js the issue I'm running into is that if we completely bypass gas estimation, the `txn.gas_limit` remains `None` which causes a missing field error later on when signing the transaction (via `eth_account.EthAccount`, which is requiring the field)
> 
> Thanks for explaining this, that totally makes sense and now I understand the problem! I think that @NotPeopling2day is on to something here- we may need a way for `EcosystemAPI` to define what its max fee value is. Perhaps a new API property?

Max fee limit, or do you mean max gas usage?

tests/functional/test_receipt.py
helloibis: Updated in f42fa62a

src/ape/api/accounts.py
helloibis: @unparalleled-js ohh true

One issue I did run into for a sec the other day was at first I had args of subclasses in a different order, the type checker was very mad 😂 should be all good now!

helloibis:
> > Thanks for explaining this, that totally makes sense and now I understand the problem! I think that @NotPeopling2day is on to something here- we may need a way for `EcosystemAPI` to define what its max fee value is. Perhaps a new API property?
> 
> Max fee limit, or do you mean max gas usage?

Seems like [Brownie grabs some configuration values for the max/priority fee if you're not specifying a gas limit](https://github.com/eth-brownie/brownie/blob/4ae5f527ea86eb95766fe225a0f67620ffd36022/brownie/network/account.py#L707-L712) for a given txn. Should we have parity with that behavior?

I also opened up a separate ticket for configuration details at https://github.com/ApeWorX/ape/issues/985. For now I just mentioned `gas_limit`, but as noted with the Brownie question there may be other, related configuration settings we'd want to add in as well

fubuloubu (dismissed):

helloibis:
@fubuloubu @unparalleled-js PR is ready for review, with the state of the feature still being what's described in https://github.com/ApeWorX/ape/pull/957#issuecomment-1215565267

fubuloubu (approved):
This looks good if @unparalleled-js gives it a bit of QA and is okay with it

helloibis:
Just pushed up 59324c0b to fix the default value for `raise_on_fail` inside `get_receipt()` / `await_confirmations()`

If we had left it to `True` this actually would've been a breaking change, but leaving it to `False` should still allow it to work as documented earlier (since `send_transaction()` will raise) without causing any breaking changes

helloibis:
Also just rebased to fix the git conflict

src/ape/contracts/base.py
```diff
@@ -208,12 +208,14 @@ def serialize_transaction(self, *args, **kwargs) -> TransactionAPI:
         )
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        raise_on_fail = kwargs.pop("raise_on_fail", True)
```
helloibis, line 211: Just fixed this in 12d9cd20 because only `get_receipt()` and `await_confirmations()` default to `False`, everything else defaults to `True` so that should include this as well

fubuloubu (dismissed):

src/ape/api/providers.py
```diff
@@ -317,12 +317,18 @@ def get_transactions_by_block(self, block_id: BlockID) -> Iterator[TransactionAP
...
         Send a transaction to the network.
 
+        Raises:
+            :class:`~ape.exceptions.TransactionError`: When the transaction fails
+              if ``raise_on_fail`` is ``True``.
+
         Args:
             txn (:class:`~ape.api.transactions.TransactionAPI`): The transaction to send.
+            raise_on_fail (bool): ``True`` will cause failed transactions to raise
+              `~ape.exceptions.TransactionError`.
```
unparalleled-js, line 331: Maybe add `
```suggestion
              `~ape.exceptions.TransactionError`.  Defaults to ``True``.
```

src/ape/api/providers.py
```diff
@@ -812,6 +822,8 @@ def get_receipt(
               to wait before returning the receipt.
             timeout (Optional[int]): The amount of time to wait for a receipt
               before timing out.
+            raise_on_fail (bool): Whether an exception should be raised if the
+              transaction failed.
```
unparalleled-js, line 826:
```suggestion
              transaction failed. Defaults to ``False``.
```

unparalleled-js (commented):
Just did some testing, this is great! I can get failed receipts easily and my projects' tests still pass (with all their revert-based testing etc).

However -- dun dun dun. I noticed this breaks `ape-hardhat` and `ape-foundry` so we may either have to hold off on merging this to main **or** do some preparation fixes there first to see if we can mitigate breaking those plugins

helloibis:
> I noticed this breaks ape-hardhat and ape-foundry so we may either have to hold off on merging this to main or do some preparation fixes there first to see if we can mitigate breaking those plugins

Ah yeah, almost forgot about those! I mentioned them earlier but there have been so many comments 😅 will get on those now and make companion PRs that can go in at the same time as this

fubuloubu (dismissed):
looking good

tests/functional/test_provider.py
```diff
@@ -79,6 +84,45 @@ def test_get_receipt_exists_with_timeout(eth_tester_provider, vyper_contract_ins
     assert receipt_from_provider.receiver == vyper_contract_instance.address
 
 
+def test_get_receipt_without_raising(eth_tester_provider, vyper_contract_instance, owner):
+    # The Contract raises empty revert when setting number to 5.
+    receipt_from_invoke = vyper_contract_instance.setNumber(
```
unparalleled-js, line 89: This part of the test might make more sense in `test_contract_instance()`

`def test_invoke_failing_method_without_raising()`

unparalleled-js (approved):
LGTM

helloibis:
Moved this to a feature branch since we'll hold til 0.6.0 for the breaking change. This PR should be good to merge now, and here are the accompanying PRs (each with their own feature branch):

https://github.com/ApeWorX/ape-hardhat/pull/99
https://github.com/ApeWorX/ape-foundry/pull/17

tests/functional/test_provider.py
helloibis: Can do, I'll take a look at cleaning up some test stuff in the upcoming PR for reverts since that will need to add in testing changes

Issue #530, opened 2022-03-02
Managing Deployments


fubuloubu:
### Overview
Allow specifying final deployments to networks, as well as referencing them.

Currently, we have limited deployment management features from the config file of a project, as well as the project manifests, but for the currently active project we need a way to track and reference deployments to create structured deployment processes (e.g. B depends on A)

This is a test of Linear app integration

### Specification
```py
from ape import project  # NOTE: Our local project


# Do deployment of `my_contract = account.deploy(project.MyContract)`
project.track_deployment(my_contract)  # NOTE: Appends to list of deployments for type
# Adds tracking for deployment of `MyContract` type on `networks.active_network`
# NOTE: This only works for our local project (deployments are immutable otherwise)

...

# Reference previous deployment
# NOTE: This works for any project, not just our local one
my_contract = project.deployments(project.MyContract)[-1]
```

Also add a way to "untrack" a deployment:
```py
project.drop_deployment(my_contract)
# NOTE: This only works for our local project (deployments are immutable otherwise)
```

### Dependencies
EthPM `dependencies` field

unparalleled-js:
Yes!!!

unparalleled-js:
@NotPeopling2day said I can take this one

unparalleled-js:
Status update!

First, was confused about some things in ethpm-types so I spent time researching and updating the documentation there: 
https://github.com/ApeWorX/ethpm-types/pull/40

After I understood more about the schema for deployments, I realized I would like to local deployments cache to be more work-able with me, so I had to do this work today: https://github.com/ApeWorX/ape/pull/940

So we are making good practice but this was a little more complex than I originally thought! But I assume it still won't take too much more effort.

fubuloubu:
@unparalleled-js as a side note: always keep in mind we can make changes to EIP-2678, we've made a few already even (mostly additions). If any substantive changes need to happen that depart from EIP-2678, we can look at publishing it as a new EIP.

PR #961, opened 2022-08-10
feat: track deployments

unparalleled-js:
### What I did

fixes: #530 

### How I did it

* Build the `EthPMContractInstance` artifact in the new `track_deployment()` method.
* Have it store the build artifact in `<project>/.build/deployments/<bip122-chain-id>/<contract-name>.json`.

### How to verify it

In an Ape project, run:

```python

acct = accounts.load("metamask0")
contract = project.MyContract.at("0x...")
project.track_deployment(contract)
```

The ensure the build artifact is present in the `deployments` list

### Checklist

### DIFF
--- a/docs/userguides/publishing.md
+++ b/docs/userguides/publishing.md
@@ -1,0 +1,33 @@
+# Publishing
+
+Publishing smart-contract packages using Ape is influenced from [EIP-2678](https://eips.ethereum.org/EIPS/eip-2678) and uses the [ethpm-types](https://github.com/ApeWorX/ethpm-types) Python package extensively (which is also managed by the ApeWorX organization).
+This guide exists to walk through the steps of publishing your project.
+
+## Compilation
+
+First, your project must compile.
+
+```bash
+ape compile
+```
+
+To learn more about project compilation, follow [this guide](./compile.html).
+Once your project has successfully compiled, you will have the start of your `PackageManifest` generated in your project's `.build/` directory.
+
+## Tracking Deployments
+
+If your project contains delpoyments that you wish to include in its package manifest, use the [track_deployment()](../methoddocs/managers.html#ape.managers.project.manager.ProjectManager.track_deployment) method.
+Example:
+
+```python
+from ape import accounts, project
+
+account = accounts.load("mainnet-account")
+
+# Assume your project has a contract named 'MyContract' with constructor that accepts argument '123'.
+contract = project.MyContract.deploy(123, sender=account)
+project.track_deployment(contract)
+```
+
+**NOTE**: If the contract is already deployed, you can use the [.at()](../methoddocs/contracts.html#ape.contracts.base.ContractContainer.at) method on the contract container to get an instance first.
+For more information on accessing contract instances, follow [this guide](./contracts.html).
--- a/docs/userguides/publishing.md
+++ b/docs/userguides/publishing.md
@@ -16,7 +16,7 @@ Once your project has successfully compiled, you will have the start of your `Pa
 
 ## Tracking Deployments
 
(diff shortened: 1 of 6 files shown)


docs/userguides/publishing.md
```diff
@@ -0,0 +1,33 @@
+# Publishing
```
unparalleled-js, line 1: Purposely not including this in `index.md` until publishing, as a feature, is complete

docs/userguides/publishing.md
```diff
@@ -0,0 +1,33 @@
...
+Example:
+
+```python
+from ape import accounts, project
+
+account = accounts.load("mainnet-account")
+
+# Assume your project has a contract named 'MyContract' with constructor that accepts argument '123'.
+contract = project.MyContract.deploy(123, sender=account)
+project.track_deployment(contract)
```
fubuloubu, line 29: Maybe also add an example fetching a pre-deployed contract using `Contract(<addr>)`

docs/userguides/publishing.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+**NOTE**: If the contract is already deployed, you can use the [.at()](../methoddocs/contracts.html#ape.contracts.base.ContractContainer.at) method on the contract container to get an instance first.
```
fubuloubu, line 32: `Contract(...)` uses `.at(...)` so probably that's easier to shout out

fubuloubu (dismissed):
Looking pretty nice!

docs/userguides/publishing.md
unparalleled-js: Yeah we can use that, as it is the more public facing approach.

But for some reason, it makes me nervous because I know `project.MyContract` is going to use the right contract type whereas `Contract()` has to first find the contract type, which it does do successfully, but it is an extra step it has to do.

unparalleled-js:
I think we need a better way to access these deployments.

- [ ] Add `manifest_deployments` property to the `ProjectManager`

docs/userguides/publishing.md
unparalleled-js: Done!

docs/userguides/publishing.md
unparalleled-js: Done!

docs/userguides/publishing.md
unparalleled-js: Changes completed!

src/ape/managers/project/manager.py
```diff
@@ -157,6 +153,35 @@ def compiler_data(self) -> List[Compiler]:
...
+        """
+        Metadata about the active project as per EIP
+        https://eips.ethereum.org/EIPS/eip-2678#the-package-meta-object
+        Use when publishing your package manifest.
+        """
+
+        return self.config_manager.meta  # type: ignore
+
+    @property
+    def package_deployments(self) -> Dict[BIP122_URI, Dict[str, EthPMContractInstance]]:
```
fubuloubu, line 167: why `package_deployments`? why not `deployments`?

fubuloubu (approved):

src/ape/managers/project/manager.py
unparalleled-js: I am nervous a user will assume this is similar to `project.MyContract.deployments` where it is everything deployed.
I want to be sure that it is obvious as possible that these are only the publishing / final-package related deployments.

That being said, I am now realizing that `package_deployments` sounds like a deployment of a full package, so maybe we need a better name.

unparalleled-js: Ideas so far:

```
disclosed_deployments
ethpm_deployments
publishable_deployments
package_deployments
build_deployments
manifest_deployments
```

unparalleled-js: Renamed to `tracked_deployments` and added docs.

fubuloubu (approved):

PR #962, opened 2022-08-10
fix: incorrect log filters, broken logs with geth

banteg:
### What I did
- fixed log fetching with geth provider, which had a `_make_request` before it was added to web3 provider with a different api. bug was introduced in #943.
- fixed `ContractEvent` instantiating a `LogFilter` with no search topics, which resulted in fetching all contract logs and not just the ones corresponding to the event.

### How to verify it
Screen Shot 2022-08-10 at 21 16 39
Screen Shot 2022-08-10 at 21 16 49

### Checklist

### DIFF
--- a/tests/functional/test_geth.py
+++ b/tests/functional/test_geth.py
@@ -127,5 +127,5 @@ def test_get_logs_when_connected_to_geth(vyper_contract_instance, eth_tester_pro
     vyper_contract_instance.setNumber(123, sender=owner)
     actual = vyper_contract_instance.NumberChange[-1]
     assert actual.event_name == "NumberChange"
-    assert actual.contract_address == "0xF7F78379391C5dF2Db5B66616d18fF92edB82022"
+    assert actual.contract_address == vyper_contract_instance.address
     assert actual.event_arguments["newNum"] == 123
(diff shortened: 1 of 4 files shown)


fubuloubu (approved):

fubuloubu (dismissed):
Not a good way to test for this kind of thing without `geth --dev` support in test infra

unparalleled-js (approved):

tests/functional/test_geth.py
```diff
@@ -98,11 +103,27 @@ def test_get_call_tree_erigon(mock_web3, geth_provider, trace_response):
...
+    geth = networks.get_provider_from_choice("ethereum:local:geth")
     assert repr(geth) == "<geth>"
 
 
 def test_repr_connected(mock_web3, geth_provider):
     mock_web3.eth.chain_id = 123
     assert repr(geth_provider) == "<geth chain_id=123>"
+
+
+def test_get_logs_when_connected_to_geth(
```
unparalleled-js, line 116: Although hacky, this test ensures the bug is resolved and that it won't return.

fubuloubu (approved):

fubuloubu (approved):

Issue #967, opened 2022-08-11
Change `receiver` and `sender` to `bytes` datatype in `TransactionAPI`


johnson2427:
### Overview

Currently `sender` and `receiver` are of type `str` in `TransactionAPI`, would like to change these to `bytes`.
Also, the `signature` is excluded from the pydantic model, we should keep this data.

### Specification

```python
class TransactionAPI(BaseInterfaceModel):
    chain_id: int = Field(0, alias="chainId")
    hash: Any = Field(None, alias="txn_hash")
    receiver: Optional[bytes] = Field(None, alias="to")
    sender: Optional[bytes] = Field(None, alias="from")
    gas_limit: Optional[int] = Field(None, alias="gas")
    nonce: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
    value: int = 0
    data: bytes = b""
    type: Union[int, bytes, str]
    max_fee: Optional[int] = None
    max_priority_fee: Optional[int] = None

    # If left as None, will get set to the network's default required confirmations.
    required_confirmations: Optional[int] = Field(None, exclude=True)

    signature: Optional[TransactionSignature] = Field(...)
...
```

### Dependencies

There will be some tests that are broken from this, need to update the tests to be able to handle the `bytes` datatype

johnson2427:
This is a light suggestion, if we do not want to implement this, then I can convert the string to bytes from the cache side

johnson2427:
@fubuloubu @unparalleled-js lmk what you think

fubuloubu:
To better store Addresses on disk/in SQL/using inside of Ape, worth a bit of a dive one what's the best option for each and what the tradeoffs might be to choose one

However, I would say the type should eventually be `AddressType` and not `str` so as to have better type safety in the library as well.

PR #680, opened 2022-04-29
feat: added ape cache plugin

johnson2427:
### What I did

Added CacheQueryProvider to ape cache and created initial database structure and methods to instantiate an sqlite database for caching provider data.

fixes: #632

### How I did it
created cli and added init, purge and query functions

### How to verify it
I used the infura plugin
```bash
ape plugins install infura
```

Then instantiate the caching database
```bash
ape cache init --network ethereum:mainnet:infura
ape console --network ethereum:mainnet:infura
```

Run this multiple times:
```python
In [1]: chain.blocks.query("number", start_block=0, stop_block=100)
```
You should see that the query occurs quicker after the first call. This will take a few seconds to grab the data from the provider

Also run transaction calls multiple times:
```python
In [2]: chain.blocks[-2].transactions
```
You should see that the query is quicker after the first call. That block number can change, so you may not see it after the first attempt. But you will have `DynamicTransactionAPI` datatypes in the return if it comes from the provider. You will just receive a dictionary if it comes from the database.

Then exit the IPython kernel
And run the below commands
```bash
ape cache query --network ethereum:mainnet:infura "SELECT *  FROM blocks"
ape cache query --network ethereum:mainnet:infura "SELECT * FROM transactions"
```

Both should return data from the database.

Also run pytests

### Checklist

- [x] All changes are completed
- [x] New test cases have been added
- [x] Documentation has been updated

PR #970, opened 2022-08-11
refactor: change datatypes and keep `signature` element in `TransactionAPI`

johnson2427:
### What I did

Add `hash` to the `TransactionAPI`. Define as a `bytes` datatype
Change `sender` and `receiver` to `AddressType` and do not exclude `signature` from `TransactionAPI`.
Forces quite a bit of refactoring to prevent breaking anything upstream.

fixes: #967 
blocks: #680 

### How I did it

We currently bring `receiver` and `sender` in as string values, but they are addresses, and they should be defined as such. And signature is important information that we would like to be able to query, so excluding isn't the best option here.

### How to verify it

All tests have to pass

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -11,7 +11,7 @@ from tqdm import tqdm  # type: ignore
 from ape.api.explorers import ExplorerAPI
 from ape.exceptions import TransactionError
 from ape.logging import logger
-from ape.types import ContractLog, TransactionSignature
+from ape.types import AddressType, ContractLog, TransactionSignature
 from ape.utils import BaseInterfaceModel, abstractmethod, raises_not_implemented
 
 if TYPE_CHECKING:
@@ -27,8 +27,8 @@ class TransactionAPI(BaseInterfaceModel):
     """
 
     chain_id: int = Field(0, alias="chainId")
-    receiver: Optional[str] = Field(None, alias="to")
-    sender: Optional[str] = Field(None, alias="from")
+    receiver: Optional[AddressType] = Field(None, alias="to")
+    sender: Optional[AddressType] = Field(None, alias="from")
     gas_limit: Optional[int] = Field(None, alias="gas")
     nonce: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
     value: int = 0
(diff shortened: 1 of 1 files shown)


johnson2427:
@helloibis since you were in the meeting, I added you as a reviewer in case you wanted to see another level above what we were working on for that cache system,

@fubuloubu, @unparalleled-js, @NotPeopling2day, I started this PR, but I think some guidance from you 3 is imperative. 1, I don't know if this is what we want at this juncture, 2, I hope there's an easier way to fix these tests than the nightmare I was starting to go down.

fubuloubu:
Hmm, I think maybe one thing confusing (and I apologize for not bringing it up sooner), it's that `TransactionAPI` is actually pre-broadcast, meaning it's before the transaction gets confirmed by the network. `ReceiptAPI` is probably what should be the data type used inside of the the cache db, because we want to explore past "confirmed" transactions.

This probably deserves a fuller write-up in our internal docs

src/ape/api/transactions.py
```diff
@@ -27,8 +27,9 @@ class TransactionAPI(BaseInterfaceModel):
     """
 
     chain_id: int = Field(0, alias="chainId")
-    receiver: Optional[str] = Field(None, alias="to")
-    sender: Optional[str] = Field(None, alias="from")
+    hash: Any = Field(None, alias="txn_hash")
```
unparalleled-js, line 30: We already have `txn_hash` computer property. We don't have a txn_hash property on a txn that is not pre-computed since you get the hash on the receipt

unparalleled-js, line 30: We already have txn_hash computed property. We don't have a txn_hash property on a txn that is not pre-computed since you get the hash on the receipt

src/ape/api/transactions.py
```diff
@@ -40,7 +41,7 @@ class TransactionAPI(BaseInterfaceModel):
     # If left as None, will get set to the network's default required confirmations.
     required_confirmations: Optional[int] = Field(None, exclude=True)
 
-    signature: Optional[TransactionSignature] = Field(exclude=True)
+    signature: Optional[TransactionSignature]
```
unparalleled-js, line 44: Why do we need the signature like this?

src/ape/api/transactions.py
johnson2427: This PR was closed last night, we have been using the TransactionAPI data in the cache database, but it looks like we actually need ReceiptAPI instead.

src/ape/api/transactions.py
johnson2427: Same as the previous comment. And I needed that signature data for the caching system.

PR #971, opened 2022-08-12
fix: warning from using LegacyVersion

unparalleled-js:
### What I did

Another distracting warning that kept appearing in my ape tests from using a dependency that used a branch.
The warning would be about the use of `LegacyVersion` which `packaging.parse()` will return if the given param is not a PEP440 version ID (such as the string `"main"`).

### How I did it

To remedy, instead of using `parse()`, we attempt to init a `Version()` directly and catch the error when it fails. Then, the warning never arrives. This also is slightly more efficient.

### How to verify it

When you run tests in a project that use a dependency that uses a branch n GitHub, you don't get the `LegacyVersion` warning.

See the test I added!

### Checklist

### DIFF
--- a/tests/functional/test_dependencies.py
+++ b/tests/functional/test_dependencies.py
@@ -1,0 +1,62 @@
+import shutil
+from pathlib import Path
+from typing import Dict
+
+import pytest
+
+
+@pytest.fixture
+def oz_dependencies_config():
+    def _create_oz_dependency(version: str) -> Dict:
+        return {
+            "name": "OpenZeppelin",
+            "version": version,
+            "github": "OpenZeppelin/openzeppelin-contracts",
+        }
+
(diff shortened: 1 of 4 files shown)


tests/functional/test_dependencies.py
```diff
@@ -0,0 +1,62 @@
...
+        "github": "apeworx/testfoobartest",
+        "name": "foobar",
+        "branch": "main",
+    }
+    dependency = dependency_manager.decode_dependency(dependency_config)
+    _ = dependency.cached_manifest
+
+    # Tests against bug where we tried creating version objects ot of branch names
+    # thus causing warnings to show in `ape test` runs.
+    assert DeprecationWarning not in [w.category for w in recwarn.list]
```
unparalleled-js, line 62: this is the new test that fails when running on main

fubuloubu (approved):

NotPeopling2day (approved):

PR #974, opened 2022-08-12
test: use x-dist in tests

unparalleled-js:
### What I did

* Speeds up test runs by utilizing x-dist
* Fixes warnings from coverage
* Fixes flakey script test

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -74,7 +74,7 @@ jobs:
             pip install .[test]
 
         - name: Run Tests
-          run: pytest -m "not fuzzing" -s --cov -n auto
+          run: pytest -m "not fuzzing" -s --cov=src -n auto
 
     fuzzing:
         runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


unparalleled-js:
The tests run more than 3x faster in CI and 10x faster locally (For me) from this PR!

No more long waits!!! Way less build times in GH. Yay and yay!

.github/workflows/test.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
             pip install .[test]
 
         - name: Run Tests
-          run: pytest -m "not fuzzing" -s --cov
+          run: pytest -m "not fuzzing" -s --cov=src -n auto
```
fubuloubu, line 77: I thought github actions only executes on single core? so this is kind of surprising

fubuloubu (approved):

.github/workflows/test.yaml
unparalleled-js: ape-hardhat has been doing this all along too

PR #976, opened 2022-08-15
feat: support nested array

yuntai:
### What I did

support for nested array when parsing return data

fixes: #

### How I did it

For type `address[3][]`, sub-type is extracted as `address` not `address[3]` which can be recursively parsed. 
All fixes are in `decode_primitive_value()` in `ecosystem.py` 
- make the subtype of a nested array to array. e.g.) subtype(uint[][3]) => uint[]
- fix to return a list instead of a tuple when parsing value when its type is list

### How to verify it

Added tests against various nested array types

### Checklist


fubuloubu (approved):
This looks good, will let @unparalleled-js doublecheck

yuntai:
> Some linting concerns and would like to also get the tests in vyper to work I can take care of these if you like!

yes, please!

unparalleled-js (approved):

fubuloubu (approved):

PR #978, opened 2022-08-15
fix: issue with editable installs of plugins

unparalleled-js:
### What I did

OK, was able to fix the environment issues preventing me from testing editable installs of things.

Found out the editable installs are suddenly weirdly prefixed and formatted in pkg utils, not sure when or how or why this is happening.. Also `ispkg` was returning `False` in this case and I am not sure why that it is either.

### How I did it

Extract the real package name from the weirdly formatted one and register the API implementation classes that way.

### How to verify it

I am not sure; no one else has reported this issue.
At the very least, nothing should be broken from this change.

If you are able to reproduce the issue of not being able to work with editable installs, you should now be able to.

### Checklist

### DIFF
--- a/src/ape_networks/_cli.py
+++ b/src/ape_networks/_cli.py
@@ -14,7 +14,7 @@ def _filter_option(name: str, options):
     return click.option(
         f"--{name}",
         f"{name}_filter",
-        multiple=Tree,
+        multiple=True,
         help=f"Filter the results by {name}",
         type=click.Choice(options),
     )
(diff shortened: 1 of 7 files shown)


src/ape_networks/_cli.py
```diff
@@ -14,7 +14,7 @@ def _filter_option(name: str, options):
     return click.option(
         f"--{name}",
         f"{name}_filter",
-        multiple=Tree,
```
fubuloubu, line 17: lol, how did this happen??

fubuloubu (approved):

src/ape_networks/_cli.py
unparalleled-js: It wasn't causing any issues, but I am guessing it happens from a text editor autocorrection overlook. Like went to type `True` but the IDE autofilled to `Tree` since it was present in the namespace. That is my guess!

PR #979, opened 2022-08-15
feat: compiler settings

sabotagebeats:
### What I did
added get_compiler_settings

fixes: #

### Checklist


src/ape/api/compiler.py
```diff
@@ -35,6 +36,19 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
             Set[str]: A set of available compiler versions.
         """
 
+    @abstractmethod
```
unparalleled-js, line 39: Instead of making this abstract, we can have it return `{}` by default? This way, it won't be a breaking change.
We can later on decide to change it to abstract if it makes sense (such as pre-0.5)

unparalleled-js, line 39: Instead of making this abstract, we can have it return {} by default? This way, it won't be a breaking change.
We can later on decide to change it to abstract if it makes sense (such as pre-0.5)

src/ape/api/compiler.py
```diff
@@ -35,6 +36,19 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
...
+    def get_compiler_settings(self, contract_filepaths: List[Path], base_path: Optional[Path]
+    ) -> String:
+        """
+        Retreive the json of compiler settings in manifest
+
+        Args:
+            all_paths (List[pathlib.Path]): The list of paths.
+
+        Returns:
+            str: A json string of the current compiler settings.
```
unparalleled-js, line 49: we should probably be returning a `Dict[Version, Dict]`

src/ape/api/projects.py
```diff
@@ -146,8 +146,11 @@ def _create_compilers_list(
                 version = versions[0]
                 filtered_paths = [p for p in source_paths if p.suffix == ext]
                 version_map = {version: filtered_paths}
+            
+            settings = compiler.get_compiler_settings(source_paths,contracts_folder)
 
             for version, paths in version_map.items():
+                settings = compiler.get_compiler_settings(paths,contracts_folder)
```
unparalleled-js, line 153: We are already getting all the settings on line 150, this is a duplicate line.
Instead, grab the settings that pertains to this version, something like:

```python
version_settings = settings[version]
```

src/ape/api/projects.py
```diff
@@ -157,7 +160,7 @@ def _create_compilers_list(
                     Compiler(
                         name=compiler.name,
                         version=str(version),
-                        settings={},
+                        settings=settings,
```
unparalleled-js, line 163: This should be like the `version_settings`, not the full settings lists for each compiler. Get the settings for _this_ version.

unparalleled-js (commented):
I added some comments I hope are helpful!

src/ape/api/compiler.py
sabotagebeats: can you let me know if I did this right in the newest commit?

sabotagebeats:
failing type check: 
```
Run mypy .
src/ape/api/projects.py:153: error: Value of type String? is not indexable
src/ape/api/compiler.py:41: error: Variable "tokenize.String" is not valid as a type
src/ape/api/compiler.py:41: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 2 errors in 2 files (checked 160 source files)
Error: Process completed with exit code 1.
```

src/ape/managers/compilers.py
```diff
@@ -110,6 +110,14 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
 
         return contract_types_dict  # type: ignore
 
+    def get_compiler_settings(self, contract_filepaths: List[Path], base_path: Optional[Path]):
+        settings = dict()
+        for compiler in self.registered_compilers:
+            settings[compiler] = self.registered_compilers[compiler].get_compiler_settings(
+                contract_filepaths, base_path=base_path
+            )
+        return settings
```
fubuloubu, line 119: Be a little more performant
```suggestion
        return {
            compiler: self.registered_compilers[compiler].get_compiler_settings(
                contract_filepaths, base_path=base_path
            )
            for compiler in self.registered_compilers
        }
```

sabotagebeats:
associated PR: https://github.com/ApeWorX/ape-solidity/pull/64

sabotagebeats:
mypy fails if traitlets version is too small, also is being used and wasn't listed in setup.py

sabotagebeats:
adding tests in solidity plugin because that's where data will be found.

setup.py
```diff
@@ -111,6 +111,7 @@
         "singledispatchmethod ; python_version<'3.8'",
         "tqdm>=4.62.3,<5.0",
         "typing-extensions ; python_version<'3.8'",
+        "traitlets>=5.3.0",
```
fubuloubu, line 114: Why was this added to dependencies?

src/ape/api/compiler.py
```diff
@@ -35,6 +35,19 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
             Set[str]: A set of available compiler versions.
         """
 
+    def get_compiler_settings(
+        self, contract_filepaths: List[Path], base_path: Optional[Path]
+    ) -> Dict[Version, Dict]:
+        """
+        Retrieve the json of compiler settings in manifest
```
unparalleled-js, line 42:
```suggestion
        Get a mapping of the settings that would be used to compile each of the sources
        by the compiler version number.
```

src/ape/api/compiler.py
```diff
@@ -35,6 +35,19 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
...
+
+        Args:
+            all_paths (List[pathlib.Path]): The list of paths.
+
+        Returns:
+            Dict[Version, Dict]: A dict of the current compiler settings.
```
unparalleled-js, line 48:
```suggestion
            Dict[Version, Dict]: A dict of compiler settings by compiler version.
```

unparalleled-js (approved):
This PR actually fixes an intermittent bug that appears across all plugins

unparalleled-js (approved):

unparalleled-js (approved):

PR #983, opened 2022-08-16
feat: required param in network option

unparalleled-js:
### What I did

Add `required` parameter to `network_option`.

### How I did it

self explanatory!

### How to verify it

You can now require the user to input a value when they use the `--network` flag

```python
import click
from ape.cli import network_option


@click.command()
@network_option(required=True)
def cmd(network)
    print(network)
```

### Checklist

### DIFF
--- a/tests/functional/test_cli.py
+++ b/tests/functional/test_cli.py
@@ -148,3 +148,14 @@ def test_network_option_adhoc(runner, network_cmd, network_input):
     result = runner.invoke(network_cmd, ["--network", network_input])
     assert result.exit_code == 0, result.output
     assert OUTPUT_FORMAT.format(network_input) in result.output
+
+
+def test_network_option_make_required(runner):
+    @click.command()
+    @network_option(required=True)
+    def cmd(network):
+        click.echo(OUTPUT_FORMAT.format(network))
+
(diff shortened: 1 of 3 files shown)


tests/functional/test_cli.py
```diff
@@ -148,3 +148,14 @@ def test_network_option_adhoc(runner, network_cmd, network_input):
     result = runner.invoke(network_cmd, ["--network", network_input])
     assert result.exit_code == 0, result.output
     assert OUTPUT_FORMAT.format(network_input) in result.output
+
+
+def test_network_option_make_required(runner):
+    @click.command()
+    @network_option(required=True)
+    def cmd(network):
```
unparalleled-js, line 156: @johnson2427 we can make the `ape cache init` and `ape cache purge` etc. commands set `required=True` so that the user _has_ to provide a network. This will lessen the confusionz because

1. The user will only get the local network error message if they explicitly try to use the local network and not when excluding the option altogether
2. The user when excluding the option will realize they need the option and then likely input them mainnet value or something and never ever hit the local-error network to begin with

fubuloubu: Was just thinking about this today

fubuloubu (approved):

PR #993, opened 2022-08-18
fix: issue with editable plugins

unparalleled-js:
### What I did

Editable plugin installation was giving me another issue with a value like this:

```
__editable___ape_starknet_0_4_0a2_dev0_gf91d85b_d20220818_finder
```

so this PR handles that.
**NOTE**: This is not really a user facing bug, only someone who develops plugins would hit this. There is not really an easy way to test this.

### Checklist


fubuloubu (approved):
Assuming you have been playing with this locally in response to issues you are seeing

PR #1003, opened 2022-08-19
chore: add CodeQL Analysis to Repo

fubuloubu:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,38 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '23 17 * * 5'
```
unparalleled-js, line 9: Does this need to run on a schedule?

unparalleled-js (commented):
It is showing one alert in `test_project.py`, do we want to also address that?

unparalleled-js (dismissed):

.github/workflows/codeql.yml
fubuloubu: no, that came with the original copy, will remove

fubuloubu:
> It is showing one alert in `test_project.py`, do we want to also address that?

yeah probably

.github/workflows/codeql.yml
fubuloubu, line 9:
```suggestion
```

fubuloubu:
> It is showing one alert in `test_project.py`, do we want to also address that?

took a further look and dismissed it as False Positive

UI for reporting and removing false positives is easy, can do it through github on the files page

unparalleled-js (approved):

PR #1004, opened 2022-08-19
chore: stop duplicate CI runs on branches

fubuloubu:
### What I did
Stop running CI twice


unparalleled-js (approved):
Do we need to do the same for the lint checks?

So it was only running twice when the user pushed their branch to the ApeWorX org, right (just making sure I am followin)?

unparalleled-js:
I think the tests may not be running

unparalleled-js (changes requested):
I don't think this will work, the logic seems off.

However, I think we can do this:

```yaml
  push:
    branches:
      - 'main'
```

and now it will only run on pushes to main instead of pushing to branches

unparalleled-js:
I'm trying it on hardhat now: https://github.com/ApeWorX/ape-hardhat/pull/92/files

fubuloubu:
> I don't think this will work, the logic seems off.
> 
> However, I think we can do this:
> 
> ```yaml
>   push:
>     branches:
>       - 'main'
> ```
> 
> and now it will only run on pushes to main instead of pushing to branches
> 
> Edit: this goes in the `on` section at the top. You can make it an object rather than a list of strings

The idea behind not restricting other branches is that in a fork, someone can just enable github actions (and add their secrets) and it should function there too, for heavier development

.github/workflows/test.yaml
```diff
@@ -5,6 +5,10 @@ name: Test
 jobs:
     linting:
         runs-on: ubuntu-latest
+        if: |
+            github.event_name != "pull_request"
+            || github.event.pull_request.head.repo.full_name
+            == github.event.pull_request.base.repo.full_name
```
fubuloubu, line 11:
```suggestion
            github.event_name == "push"
            && github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -30,6 +34,10 @@ jobs:
 
     type-check:
         runs-on: ubuntu-latest
+        if: |
+            github.event_name != "pull_request"
+            || github.event.pull_request.head.repo.full_name
+            == github.event.pull_request.base.repo.full_name
```
fubuloubu, line 40:
```suggestion
            github.event_name == "push"
            && github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -49,6 +57,10 @@ jobs:
 
     functional:
         runs-on: ${{ matrix.os }}
+        if: |
+            github.event_name != "pull_request"
+            || github.event.pull_request.head.repo.full_name
+            == github.event.pull_request.base.repo.full_name
```
fubuloubu, line 63:
```suggestion
            github.event_name == "push"
            && github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -78,6 +90,10 @@ jobs:
 
     fuzzing:
         runs-on: ubuntu-latest
+        if: |
+            github.event_name != "pull_request"
+            || github.event.pull_request.head.repo.full_name
+            == github.event.pull_request.base.repo.full_name
```
fubuloubu, line 96:
```suggestion
            github.event_name == "push"
            && github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -5,6 +5,9 @@ name: Test
 jobs:
     linting:
         runs-on: ubuntu-latest
+        if: |
+            github.event_name == "push"
+            && github.repository_owner != "ApeWorX"
```
fubuloubu, line 10:
```suggestion
            github.event_name == "pull_request"
            || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -30,6 +33,8 @@ jobs:
 
     type-check:
         runs-on: ubuntu-latest
+            github.event_name == "push"
+            && github.repository_owner != "ApeWorX"
```
fubuloubu, line 37:
```suggestion
            github.event_name == "pull_request"
            || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -49,6 +54,8 @@ jobs:
 
     functional:
         runs-on: ${{ matrix.os }}
+            github.event_name == "push"
+            && github.repository_owner != "ApeWorX"
```
fubuloubu, line 58:
```suggestion
            github.event_name == "pull_request"
            || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -78,6 +85,8 @@ jobs:
 
     fuzzing:
         runs-on: ubuntu-latest
+            github.event_name == "push"
+            && github.repository_owner != "ApeWorX"
```
fubuloubu, line 89:
```suggestion
            github.event_name == "pull_request"
            || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -5,6 +5,9 @@ name: Test
 jobs:
     linting:
         runs-on: ubuntu-latest
+        if: |
+            github.event_name == "pull_request"
+            || github.repository_owner != "ApeWorX"
```
fubuloubu, line 10:
```suggestion
        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```

.github/workflows/test.yaml
```diff
@@ -30,6 +33,8 @@ jobs:
 
     type-check:
         runs-on: ubuntu-latest
+            github.event_name == "pull_request"
+            || github.repository_owner != "ApeWorX"
```
fubuloubu, line 37:
```suggestion
        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```

.github/workflows/test.yaml
```diff
@@ -49,6 +54,8 @@ jobs:
 
     functional:
         runs-on: ${{ matrix.os }}
+            github.event_name == "pull_request"
+            || github.repository_owner != "ApeWorX"
```
fubuloubu, line 58:
```suggestion
        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```

.github/workflows/test.yaml
```diff
@@ -78,6 +85,8 @@ jobs:
 
     fuzzing:
         runs-on: ubuntu-latest
+            github.event_name == "pull_request"
+            || github.repository_owner != "ApeWorX"
```
fubuloubu, line 89:
```suggestion
        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```

unparalleled-js (dismissed):
LGTM

to translate, actions will run either if it's a PR (no matter if from fork or not) or if it is a push to a fork. 

Thus, a push to ApeWorX/ape won't trigger unless it's on a PR
Thus, there are no more duplicate CI runs on branches pushed to ApeWorX/ape.

fubuloubu:
> LGTM
> 
> to translate, actions will run either if it's a PR (no matter if from fork or not) or if it is a push to a fork. 
> 
> Thus, a push to ApeWorX/ape won't trigger unless it's on a PR
> Thus, there are no more duplicate CI runs on branches pushed to ApeWorX/ape.

Yeah, still not quite perfect (would prefer branches to run if and only if they are not in an active PR), but this is the best I can do.

I also am not a huge fan of how they say "ignored"

fubuloubu:
Also appears that linting, type checking, and fuzzing didn't trigger...

.github/workflows/test.yaml
```diff
@@ -5,6 +5,7 @@ name: Test
 jobs:
     linting:
         runs-on: ubuntu-latest
+        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```
fubuloubu, line 8:
```suggestion
        if: github.event_name == "pull_request" || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -30,6 +31,7 @@ jobs:
 
     type-check:
         runs-on: ubuntu-latest
+        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```
fubuloubu, line 34:
```suggestion
        if: github.event_name == "pull_request" || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -49,6 +51,7 @@ jobs:
 
     functional:
         runs-on: ${{ matrix.os }}
+        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```
fubuloubu, line 54:
```suggestion
        if: github.event_name == "pull_request" || github.repository_owner != "ApeWorX"
```

.github/workflows/test.yaml
```diff
@@ -78,6 +81,7 @@ jobs:
 
     fuzzing:
         runs-on: ubuntu-latest
+        if: ${{ github.event_name == 'pull_request' || github.repository_owner != 'ApeWorX' }}
```
fubuloubu, line 84:
```suggestion
        if: github.event_name == "pull_request" || github.repository_owner != "ApeWorX"
```

fubuloubu (commented):
Don't need the escapes with `if`

unparalleled-js:
See https://github.com/ApeWorX/ape/pull/1171, including the stackoverflow link

PR #1007, opened 2022-08-19
fix: issue with decoding None block hashes

unparalleled-js:
### What I did

Regression from cache PR where could not decode pending blocks in `ape-hardhat`

### How I did it

Handle when hash is None in data when decoding block

### How to verify it

Launch hardhat console

```sh
ape console --network ethereum:local:hardhat
```

Get pending block

```python
networks.provider.get_block("pending")
```

It should work

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -272,7 +272,7 @@ class Ethereum(EcosystemAPI):
         return receipt
 
     def decode_block(self, data: Dict) -> BlockAPI:
-        data["hash"] = HexBytes(data["hash"])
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -272,7 +272,7 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
         return receipt
 
     def decode_block(self, data: Dict) -> BlockAPI:
```
johnson2427, line 274: Good find! Works for me!

johnson2427 (approved):

PR #1012, opened 2022-08-25
feat: add `ape` to console namespace

delaaxe:
### What I did

As it is good practice to explicitly use the `np.` and `pd.` namespaces in larger scripts using numpy/pandas, some users prefer always using the ape prefix as in `ape.networks`, `ape.accounts` etc. for more clarity.

Having the `ape` name available by default in the console allows for sending code from script to console easily and using the same syntax/habits in both.

### Checklist

### DIFF
--- a/src/ape_console/_cli.py
+++ b/src/ape_console/_cli.py
@@ -113,6 +113,7 @@ def console(project=None, verbose=None, extra_locals=None):
             faulthandler.enable()  # NOTE: In case we segfault
 
     namespace = {component: getattr(ape, component) for component in ape.__all__}
+    namespace["ape"] = ape
 
     if extra_locals:
         namespace.update(extra_locals)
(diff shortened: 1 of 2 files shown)


src/ape_console/_cli.py
```diff
@@ -113,6 +113,7 @@ def console(project=None, verbose=None, extra_locals=None):
             faulthandler.enable()  # NOTE: In case we segfault
 
     namespace = {component: getattr(ape, component) for component in ape.__all__}
+    namespace["ape"] = ape
```
unparalleled-js, line 116: Thoughts on this:

```python

class Ape:
  def __getattr__(self, item):
      return getattr(ape, item)

  def __repr__(self):
      return "<ape>"

namespace["ape"] = Ape()
```

This makes it look like this in the console:

```
(dev) ➜  ape git:(patch-1) ✗ ape console

In [1]: ape
Out[1]: <ape>
```

unparalleled-js (commented):
I am good with this! You make a compelling argument about `pd` and such.

However, we also need to update `docs/userguides/console.md` to reflect this addition.
And I added a suggestion regarding `__repr__` and making a wrapper... Let me know thoughts on that, just an idea to make it look nicer.

src/ape_console/_cli.py
delaaxe: This is nice but doing it like so breaks tab-discovery, which is crucial for new users:

image

Maybe there's a way to set the repr on the actual module?

delaaxe: Actually yes, in case we really want to go down that path: https://stackoverflow.com/a/1727209/1110703

unparalleled-js: Sounds like this is a bad idea. Let's skip it! Was a fun thought experiment though. But probably best to keep modules as modules

delaaxe:
Updated the userguide

unparalleled-js (approved):

NotPeopling2day (approved):

Issue #996, opened 2022-08-18
Refactor ReceiptAPI into base model


NotPeopling2day:
### Overview

To prepare for refactoring Query and Cache to use `ReceiptAPI`, we need to slim down the current `ReceiptAPI` to a base model and move any additional fields to their respective subclasses and ecosystems. We should also look at any fields we need to compute and add to these subclassed models to better support data scientist and/or to add value when querying.

Also, we need to add the `raw_transaction` field so transactions can be recreated from the receipt object. Many transaction types will result in the same receipt structure.

### Specification

- [ ] Refactor `ReceiptAPI` to be a base model with minimal fields.
- [ ] Include a raw_transaction field to save the original transaction
- [ ] Add computed properties in subclassed models of `ReceiptAPI` in ecosystem plugins

PR #1015, opened 2022-08-26
refactor: optimize ReceiptAPI for better management of transaction data

johnson2427:
### What I did

Refactored ReceiptAPI, and created a new `transaction` attribute to better handle transaction data.

fixes: #996 
fixes: ape-22

### How I did it

A lot of design work, then a lot of changing tests, making the data fall into place so ReceiptAPI could parse the data.

### How to verify it

pytest

### Checklist


johnson2427:
@fubuloubu and @NotPeopling2day I know you are both traveling. I made the changes that we talked about, and then had to make the fixes required in the tests to format the transactions to work inside of the receipt api. I have a few mypy issues, but it's because of the allowance of nulls and optional fields. I can fix those after you guys do your initial review. I figure I don't have this the way you two would like to see on the first go around.

tests/functional/test_chain.py
```diff
@@ -95,8 +95,8 @@ def test_account_history(sender, receiver, chain):
     assert len(transactions_from_cache) == length_at_start + 1
 
     txn = transactions_from_cache[-1]
-    assert txn.sender == receipt.sender == sender
-    assert txn.receiver == receipt.receiver == receiver
+    assert txn.raw_transaction.sender == receipt.raw_transaction.sender == sender
```
fubuloubu, line 98: Wonder if you could make properties in `ReceiptAPI` that re-expose properties in `TransactionAPI` via `__getattr__` for exactly these use cases

tests/functional/test_receipt.py
```diff
@@ -124,17 +124,28 @@ def test_get_failed_receipt(owner, vyper_contract_instance, eth_tester_provider)
 
 def test_receipt_raise_for_status_out_of_gas_error(mocker):
     gas_limit = 100000
+    txn = BaseTransaction(
```
fubuloubu, line 127: Shouldn't use `BaseTransaction` directly (probably should be `_BaseTransaction` to communicate that its private)

src/ape/api/transactions.py
```diff
@@ -147,18 +147,11 @@ class ReceiptAPI(BaseInterfaceModel):
...
     gas_used: int
     logs: List[dict] = []
-    nonce: Optional[int] = None
-    receiver: str
-    required_confirmations: int = 0
-    sender: str
     status: int
     txn_hash: str
-    value: int = 0
+    raw_transaction: TransactionAPI
```
fubuloubu, line 154: Wonder if it should just be `.transaction`, adding raw seems redundant

src/ape_ethereum/ecosystem.py
```diff
@@ -252,22 +252,31 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
         if isinstance(input_data, str):
             input_data = bytes(HexBytes(input_data))
 
+        transaction = BaseTransaction(
```
fubuloubu, line 255: Should not be `BaseTransaction`, needs a method to parse into a real transaction type via `type`

fubuloubu (commented):
Overall, it is looking okay, I think a decent amount of QA is needed here, and also this is a breaking change

Separately, I think I have a solution to the "Pandas reads a dict, not a DataModel" problem that we have. I think instead of serializing to dict via pydantic, we should probably use `getattr` to fetch properties directly from the data returned through query API, then we can have a more advanced handing of `item.property.sub_property` when querying that can unlock some more powerful query statements. A further investigation for how to load `cached_property` values from dict so they don't have to be recomputed is also a possibility 🤔

Should be a new issue though, tracked separately @NotPeopling2day

src/ape/api/transactions.py
```diff
@@ -253,24 +246,24 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+        while sender_nonce == self.raw_transaction.nonce:  # type: ignore
             time.sleep(1)
-            sender_nonce = self.provider.get_nonce(self.sender)
+            sender_nonce = self.provider.get_nonce(self.raw_transaction.sender)
             iteration += 1
             if iteration == iterations_timeout:
                 raise TransactionError(message="Timeout waiting for sender's nonce to increase.")
 
-        if self.required_confirmations == 0:
+        if self.raw_transaction.required_confirmations == 0:
```
NotPeopling2day, line 260: I'm good with this as an intermediate solution and later refactoring `required_confirmations` on a future ticket. I think ultimately we want to try to move this field to `NetworkAPI` as part of the network configuration, and then you'd have access to it through `self.network.required_confirmations`. 

Is that a correct assumption @fubuloubu ?

fubuloubu: Yes, it should basically be a config variable per-network e.g.:

```yaml
ethereum:
  mainnet:
    required_confirmations: 10
```

src/ape/api/transactions.py
johnson2427: [bc1d0d7](https://github.com/ApeWorX/ape/pull/1015/commits/bc1d0d74f8c9227c80f88b2c923fcf2971a59b04)

src/ape/api/transactions.py
johnson2427: I am marking this as resolved, if you have other questions on this comment, just mark it as unresolved and continue the conversation.

johnson2427:
@fubuloubu I have some mypy issues, and it all has to do with the optional datatypes. I can solve this with conditions, but I'm not sure we want to do that.

tests/functional/test_receipt.py
johnson2427: should I import BaseTransaction as _BaseTransaction? And this is a test, we do need tests for data types that aren't supposed to be used directly, is this something that I should change, or is having this tested directly okay?

tests/functional/test_chain.py
johnson2427: we can make `transaction` hidden by defining the attribute as `_transaction`. Then we can define a method for each value in `_transaction` to get unpacked and define each attribute that is defined in `transaction` into `receipt`

src/ape_ethereum/ecosystem.py
johnson2427: [096e8d7](https://github.com/ApeWorX/ape/pull/1015/commits/096e8d76a412bc6d8e8e115de676c3c09cf3b295)

src/ape_ethereum/ecosystem.py
```diff
@@ -248,26 +248,22 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
...
             txn_hash = data["hash"].hex() if isinstance(data["hash"], HexBytes) else data["hash"]
 
-        input_data = data.get("data") or data.get("input", b"")
-        if isinstance(input_data, str):
-            input_data = bytes(HexBytes(input_data))
+        if isinstance(data.get("data") or data.get("input", b""), str):
+            if data.get("data"):
+                data["data"] = bytes(HexBytes(data.get("data")))  # type: ignore
+            elif data.get("input", b""):
+                data["input"] = bytes(HexBytes(data.get("input", b"")))
```
NotPeopling2day, line 255: I think this can be simplified, I noticed in the old code that `input_data` doesn't seem to be used?

NotPeopling2day (commented):
Looking good!

src/ape_ethereum/ecosystem.py
johnson2427: input data was never used, so I removed it. I do not know why that was there originally, and I didn't want to just remove the conversion in case we had plans for it in the future.

As for simpler/cleaner. Maybe something like this:

```python
if data.get("data") and isinstance(data.get("data"), str):
    data["data"] = bytest(HexBytes(data.get("data")))

elif data.get("input") and isinstance(data.get("input", b""), str):
    data["input"] = bytes(HexBytes(data.get("input")))
```
If it goes to get "data" from the dictionary, and it does not exist, it never checks the instance. Not sure how to one-line this, and even if we did, I'm not sure it's cleaner

tests/functional/test_chain.py
fubuloubu: yes, I think let's do this. add a `__getattr__` method to `ReceiptAPI` that looks up fields via `getattr` in `.transaction`

tests/functional/test_receipt.py
fubuloubu: No, it's more like we need a way to convert transaction data to the relevant subclass via `EcosystemAPI`

It might be [`create_transaction`](https://github.com/ApeWorX/ape/blob/main/src/ape/api/networks.py#L348) or it might need to be a new method

src/ape/api/transactions.py
```diff
@@ -156,6 +156,9 @@ class ReceiptAPI(BaseInterfaceModel):
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.txn_hash}>"
 
+    def __getattr__(self, item):
+        return getattr(self.transaction, item, None)
```
fubuloubu, line 160: Really important thing about `__getattr__` is that it *must* raise `AttributeError` if the item does not exist, in order to allow the fallback effect of doing normal Python `__getattribute__` method, or else you end up with infinite loops

would suggest just changing it to this:
```suggestion
    def __getattr__(self, item: str):
        return getattr(self.transaction, item)
```

src/ape/api/transactions.py
```diff
@@ -248,27 +251,27 @@ def await_confirmations(self) -> "ReceiptAPI":
         iterations_timeout = 20
         iteration = 0
         # Wait for nonce from provider to increment.
-        if self.transaction.sender:
-            sender_nonce = self.provider.get_nonce(self.transaction.sender)
```
fubuloubu: One rule of thumb we've committed to in the past is that `__getattr__` methods should never be used inside of ape core directly, because when they fail, they fail quite badly (infinite loop corruption bug)

It's okay to use it for tests however

johnson2427: Yep, I found that out the hard way. Also, getattr doesn't like hidden attributes to be used either, took a while to figure out what I was doing wrong

src/ape/api/transactions.py
johnson2427: [4f12806](https://github.com/ApeWorX/ape/pull/1015/commits/4f1280606c1c84d2372599fd4252f3dd07c5cb28)

src/ape_ethereum/ecosystem.py
johnson2427: [9b6cbf0](https://github.com/ApeWorX/ape/pull/1015/commits/9b6cbf058db760e44302ccb900e879443ed99bc9)

I think this was the commit?

tests/functional/test_chain.py
johnson2427: [4f12806](https://github.com/ApeWorX/ape/pull/1015/commits/4f1280606c1c84d2372599fd4252f3dd07c5cb28)

tests/functional/test_receipt.py
johnson2427: If I'm reading this correctly, this is another ticket that has to be made?

src/ape/api/transactions.py
```diff
@@ -147,22 +147,18 @@ class ReceiptAPI(BaseInterfaceModel):
...
-    sender: str
     status: int
     txn_hash: str
-    value: int = 0
+    transaction: TransactionAPI
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.txn_hash}>"
 
+    def __getattr__(self, item: str) -> Optional[Any]:
```
fubuloubu, line 159:
```suggestion
    def __getattr__(self, item: str) -> Any:
```

tests/functional/test_receipt.py
fubuloubu: Yes

johnson2427: [fae8941](https://github.com/ApeWorX/ape/pull/1015/commits/fae8941b98373a3c8ce0e7db9546dfa96fcdc055)

Well I solved it like this, if you don't like this, we can roll back to the commit before I did this nonsense

src/ape/api/transactions.py
johnson2427: [63bb8b0](https://github.com/ApeWorX/ape/pull/1015/commits/63bb8b06479ffe17f672f50612fba6a058986d8f)

src/ape/api/transactions.py
```diff
@@ -151,14 +151,21 @@ class ReceiptAPI(BaseInterfaceModel):
     logs: List[dict] = []
     status: int
     txn_hash: str
-    transaction: TransactionAPI
+    transaction: TransactionAPI | dict
```
fubuloubu: Not acceptable syntax <3.10

Also, I want this to be `TransactionAPI` only, so no union as a dict

src/ape/api/transactions.py
```diff
@@ -151,14 +151,21 @@ class ReceiptAPI(BaseInterfaceModel):
...
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.txn_hash}>"
 
     def __getattr__(self, item: str) -> Optional[Any]:
         return getattr(self.transaction, item)
 
+    def __post_init__(self):
+        if isinstance(self.transaction, dict):
+            self.transaction = self.create_transaction(self.transaction)
```
fubuloubu, line 164: You should be able to do this as a validator instead

src/ape/api/transactions.py
johnson2427: Damn linter I built auto fixed my original syntax. That's fixed now

src/ape/api/transactions.py
johnson2427: Will do tonight

fubuloubu: @NotPeopling2day's example, but first check if it's a `dict` then do `TransactionAPI.parse_obj`, else just return `v`

fubuloubu (dismissed):

johnson2427 (commented):
I think that covers everything. Running IRL tests now

src/ape/api/transactions.py
```diff
@@ -152,7 +152,7 @@ class ReceiptAPI(BaseInterfaceModel):
     logs: List[dict] = []
     status: int
     txn_hash: str
-    transaction: TransactionAPI
+    transaction: Optional[TransactionAPI]
```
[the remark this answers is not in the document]
johnson2427: [49ca16a](https://github.com/ApeWorX/ape/pull/1015/commits/49ca16a7d4d3bceaaf8bede20dc79b64c0209685)

fubuloubu (approved):

NotPeopling2day (approved):
LGTM!!

unparalleled-js:
This PR contains breaking changes so the title should be `refactor!`.
Also we may have wanted to release before merging these

src/ape/api/transactions.py
```diff
@@ -147,22 +148,25 @@ class ReceiptAPI(BaseInterfaceModel):
...
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.txn_hash}>"
 
+    def __getattr__(self, item: str) -> Any:
+        return getattr(self.transaction, item)
+
+    @validator("transaction", pre=True)
+    def confirm_transaction(cls, value):
+        if isinstance(value, dict):
+            value = TransactionAPI.parse_obj(value)
```
unparalleled-js, line 166: This class will have unimplemented API methods, no?

unparalleled-js (commented):
Nice! Ah, sorry I missed reviewing this one.

Just one worry:
I am not sure you can create instances of `TransactionAPI` since it is like an abstract class.

src/ape/api/transactions.py
johnson2427: I originally had that isinstance in there for when we built dictionaries for tests. This might be better off being removed. I'm honestly not so sure we need it anymore after all the changes that were made to the tests. If you give TransactionAPI the data it is looking for, it is able to parse and build a transaction though. I rigorously tested it. It is not currently being tested now. I will make a ticket to add a test for this section of the validator to make sure this works correctly.

Issue #717, opened 2022-05-13
feat: add return value for transactions


Ninjagod1251:
### Overview

Provide a simple overview of what you wish to see added. Please include:

Upon making a test for ERC20 Smart contract. We should add a return value for accounts.

```py
tx = contract.myMutableMethod(...)
assert tx.return_value == ...  # .return_value doesn't exist

```

* What you are trying to do
* Why Ape's current functionality is inadequate to address your goal

### Specification

Describe the syntax and semantics of how you would like to see this feature implemented. The more detailed the better!

Remember, your feature is much more likely to be included if it does not involve any breaking changes.

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

fubuloubu:
Wait, this issue isn't what I wanted you to open, you can solve this by doing the following (note that list slices are non-inclusive on the right hand side):

```py
receiver, spender = accounts[1:3]
```

fubuloubu:
the actual issue is the following:
```py
tx = contract.myMutableMethod(...)
assert tx.return_value == ...  # .return_value doesn't exist
```

unparalleled-js:
You have to use events for transactions I believe.

https://ethereum.stackexchange.com/questions/73787/solidity-return-value-of-smart-contract-transact

I showed @Ninjagod1251 how to get the events from a receipt!

fubuloubu:
> You have to use events for transactions I believe.
> 
> https://ethereum.stackexchange.com/questions/73787/solidity-return-value-of-smart-contract-transact
> 
> I showed @Ninjagod1251 how to get the events from a receipt!

Not sure if this is event related, might require traces actually... I think the trick is to decode the `debug_traceTransaction` one level and obtain the last structLog item which should contain the return value. Can also use `evm-trace` to return

https://github.com/eth-brownie/brownie/blob/3aecd87f47c9c316c85b0b0c6252ff7d900cca74/brownie/network/transaction.py#L274-L275

Ninjagod1251:
Look like we should research more and I think I might be in over my head. I would love to listen in and test the feature

fubuloubu:
> Look like we should research more and I think I might be in over my head. I would love to listen in and test the feature

Yeah, please hold off on this for now

banteg:
this can be quite easily done now
```python
call_tree = chain.provider.get_call_tree(tx)
method_abi = Contract(tree.address).contract_type.mutable_methods[call_tree.calldata]
[...]

PR #1018, opened 2022-08-29
feat: add return value

fubuloubu:
### What I did
Added `.return_value` to `ReceiptAPI` so that you can detect what value is returned by a call. The feature requires call tree tracing to function

fixes: #717

### How I did it
Copied @banteg's suggestion from here: https://github.com/ApeWorX/ape/issues/717#issuecomment-1179771024

NOTE: does not function yet, it appears that the `ethpm-types` lookup of `ContractType.mutable_methods.__getitem__` doesn't function properly, getting a `KeyError`

### How to verify it
Try mainnet forking with hardhat/foundry and pulling a receipt with a return value, then decoding it

### Checklist


helloibis (dismissed):
Looks good to me aside from that ethpm-types issue you mentioned

src/ape/managers/converters.py
```diff
@@ -83,6 +83,18 @@ def convert(self, value: str) -> AddressType:
         return to_checksum_address(value)
 
 
+class BytesAddressConverter(ConverterAPI):
+    """
+    A converter that converts a raw bytes address to an ``AddressType``.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, bytes) and len(value) == 20
```
unparalleled-js, line 92: Is there a potential unintended conversion here? Like can these conditions be true without it being an address?

Is there any bad that could happen from unintended conversion here?

fubuloubu: Can update to use `eth_utils.is_canonical_address`:
https://github.com/ethereum/eth-utils/blob/master/eth_utils/address.py#L88

fubuloubu: On second thought, this is okay. Any random string of 20 bytes is a valid address.

unparalleled-js: since bytes expecting args always expect bytes, the unnecessary conversion I was worried about would not happen

src/ape/api/transactions.py
```diff
@@ -295,6 +295,26 @@ def await_confirmations(self) -> "ReceiptAPI":
 
         return self
 
+    @property
+    def return_value(self) -> Any:
+        """
+        Obtain
```
fubuloubu, line 301:
```suggestion
        Obtain the final return value of the call. Requires tracing to function,
        since this is not available from the receipt object.
        
        Returns:
            Any: The return value (ABI Decoded) of the call
```

src/ape/api/transactions.py
```diff
@@ -295,6 +295,30 @@ def await_confirmations(self) -> "ReceiptAPI":
 
         return self
 
+    @property
+    def return_value(self) -> Any:
+        """
+        Obtain the final return value of the call. Requires tracing to function,
+        since this is not available from the receipt object.
+
```
fubuloubu, line 303:
```suggestion

```

fubuloubu:
Just needs a test case or two, if someone has some free time and wants to add

fubuloubu:
Note: can't test this in the core test suite because we don't have a provider that allows checking the call tree during normal operation

src/ape/api/transactions.py
```diff
@@ -295,6 +295,29 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+        Returns:
+            Any: The return value (ABI Decoded) of the call
+        """
+        call_tree = self.provider.get_call_tree(self.txn_hash)
+
+        contract_type = self.chain_manager.contracts.get(call_tree.address)
+        if not contract_type:
+            raise ContractError("Cannot find contract type to decode with. Is it published?")
+
+        method_abi = contract_type.mutable_methods[call_tree.calldata]
```
unparalleled-js, line 312: If transacting on a view method, this will fail

unparalleled-js, line 312: I think this may need to be `calldata[:4]` ?

unparalleled-js (dismissed):
LGTM - but I added some commits so need those reviewed.

unparalleled-js:
> Note: can't test this in the core test suite because we don't have a provider that allows checking the call tree during normal operation

Testing PR: https://github.com/ApeWorX/ape-hardhat/pull/94

src/ape/api/transactions.py
```diff
@@ -295,6 +295,33 @@ def await_confirmations(self) -> "ReceiptAPI":
...
+
+        contract_type = self.chain_manager.contracts.get(call_tree.address)
+        if not contract_type:
+            raise ContractError("Cannot find contract type to decode with. Is it published?")
+
+        selector = call_tree.calldata
+        if selector in contract_type.mutable_methods:
+            method_abi = contract_type.mutable_methods[selector]
+        elif selector in contract_type.view_methods:
+            method_abi = contract_type.view_methods[selector]
```
fubuloubu, line 316: Probably needs an `else` case that raises

unparalleled-js (approved):

PR #1019, opened 2022-08-30
docs: update contracts user guide to include `at()` method reference

unparalleled-js:
### What I did

@Ninjagod1251 pointed out to me that there was gap in the user guide regarding the at() method.
This PR fills in the gap to assist the education content we are producing!

### How I did it

words words words

### How to verify it

Link checks?

### Checklist

### DIFF
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -23,7 +23,7 @@ contract = project.MyContract.deploy(sender=dev)
 
 ## From Project Contract Address
 
-You can also use the [at() method](../methoddocs/contracts.html#ape.contracts.base.ContractContainer.at) method from the same top-level project manager when you know the address of an already-deployed contract:
+You can also use the [at() method](../methoddocs/contracts.html#ape.contracts.base.ContractContainer.at) from the same top-level project manager when you know the address of an already-deployed contract:
 
 ```python
 from ape import project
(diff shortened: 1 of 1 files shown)


docs/userguides/contracts.md
```diff
@@ -16,7 +21,17 @@ dev = accounts.load("dev")
...
 ```
 
-## From Contract
+## From Project Contract Address
+
+You can also use the [at() method](../methoddocs/contracts.html#ape.contracts.base.ContractContainer.at) method from the same top-level project manager when you know the address of an already-deployed contract:
+
+```python
+from ape import project
+
```
Ninjagod1251, line 30: do you need a load an account that owns the address first to do any owner privilege methods on the contract? 
`dev = accounts.load('dev')`
`contract.mint("vitalik.eth", sender=dev)`

Should we make a side note for that?

unparalleled-js: there is a section at the bottom of this guide about contract interaction where we show how to make transactions from accounts.

this part is only about getting the contract instance

fubuloubu: > do you need a load an account that owns the address first to do any owner privilege methods on the contract? 

No you don't

fubuloubu (approved):

fubuloubu (approved):

PR #1024, opened 2022-08-31
refactor: make stdout-logger only log debug, not info

fubuloubu:
Otherwise puts out a lot of noise:
image


NotPeopling2day (approved):

unparalleled-js (approved):
This is good.  If I understand correctly, we can do anything with lines from the process queue.

Testing is good but foundry is not working right now so can't test there atm.

Issue #901, opened 2022-07-21
docker images need to be tested for running properly before pushing image to dockerhub


sabotagebeats:
### Elevator pitch:
ape docker image was failing when being ran and we were not able to know this because it was building fine. we should fix this.

### Value:
prevents bad docker images from being pushed 

### Dependencies:
n/a

### Design approach:
test docker image after building and make sure it runs successfully before pushing
detailed here: https://github.com/ApeWorX/ape/pull/891#issuecomment-1190387831

### Task list:
* [ ] Tasks go here

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

sabotagebeats:
related: https://github.com/ApeWorX/ape/pull/891

Issue #456, opened 2022-02-01
Custom plugins in Dockerfile


sabotagebeats:
### Overview

Allow user to specify which plugins they want in the dockerfile via Dockerfile arguments

# What you are trying to do

* Reduce docker build time
* Allow user to dynamically set which plugins via using a config file
* Check if we still need to pre-install `typing-extensions`

# Why Ape's current functionality is inadequate to address your goal

- [ ] Dockerfile installs from plugins separately based on Dockerfile not on config file.
- [ ] If we use a config file it allows user to configure their image more.
- [ ] See you can use ape's base image to make a more project-focused image
- [ ] Make a script that pulls for latest block but package it as a Docker image (as a proof of concept)

Issue #894, opened 2022-07-20
remove `RUN pip install eth-rlp==0.3.0` from dockerfile (blocked)


sabotagebeats:
### Elevator pitch:
this line shouldn't need to be here but is because of a dependency mismatch. if we can resolve that by updating the pins then we can remove that line. This work item is to track that. 

### Value:
cleaner dockerfile, faster build time for docker

### Dependencies:
related: https://github.com/ApeWorX/ape/pull/891
blocked by: https://github.com/ApeWorX/ape/issues/892

### Design approach:
remove that line from dockerfile after pins have been updated

### Task list:
* [ ] https://github.com/ApeWorX/ape/issues/892
* [ ] remove that line from dockerfile after pins have been updated

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

unparalleled-js:
I tried putting this in setup.py, as this is the version that gets installed anyway. However - it still fails unless present in the Dockerfile so I was not able to get rid of it.

unparalleled-js:
I am guessing `pip install eth-rlp` does something magical...

unparalleled-js:
I researched this more.

In site packages, the install of eth-rlp is corrupt. Everything is fine except the `__init__.py` file is unexpected.
Clearing the install and reinstalling fixes the issue.

However - this smells like an issue somewhere in eth-acount or pyrlp, not in ape.

unparalleled-js:
Was done in python 3.7 removal PR

PR #1029, opened 2022-09-02
feat: allow custom plugin specification in Docker file and other Docker / plugin improvements

unparalleled-js:
### What I did

fixes: #901 
fixes: #456 

* Verify Ape works on Docker build
* Adds metadata
* Uses python image instead of Ubuntu
* Allows specifying plugins to install in Dockerfile

### How I did it

1. Fix all broken plugins
2. Uninstall rlp before installing (because when already installed as same version, won't re-install and thus never re-generates the correct files).

### Checklist


fubuloubu (dismissed):

Dockerfile
```diff
@@ -3,9 +3,6 @@ RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.9 pyt
 ADD . /code
 WORKDIR /code
 
-# need to use typing-extensions<4 to deal with issue 387 bug: Docker will not build
-# TODO: Figure out a better solution or wait for it to resolve itself.
-RUN pip install typing-extensions==3.10.0.2
 RUN pip install .
 RUN ape plugins install alchemy ens etherscan foundry hardhat infura ledger solidity template tokens trezor vyper
 RUN pip install eth-rlp==0.3.0
```
fubuloubu, line 8: Is this still needed?

unparalleled-js: yes :(

I updated the ticket with what I have learned https://github.com/ApeWorX/ape/issues/894

Dockerfile
```diff
@@ -1,9 +1,43 @@
...
+ARG PYTHON_VERSION="3.9"
+FROM python:${PYTHON_VERSION}
+
+RUN apt-get update && apt-get upgrade --yes && apt-get install git
+
+# See http://label-schema.org for metadata schema
+# TODO: Add `build-date` and `version`
+LABEL maintainer="ApeWorX" \
+      org.label-schema.name="ape" \
+      org.label-schema.license="Apache" \
```
fubuloubu, line 15: is apache 2.0 an option?

Dockerfile
```diff
@@ -1,9 +1,43 @@
-FROM ubuntu:22.04
-RUN apt-get update && apt-get upgrade --yes && apt-get install git python3.9 python3-pip --yes
-ADD . /code
-WORKDIR /code
-
-RUN pip install .
-RUN ape plugins install alchemy ens etherscan foundry hardhat infura ledger solidity template tokens trezor vyper
-RUN pip install eth-rlp==0.3.0
+#---------------------------------------------------------------------------------------------
+# Copyright (c) ApeWorX. All rights reserved.
```
fubuloubu, line 2: Usually has a copyright date, doesn't it?

Dockerfile
```diff
@@ -1,9 +1,43 @@
...
+LABEL maintainer="ApeWorX" \
+      org.label-schema.name="ape" \
+      org.label-schema.license="Apache" \
+      org.label-schema.description="Ape Ethereum Framework." \
+      org.label-schema.url="https://docs.apeworx.io/ape/stable/" \
+      org.label-schema.usage="https://docs.apeworx.io/ape/stable/userguides/quickstart.html#via-docker" \
+      org.label-schema.vcs-url="https://github.com/ApeWorX/ape" \
+      org.label-schema.docker.cmd="docker run --volume $HOME/.ape:/root/.ape --volume $HOME/.vvm:/root/.vvm --volume $HOME/.solcx:/root/.solcx --volume $PWD:/root/project --workdir /root/project apeworx/ape compile"
+
+RUN useradd --create-home --shell /bin/bash harambe
```
fubuloubu, line 22: lol nice

Dockerfile
```diff
@@ -1,9 +1,43 @@
...
+      etherscan \
+      foundry \
+      hardhat \
+      infura \
+      ledger \
+      solidity \
+      template \
+      tokens \
+      trezor \
+      vyper
```
fubuloubu: wonder if it should just be replaced with `pip install --no-cache-dir .[recommended-plugins]`

I know that it doesn't work by itself, but if it's the second command it might work

We were more explicit here because it was that (without the previous install) and thus was failing, but it would be great to control this all from one place

fubuloubu (approved):

Dockerfile
unparalleled-js: I tried that and it didn't work unfortunately. Same issue.

unparalleled-js: Additionally, I like the idea of possibly parametrizing this list as Dockerfile args so I can make a Starknet one easier.

fubuloubu: game for that. would be really slick to allow making a set of images that people can pull from with the right sets of plugins for their use cases

unparalleled-js: I did think of a way to do both.
We put a `recommended_plugins.txt` file and have `setup.py` create its list from it.
Then, we allow the Dockerfile an `ARG plugins_list="./recommended_plugins.txt"` so you can give Dockerfile custom lists but by default it uses the same values from `pip install .'[recommended_plugins]'`

sabotagebeats (approved):
Thank you for finding this fix!

Dockerfile
fubuloubu: Hmm, maybe we can read that file in to set the extra directly, so it's from one source (don't forget to add to project data files under `setup` as well)

docs/userguides/installing_plugins.md
```diff
@@ -49,6 +49,25 @@ To install plugins individually, run the following command:
...
+# These will be installed from the ape command
+ape-trezor==0.5.1
+ape-infura
+ape-hardhat
+
+# These will NOT be installed from the ape command
+other-non-ape-plugin-dependency
+```
+
+Then, point the command at the file to install them
```
unparalleled-js, line 65:
```suggestion
Then, point the `install` command at the file to install them:
```

docs/userguides/installing_plugins.md
```diff
@@ -49,6 +49,25 @@ To install plugins individually, run the following command:
 ape plugins install vyper solidity
 ```
 
+You can also install plugins from [requirements files](https://pip.pypa.io/en/latest/reference/requirements-file-format/).
+To do this, set-up your requirements file to include ape plugin package names (including `ape-` prefix):
```
fubuloubu, line 53: Not sure I'm fully on board with this change, basically it'll make two different ways of installing plugins, which both use different naming conventions for the plugins themselves

Perhaps the current way of installing plugins is deficient or overly restrictive.

Also gets me thinking if we'd want to switch over somehow to using `pyproject.toml` for ape config files... but that's a way bigger question

unparalleled-js: There are many ways to install plugins already. You can use requirements files today (I do) and use `pip` to install them. This allows you to use the `ape plugins install` command instead of `pip`

unparalleled-js: Perhaps we can switch the dockerfile to use `pip install` instead of `ape plugins install`

fubuloubu: Yeah, I like the distinction of `pip install -r requirements.txt` vs. `ape plugins install .` because the later only works with plugins so it's a bit safer, and also has the handling for when plugins are "trusted" vs. "community"

sabotagebeats (commented):
This is really cool!

Dockerfile
```diff
@@ -1,9 +1,37 @@
...
+      org.label-schema.vcs-url="https://github.com/ApeWorX/ape" \
+      org.label-schema.docker.cmd="docker run --volume $HOME/.ape:/root/.ape --volume $HOME/.vvm:/root/.vvm --volume $HOME/.solcx:/root/.solcx --volume $PWD:/root/project --workdir /root/project apeworx/ape compile"
+
+RUN useradd --create-home --shell /bin/bash harambe
+WORKDIR /home/harambe
+COPY . .
+
+RUN pip install --upgrade pip \
+    && pip install --no-cache-dir . \
+    && pip install -r recommended-plugins.txt \
```
unparalleled-js, line 29: The recommended plugins get installed using the `pip` approach now, so now change to `ape plugins install` is needed, as @fubuloubu pointed out that was not a good idea.

I am thinking this may not be the final implementation of this, but it is closer to what we want and achieves the goals.

recommended-plugins.txt
```diff
@@ -0,0 +1,12 @@
+ape-alchemy
```
unparalleled-js, line 1: The is shared between `Dockerfile` and `setup.py`

fubuloubu (approved):
Not quite a solution to #456, but gets us closer

Might need a docker build workflow for solution to #901

fubuloubu:
> > Not quite a solution to #456, but gets us closer
> 
> #456 says `Allow user to specify which plugins they want in the dockerfile via Dockerfile arguments`. The PR adds an ARG to the Dockerfile that allows you to specify plugins to install.. Can you help me understand what is missing in your eyes?
> 

Hmm, I guess just reading it differently. I agree with you now.

> > Might need a docker build workflow for solution to #901
> 
> Before this PR, there were times where if you ran `ape --version` (or any command), it would throw a stacktrace. That was why #901 got created. Now, with this PR, the docker build process runs `ape --version`. If that fails, the image's build will fail and thus it won't publish. Thus, I believe this fixes issue #901 as well.

Okay sweet, well that is good. Would be nice to have that validation in a PR, if for whatever reason the docker build doesn't work but the tests do

PR #1033, opened 2022-09-06
feat!: make `ProviderAPI.is_connected()` abstract

unparalleled-js:
### What I did

This should have been an API method but was avoided due to not wanting to have breaking changes.
In prep for 0.5, we want this now.

### Checklist


src/ape/api/providers.py
```diff
@@ -143,12 +143,11 @@ def disconnect(self):
         Disconnect from a provider, such as tear-down a process or quit an HTTP session.
         """
 
-    @property
+    @abstractmethod
```
unparalleled-js, line 146: - [ ] Change places where we catch `ProviderNotConnectedError` to first checking `is_connected()` instead.

fubuloubu (approved):

NotPeopling2day (approved):

PR #1035, opened 2022-09-06
refactor!: rename  `LogIInputABICollection.data` and `.topics` to `.data_abi_types` and `.topic_abi_types`

unparalleled-js:
### What I did

I had gotten really confused reading the `decode()` method one time recently and made a note that I wanted to rename this member variables so that I would not get confused again in the future.
It is because `decode` also takes topics and values as params but those are the actual values whereas the members vars are the ABI types.

**This is in prep for 0.5**

### How I did it

renames

### Checklist


fubuloubu (approved):

NotPeopling2day (approved):
LGTM

PR #1036, opened 2022-09-06
refactor!: warn instead of err when appending a known txn

unparalleled-js:
### What I did

It can be annoying when appending a known transaction and it error.
I think a warning is better so you don't fail everything.

### Checklist


fubuloubu (approved):
Yes, that seems appropiate!

PR #1037, opened 2022-09-06
refactor!: rename `ProviderAPI.get_transaction()` to `get_receipt()`

unparalleled-js:
### What I did

I always find it confusing that `get_transaction` returns a `ReceiptAPI`.
I think it would be easier to have the method named `get_receipt()`
Thoughts?

### Checklist


fubuloubu (approved):
Sure, I think that seems appropiate

Needs to analyze downstream effects on plugins/major projects, but I bet it's not too bad with some shoutouts in the release notes

Issue #1032, opened 2022-09-06
check ape has connected to the intended network


banteg:
### What went wrong?

similar to #1030, ape geth plugin should check it has indeed connected to the desired network and not just some network

<img width="478" alt="Screen Shot 2022-09-06 at 20 06 34" src="https://user-images.githubusercontent.com/4562643/188683707-d40bae90-edc5-448e-a0b6-d28c11c50533.png">

### How can it be fixed?

assert chain id matches the hardcoded value

fubuloubu:
suggested fix: detect if there is a mismatch during `.connect()`

Issue #1030, opened 2022-09-04
check ape has connected to the intended client


banteg:
### What went wrong?

ape just sent a mainnet transaction from a mainnet-fork console (or so i thought). i had mainnet set up as a remote in the config, but also had it running locally at `:8545`, which was also the default port for foundry. so it seems it connected to that and didn't check it wasn't a foundry client running there.

### How can it be fixed?

we could check `chain.provider.client_version` or call a fork-only method to make sure we are connected to the right thing and catch such misconfiguration issues early.

PR #1038, opened 2022-09-07
fix: validate Ape has connected to the right network and client

fubuloubu:
### What I did
Reorder how `Web3Provider` determines chain ID so that a fault is detected with mismatch of chain ID

fixes: #1030
fixes: #1032

### How I did it
We have some validation already in place for chain ID, and this solves a corner case in the logic for validating that the correct network and client is connecting when a connection already exists.

### How to verify it
1. Start anvil
2. `ape console --network :mainnet` (assumes `geth` is default)

Prior to change, (2) will work just fine, and fails after this change

### Checklist


src/ape/api/providers.py
```diff
@@ -681,16 +681,16 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
 
     @property
     def chain_id(self) -> int:
-        if self.network.name not in (
+        if hasattr(self.web3, "eth"):
```
unparalleled-js, line 684: `self.web3` may raise `ProviderNotConnectedError`

unparalleled-js: I think this is fine as long as `__repr__` does not error when disconnected, I think `__repr__` handles that though and we are tests.

I believe this works.
I am not sure what is with the tests though.

src/ape/api/providers.py
```diff
@@ -681,18 +681,27 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
...
+        except ProviderNotConnectedError:
+            if default_chain_id:
+                return default_chain_id
+
+            raise  # Original error
+
+        if default_chain_id:
+            return default_chain_id
+
+        raise ProviderNotConnectedError()
```
fubuloubu, line 705: Wouldn't this be unreachable code now?

unparalleled-js: This would happen if there was a `web3` (connected), but it had no `eth` property

banteg:
i think this doesn't address the problem, we need to check that the associated chain id in the network spec matches the chain id from the provider we connected to. this needs to be done on connection to prevent user error of accidentally connecting to the wrong endpoint. such error could have dire results such as sending test transactions on live network or caching the wrong abis.

unparalleled-js:
> i think this doesn't address the problem, we need to check that the associated chain id in the network spec matches the chain id from the provider we connected to. this needs to be done on connection to prevent user error of accidentally connecting to the wrong endpoint. such error could have dire results such as sending test transactions on live network or caching the wrong abis.

If you look at the connect method in `ape-geth`, you will see it already has logic for this, but the chain ID here was causing it  not to work.

With these changes, if you do something like:

```bash
ape console --network ethereum:rinkeby:geth
```

and have a URI for mainnet in your config file, you will get this output (including the error):

```bash
INFO: Connecting to existing Erigon node at 'https://erigon.yearn.science/'.
ERROR: (ProviderError) HTTP Connection does not match expected chain ID. Are you connected to 'rinkeby'?
```

However, if you switch to `main` branch, this bug is still present. Wen you try the same thing, it allow you to connect to the console.

Thus, I think this PR fixes the bug. Or no??

banteg (approved):
checked this pr and confirmed it works for me

banteg:
confirmed it works. maybe one suggestion is to also add numeric chain ids to the error.
```
INFO: Connecting to existing Erigon node at 'http://localhost:8545'.
ERROR: (ProviderError) HTTP Connection does not match expected chain ID. Are you connected to 'goerli'?
```
becomes
```
ERROR: (ProviderError) HTTP Connection chain ID (1) does not match expected chain ID (5). Are you connected to 'goerli'?
```

src/ape/api/providers.py
```diff
@@ -682,18 +682,28 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
...
-        elif hasattr(self.web3, "eth"):
-            return self.web3.eth.chain_id
+        try:
+            if hasattr(self.web3, "eth"):
+                return self.web3.eth.chain_id
 
-        else:
-            raise ProviderNotConnectedError()
+        except ProviderNotConnectedError:
+            if default_chain_id:
```
helloibis, line 698: I'm assuming there's never a case where `chain_id == 0` right?

helloibis (approved):
LGTM

src/ape/api/providers.py
unparalleled-js: I changed it to `if default_chain_id is not None` just in case!

helloibis (approved):
🚀

PR #1039, opened 2022-09-07
refactor: Use `AddressType` instead of `str` when possible

unparalleled-js:
### What I did

* Changes `txn.sender` and `txn.receiver` to be `AddressType` instead of `str`
* Changes `chain.contract.instance_at()` to only accept `AddressType` and not `str`
* Changes `ContractInstance` to only accept `AddressType` and not `str`
* Removed unneeded calls to `decode_address()`

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -705,7 +705,7 @@ class ContractCache(BaseManager):
         """
 
         if self.conversion_manager.is_type(address, AddressType):
-            contract_address: AddressType = address  # type: ignore
+            contract_address = cast(AddressType, address)
         else:
             try:
                 contract_address = self.conversion_manager.convert(address, AddressType)
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -680,7 +680,7 @@ class ContractCache(BaseManager):
 
     def instance_at(
         self,
-        address: Union[str, "AddressType"],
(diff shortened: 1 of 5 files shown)


src/ape/managers/chain.py
```diff
@@ -680,7 +680,7 @@ def get_container(self, contract_type: ContractType) -> ContractContainer:
 
     def instance_at(
         self,
-        address: Union[str, "AddressType"],
+        address: AddressType,
```
fubuloubu, line 683: This one is actually right, because this can accept string values that convert to `AddressType`

fubuloubu (approved):

src/ape/managers/chain.py
```diff
@@ -704,17 +704,17 @@ def instance_at(
...
 
-        try:
-            # Handles ENS domain names
-            address = self.conversion_manager.convert(address, AddressType)
-        except ConversionError:
-            address = address
-
-        address = self.provider.network.ecosystem.decode_address(address)
+        if self.conversion_manager.is_type(address, AddressType):
+            contract_address: AddressType = address  # type: ignore
```
fubuloubu, line 708: correct thing to do here is `cast`

NotPeopling2day (dismissed):

fubuloubu (approved):

NotPeopling2day (approved):

Issue #997, opened 2022-08-18
Cache: Refactor contract event args into separate model


NotPeopling2day:
### Overview

When a user wants to query contract events, the Dataframe should return event args as columns instead of a single dict. Currently the contract event args are stored at a singular JSON field. We need to research the proper structure for storing and returning this data such that they are represented to a user as columns for each argument. 

I.e. | to | from | value | instead of {"to": ..., "from": ..., "value": ...}
      | ... | ...      | ...       |

A consideration is to create a dynamic model that can become an intermediary for parsing the values into columns.

### Specification

- [ ] Research the best method of storing values and retrieving values
- [ ] Implement solution (add more detail once research is completed)

fubuloubu:
ideas: more abstractly, let `EventAPI` be something driven by `EcosystemAPI` which let's them specify different event_ids and topics

PR #1042, opened 2022-09-09
feat: adding new `BaseModel` for `event_arguments` in `ContractLog`

johnson2427:
### What I did

For us to be able to better manage `ContractEvents` queries for the caching system, creating a new `BaseModel` that `ContractLog` will inherit so we can avoid storing a JSON datatype.

fixes: #997 

### How I did it

Have to run through many contract deployments to see what datatypes can be available. Then when that is better understood, applying the `EventArguments` model will be better defined. Will also allow for additional tests to be made to better test fringe events.

### How to verify it

To be determined

### Checklist


johnson2427:
Very preliminary, more spike research to be done the rest of the week for this

johnson2427:
Hey all, I made quite a few changes that I was getting ready to commit, but the test suite is broken from the command line. And it looks like the main branch has broken tests as well. I'm trying to resolve these issues at the moment, but I'm a bit hung up. When I run these files individually from pytest, they all pass.

tests/functional/test_ecosystem.py
```diff
@@ -75,6 +75,7 @@ def test_transaction_acceptance_timeout(temp_config, config, networks):
 def test_decode_logs(ethereum, vyper_contract_instance):
     abi = vyper_contract_instance.NumberChange.abi
     result = [x for x in ethereum.decode_logs([LOG], abi)]
+    breakpoint()
```
unparalleled-js:
```suggestion
```

src/ape/types/__init__.py
```diff
@@ -227,8 +186,47 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
         return self.event_arguments.get(item, default)
 
 
-class ContractLogEventArguments(BaseModel):
-    pass
+class ContractLog(EventArguments):
```
unparalleled-js, line 189: What happens in the `dict` / `JSON` version of this class when there are naming collisions between event arguments and the normal properties on this class (e.g. `contract_address`)?

For example, if I defined a topic named `contract_address` in my contract Event declaration, there would **not** be a way to define both values in a raw dict of a `ContractLog`, thus causing serializations issues with this class.

Or am I misunderstanding?

unparalleled-js (changes requested):
I am worried about the effects of naming collisions in the raw model's keys between the custom event arguments and the normal class properties of `ContractLog`. That is why we kept them separate.

src/ape/types/__init__.py
johnson2427: you are absolutely correct here, this is why I'm so stuck in the weeds with this task. I'm doing a really deep recon here, trying to find all possible values, and then I'm trying to figure out how to handle collisions. I'm starting to lean on having 2 `BaseModel`'s and then pulling in `EventArguments` as a pydantic model instead of as a dict like we were before. Then I would still be able to manage the cache database for `ContractEvents`. Idk, I almost want to use a document database for this. It almost doesn't make sense to attempt to flatten this because `event_arguments` are so variable. @fubuloubu what do you think?

johnson2427:
the cached_property just seems like event_arguments with extra steps

src/ape/types/__init__.py
johnson2427: This is a stale conversation, and we have addressed this topic a few times. I am going to move this to resolved and continue working on this issue here.

src/ape/types/__init__.py
```diff
@@ -143,50 +144,16 @@ def encode_topic_value(abi_type, value):
...
-
-        return value
-
-    @validator("contract_address", pre=True)
-    def validate_address(cls, value):
-        from ape import convert
-
-        return convert(value, AddressType)
+    @cached_property
+    def topics(self):
```
unparalleled-js, line 152: I don't see `topics` being used anywhere.

I would like to know more about why we need the data structure this way. Perhaps by seeing it in action that will clear things up

unparalleled-js, line 152: - [ ] add test for `topics` property

src/ape/types/__init__.py
```diff
@@ -226,6 +193,57 @@ def __getitem__(self, item: str) -> Any:
...
+    @validator("block_number", "log_index", "transaction_index", pre=True)
+    def validate_hex_ints(cls, value):
+        if not isinstance(value, int):
+            return to_int(value)
+
+        return value
+
+    @validator("contract_address", pre=True)
+    def validate_address(cls, value):
+        from ape import convert
```
unparalleled-js, line 243: Thought (not PR feedback): It might be nice to have an easier way to use Ape's conversion functionality without having to have localized imports like this.

src/ape/types/__init__.py
```diff
@@ -226,6 +193,57 @@ def __getitem__(self, item: str) -> Any:
...
+    class Config:
+        # NOTE: Due to https://github.com/samuelcolvin/pydantic/issues/1241 we have
+        # to add this cached property workaround in order to avoid this error:
+
+        #    TypeError: cannot pickle '_thread.RLock' object
+
+        keep_untouched = (cached_property,)
+
+
+class ContractLog(EventArguments):
```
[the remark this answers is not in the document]
johnson2427: I had originally planned a completely different object in my head and started there. This could really just be moved to the ContractLog, and keep ContractLog a BaseModel. Let me know if you want to do that after looking at the latest changes

src/ape/types/__init__.py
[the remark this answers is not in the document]
johnson2427: This has changed so I will close this conversation and we can review the newer stuffs

src/ape/types/__init__.py
```diff
@@ -143,50 +144,16 @@ def encode_topic_value(abi_type, value):
...
-
-    @validator("contract_address", pre=True)
-    def validate_address(cls, value):
-        from ape import convert
-
-        return convert(value, AddressType)
+    @cached_property
+    def topics(self):
+        topics = dict()
+        for key in enumerate(self.event_arguments):
```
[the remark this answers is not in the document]
johnson2427: Obsolete with latest changes

src/ape/types/__init__.py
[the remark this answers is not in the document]
johnson2427: obsolete with latest changes

src/ape/types/__init__.py
```diff
@@ -184,6 +183,54 @@ class ContractLog(BaseModel):
...
+
+    topics: List = Field(default=[])
+    """The hashed topics of the event."""
+
+    @property
+    def event_name(self) -> str:
+        return self.abi.name
+
+    @cached_property
+    def event_arguments(self) -> Dict[str, Any]:
```
johnson2427, line 197: I like this implementation. I wanted something like this but wasn't sure how to attack it. This is perfect

src/ape/types/__init__.py
```diff
@@ -184,6 +183,54 @@ class ContractLog(BaseModel):
...
+        """The arguments to the event, including both indexed and non-indexed data."""
+        abis = LogInputABICollection(abi=self.abi)
+        try:
+            return abis.decode(self.topics, self.data)
+        except InsufficientDataBytes:
+            logger.debug("failed to decode log data for %s", self.json(), exc_info=True)
+            return {}
+
+    @cached_property
+    def topic_0(self) -> Optional[str]:
```
johnson2427, line 207: I know I did this, I just don't know if this is what @fubuloubu wanted

src/ape/types/__init__.py
```diff
@@ -197,15 +244,21 @@ def validate_address(cls, value):
 
         return convert(value, AddressType)
 
-    @property
+    @cached_property
```
johnson2427, line 247: I think we have a lot of `@property` calls that should probably be `@cached_property`. A problem for another day

src/ape/utils/abi.py
```diff
@@ -5,7 +5,11 @@
 from eth_abi import decode, grammar
 from eth_utils import decode_hex, to_checksum_address
 from ethpm_types import HexBytes
-from ethpm_types.abi import ABIType, EventABI, EventABIType, MethodABI
+from ethpm_types.abi import ABIType, EventABI, MethodABI
+from pydantic import validator
+
+from ape.utils.basemodel import BaseModel
+from ape.utils.misc import cached_property
```
johnson2427, line 12: This is a functools method, am I using the wrong wrapper?

src/ape/utils/abi.py
```diff
@@ -245,15 +249,27 @@ def is_dynamic_sized_type(abi_type: Union[ABIType, str]) -> bool:
...
 
-class LogInputABICollection:
-    def __init__(self, abi: EventABI):
-        self.abi = abi
-        self.topic_abi_types = [i for i in abi.inputs if i.indexed]
-        self.data_abi_types: List[EventABIType] = [i for i in abi.inputs if not i.indexed]
+class LogInputABICollection(BaseModel):
+    abi: EventABI
+
+    class Config:
```
johnson2427, line 255: I only suggest moving this to the bottom of the object. I looks better IMO. But it doesn't bother me either way

tests/functional/test_ecosystem.py
```diff
@@ -76,24 +76,38 @@ def test_decode_logs(ethereum, vyper_contract_instance):
...
-            "b": HexBytes("0xabffd4675206dab5d04a6b0d62c975049665d1f512f29f303908abdd20bc08a1"),
-            "prevNum": 0,
-            "dynData": "Dynamic",
-        },
-        "transaction_hash": "0x74dd040dfa06f0af9af8ca95d7aae409978400151c746f55ecce19e7356cfc5a",
-        "block_number": 11093676,
-        "block_hash": "0x2c99950b07accf3e442512a3352a11e6fed37b2331de5f71b7743b357d96e4e8",
-        "log_index": 0,
-        "transaction_index": 0,
+    actual = result[0]
```
johnson2427, line 79: Such a much cleaner implementation. We should apply this everywhere in our test suite

johnson2427 (commented):
I'm not allowed to approve, but this is so well done @unparalleled-js. Thank you for all your help!

src/ape/utils/abi.py
unparalleled-js: I put it here for consistency's sake because all the other models in the codebase put it directly below the fields

src/ape/utils/abi.py
unparalleled-js: We export `@cached_property` from `ape.utils` for historical reasons, possibly because of Python 3.7 complications and to encourage its usage.

src/ape/utils/abi.py
unparalleled-js: I put it here for consistency's sake because all the other models in the codebase put it directly below the fields

src/ape/contracts/base.py
```diff
@@ -391,7 +391,7 @@ def __len__(self):
 
     def query(
         self,
-        *columns: List[str],
+        *columns: str,
```
fubuloubu, line 394: TIL https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

src/ape/types/__init__.py
```diff
@@ -184,6 +183,54 @@ class ContractLog(BaseModel):
...
+            return self.topics[2]
+
+        return None
+
+    @cached_property
+    def topic_3(self) -> Optional[str]:
+        if self.topics and len(self.topics) > 3:
+            return self.topics[3]
+
+        return None
```
fubuloubu, line 232: Probably don't need this

johnson2427: duplicate of the comment above

src/ape/contracts/base.py
johnson2427: This will remain the same as it currently is defined

src/ape/types/__init__.py
```diff
@@ -157,32 +158,62 @@ class ContractLog(BaseModel):
...
     The index of the transaction's position when the log was created.
     Is `None` when from the pending block.
     """
+    transaction_index: Optional[int] = None
+
+    """The data hash of the event."""
+    data: Any = Field(default=None)
+
+    """The hashed topics of the event."""
+    topics: List = Field(default=[])
```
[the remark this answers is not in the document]
johnson2427:
```python
def decode(self, topics: List[str], data: str) -> Dict:
```
This decode method, which is used on line 201 to define the `event_arguments` requires that the `List` values be of type `str`

src/ape/types/__init__.py
```diff
@@ -157,32 +158,62 @@ class ContractLog(BaseModel):
...
 
-    transaction_index: Optional[int] = None
     """
     The index of the transaction's position when the log was created.
     Is `None` when from the pending block.
     """
+    transaction_index: Optional[int] = None
+
+    """The data hash of the event."""
+    data: Any = Field(default=None)
```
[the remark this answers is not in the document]
johnson2427:
```python
def decode(self, topics: List[str], data: str) -> Dict:
```
This decode method, which is used on line 201 to define the `event_arguments` requires that `data` be of type `str`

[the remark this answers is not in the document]
unparalleled-js: I made it bytes and fixed it led to fining a bug that got fixed! Thanks!

src/ape/contracts/base.py
fubuloubu: Yup, I had thought the proper way to specify it was `*args: List[type]` but `*args: type` makes more sense

src/ape/contracts/base.py
```diff
@@ -414,6 +414,7 @@ def query(
         Returns:
             pd.DataFrame
         """
+        column_list = list(columns)
```
fubuloubu, line 417: Might need this to handle empty columns
```suggestion
        column_list = list(columns or [])
```

unparalleled-js: `columns` is always a tuple since it is the `*args` thing and

```python
>>> list(())
[]
```

so I think this is fine.

fubuloubu: wonder if the data type should have just been tuple in `QueryType`

unparalleled-js: For `_BaseQuery.columns`? Yes, it can probably be a `Tuple[...]` because after the columns list is built, it doesn't need to change in the `_BaseQuery`. However, here we are building the list from the `__fields__` so it is dynamic at this point, but it can be frozen later.

fubuloubu: Can think about this later lol

johnson2427:
@fubuloubu are we good to go here?

fubuloubu (approved):

fubuloubu (commented):
It might be nice to have the indexed arguments in a separate structure from `event_arguments` (or a sub-field) so that they can be searched more easily from this object.

johnson2427:
> It might be nice to have the indexed arguments in a separate structure from `event_arguments` (or a sub-field) so that they can be searched more easily from this object.

Isn't that what topics and data are? I thought that's what we were doing with those two new datatypes

fubuloubu:
> > It might be nice to have the indexed arguments in a separate structure from `event_arguments` (or a sub-field) so that they can be searched more easily from this object.
> 
> Isn't that what topics and data are? I thought that's what we were doing with those two new datatypes

Might need a higher level discussion, basically when I want to filter entries when performing a query (e.g. `ContractInstance.Event.query(..., filter_args={"argname": 1, ...})`) I want to do it using the direct argument names of the event, but when it gets stored in the database we are storing in this tighter `topics + data` format, so some code is needed to help translate `filter_args` into the right query syntax so basically it's like `WHERE topic0 == b"\x00...\x01"` instead of what the user specifies (filter by `argname == 1`)

johnson2427:
> > > It might be nice to have the indexed arguments in a separate structure from `event_arguments` (or a sub-field) so that they can be searched more easily from this object.
> > 
> > 
> > Isn't that what topics and data are? I thought that's what we were doing with those two new datatypes
> 
> Might need a higher level discussion, basically when I want to filter entries when performing a query (e.g. `ContractInstance.Event.query(..., filter_args={"argname": 1, ...})`) I want to do it using the direct argument names of the event, but when it gets stored in the database we are storing in this tighter `topics + data` format, so some code is needed to help translate `filter_args` into the right query syntax so basically it's like `WHERE topic0 == b"\x00...\x01"` instead of what the user specifies (filter by `argname == 1`)

Okay, can we get on a call with @unparalleled-js so we can map this out a little further?

johnson2427:
@fubuloubu I've been thinking about this filter_args issue, and after going through all the necessary changes to get to this point, that filter_args argument is going to have to be handled with the web3 provider. Also, because we are using sqlite, I cannot store a list. And since its a 1 to many relationship, we could just make another table. Or we have to encode this to be a string. Or I need a better understanding of what I need to do with topics

fubuloubu:
There should only ever be 4 elements to the list (limitation of the chain itself), so I would say you can flatten the list to 4 bytes32 columns

PR #1049, opened 2022-09-13
fix: allow ecosystems to still load when default provider or network not found

unparalleled-js:
### What I did

If a provider plugin has a single bug in it currently, every core plugin that uses the `network_option` fails to load.
This is because the config loads in the plugin attempting to set the default but the default provider "does not exist" but it fails to load.

### How I did it

Instead of failing to return any ecosystems at all when attempting to set the default provider fails with a `NetworkError`, log the error and move on.

### Checklist

### DIFF
--- a/src/ape_geth/providers.py
+++ b/src/ape_geth/providers.py
@@ -215,20 +215,19 @@ class GethProvider(Web3Provider, UpstreamProvider):
             if not self._web3.isConnected():
                 self._geth.disconnect()
                 raise ConnectionError("Unable to connect to locally running geth.")
+        elif "geth" in self.client_version.lower():
+            logger.info(f"Connecting to existing Geth node at '{self._clean_uri}'.")
+        elif "erigon" in self.client_version.lower():
+            logger.info(f"Connecting to existing Erigon node at '{self._clean_uri}'.")
+            self.concurrency = 8
+            self.block_page_size = 40_000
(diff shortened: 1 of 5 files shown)


src/ape_geth/providers.py
```diff
@@ -215,20 +215,19 @@ def connect(self):
             if not self._web3.isConnected():
                 self._geth.disconnect()
                 raise ConnectionError("Unable to connect to locally running geth.")
+        elif "geth" in self.client_version.lower():
+            logger.info(f"Connecting to existing Geth node at '{self._clean_uri}'.")
```
sabotagebeats, line 219: I am wondering if this could be reused in this way to simplify this if statement at all?
`logger.info(f"Connecting to existing '{self.client_version.lower()}' node at '{self._clean_uri}'.")`

fubuloubu (dismissed):

src/ape_geth/providers.py
unparalleled-js: I messed around this. It is kind of tricky because of the `else:` statement where the client is unknown as well as the settings that get set for Nethermind and Erigon. However, we can share a helper method to enforce a consistent log

fubuloubu (approved):

sabotagebeats (approved):
changes look good to me

PR #1050, opened 2022-09-15
feat: show gas reports from Ethereum transaction receipts

unparalleled-js:
### What I did

First step to showing gas reports in tests.
Adds a way to show one on a receipt.

### How I did it

* Share code with tracing stuff.

### How to verify it

Get a txn receipt from somewheres, like gran a txn hash from Etherscan and get the receipt via `provider.get_receipt()` and then call `show_gas_usage()`.

See my trace project as an example and could even use that, that is what I do: https://github.com/unparalleled-js/ape-trace-project#gas

### Checklist


docs/userguides/transactions.md
```diff
@@ -231,6 +231,34 @@ receipt = networks.provider.get_receipt(txn_hash)
...
+                           DAI Gas
+┏━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┓
+┃ Method    ┃ Times called ┃ Min.  ┃ Max.  ┃ Mean  ┃ Median ┃
+┡━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━┩
+│ balanceOf │ 4            │ 1302  │ 1302  │ 1302  │ 1302   │
+│ transfer  │ 5            │ 6974  │ 30374 │ 20174 │ 26174  │
+│ allowance │ 1            │ 1377  │ 1377  │ 1377  │ 1377   │
+│ approve   │ 1            │ 22414 │ 22414 │ 22414 │ 22414  │
+│ burn      │ 1            │ 11946 │ 11946 │ 11946 │ 11946  │
+│ mint      │ 1            │ 25845 │ 25845 │ 25845 │ 25845  │
```
fubuloubu, line 258: Does this only show methods called of the given target contract? or all call node targets inside of the call?

If the latter, it seems like a column for contract type would be necessary

fubuloubu (approved):

docs/userguides/transactions.md
unparalleled-js: It is `1` table per contract, where the title of the contract is above the table. This table here is only for the `DAI `contract. Every target gets its own table. So if you had:

```
ContractA calls ContractB
ContractA calls ContractC
ContractC calls ContractB
```

, there would only be `1` table for `ContractB` containing data from both `ContractA` calling it as we al as `ContractC` calling it. And `ContractA` and `ContractC` would have their own tables with their own methods on it.

That is the intent anyway!

unparalleled-js: I just did some manual testing:

If you run `ape run sim_gas` from this project: https://github.com/unparalleled-js/ape-trace-project, it calls a method in `ContractA` that calls a method in `ContractB` that calls a method in `ContractC. Then, it directly calls the method in `ContractC`. The resulting report looks like this:

```
                            ContractA Gas                             
┏━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┓
┃ Method             ┃ Times called ┃ Min.  ┃ Max.  ┃ Mean  ┃ Median ┃
┡━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━┩
│ mergeGasReportTest │ 1            │ 13166 │ 13166 │ 13166 │ 13166  │
└────────────────────┴──────────────┴───────┴───────┴───────┴────────┘
                              SYMBOL Gas                              
┏━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━━━┓
┃ Method                ┃ Times called ┃ Min. ┃ Max. ┃ Mean ┃ Median ┃
┡━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━━━┩
│ callContractCMethodC3 │ 1            │ 5327 │ 5327 │ 5327 │ 5327   │
└───────────────────────┴──────────────┴──────┴──────┴──────┴────────┘
                      ContractC Gas                      
┏━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━┳━━━━━━━━┓
┃ Method   ┃ Times called ┃ Min. ┃ Max. ┃ Mean ┃ Median ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━━━┩
│ methodC3 │ 2            │ 166  │ 166  │ 166  │ 166    │
└──────────┴──────────────┴──────┴──────┴──────┴────────┘
```

so the resulting tables are per-contract as intended!

fubuloubu (approved):

PR #1053, opened 2022-09-15
chore: update gas limit type hints

helloibis:


src/ape_ethereum/ecosystem.py
```diff
@@ -88,7 +88,7 @@ class Config:
         smart_union = True
 
     @validator("gas_limit", pre=True)
-    def validate_gas_limit(cls, value: GasLimit) -> Union[Literal["auto", "max"], int]:
+    def validate_gas_limit(cls, value: GasLimit) -> GasLimit:
```
helloibis, line 91: Technically the input and output types are different since the output is only `Union[Literal["auto", "max"], int]` 

Does that need to be reflected here with perhaps two type definitions?

unparalleled-js: I think it is ok to share the 1 type

unparalleled-js (approved):
Lgtm!

PR #1060, opened 2022-09-20
fix: strange error when getting logs from multiple addresses when addresses list empty

unparalleled-js:
### What I did

Ran into a concurrency failure because the address list was empty.
Make fail early to prevent such failure.
Also never let `max_thread` be < 0 just to be extra safe.
Additionally, only convert non-AddressTypes as a perf

### Checklist


fubuloubu (approved):
side note: will probably have to revisit thread safety if we make change to async for v1

PR #1069, opened 2022-09-22
refactor: use `is_file()` and `is_dir()` over `exists()` whenever makes sense

unparalleled-js:
### What I did

Was debugging an issue with compiling and I fixed it in `ape-solidity` (about to open PR). Except while fixing that bug, also fixed all of these, so this is small PR to change `exists()` usage to be `is_file()` or `is_dir()` depending on the situation.

### Checklist


fubuloubu (approved):
Good catch!

PR #1074, opened 2022-09-23
feat: adds `chain.get_receipt()` method

unparalleled-js:
### What I did

**I almost have a solid end-to-end on gas reporting in tests.**
However - this is one bit of logic I'd like to extract to its own PR to have it reviewed on its own...

Adds another index to the `AccountHistory` such that we can **rapidly** look up a cached `Receipt` by its `transaction_hash`. Then, attempts to do this in `chain.get_receipt()` so save an RPC call.

**NOTE**: The transactions presently in the `AccountHistory` should all be valid on exist on the chain. Because, when we restore to an older snapshot, we remove those transactions. If the provider cannot find the receipt, it will use the RPC call method instead. Any receipts send by your accounts will for sure be already cached!!! This greatly speeds up testing and will **definitely speed up the gas reporting work I am doing**.

### How I did it

* Add a second map that allows looking up receipts by txn hash in `AccountHistory`.
* Check this map during `get_receipt()` call in `Web3Provider`.

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -424,19 +424,27 @@ class AccountHistory(BaseManager):
             h: r for h, r in self._hash_to_receipt_map.items() if r.block_number <= block_number
         }
 
-    def get_receipt(self, transaction_hash: str) -> Optional[ReceiptAPI]:
+    def get_receipt(self, transaction_hash: str) -> ReceiptAPI:
         """
         Get a receipt from the history by its transaction hash.
+        If the receipt is not currently cached, will use the provider
+        to retrieve it.
 
         Args:
             transaction_hash (str): The hash of the transaction.
 
         Returns:
-            Optional[:class:`~ape.api.transactions.ReceiptAPI`]: The receipt, if it
-              is present in the transaction history. Else, ``None``.
+            :class:`~ape.api.transactions.ReceiptAPI`: The receipt.
         """
 
-        return self._hash_to_receipt_map.get(transaction_hash)
+        receipt = self._hash_to_receipt_map.get(transaction_hash)
+        if not receipt:
+            # TODO: Replace with query manager once supports receipts
+            #  instead of transactions.
+            receipt = self.provider.get_receipt(transaction_hash)
+            self.append(receipt)
+
+        return receipt
 
 
 class ContractCache(BaseManager):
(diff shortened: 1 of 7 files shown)


unparalleled-js:
Another perf stat:  In ape core's test suite locally for me:

Before changes:  58.46s 
After changes: 44.84s

unparalleled-js:
Perf:

Getting a large trace _with_ changes:

```
ape -v error run sim --network ethereum:local:hardhat  15.14s user 1.35s system 118% cpu 13.966 total
```

Getting a large trace _without_ changes:

```
ape -v error run sim --network ethereum:local:hardhat  16.20s user 1.47s system 117% cpu 14.995 total
```

src/ape/managers/chain.py
```diff
@@ -1075,3 +1092,12 @@ def set_balance(self, account: Union[BaseAddress, AddressType], amount: Union[in
...
 
         return self.provider.set_balance(account, amount)
+
+    def get_receipt(self, transaction_hash: str) -> ReceiptAPI:
+        cached_receipt = self.chain_manager.account_history.get_receipt(transaction_hash)
+        if cached_receipt:
+            return cached_receipt
+
+        receipt = self.provider.get_receipt(transaction_hash)
+        self.chain_manager.account_history.append(receipt)
```
unparalleled-js, line 1102: Once the query manager gives us receipts instead of transactions, we can use the query manager here.

src/ape/managers/chain.py
```diff
@@ -1075,3 +1092,12 @@ def set_balance(self, account: Union[BaseAddress, AddressType], amount: Union[in
             amount = int(amount, 16)
 
         return self.provider.set_balance(account, amount)
+
+    def get_receipt(self, transaction_hash: str) -> ReceiptAPI:
```
unparalleled-js, line 1096: Note that this was originally intended to be here in the original ticket! (was called `get_transaction`) then:
https://github.com/ApeWorX/ape/issues/129

Just thought I'd add that, it feels validating

src/ape/managers/chain.py
```diff
@@ -416,10 +416,27 @@ def revert_to_block(self, block_number: int):
...
+
+        Args:
+            transaction_hash (str): The hash of the transaction.
+
+        Returns:
+            Optional[:class:`~ape.api.transactions.ReceiptAPI`]: The receipt, if it
+              is present in the transaction history. Else, ``None``.
+        """
+
+        return self._hash_to_receipt_map.get(transaction_hash)
```
fubuloubu, line 439: Curious if this is better 
```suggestion
        return self._hash_to_receipt_map.get(transaction_hash) or self.provider.get_receipt(transaction_hash)
```

src/ape/managers/chain.py
fubuloubu: One of the things I wanted to say

fubuloubu (approved):

fubuloubu (approved):

PR #1075, opened 2022-09-23
feat!: assert on dev messages

helloibis:
Opening this preemptively to track changes, git rebases, etc

#957

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -73,20 +73,28 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
           :class:`~ape.types.signatures.TransactionSignature` (optional): The signed transaction.
         """
 
-    def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
+    def call(
+        self,
+        txn: TransactionAPI,
+        send_everything: bool = False,
+        raise_on_fail: bool = True,
+    ) -> ReceiptAPI:
         """
         Make a transaction call.
 
         Raises:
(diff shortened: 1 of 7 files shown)


helloibis:
Currently tracking the breaking changes list in Linear for the 0.6.0 release, I'll update this PR's description once the subtasks are done and we're ready to merge

fubuloubu (approved):
What makes this a breaking change?

helloibis:
> What makes this a breaking change?

@fubuloubu quick summary is: no behavior is changing in a breaking way (since the default `raise_on_fail` for `get_receipt()` and `await_confirmations()` will be `False`) but because of the added kwarg it breaks provider plugins

Once this PR is out of draft I'll be updating its description with a full run-down

Jules and I discussed a bit on maybe using `**kwargs` or similar to mitigate the breaking signature change, but we ended up deciding to hold until a v0.6.0 because of the break and how many repos this involves. The dependent pieces that will all go out for v0.6.0 are:

- `ape-vyper` updates to add the source map (https://github.com/ApeWorX/ape-vyper/pull/51), and same for solidity I think
- `ape` updates to add the feature (which will be supported by the aforementioned compiler plugins adding sourcemap)
- The remaining provider plugins are updated to reflect the new `ape` core version dependency (the ones that just had those feature branches created: Hardhat, Foundry, etc.)

The changes in the open `ape` PR draft aren't breaking changes so far but I'll have to see where I land. I think though that those changes should remain non-breaking

helloibis:
Closing in favor of https://github.com/ApeWorX/ape/pull/1125

Issue #350, opened 2021-12-21
Testing: Gas reporting in tests


NotPeopling2day:
### Overview

Create a gas report when running tests

### Specification

```python
def get_gas_usage(self):
        """
        """

        calltree = self._calltree(show_internal=True)
        # TODO: gas report logic
        # {Address: {Calldata: [GasUsed, ...]}}
        return calltree
```

https://github.com/eth-brownie/brownie/blob/master/brownie/test/output.py#L62

### Dependencies

#345
#347

fubuloubu:
depends on #700

fubuloubu:
ipfs://bafybeihapxqz43vadpqicxulwihrj6pqnvrxavf7hruk6xk45es57ds3hi/forge/gas-reports.html

unparalleled-js:
I am guessing we want this on the receipt?

```python
receipt.show_gas_report()
```

This ticket could use some more acceptance criteria

unparalleled-js:
> ipfs://bafybeihapxqz43vadpqicxulwihrj6pqnvrxavf7hruk6xk45es57ds3hi/forge/gas-reports.html

Ah interesting. Do we want something similar here?

```yaml
test:
    gas_reports: [MyContract]
```

```bash
ape test --gas-reports
```

unparalleled-js:
> ipfs://bafybeihapxqz43vadpqicxulwihrj6pqnvrxavf7hruk6xk45es57ds3hi/forge/gas-reports.html

Does this actually require tracing? This table looks like it tracks the calls and `gas_used` from the receipts and calculates some stats on them.

fubuloubu:
> > ipfs://bafybeihapxqz43vadpqicxulwihrj6pqnvrxavf7hruk6xk45es57ds3hi/forge/gas-reports.html
> 
> Does this actually require tracing? This table looks like it tracks the calls and `gas_used` from the receipts and calculates some stats on them.

What you want is all internal and external calls that were made, and their relative gas usage

unparalleled-js:
Make progress here but haven't gotten to the hard part yet I don't think.

PR #1083, opened 2022-09-27
feat: add `--gas` flag to `ape test` command to output gas reports after tests

unparalleled-js:
### What I did

fixes: #350 

### How I did it

* Add `--gas` flag to `ape test` command
* In function isolation fixture, build the tree on each incoming receipt
* In terminal reporting pytest hook, output the gas report

### How to verify it

1. Make sure to use a provider like `ape-hardhat` or `ape-foundry` that supports tracing.
2. Go to your project with tests
3. Run `ape test --gas` or however you run your tests, just include `--gas`.

### Checklist

### DIFF
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -279,7 +279,7 @@ ape test --network ethereum:local:hardhat --gas
 At the end of test suite, you will see tables such as:
 
 ```sh
-                            FundMe Gas
+                            FundMe.vy Gas
 
   Method           Times called    Min.    Max.    Mean   Median
  ────────────────────────────────────────────────────────────────
@@ -294,7 +294,7 @@ At the end of test suite, you will see tables such as:
  ───────────────────────────────────────────────────────
   to:test0              2   2400   9100   5750     5750
 
-                     TestContractVy Gas
+                     TestContract.Vy Gas
 
   Method      Times called    Min.    Max.    Mean   Median
  ───────────────────────────────────────────────────────────
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -266,3 +266,37 @@ When you exit a provider's context, Ape **does not** disconnect the provider.
 When you re-enter that provider's context, Ape uses the previously-connected provider.
 At the end of the tests, Ape disconnects all the providers.
 Thus, you can enter and exit a provider's context as much as you need in tests.
+
+## Gas Reporting
+
+To include a gas report at the end of your tests, you can use the `--gas` flag.
+**NOTE**: This feature requires using a provider with tracing support, such as [ape-hardhat](https://github.com/ApeWorX/ape-hardhat).
+
+```bash
+ape test --network ethereum:local:hardhat --gas
+```
+
+At the end of test suite, you will see tables such as:
+
+```sh
+                            FundMe Gas
+
+  Method           Times called    Min.    Max.    Mean   Median
+ ────────────────────────────────────────────────────────────────
+  fund                        8   57198   91398   82848    91398
+  withdraw                    2   28307   38679   33493    33493
+  changeOnStatus              2   23827   45739   34783    34783
+  getSecret                   1   24564   24564   24564    24564
+
+                  Transferring ETH Gas
+
+  Method     Times called   Min.   Max.   Mean   Median
+ ───────────────────────────────────────────────────────
+  to:test0              2   2400   9100   5750     5750
+
+                     TestContractVy Gas
+
+  Method      Times called    Min.    Max.    Mean   Median
+ ───────────────────────────────────────────────────────────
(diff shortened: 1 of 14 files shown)


unparalleled-js:
- [ ] Fix regression in trace output (failing in hardhat tests)

docs/userguides/testing.md
```diff
@@ -294,7 +294,7 @@ At the end of test suite, you will see tables such as:
  ───────────────────────────────────────────────────────
   to:test0              2   2400   9100   5750     5750
 
-                     TestContractVy Gas
+                     TestContract.Vy Gas
```
fubuloubu, line 297:
```suggestion
                     TestContract.vy Gas
```

unparalleled-js:
Tests: https://github.com/ApeWorX/ape-hardhat/pull/98

src/ape/pytest/fixtures.py
```diff
@@ -69,17 +123,72 @@ def _isolation(self) -> Iterator[None]:
...
-    _module_isolation = pytest.fixture(_isolation, scope="module")
-    _class_isolation = pytest.fixture(_isolation, scope="class")
-    _function_isolation = pytest.fixture(_isolation, scope="function")
+        _silence_connection_failure(lambda: self.chain_manager.restore(snapshot_id))
+
+    def _get_block_number(self) -> Optional[int]:
+        return _silence_connection_failure(lambda: self.provider.get_block("latest").number)
+
+
+def _silence_connection_failure(fn: Callable) -> Optional[Any]:
```
unparalleled-js, line 138: - [ ] making a more useful decorator for this

src/ape/pytest/runners.py
```diff
@@ -153,14 +156,20 @@ def pytest_collection_finish(self, session):
...
+        if self.pytest_config.getoption("--gas"):
+            terminalreporter.section("Gas Profile")
+            gas_report = self.receipt_capture.gas_report
+            if gas_report:
+                tables = parse_gas_table(gas_report)
+                rich_print(*tables)
+            else:
+                terminalreporter.write_line(f"{LogLevel.WARNING.name}: No gas usage data found.")
+
         if self._provider_is_connected:
```
fubuloubu, line 173: Do we want this under `pytest_terminal_summary` too?

unparalleled-js: It at least needs to happen _after_ outputting the tables, and before it wasn't. There is probably a better a hook for it though.

fubuloubu: Well, my worry is if there is somehow a flag to skip this hook from operating, it might break the cleanup code

unparalleled-js: Resolved in 3d36109

fubuloubu: Nice!

fubuloubu (dismissed):

unparalleled-js:
- [ ] Fix issue where sessional and module-scoped fixtures containing receipts don't show up in gas report

src/ape/pytest/fixtures.py
```diff
@@ -54,32 +69,114 @@ def project(self) -> ProjectManager:
...
+
+    @allow_disconnected
+    def _restore(self, snapshot_id: SnapshotID):
+        if snapshot_id not in self.chain_manager._snapshots:
+            return
+
+        self.chain_manager.restore(snapshot_id)
+
+
+class ReceiptCapture(ManagerAccessMixin):
```
unparalleled-js, line 115: Idea: was thinking we can potentially expose this class as a fixture for times when users want to track gas and manager their own snapshots:

```python
def test_something(receipt_capture, chain, contract, networks, owner):
    before_changing_code_snapshot = chain.snapshot()
    networks.provider.set_code(contract, HexBytes("0x123"))
    
    # Use receipt_capture so transaction not lost in gas report
    with receipt_capture:
        contact.doThing(sender=owner)
    
    chain.restore(before_changing_code_snapshot)
```

fubuloubu (approved):

src/ape/pytest/fixtures.py
fubuloubu: Was hoping that this would always be the case. I could see wanting to opt-out of the behavior, but prefer it not be opt-in (because that will be confusing to have to opt-in to the behavior, instead you will have a disconnect in what's being captured)

unparalleled-js: The other approach may be to add hooks to the `chain.snapshot()` call or something. Otherwise, not sure how you would be able to capture receipts if you you are managing snapshots within a test, and those receipts will get lost.

unparalleled-js: Either way - I didn't do any of this  yet! Was just an idea. I think it is pretty rare users manager their own snapshots this way? I have seen it though.

PR #1084, opened 2022-09-27
feat: assert on dev revert messages

helloibis:
### What I did

Add dev message assertions to `ape.reverts()`

fixes: APE-43

### How I did it

Adds `txn` property to `TransactionError` exceptions. This transaction is then utilized by the `RevertsContextManager`to fetch the transaction trace and find out the originating contract (receiver).

A `dev_message` (mutually exclusive with `expected_message`) was added to `RevertsContextManager`. When supplied, the exception is inspected for a transaction and the transaction trace is inspected. If the trace or contract source cannot be retrieved for any reason, the manager raises `AssertionError`. If a dev message is not found, `AssertionError` is raised. Finally, if a dev message is found but does not match the expectation, this also raises `AssertionError`

### How to verify it

Test project available at https://github.com/helloibis/ape-revert-test-project

### Caveats

- Does not currently work with Solidity (APE-298) or Cairo (APE-299)
- Requires use of a provider that supports tracing (such as Hardhat)
  - Cannot be used with the test `LocalProvider`
- May not work when JSON format contracts are supplied to Vyper, but still testing this
- Cannot support non-local sources currently

### Checklist

### DIFF
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -1,11 +1,16 @@
+from __future__ import annotations
+
 import time
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Iterator, Optional
 
 from eth_utils import humanize_hash
 
 if TYPE_CHECKING:
+    from evm_trace import TraceFrame
+
     from ape.api.networks import NetworkAPI
     from ape.api.providers import SubprocessProvider
+    from ape.api.transactions import TransactionAPI
     from ape.types import BlockID, SnapshotID
 
 
@@ -88,6 +93,7 @@ class TransactionError(ContractError):
         base_err: Optional[Exception] = None,
         message: Optional[str] = None,
         code: Optional[int] = None,
+        txn: Optional[TransactionAPI] = None,
     ):
         self.base_err = base_err
         if not message:
@@ -95,9 +101,17 @@ class TransactionError(ContractError):
 
         self.message = message
         self.code = code
+        self.txn = txn
 
         ex_message = f"({code}) {message}" if code else message
         super().__init__(ex_message)
+
+    @property
+    def trace(self) -> Optional[Iterator[TraceFrame]]:
+        if self.txn is not None:
+            yield from self.txn.provider.get_transaction_trace(self.txn.txn_hash.hex())
+
+        return None
 
 
 class VirtualMachineError(TransactionError):
@@ -112,8 +126,8 @@ class ContractLogicError(VirtualMachineError):
     such as from an assert/require statement.
     """
 
-    def __init__(self, revert_message: Optional[str] = None):
-        super().__init__(message=revert_message)
+    def __init__(self, revert_message: Optional[str] = None, *args, **kwargs):
+        super().__init__(message=revert_message, *args, **kwargs)
 
     @property
     def revert_message(self):
--- a/src/ape/pytest/contextmanagers.py
+++ b/src/ape/pytest/contextmanagers.py
@@ -1,26 +1,159 @@
+import re
+from collections import deque
 from typing import Optional, Type
 
-from ape.exceptions import ContractLogicError, TransactionError
+import ape
+from ape.exceptions import APINotImplementedError, ContractLogicError, TransactionError
 
 
 class RevertsContextManager:
-    def __init__(self, expected_message: Optional[str] = None):
+    def __init__(self, expected_message: Optional[str] = None, dev_message: Optional[str] = None):
+        if dev_message is not None and expected_message is not None:
+            raise ValueError("Cannot supply both 'expected_message' and 'dev_message'.")
+
         self.expected_message = expected_message
+        self.dev_message = dev_message
 
-    def __enter__(self):
-        pass
+    def _check_dev_message(self, exception: ContractLogicError):
+        """
+        Attempts to extract a dev-message from the contract source code by inspecting what
+        instruction(s) led to a transaction revert.
 
-    def __exit__(self, exc_type: Type, exc_value: Exception, traceback) -> bool:
-        if exc_type is None:
-            raise AssertionError("Transaction did not revert.")
+        Raises:
+            AssertionError: When the trace or source can not be retrieved, the trace does not end in
+            a REVERT instruction, or the dev message cannot be found or does not match the expected
+            dev message.
+        """
+        if exception.txn is None:
+            raise AssertionError("Could not fetch transaction information to check dev message.")
 
-        if not isinstance(exc_value, ContractLogicError):
+        receipt = exception.txn.provider.get_receipt(
+            str(exception.txn.txn_hash),
+            raise_on_fail=False,
+        )
+
+        if receipt.receiver is None:
+            raise AssertionError("Could not fetch transaction information to check dev message.")
+
+        try:
+            contract = ape.Contract(receipt.receiver)
+        except ValueError as exc:
             raise AssertionError(
-                f"Transaction did not revert.\n"
-                f"However, an exception of type {type(exc_value)} occurred: {exc_value}"
-            ) from exc_value
+                f"Could not fetch contract at {receipt.receiver} to check dev message."
+            ) from exc
 
-        actual = exc_value.revert_message
+        if contract.contract_type.pcmap is None:
+            raise AssertionError("Compiler does not support source code mapping.")
+
+        # Trace should not be None, but need to check for Mypy
+        if exception.trace is None:
+            raise AssertionError(
+                "Cannot check dev message; provider must support transaction tracing."
+            )
+
+        try:
+            trace = deque(exception.trace)
+        except APINotImplementedError as exc:
+            raise AssertionError(
+                "Cannot check dev message; provider must support transaction tracing."
+            ) from exc
+
+        if trace[-1].op != "REVERT":
+            raise AssertionError("Transaction trace did not end with REVERT.")
+
+        pc = None
+        pcmap = contract.contract_type.pcmap.parse()
+
+        # To find a suitable line for inspecting dev messages, we must start at the revert and work
+        # our way backwards. If the last frame's PC is in the PC map, the offending line is very
+        # likely a 'raise' statement.
+        if trace[-1].pc in pcmap:
+            pc = trace[-1].pc
+
+        # Otherwise we must traverse the trace backwards until we find our first suitable candidate.
+        # It should be noted that when an error happens, typically the contract will execute:
+        #
+        #   <...>
+        #   JUMPI    <- jump to revert routine on a failing condition
+        #
+        #   <...>
+        #   JUMPDEST <- beginning of revert routine
+        #   PUSH1
+        #   DUP1
+        #   REVERT
+        #
+        # So we must step backwards until we find the initial instructions that caused the failing
+        # condition. This will not be the JUMPI or even an arithmetic instruction like DIV, but
+        # rather the starting instruction for a given line.
+        else:
+            while len(trace) > 0:
+                frame = trace.pop()
+                if frame.pc in pcmap:
+                    pc = frame.pc
+                    break
+
+        # We were unable to find a suitable PC that matched the compiler's map.
+        if pc is None:
+            raise AssertionError("Could not find line that caused revert.")
+
+        offending_source = pcmap[pc]
+
+        # The compiler PC map had PC information, but not source information.
+        if offending_source is None or offending_source.line_start is None:
+            raise AssertionError("Could not find line that caused revert.")
+
+        sources = ape.project.extract_manifest().sources or {}
+        source_id = contract.contract_type.source_id or ""
+
+        if not (sources and source_id):
+            AssertionError("Could not retrieve contract source.")
+
+        if source_id not in sources:
+            AssertionError("Could not retrieve contract source.")
+
+        content = sources[source_id].content or ""
+
+        if not content:
+            AssertionError("Could not retrieve contract source.")
+
+        contract_source_lines = content.split("\n")
+
+        if offending_source.line_start > len(contract_source_lines):
+            raise AssertionError(
+                "Could not find line that caused revert (line outside of contract range)."
+            )
+
+        # Line numbers will not be zero-indexed
+        offending_line = contract_source_lines[offending_source.line_start - 1]
+
+        # Match '# dev: (...)' with any variation of whitespace
+        match = re.search(r"#\s*dev\s*:\s*(.+)", offending_line)
+
+        if not match:
+            raise AssertionError(
+                f"Could not find dev message '{self.dev_message}' in line:\n\t{offending_line}."
+            )
+
+        contract_dev_message = match.group(1)
+        assertion_error_prefix = f"Expected dev revert message '{self.dev_message}'."
+
+        if contract_dev_message is None:
+            raise AssertionError(f"{assertion_error_prefix} but there was none.")
+
+        if contract_dev_message != self.dev_message:
+            raise AssertionError(f"{assertion_error_prefix} but got '{contract_dev_message}'.")
+
+        return True
+
+    def _check_expected_message(self, exception: ContractLogicError):
+        """
+        Compares the revert message given by the exception to the expected message.
+
+        Raises:
+            AssertionError: When the exception message is ``None`` or if the message does not match
+            the expected message.
+        """
+        actual = exception.revert_message
 
         # Validate the expected revert message if given one.
         if self.expected_message is not None and self.expected_message != actual:
@@ -32,6 +165,24 @@ class RevertsContextManager:
                 raise AssertionError(f"{assertion_error_prefix} but there was none.")
 
             raise AssertionError(f"{assertion_error_prefix} but got '{actual}'.")
+
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type: Type, exc_value: Exception, traceback) -> bool:
+        if exc_type is None:
+            raise AssertionError("Transaction did not revert.")
+
+        if not isinstance(exc_value, ContractLogicError):
+            raise AssertionError(
+                f"Transaction did not revert.\n"
+                f"However, an exception of type {type(exc_value)} occurred: {exc_value}."
+            ) from exc_value
+
+        if self.dev_message is not None:
+            self._check_dev_message(exc_value)
+        else:
+            self._check_expected_message(exc_value)
 
         # Returning True causes the expected exception not to get raised
         # and the test to pass
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -315,7 +315,7 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def get_receipt(self, txn_hash: str) -> ReceiptAPI:
+    def get_receipt(self, txn_hash: str, raise_on_fail: bool = False) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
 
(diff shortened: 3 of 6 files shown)


helloibis:
Keep running into this weird thing where Python does not actually show me where errors occur if they are occurring __inside__ the reverts context manager. Took me a while because my initial error was a missing "ape" import and this bug/issue caused it to look like my __project__ was the thing missing `import ape` when really it was the `contextmanagers.py` file.

For example, when the context manager raises an exception the stack trace output looks like this:

image

But actually the error originated from `contextmanagers.py`:

image

...which is not shown in the stack trace.

Has anyone ever run into this? I'm completely stumped and haven't ever seen this from a Python project before

Here's a tiny example just running Python:

image

But when I run that example the output makes sense and shows me exactly where the error occurs:

image

Leads me to believe something with ape / plugins / ape-test / pytest is causing an issue?

cc @NotPeopling2day @unparalleled-js @fubuloubu

helloibis:
Here's the original problem I was running into to show how confusing this was:

image

The test file itself does `import ape` as you can see:

image

But the `contextmanagers.py` file does not `import ape` which is the actual cause of the `NameError` (see line 27 here):

image

helloibis:
Deleted my other comments, didn't realize that our pytest configuration hides stack trace info from ape code unless you use `--showinternals`

src/ape/exceptions.py
```diff
@@ -1,10 +1,13 @@
+from __future__ import annotations
```
fubuloubu, line 1: Is this necessary to do?

src/ape/pytest/contextmanagers.py
```diff
@@ -4,23 +4,55 @@
...
+
+        breakpoint()
+
+        # for item in contract.source_map.parse():
+
+        # contract.sourceMap["pc_pos_map"]trace.pc
+
+        # content = ape.project.extract_manifest().sources[contract.name].content
+
+        # trace.pc...
```
fubuloubu, line 42: Still in work?

helloibis: Yeah the PR is in draft

src/ape/exceptions.py
helloibis: Still WIP, something wasn't working with a type and trying to place it under the `if TYPING` so I am just bypassing it for now until the PR is ready

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,118 @@
...
+        # So we must step backwards until we find the initial instructions that caused the failing
+        # condition. This will not be the JUMPI or even an arithmetic instruction like DIV, but
+        # rather the starting instruction for a given line.
+        else:
+            while len(trace) > 0:
+                frame = trace.pop()
+                if str(frame.pc) in contract.contract_type.pcmap:
+                    pc = str(frame.pc)
+
+                    break
```
helloibis, line 76: @unparalleled-js @fubuloubu @NotPeopling2day 

So I noted this in the Discord as well, but the `pc_pos_map` seems to align with the trace for the _start_ of a line. In the case of my zero-division example I was trying, the instruction I was actually finding in the PC map was `DUP1`

I originally was looking for the last `JUMPI` or so but that won't align with the PC map so I decided on this instead: stepping backwards to the first matching line.

Thoughts, and are there any caveats there may be from this?

I think for transparency's sake if we can't find the dev message, we should at least print out the candidate line we looked at (if we had such a line)

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,118 @@
...
 
-        actual = exc_value.revert_message
+        # Line numbers will not be zero-indexed
+        offending_line = contract_source_lines[offending_line_num - 1]
+
+        # Match '# dev: (...)' with any variation of whitespace
+        match = re.search(r"#\s*dev\s*:\s*([\w\d\s]+)$", offending_line)
+
+        if not match:
+            raise AssertionError(f"Could not find dev message '{self.dev_message}'")
```
helloibis, line 101: As noted in https://github.com/ApeWorX/ape/pull/1084/files#r1006224591 I think I may change this to actually print out the offending line (granted it's not too large...)

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,118 @@
...
-                f"Transaction did not revert.\n"
-                f"However, an exception of type {type(exc_value)} occurred: {exc_value}"
-            ) from exc_value
+                f"Could not fetch contract at {receipt.receiver} to check dev message"
+            ) from exc
+
+        if contract.contract_type.sourcemap is None:
+            raise AssertionError("Could not fetch contract source to check dev message")
+
+        trace = deque(exc.trace)
```
helloibis, line 42: Need to swap this to `list`, originally I was thinking I could just look at the last frame so I was using `deque(exc.trace, maxlen=1)`

Going to check out if maybe we could at least truncate the trace here to something like 10 or so, since it's only ever going to be one Vyper statement that causes a revert. I'll have to play around and see though if there's a suitable `maxlen` or not

helloibis:
Spoke with @NotPeopling2day yesterday evening about the release strategy, I think for now we're thinking we can do this:

- After we code review this and it's ready, merge it to the feature branch and then merge the feature branch right away
- Make an Ape core _breaking_ release as an alpha release to start, like a `0.6.0a1`
- We can leave the feature branches for the provider plugins open, because now those plugins can import the Ape alpha release directly
- Keep manual tests in their own project for now (link in description), because we'll need to figure out how to adequately test this in core at an integration test level (and also not duplicate all the tests in the various provider plugins)
- Figure out some of the caveats that exist here (I can note these in this PR description, but things like Solidity's PC map problem) and work them out in subsequent alpha versions
- Beef up automated tests around this
- Release a _breaking_ non-alpha release after all the dust is settled (this is also when we would release the new provider plugin versions)

Thoughts? cc @unparalleled-js @fubuloubu

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
 
-        actual = exc_value.revert_message
+        # Line numbers will not be zero-indexed
+        offending_line = contract_source_lines[offending_source.line_start - 1]
+
+        # Match '# dev: (...)' with any variation of whitespace
+        match = re.search(r"#\s*dev\s*:\s*(.+)", offending_line)
+
+        if not match:
+            raise AssertionError(f"Could not find dev message '{self.dev_message}'.")
```
helloibis, line 117: @NotPeopling2day @unparalleled-js @fubuloubu thoughts on the error messaging - should we print out the offending line (up to a certain number of characters perhaps)?

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
-                f"However, an exception of type {type(exc_value)} occurred: {exc_value}"
-            ) from exc_value
+                f"Could not fetch contract at {receipt.receiver} to check dev message."
+            ) from exc
+
+        if contract.contract_type.pcmap is None:
+            raise AssertionError("Compiler does not support source code mapping.")
+
+        try:
+            trace = deque(exc.trace)
```
helloibis, line 46: @unparalleled-js @NotPeopling2day @fubuloubu 

Depending on the contract and transaction being made, you could end up with a very long trace

I'm not fully utilizing `deque`'s features right now but one of them is that it provides a `maxlen`. Is this something that we should implement and limit trace traversal to a certain number of instructions or can we just allow to examining an entire trace to try and find a dev message?

Unfortunately we do have to fetch the whole trace since we start at the bottom and work backwards. IIRC the EVM trace does like a streaming fetch - we won't get that benefit since we have to grab all of it first.

unparalleled-js: For display purposes, could we only show the last _n_ lines unless the user does `-v debug` in which case, show the full trace? Where `n` is like 10 or 15.

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
 
-from ape.exceptions import ContractLogicError, TransactionError
+import ape
+from ape.exceptions import APINotImplementedError, ContractLogicError, TransactionError
 
 
 class RevertsContextManager:
-    def __init__(self, expected_message: Optional[str] = None):
+    def __init__(self, expected_message: Optional[str] = None, dev_message: Optional[str] = None):
+        if dev_message is not None and expected_message is not None:
```
unparalleled-js, line 11: `brownie` does let you use both

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
+        # The compiler PC map had PC information, but not source information.
+        if offending_source is None or offending_source.line_start is None:
+            raise AssertionError("Could not find line that caused revert.")
+
+        try:
+            contract_source_lines = (
+                ape.project.extract_manifest()
+                .sources[contract.contract_type.source_id]
+                .content.split("\n")
+            )
```
helloibis, line 101: Mypy questions:

```
Value of type "Optional[Dict[str, Source]]" is not indexable

Item "None" of "Union[str, None, Any]" has no attribute "split"`

Invalid index type "Optional[str]" for "Optional[Dict[str, Source]]"; expected type "str"
```

Do I need to be saving `sources` to like a typed variable? From what I can tell `sources` should always be `Dict[str, Source]`. Is there a situation where it's `None` or has null values?

cc @NotPeopling2day @unparalleled-js

src/ape/exceptions.py
```diff
@@ -88,17 +91,26 @@ def __init__(
...
         self.code = code
+        self.txn = txn
 
         ex_message = f"({code}) {message}" if code else message
         super().__init__(ex_message)
 
+    @property
+    def trace(self) -> Optional[Iterator["TraceFrame"]]:
+        if self.txn is not None:
+            yield from self.txn.provider.get_transaction_trace(self.txn.txn_hash.hex())
```
unparalleled-js, line 112: I believe the existence of a `trace` implies there is a receipt. I am wondering if we should have `receipt` attached to this exception instead of `txn`. This way, we can use the `trace` property on `receipt` (which does the same as this).

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
+            raise AssertionError("Could not fetch transaction information to check dev message.")
 
-        if not isinstance(exc_value, ContractLogicError):
+        receipt = exception.txn.provider.get_receipt(
+            exception.txn.txn_hash,
+            raise_on_fail=False,
+        )
+
+        try:
+            contract = ape.Contract(receipt.receiver)
```
unparalleled-js, line 36: `receipt.receiver` is optional I think, we may need to handle that.

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
+            raise AssertionError(f"{assertion_error_prefix} but got '{contract_dev_message}'.")
+
+        return True
+
+    def _check_expected_message(self, exception: ContractLogicError):
+        """
+        Compares the revert message given by the exception to the expected message.
+
+        Raises:
+            AssertionError: When the exception message is null or if the message does not match the
```
unparalleled-js, line 135:
```suggestion
            AssertionError: When the exception message is ``None`` or if the message does not match the
```

src/ape/pytest/contextmanagers.py
helloibis: On the index one: when is `source_id` null? Would I run into that in this scenario somehow after already constructing the `Contract` earlier up in this function?

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,141 @@
...
-                f"However, an exception of type {type(exc_value)} occurred: {exc_value}"
-            ) from exc_value
+                f"Could not fetch contract at {receipt.receiver} to check dev message."
+            ) from exc
+
+        if contract.contract_type.pcmap is None:
+            raise AssertionError("Compiler does not support source code mapping.")
+
+        try:
+            trace = deque(exception.trace)
```
helloibis, line 46: Another Mypy question:

I've technically already established the pre-condition earlier on by bailing early if `exception.txn is None`

So here would I just use `# type: ignore`? Want to make sure that's the acceptable way to solve these issues with the type checking when it's complaining about situations that can't actually happen

cc @NotPeopling2day @unparalleled-js

src/ape/pytest/contextmanagers.py
unparalleled-js: `source_id` can be `None` if the source is not a local source, like you just got it from somewhere else and it wasn't included.

This happens sometimes when using hardcoded ABIs especially because there is no reason to use a source ID (until there is).

src/ape/pytest/contextmanagers.py
helloibis: Updated in force push to e5e6ffb4

src/ape/pytest/contextmanagers.py
helloibis: Updated in force push to 782e88fe

src/ape/pytest/contextmanagers.py
unparalleled-js: `sources` is `None` is not using any local sources, as in you are only using dependencies.
The spec supports this, and so do we in Ape, that is why the type of `sources` is optional.

src/ape/pytest/contextmanagers.py
helloibis: It was noted in the [original issue discussion](https://github.com/ApeWorX/ape/issues/251) that it should be mutually exclusive - is this somewhere where we can/should break with Brownie's implementation?

cc @NotPeopling2day @fubuloubu

unparalleled-js: It may be a mistake that it was written that way! That is my impression. I don't see a technical reason why you couldn't assert both.

src/ape/exceptions.py
helloibis: Yeah I do like that idea more, and it would also simplify a lot of the call sites. Currently since I made it `txn` I ended up just passing `txn` like everywhere that I had one, but if it's switched to `receipt` then I really only have to care about supplying it where it matters (after a transaction invocation).

I'll get this updated, though it will probably mean some changes in the provider plugins since I wrote those in a way that they supply `txn` when raising transaction errors

src/ape/pytest/contextmanagers.py
unparalleled-js: One way I prefer is save it is as a local variable first.

```python
txn = exception.txn
```

From a typing-checking perspective, the implementation of the calculated property could produce different results each time it was called, thus the type is not ever considered fully non-optional.

src/ape/pytest/contextmanagers.py
helloibis: Yeah and if we do allow for both, I think the only code change needed here will just be removing the `ValueError` case and then further down just using two `if` statements for message checking instead of `if/else`

src/ape/pytest/contextmanagers.py
unparalleled-js: And then do the type-checking on the local variable

src/ape/pytest/contextmanagers.py
helloibis: @unparalleled-js Gotcha, thanks! Just updated to 53dce52c

I opted for just checking them first rather than the long statement in a try/except that I had:

image

I noted in the PR description as well that this currently will only work with local sources, I think I could take a stab at non-local sources (if that's something we should be supporting) in a second PR or a second ticket

src/ape/pytest/contextmanagers.py
helloibis: @unparalleled-js in this case it's not `txn`, but rather that the `trace` property would only return `None` if the `txn` happened to be `None` as well. So by checking the transaction first the prerequisite is already set that `trace` cannot be returning `None`. Would I just `# type: ignore` in that case?

(this applies whether [we change to `receipt`](https://github.com/ApeWorX/ape/pull/1084#discussion_r1010953444) or not)

src/ape/pytest/contextmanagers.py
unparalleled-js: One use-case I can think of for wanting both is to expect a certain shorter message as the actual message and a different more verbose message as the dev string.

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,145 @@
...
+        # our way backwards. If the last frame's PC is in the PC map, the offending line is very
+        # likely a 'raise' statement.
+        if trace[-1].pc in contract.contract_type.pcmap:
+            pc = trace[-1].pc
+
+        # Otherwise we must traverse the trace backwards until we find our first suitable candidate.
+        # It should be noted that when an error happens, typically the contract will execute:
+        #
+        #   <...>
+        #   JUMPI    <- jump to revert routine on a failing condition
```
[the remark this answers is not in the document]
helloibis: @unparalleled-js what changes between them, wouldn't it always just be traversing the trace and just comparing the PC values?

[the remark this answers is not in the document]
unparalleled-js: Maybe? I don't know.

But this line:

```
if trace[-1].op != "REVERT"
```

refers to an EVM opcode.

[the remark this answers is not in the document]
unparalleled-js: an maybe an EVM behavior? It is hard to say because there is not a that level of tracing yet in Starknet, AFAIK.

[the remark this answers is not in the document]
unparalleled-js: Ecosystems like Cosmos and Near use a WASM system though:

https://cosmwasm.com/

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,151 @@
...
+        if offending_source.line_start > len(contract_source_lines):
+            raise AssertionError(
+                "Could not find line that caused revert (line outside of contract range)."
+            )
+
+        # Line numbers will not be zero-indexed
+        offending_line = contract_source_lines[offending_source.line_start - 1]
+
+        # Match '# dev: (...)' with any variation of whitespace
+        match = re.search(r"#\s*dev\s*:\s*(.+)", offending_line)
```
unparalleled-js, line 124: This part seems tricky because we don't know anything about the language, like does it use `#` or `//` for comments or something entirely different?

We need to delegate some of this to the compiler plugins?

src/ape/pytest/contextmanagers.py
unparalleled-js: Sure! That sounds like it goes with our initiative to show traces in place of python exceptions during uncaught reverts.

src/ape/pytest/contextmanagers.py
unparalleled-js: If you are not too worried, you can use `type: ignore` with an additional comment explaining it.
My aim is suggesting not to do that is preventing possible future bugs but complicating the implementation of `trace()` such that `None` is returned in other scenarios besides when the txn is `None`.

src/ape/pytest/contextmanagers.py
helloibis: Yeah good point

I think maybe for now since this really only works with Vyper, we could stick with this for the alpha

Then as part of APE-298/299 there could be a step to break some of this stuff out and delegate it to the plugins. I think that breaking this out would also kind of tie in with https://github.com/ApeWorX/ape/pull/1084#discussion_r1012180056 as well

Perhaps that "breaking out" step could be its own thing done before APE-298/299

src/ape/pytest/contextmanagers.py
helloibis: Updated to specifically check `if exception.trace is None`

helloibis:
Closing in favor of https://github.com/ApeWorX/ape/pull/1125

PR #1090, opened 2022-10-05
feat: Add `set_code` method to `ProviderAPI`

sabotagebeats:
### What I did
added set-code in providers api

fixes: APE-146

### How I did it
as per the spec

Specification

In `ProviderAPI`, add method:

```
@raises_not_implemented
def set_code(address, AddressType, code: bytes):
    """
    Change the code of a smart contract, for development purposes.
    Test providers implement this method when they support it.
    """
```

Then, in ape-hardhat, we can support it be calling the RPC https://hardhat.org/hardhat-network/docs/reference#hardhat_setcode

### Checklist


src/ape/api/providers.py
```diff
@@ -418,6 +418,18 @@ def __repr__(self) -> str:
...
+    @raises_not_implemented
+    def set_code(self, address: AddressType, code: bytes):
+        """
+        Change the code of a smart contract, for development purposes.
+        Test providers implement this method when they support it.
+
+        Args:
+            address (AddressType): An address on the network.
+            code(bytes): The new bytecode
+
```
unparalleled-js, line 430:
```suggestion
```

src/ape/api/providers.py
```diff
@@ -418,6 +418,18 @@ def __repr__(self) -> str:
...
 
+    @raises_not_implemented
+    def set_code(self, address: AddressType, code: bytes):
+        """
+        Change the code of a smart contract, for development purposes.
+        Test providers implement this method when they support it.
+
+        Args:
+            address (AddressType): An address on the network.
+            code(bytes): The new bytecode
```
unparalleled-js, line 429:
```suggestion
            code(bytes): The new bytecode.
```

unparalleled-js (commented):
Awesome.

Before merging, let's add support to `ape-hardhat` as well as tests there, to prove the API method suffices in its signature.

sabotagebeats:
> Awesome.
> 
> Before merging, let's add support to `ape-hardhat` as well as tests there, to prove the API method suffices in its signature.

what do we need to do to add `ape-hardhat` support for this feature?

src/ape/types/__init__.py
```diff
@@ -31,6 +31,11 @@
 as well as a block number or hash (HexBytes).
 """
 
+ContractCode = Union[str, bytes, HexBytes]
+"""
+An ID that represents contract code, which can be represented in string, bytes, or HexBytes.
```
unparalleled-js, line 36: It is not really an ID here

```suggestion
A type that represents contract code, which can be represented in string, bytes, or HexBytes.
```

unparalleled-js (dismissed):

src/ape/api/providers.py
```diff
@@ -180,7 +180,7 @@ def get_balance(self, address: str) -> int:
         """
 
     @abstractmethod
-    def get_code(self, address: str) -> bytes:
+    def get_code(self, address: AddressType) -> bytes:
```
unparalleled-js, line 183:
```suggestion
    def get_code(self, address: AddressType) -> ContractCode:
```

src/ape/api/providers.py
```diff
@@ -180,7 +180,7 @@ def get_balance(self, address: str) -> int:
...
         """
         Get the bytes a contract.
```
unparalleled-js, line 186: The arg doc needs to be updated:

```python
Args:
    address (``AddressType``): The address of the contract.

Returns:
    :class:`~ape.types.ContractCode`: The contract bytecode.
```

src/ape/api/providers.py
sabotagebeats: I think this should be resolved in newest commit.

src/ape/api/providers.py
sabotagebeats: I think this should be resolved in newest commit.

src/ape/api/providers.py
```diff
@@ -428,6 +428,17 @@ def __repr__(self) -> str:
...
 
+    @raises_not_implemented
+    def set_code(self, address: AddressType, code: ContractCode) -> bool:
+        """
+        Change the code of a smart contract, for development purposes.
+        Test providers implement this method when they support it.
+
+        Args:
+            address (AddressType): An address on the network.
+            code(bytes): The new bytecode.
```
unparalleled-js, line 439:
```suggestion
            code (:class:`~ape.types.ContractCode`): The new bytecode.
```

unparalleled-js (commented):
1 issue in doc str

unparalleled-js (approved):

sabotagebeats:
@unparalleled-js is there any more changes that need to be made or can this be merged?

PR #1097, opened 2022-10-17
feat: set local networks to max gas_limit

helloibis:
### What I did

Updated local config to use max gas limit

Also updated documentation to reflect the two defaults

### How I did it
Set `EthereumConfig.local` to `gas_limit="max"`

### How to verify it
Shown in tests that the local config uses "max" by default

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -146,6 +146,8 @@ You may use one of:
 - `"max"` - the maximum block gas limit is used
 - A number or numeric string, base 10 or 16 (e.g. `1234`, `"1234"`, `0x1234`, `"0x1234"`)
 
+For the local network configuration, the default is `"max"`. Otherwise it is `"auto"`.
+
 ## Plugins
 
 Set which plugins you want to always use:
--- a/docs/userguides/networks.md
+++ b/docs/userguides/networks.md
@@ -52,6 +52,8 @@ You may use one of:
 - `"max"` - the maximum block gas limit is used
 - A number or numeric string, base 10 or 16 (e.g. `1234`, `"1234"`, `0x1234`, `"0x1234"`)
 
+For the local network configuration, the default is `"max"`. Otherwise it is `"auto"`.
+
 ## Local Network
 
 The default network in Ape is the local network (keyword `"local"`).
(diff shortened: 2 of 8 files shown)


helloibis:
I'm getting test failures but for stuff unrelated to the change, for instance in `test_cli.py`:

image

Any ideas? I don't think it's a blocker to this PR though necessarily

fubuloubu (approved):

helloibis:
image

So for these tests I think we're in a bit of a weird state between the gas config changes and the reverts changes

The current behavior - without any of the reverts changes - does not raise an exception on a failed committed transaction. The tests in the screenshot will fail because of that, but not in a way where we should change them in this PR.

I think these changes need to wait until after the reverts stuff is completed so that the behavior can be fixed without needing to adjust these tests

cc @unparalleled-js @NotPeopling2day

unparalleled-js:
You may be hitting the issue fixed here: https://github.com/ApeWorX/ape/pull/1100

unparalleled-js:
- The test `test_send_transaction_without_enough_funds` can check for `AccountError` instead of `TransactionError`, that makes more sense anyway.

For the other tests, I think we need to merge https://github.com/ApeWorX/ape/pull/1100

unparalleled-js:
hoping to find more clarity after rebasing from this: https://github.com/ApeWorX/ape/pull/1103

unparalleled-js:
OK I figured out why the test is failing. There is no bug but it is actually behaving like it is supposed.
More info here: https://github.com/ApeWorX/ape/pull/1104

tldr: Because gas is so gas on the max, there is enough money left over because the value was set very low to safely assume more gas used.

fubuloubu (approved):

unparalleled-js:
One pending issue is that if the user configured a specific gas limit but wants to use `"auto"` or `"max"` in an adhoc sense (from txn kwargs), it currently won't work.

unparalleled-js:
> One pending issue is that if the user configured a specific gas limit but wants to use `"auto"` or `"max"` in an adhoc sense (from txn kwargs), it currently won't work.

https://github.com/helloibis/ape/pull/1

fubuloubu (approved):

unparalleled-js (approved):

helloibis:
Thanks again @unparalleled-js!

Issue #1098, opened 2022-10-18
Revert errors not working when specifying gas


unparalleled-js:
### Environment information

* OS: macOS
* Python Version: 3.9.7
* `ape` and plugin versions:

Latest ape

### What went wrong?

When specfying gas, ape reverts do not appear and no way to get revert message:

```python
    with ape.reverts():
        token.batchTransfer(payments, sender=owner, gas='1000000')
```

### How can it be fixed?

TBD

unparalleled-js:
Hmm, I am actually wondering why `web3.py` is not raising any error when specifying gas.
The difference is that before it was happening in the estimate RPC but now it is happening in the `send_transaction()`.
I wonder if it is related to the web3.py version we are using!

PR #1100, opened 2022-10-18
fix: issue where revert message missing when specify gas using EthTester

unparalleled-js:
### What I did

fixes: #1098 

### Checklist

### DIFF
--- a/src/ape_test/provider.py
+++ b/src/ape_test/provider.py
@@ -136,11 +136,13 @@ class LocalProvider(TestProviderAPI, Web3Provider):
         )
 
         if receipt.failed:
+            txn_dict = receipt.transaction.dict()
+            txn_dict["nonce"] += 1
+            txn_params = cast(TxParams, txn_dict)
+
             # Replay txn to get revert reason
             try:
-                txn_dict = receipt.transaction.dict()
-                txn_dict["nonce"] += 1
-                self.web3.eth.call(cast(TxParams, txn_dict))
+                self.web3.eth.call(txn_params)
             except (ValidationError, TransactionFailed) as err:
                 raise self.get_virtual_machine_error(err, sender=txn.sender) from err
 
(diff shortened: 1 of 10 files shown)


src/ape_test/provider.py
```diff
@@ -133,8 +134,15 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         receipt = self.get_receipt(
             txn_hash.hex(), required_confirmations=txn.required_confirmations or 0
         )
-        if txn.gas_limit is not None and receipt.ran_out_of_gas:
-            raise OutOfGasError()
+
+        if receipt.failed:
```
unparalleled-js, line 138: Here is my hacky fix, but perhaps this is a bug that should be resolved in EthTester

src/ape_test/provider.py
```diff
@@ -133,8 +134,15 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
+            # Replay txn to get revert reason
+            try:
+                txn_dict = receipt.transaction.dict()
+                txn_dict["nonce"] += 1
+                self.web3.eth.call(cast(TxParams, txn_dict))
+            except (ValidationError, TransactionFailed) as err:
+                raise self.get_virtual_machine_error(err, sender=txn.sender) from err
```
helloibis, line 145: Do the `ValidationError` and `TransactionFailed` only get raised via `web3.eth.call()`? If so I think we can just put that line only in the `try` so it doesn't imply that the other two lines could raise either of those

src/ape_test/provider.py
helloibis, line 138: Can transactions fail for reasons that cause it to not be idempotent? That is to say, something like a network connection error or similar would it could fail the first time but pass the second time if sent again?

unparalleled-js: If we get a receipt at all at this point, the transaction was published.

helloibis: Oh right, was confusing this with a preemptive failure like gas estimation

Also was forgetting this is `call()` and not a second sending of the transaction 

I think we're good then 👍

helloibis (approved):

PR #1104, opened 2022-10-19
test: demonstrate transfer with `send_everything=True` and setting a very high gas

unparalleled-js:
### What I did

So this PR came about while debugging a failing test from https://github.com/ApeWorX/ape/pull/1097.
Turns out, using max gas is so high that there is plenty of remainder in that situation to afford attempting to send full balance again. This makes sense but throw @helloibis and I for a head-spin for a while.

### How I did it

* Refactor code to make sense of what the heck is going on
* Set breakpoints to compare estimated gas limit and max gas limit
* Notice that max is magnitudes higher and thus does not drain balance sufficiently to cause user unable to transact afterwards because the value set from `set_everything=True` is safely assuming usage of full gas, as it has to.

### How to verify it

We need to update the test is all in @helloibis 's branch. We can rebase off this PR to have the insight "previously" known. Sorry for the confusions but I feel pretty good about my understanding now of what is going on.

### Checklist

### DIFF
--- a/tests/functional/test_accounts.py
+++ b/tests/functional/test_accounts.py
@@ -64,11 +64,22 @@ def test_sign_message_with_prompts(runner, keyfile_account):
         assert signature is None
 
 
-def test_transfer(sender, receiver):
-    initial_balance = receiver.balance
-    sender.transfer(receiver, "1 gwei")
-    expected = initial_balance + convert("1 gwei", int)
-    assert receiver.balance == expected
+def test_transfer(sender, receiver, eth_tester_provider):
+    initial_receiver_balance = receiver.balance
+    initial_sender_balance = sender.balance
+    value_str = "2 gwei"
+    value_int = convert(value_str, int)
+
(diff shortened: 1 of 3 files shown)


tests/functional/test_accounts.py
```diff
@@ -81,15 +92,46 @@ def test_transfer_without_value_send_everything_false(sender, receiver):
...
 
 
+def test_transfer_without_value_send_everything_true_with_gas_specified(
+    sender, receiver, eth_tester_provider
+):
+    initial_receiver_balance = receiver.balance
+    initial_sender_balance = sender.balance
+
+    # The gas selected here is very high compared to what actually gets used.
+    gas = 25000000
```
unparalleled-js, line 120: if you change this to `2100`, it will have the same effect as the test above and the second transfer will fail

fubuloubu: good find!

fubuloubu (approved):

PR #1105, opened 2022-10-19
feat: add `--watch` flag to `ape test` (and associated options)

fubuloubu:
### What I did
Allows users to run a continuous watch loop using watchdog file change events, and re-trigger the test suite until the user exits.

### How I did it
Copied `pytest-watcher`

### How to verify it
Run it in a project

### Checklist

- [ ] ~~New test cases have been added~~


fubuloubu:
Not sure what the test failure is about, it works locally

src/ape_test/_cli.py
```diff
@@ -12,9 +80,57 @@
...
+    "--watch-folders",
+    multiple=True,
+    type=Path,
+    default=[Path("contracts"), Path("tests")],
+    help=(
+        "Folders to watch for changes using `ape test --watch`."
+        " Defaults to `contracts/` and `tests/`"
+    ),
+)
+@click.option(
```
unparalleled-js, line 100: We should add the CLI configuration + the implementation code to the `pytest` integration module(s). Otherwise, there is a discrepancy between `pytest` and `ape test` all of a sudden. Currently, we are able to run either `pytest` or `ape test`. Otherwise `pytest` will only half-work, which causes confusion. 

History: shortly after we released 0.1.0, this confused one user in particular and we fixed it by allowing `pytest` to work fully with the `ape_test` integration when running `pytest` as the command. I was looking for the message or PR but couldn't find either though.

Additionally, it will allow us to test this using `pytester`, but that is really an aside.

unparalleled-js: My main concern is that now `pytest` and `ape test` are different when before they weren't, but I suppose if you are okay with that then I won't make too big of a fuss bout it.

unparalleled-js (commented):
It seems to work, at my first pass!

I mostly would like some docs, I don't see this code changing much so I am okay with skipping tests for now, also due to the complicated nature of it.

I know it is testable though so if you are feeling motivated there, that'd be a nice extra!

fubuloubu:
> It seems to work, at my first pass!
> 
> I mostly would like some docs, I don't see this code changing much so I am okay with skipping tests for now, also due to the complicated nature of it.
> 
> I know it is testable though so if you are feeling motivated there, that'd be a nice extra!

see [4b2a1a7](https://github.com/ApeWorX/ape/pull/1105/commits/4b2a1a7081d3f3c0997a1609ee59de0f83f0c535)

unparalleled-js (approved):

src/ape_test/_cli.py
fubuloubu: Yeah, I wouldn't want this to work with `pytest` because I would use `pytest-watch` for this functionality instead

src/ape_test/_cli.py
```diff
@@ -1,9 +1,77 @@
...
+        events.EVENT_TYPE_MOVED,
+    )
+
+    def dispatch(self, event: events.FileSystemEvent) -> None:
+        if event.event_type in self.EVENTS_WATCHED:
+            self.process_event(event)
+
+    @cached_property
+    def _extensions_to_watch(self) -> List[str]:
+        return [".py", *self.compiler_manager.registered_compilers.keys()]
```
unparalleled-js, line 47: I believe the `.sol` / `.vy` files should be here as well?
For some reason, not working when I make changes to files in `contracts/`.
The tests never re-trigger.

src/ape_test/_cli.py
```diff
@@ -12,9 +80,60 @@
...
+    "--watch",
+    is_flag=True,
+    default=False,
+    help="Watch for changes to project files and re-run the test suite with the given options.",
+)
+@click.option(
+    "--watch-paths",
+    multiple=True,
+    type=Path,
+    default=[Path("contracts"), Path("tests")],
```
unparalleled-js, line 94: We will need to include compiling in the re-trigger and.or possibly put `.build/` here instead

src/ape_test/_cli.py
```diff
@@ -12,9 +80,60 @@
...
+    help=(
+        "Folders to watch for changes using `ape test --watch`."
+        " Defaults to `contracts/` and `tests/`"
+    ),
+)
+@click.option(
+    "--watch-delay",
```
unparalleled-js, line 101: This is currently allowing negative values

unparalleled-js (changes requested):
We need more magic sauce on the contracts-related changes, that is the main thing.

src/ape_test/_cli.py
unparalleled-js: Could perhaps fail earlier though!

src/ape_test/_cli.py
unparalleled-js: Ok, it seems to be working for me now... hmm, strange!

src/ape_test/_cli.py
unparalleled-js: This works, my issue was unrelated to this.

unparalleled-js:
OK - the issues I described in my resolved comments seem to only happen during specifying a single file (and make sures to that file).

This is what I am doing:

```bash
ape test --watch-paths contracts/None.vy --watch
```

Yes, I have a Vyper contract called `None.vy` (QA!).

Then I make changes to `None.vy` and The Watcher never notices.

unparalleled-js (approved):
Still more feedback, but things seem good now with the folder check!

src/ape_test/_cli.py
```diff
@@ -1,9 +1,77 @@
...
+
+    @cached_property
+    def _extensions_to_watch(self) -> List[str]:
+        return [".py", *self.compiler_manager.registered_compilers.keys()]
+
+    def _is_path_watched(self, filepath: str) -> bool:
+        """
+        Check if file should trigger pytest run
+        """
+        return any(map(filepath.endswith, self._extensions_to_watch))
```
unparalleled-js, line 53: We can handle individual files by using their parent as the watched directory and checking against a list of files to watch here and only emitting the event if the src_path matches one of those files

Ninjagod1251:
The test does not stop until I force stop Ctrl-C. So it work as intended.

```
(ape) chris@DESKTOP-ID4V0R6 ~/demo/ape-nft-project (main)$ ape test --watch
WARNING: Found 2 packages named 'eth-ape'.
Installation paths:
        /home/chris/virtualenvs/ape/lib/python3.8/site-packages/eth_ape-0.5.3.dev9+g0a91b3e2.dist-info,
        /home/chris/ape/src/eth_ape.egg-info
WARNING: Found 2 packages named 'eth-ape'.
Installation paths:
        /home/chris/virtualenvs/ape/lib/python3.8/site-packages/eth_ape-0.5.3.dev9+g0a91b3e2.dist-info,
        /home/chris/ape/src/eth_ape.egg-info
================================================ test session starts ================================================
platform linux -- Python 3.8.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/chris/demo/ape-nft-project
plugins: eth-ape-0.5.3.dev9+g0a91b3e2, web3-6.0.0b6, forked-1.4.0, cov-3.0.0, mock-3.7.0, xdist-2.5.0, hypothesis-6.39.0, asyncio-0.18.3, typeguard-2.13.3
asyncio: mode=legacy
collected 9 items

tests/test_token.py .........                                                                                 [100%]

================================================= warnings summary ==================================================
../../virtualenvs/ape/lib/python3.8/site-packages/pytest_asyncio/plugin.py:191
  /home/chris/virtualenvs/ape/lib/python3.8/site-packages/pytest_asyncio/plugin.py:191: DeprecationWarning: The 'asyncio_mode' default value will change to 'strict' in future, please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.
    config.issue_config_time_warning(LEGACY_MODE, stacklevel=2)

tests/test_token.py::test_incorrect_signer_minter
  /home/chris/virtualenvs/ape/lib/python3.8/site-packages/eth_abi/codec.py:166: DeprecationWarning: abi.decode_single() is deprecated and will be removed in version 4.0.0 in favor of abi.decode()
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================================== 9 passed, 2 warnings in 2.14s ===========================================
```

Ninjagod1251:
Why do we want to keep the test suspended?

Ninjagod1251:
it also works with 
`ape test -I -s --watch` ❤️

unparalleled-js:
> Why do we want to keep the test suspended?

The idea is that you make changes to some of the files in`contracts/` or `tests/` and then it auto-runs the tests.

Try changing the `--watch-delay` and `--watch-folders` options and see what happens!

Ninjagod1251:
LGTM :)

PR #1107, opened 2022-10-21
fix: use sourceId rather than contractType name for error message

hhamud:
### What I did

Using this PR to open up a discussion on creating a clearer compiler message.

Slightly edited the compiler message to be clearer about what is causing the contract type clash. I was having an issue where ape compile would fail if there was a JSON file that had the same name as a contract and I was confused until @fubuloubu pointed out that JSON files (ABIs) are treated as contract types. 

fixes: #

### How I did it

Edited the compiler error message
### Checklist


fubuloubu (approved):
Nice, this seems clearer yes

unparalleled-js (approved):

Issue #1092, opened 2022-10-06
Skip reporting gas for specific methods/contracts using a regex


fubuloubu:
### Overview

Sometimes I might have additional test-only contracts or methods where I don't necessarily want them to be reported with the reporting feature. As an example, here's a capture I made with a project I am working on:

```sh
                       test/Token.vy Gas                        
                                                                
  Method         Times called    Min.    Max.    Mean   Median  
 ────────────────────────────────────────────────────────────── 
  DEBUG_mint                4   10473   44673   27573    27573  
  approve                   2   24273   24273   24273    24273  
  balanceOf                 2    2509    2509    2509     2509  
  transferFrom              2   26112   26112   26112    26112  
  transfer                  4   12263   29363   20813    20813  
                                                                
                          VyperVault.vy Gas                           
                                                                      
  Method               Times called    Min.    Max.    Mean   Median  
 ──────────────────────────────────────────────────────────────────── 
  deposit                         1   77727   77727   77727    77727  
  DEBUG_steal_tokens              2   17799   34899   26349    26349  
                                                                      
                        SolidityVault.sol Gas                         
                                                                      
  Method               Times called    Min.    Max.    Mean   Median  
 ──────────────────────────────────────────────────────────────────── 
  deposit                         1   80248   80248   80248    80248  
  DEBUG_steal_tokens              2   17651   34751   26201    26201  
```

### Specification

I'd like to be able to specify via config option and/or command-line flag when running `ape test` a specific set of contracts and/or contract methods to skip when executing the gas report feature

Using the above example, I would want to skip contracts under `test/*` and methods that start with `DEBUG_*`:

```yaml
test:
  skip_contract_types:
  - Token
  skip_method_names:
  - DEBUG_[a-z_]*
```

which would make the report look like:

```sh
                          VyperVault.vy Gas                           
                                                                      
  Method               Times called    Min.    Max.    Mean   Median  
 ──────────────────────────────────────────────────────────────────── 
  deposit                         1   77727   77727   77727    77727  
                                                                      
                        SolidityVault.sol Gas                         
                                                                      
  Method               Times called    Min.    Max.    Mean   Median  
 ──────────────────────────────────────────────────────────────────── 
  deposit                         1   80248   80248   80248    80248  
```

NOTE: Example not final

https://linear.app/apeworx/issue/APE-286/skip-reporting-gas-for-specific-methodscontracts-using-a-regex

### Dependencies

no dependencies

[...]

PR #1111, opened 2022-10-28
feat: configure gas report, including skip files

unparalleled-js:
### What I did

fixes: #1092 - Note uses glob instead of regex though, is that okay?

Additionally fixes bug where if two receipts had same transaction hash, the second one would not show in the gas report.

### How I did it

* Add config wrapper to handle both CLI args and `ape-config.yaml` as one and pass that as a dependency in the `pytest-plugin` code.
* Change `receipt_map` in `ReceiptCapture` to first use a key of the source ID so can trace transactions with same hash from different sources.

### How to verify it

https://github.com/fubuloubu/ERC4626/pull/18

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -71,19 +71,19 @@ class ContractFunctionPath:
     Useful for identifying a method in a contract.
     """
 
-    contract: str
-    method: Optional[str] = None
+    contract_id: str
+    method_id: Optional[str] = None
 
     @classmethod
     def from_str(cls, value: str) -> "ContractFunctionPath":
         if ":" in value:
             contract_name, method_name = value.split(":")
-            return cls(contract=contract_name, method=method_name)
+            return cls(contract_id=contract_name, method_id=method_name)
 
-        return cls(contract=value)
+        return cls(contract_id=value)
 
     def __str__(self) -> str:
-        return f"{self.contract}:{self.method}"
+        return f"{self.contract_id}:{self.method_id}"
 
     def __repr__(self) -> str:
         return f"<{self}>"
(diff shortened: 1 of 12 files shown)


fubuloubu (dismissed):

fubuloubu (approved):

fubuloubu (approved):

fubuloubu:
Would love at least one other person do QA on this feature, but looking very good!

unparalleled-js:
tests here: https://github.com/ApeWorX/ape-hardhat/pull/108

src/ape_test/__init__.py
```diff
@@ -9,9 +11,51 @@
 from .provider import LocalProvider
 
 
+class GasExclusion(PluginConfig):
+    contract: str = "*"  # If only given method, searches across all contracts.
+    method: Optional[str] = None
```
fubuloubu, line 16: curious if these should be `contract_name` and `method_name`

fubuloubu, line 16:
```suggestion
    contract: str = "*"  # If only given method, searches across all contracts.
    method: Optional[str] = None  # By default, match all methods in a contract
```

fubuloubu:
> Would love at least one other person do QA on this feature, but looking very good!

Did some QA, works great!

fubuloubu (dismissed):

src/ape/types/__init__.py
```diff
@@ -64,6 +65,30 @@
...
 
 
+@dataclass
+class ContractFunctionPath:
+    """
+    Useful for identifying a method in a contract.
+    """
+
+    contract_id: str
+    method_id: Optional[str] = None
```
fubuloubu, line 75: I wouldn't use `method_id` because that has a particular contextual meaning which doesn't match here

fubuloubu (approved):

PR #1115, opened 2022-11-01
fix: issue preventing use of geth with non parity traces

unparalleled-js:
### What I did

Issue preventing the use of local geth (like `geth --dev`) from working with traces.
Parity traces have been tested to death but regular geth hadn't.
This makes it so you can run tests using `etheruem:local:geth` and get tracing support.

**We have tracing support within core ape. We have really had it this whole time! Let's use it more.**

### Checklist


unparalleled-js:
Partner PR: https://github.com/ApeWorX/evm-trace/pull/23

fubuloubu (approved):

PR #1118, opened 2022-11-02
feat: add `snapshot()` and `revert()` implementations to `ape-geth` local provider

unparalleled-js:
### What I did

While working on my trace call work, I realized I needed to separate the Geths in order to subscribe to `TestProviderAPI`.
While making the `GethDev` class, I noticed there were a couple Test APIs that  could easily be implemented, namely `snapshot()` and `revert()`.

### How I did it

Use `debug_setHead` to return to an older block.
For the `SnapshotID`, use the block number of the latest block.
Now, you have the same feature implemented in the Geth Dev provider we have.
This is 1 step closer to making GethDev great, because it has amazing tracing features so I want to encourage its use.

The work related to the refactor is needed for my trace work.

### How to verify it

**NOTE: Cannot add very interesting tests because the logic is rather simple and we don't have a `geth` installed on CI/CD yet to test on.**

Here is the test if I were to add one:

```python
def test_snapshot_and_revert(geth_dev, account_0, account_1):
    block_number = geth_dev.snapshot()
    # Increase blocks
    account_0.transfer(account_1, "1 ETH")
    assert geth_dev.get_block("latest") == block_number + 1
    geth_dev.revert(block_number)
    assert geth_dev.get_block("latest") == block_number
```

You can use or something like it to manually verify the feature.

### Checklist

### DIFF
--- a/src/ape_geth/provider.py
+++ b/src/ape_geth/provider.py
@@ -349,8 +349,6 @@ class GethDev(TestProviderAPI, BaseGethProvider):
 
         super().disconnect()
 
-        super().disconnect()
-
     def revert(self, snapshot_id: SnapshotID):
         if isinstance(snapshot_id, int):
             block_number = to_hex(snapshot_id)
--- a/src/ape_geth/provider.py
+++ b/src/ape_geth/provider.py
@@ -4,7 +4,10 @@ from typing import Any, Dict, Iterator, List, Optional, Union
 
 import ijson  # type: ignore
 import requests
-from eth_utils import to_wei
+from eth_typing import HexStr
+from eth_utils import to_wei, to_hex, add_0x_prefix
+
+from ape.types import SnapshotID
 from evm_trace import (
     CallTreeNode,
     CallType,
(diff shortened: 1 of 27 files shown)


fubuloubu:
> **NOTE: Cannot add very interesting tests because the logic is rather simple and we don't have a `geth` installed on CI/CD yet to test on.**

 Is it possible to simple download the binary from the releases page?

src/ape_geth/provider.py
```diff
@@ -249,35 +205,27 @@ def is_likely_poa() -> bool:
                 else "Error getting chain id."
             )
 
-        if is_likely_poa():
+        try:
```
unparalleled-js, line 208: Shoutout to @helloibis for this part

unparalleled-js:
> Is it possible to simply download the binary from the releases page?

Standby for absolutely better local geth tests, I hope. So far so good but using a rather new GH action. It wasn't this simple last I tried it but now thanks to https://github.com/marketplace/actions/install-geth-go-ethereum-and-tools , it is.
I just hope it is not a scam, it seems like this org just popped up recently with many actions.

unparalleled-js:
I will open a pre-cursor PR to this one.

unparalleled-js:
This PR is needed for adding tests: https://github.com/ApeWorX/ape/pull/1120

unparalleled-js:
> Is it possible to simply download the binary from the releases page?

It actually is not that simple, but some people made it simple for us using a GH action.
PR here: https://github.com/ApeWorX/ape/pull/1120

Issue #627, opened 2022-04-01
Unable to have contract attribute named 'balance'


unparalleled-js:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.4.dev19+g66743a22.d20220401

$ ape plugins list
Installed Plugins:
  solidity     0.1.0
  etherscan    0.1.1.dev4+g0edd4eb.d20220329
  fantom       0.1.1
  alchemy      0.1.0
  ens          0.1.1.dev1+g9516d0b
  infura       0.1.1.dev1+g0665321
  starknet     0.1.dev16+g96cb099.d20220401
  vyper        0.1.0
  tokens       0.1.0
  hardhat      0.1.3.dev1+g27fc667
  cairo        0.1.dev6+g36e9667
  jules        0.1.dev29+gbef3bf6
```

* Python Version: 3.7.9
* OS: macOS 12.3

### What went wrong?

My Cairo contract has a storage var named `balance()` but I can't access it because that is also a member of `AddressAPI`.

### How can it be fixed?

Have a way to call methods by string name, such as:

```python
contract.methods["nonce"]()
```

evmBrahmin:
I've run into a similar issue as the above. I am attempting to call the `nonce()` attribute of a gnosis-safe proxy contract. The safe contract stores the number of multi-sig transactions in a public variable called `nonce` which clashes with the `nonce` member of the `AddressAPI` 

Agreed that the solution proposed by @unparalleled-js would be ideal for resolving this issue. Happy to take a crack at this.

PR #1121, opened 2022-11-04
feat: ability to interact with contracts by method string name

evmBrahmin:
### What I did
Added a function, `method_call(method_name: str, *args, **kwargs)`, to the `ContractInstance` class. This makes it possible to directly call a method of the contract using `<contract_name>.method_call('<method_name>', *args, **kwargs)`.

###  Fixes
fixes: #627 

### How I did it
Instead of relying on `__getattr__` to place the unknown attributes, the `method_call` function uses the `method_name` input to look up the `ContractCallHandler` object associated with that method name. It then executes the smart contract function by executing `ContractCallHandler` with `*args` & `**kwargs`. It then returns the output of the smart contract call.

### How to verify it
This is easily verifiable by calling any smart contract function using `<contract_name>.method_call('method_name', *args, *kwargs)`. For examples you can look at lines 26-35 of `test_contract_instance.py`

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -677,9 +677,42 @@ class ContractInstance(BaseAddress):
             # NOTE: Must raise AttributeError for __attr__ method or will seg fault
             raise AttributeError(str(err)) from err
 
-    def method_call(self, method_name: str, *args, **kwargs) -> Any:
+    def call_view_method(self, method_name: str, *args, **kwargs) -> Any:
         """
-        Call a contract function directly using the method_name.
+        Call a contract's view function directly using the method_name.
+        This is helpful in the scenario where the contract has a
+        method name matching an attribute of the BaseAddress class,
+        such as 'nonce' or 'balance'
+
+        Args:
+            method_name (str): The contract method name to be called
+
+        Returns:
+            Output of smart contract interaction
+
+        """
+
+        if method_name in self._view_methods_:
+            view_handler = self._view_methods_[method_name]
+            output = view_handler(*args, **kwargs)
+            return output
+
+        elif method_name in self._mutable_methods_:
+            mutable_handler = self._mutable_methods_[method_name]
+            output = mutable_handler(*args, **kwargs)
+            return output
+
+        else:
+            # Didn't find anything that matches
+            # NOTE: `__getattr__` *must* raise `AttributeError`
+            name = self.contract_type.name or self.__class__.__name__
+            raise AttributeError(f"'{name}' has no attribute '{method_name}'.")
+
+    def invoke_transaction(self, method_name: str, *args, **kwargs) -> ReceiptAPI:
+        """
+        Call a contract's function directly using the method_name.
+        This function is for non-view function's which may change
+        contract state and will execute a transaction.
         This is helpful in the scenario where the contract has a
         method name matching an attribute of the BaseAddress class,
         such as 'nonce' or 'balance'
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -686,9 +686,11 @@ class ContractInstance(BaseAddress):
 
         Args:
             method_name (str): The contract method name to be called
+            *args: Contract method arguments.
+            **kwargs: Transaction values, such as ``value`` or ``sender``
 
         Returns:
-            Output of smart contract interaction
+            Any: Output of smart contract view call.
 
         """
 
@@ -719,9 +721,11 @@ class ContractInstance(BaseAddress):
 
         Args:
             method_name (str): The contract method name to be called
+            *args: Contract method arguments.
+            **kwargs: Transaction values, such as ``value`` or ``sender``
 
         Returns:
-            Output of smart contract interaction
+            ReceiptAPI: Output of smart contract interaction.
 
         """
 
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -677,6 +677,36 @@ class ContractInstance(BaseAddress):
             # NOTE: Must raise AttributeError for __attr__ method or will seg fault
             raise AttributeError(str(err)) from err
 
+    def method_call(self, method_name: str, *args, **kwargs) -> Any:
+        """
+        Call a contract function directly using the method_name.
+        This is helpful in the scenario where the contract has a
+        method name matching an attribute of the BaseAddress class,
+        such as 'nonce' or 'balance'
+
+        Args:
+            method_name (str): The contract method name to be called
+
+        Returns:
+            Output of smart contract interaction
+
+        """
+        if method_name not in {*self._view_methods_, *self._mutable_methods_}:
+            # Didn't find anything that matches
+            # NOTE: `__getattr__` *must* raise `AttributeError`
+            name = self.contract_type.name or self.__class__.__name__
+            raise AttributeError(f"'{name}' has no attribute '{method_name}'.")
+
+        if method_name in self._view_methods_:
+            view_handler = self._view_methods_[method_name]
+            output = view_handler(*args, **kwargs)
+            return output
+
+        elif method_name in self._mutable_methods_:
+            mutable_handler = self._mutable_methods_[method_name]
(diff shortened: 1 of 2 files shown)


src/ape/contracts/base.py
```diff
@@ -677,6 +677,36 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+        such as 'nonce' or 'balance'
+
+        Args:
+            method_name (str): The contract method name to be called
+
+        Returns:
+            Output of smart contract interaction
+
+        """
+        if method_name not in {*self._view_methods_, *self._mutable_methods_}:
```
unparalleled-js, line 694: This section could be an `else:` after the next set if `if / elif` to save a comparison

src/ape/contracts/base.py
```diff
@@ -677,6 +677,36 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
             # NOTE: Must raise AttributeError for __attr__ method or will seg fault
             raise AttributeError(str(err)) from err
 
+    def method_call(self, method_name: str, *args, **kwargs) -> Any:
```
unparalleled-js, line 680: I am thinking we may want two methods instead.. One for view calls and one for transactions.
Reason: The return types are really different. Like one returns a `ReceiptAPI` and one returns the return value.

To sound more action-y:

```suggestion
    def call_view_method(self, method_name: str, *args, **kwargs) -> Any:
```

And the one for transactions could be:

```python
def invoke_transaction(self, method_name: str, *args, **kwargs) -> ReceiptAPI:
```

src/ape/contracts/base.py
```diff
@@ -677,6 +677,36 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+        """
+        Call a contract function directly using the method_name.
+        This is helpful in the scenario where the contract has a
+        method name matching an attribute of the BaseAddress class,
+        such as 'nonce' or 'balance'
+
+        Args:
+            method_name (str): The contract method name to be called
```
unparalleled-js, line 688: It is still good to document the `*args` and `**kwargs`:

```python
*args: Contract method arguments.
**kwargs: Transaction values, such as ``value`` or ``sender``.
```

src/ape/contracts/base.py
```diff
@@ -677,6 +677,36 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+
+        Returns:
+            Output of smart contract interaction
```
unparalleled-js, line 691:
```suggestion
            Any: Output of smart contract view call.
```

tests/functional/test_contract_instance.py
```diff
@@ -40,6 +40,12 @@ def test_contract_calls(owner, contract_instance):
     assert contract_instance.myNumber() == 2
 
 
+def test_contract_mehtod_calls(owner, contract_instance):
```
unparalleled-js, line 43: The name of this strikes the same intent as the test above and encompasses the same. Can they be combined?

Also, `method` is spelled wrong here.

evmBrahmin:
@unparalleled-js Thanks for the feedback, implemented in the latest push!

src/ape/contracts/base.py
```diff
@@ -677,6 +677,74 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+            *args: Contract method arguments.
+            **kwargs: Transaction values, such as ``value`` or ``sender``
+
+        Returns:
+            ReceiptAPI: Output of smart contract interaction.
+
+        """
+
+        if method_name in self._view_methods_:
+            view_handler = self._view_methods_[method_name]
```
unparalleled-js, line 733: I think we need to handle these as transactions since they are in the `invoke_transaction()` method.
To do this, it is something like:

```python
handler = self._view_methods_[method_name].transact
```

To ensure we are correctly doing this, we should add more tests for the following scenarios:

1. Test that shows we can call a mutable method by name in `call_view_method()` and it is treated like a call (a simulation)
2. Test that shows we can call a view method in `invoke_transaction()` and it will transact and return us a receipt, even though it is only a view method.

src/ape/contracts/base.py
```diff
@@ -677,6 +677,74 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+            output = view_handler(*args, **kwargs)
+            return output
+
+        elif method_name in self._mutable_methods_:
+            mutable_handler = self._mutable_methods_[method_name]
```
unparalleled-js, line 703: I think we have to do something like this:

```suggestion
            handler = self._mutable_methods_[method_name].call
```

in order to handle mutable methods as calls.

Let's add tests to be sure.

src/ape/contracts/base.py
```diff
@@ -677,6 +677,74 @@ def _mutable_methods_(self) -> Dict[str, ContractTransactionHandler]:
...
+            method_name (str): The contract method name to be called
+            *args: Contract method arguments.
+            **kwargs: Transaction values, such as ``value`` or ``sender``
+
+        Returns:
+            Any: Output of smart contract view call.
+
+        """
+
+        if method_name in self._view_methods_:
```
unparalleled-js, line 697: We could create a helper method to share some of this code between the two new methods, but see some of the other feedback before doing that!

It may look like this:

```python
def _call_contract(method_name: str, *args, **kwargs):`
```

tests/functional/test_contract_instance.py
```diff
@@ -38,6 +38,9 @@ def test_init_specify_contract_type(
 def test_contract_calls(owner, contract_instance):
     contract_instance.setNumber(2, sender=owner)
     assert contract_instance.myNumber() == 2
+    contract_instance.invoke_transaction("setNumber", 3, sender=owner)
```
unparalleled-js, line 41: This part should probably in a separate test, like `test_invoke_transaction()` or something.
Also, let's assert something:

```suggestion
    receipt = contract_instance.invoke_transaction("setNumber", 3, sender=owner)
    assert not receipt.failed
```

src/ape/contracts/base.py
evmBrahmin: Great catch, I was actually unaware of the difference between passing back a `ContractTransactionHandler` object with and without the `.transact` attached.  Will implement this change

src/ape/contracts/base.py
evmBrahmin: We could do this, it comes down to how we want the `call_view_methods()` function to behave. Right now I have it that the `_view_methods_` handler in `call_view_methods()` function does NOT use `.transact`. This is because `.transact` requires the user to define the `sender` `**kwarg`. View functions imo shouldn't require defining the `sender` because state isn't being changed. 

If we did add the `.transact` to the handler in `call_view_methods()` then we could share the code between the two functions

evmBrahmin: Okay great, thats how I implemented it in the latest changes :D Given the difference between the functions, I don't think it makes sense to use helper code here, unless i'm missing something.

evmBrahmin:
@unparalleled-js Workflows should all pass now, had to remove a print from a test to get the flake8 to pass

tests/functional/test_contract_instance.py
```diff
@@ -38,9 +38,27 @@ def test_init_specify_contract_type(
 def test_contract_calls(owner, contract_instance):
     contract_instance.setNumber(2, sender=owner)
     assert contract_instance.myNumber() == 2
-    contract_instance.invoke_transaction("setNumber", 3, sender=owner)
+
+
+def test_invoke_transaction(owner, contract_instance):
+    # Test mutable method call with invoke_transaction
+    receipt = contract_instance.invoke_transaction("setNumber", 3, sender=owner)
+    print(receipt)
```
unparalleled-js:
```suggestion
```

evmBrahmin: @unparalleled-js implemented in the last commit 😄

unparalleled-js (approved):

Issue #1122, opened 2022-11-06
`parse_type` is buggy


nazariyv:
### Environment information

* OS: linux
* Python Version: 3.10.6
* `ape` and plugin versions:
`ape`: 0.5.3. But I could reproduce the same issue in 0.5.4

```
Installed Plugins:
  etherscan      0.5.2
  ens            0.5.1
  addressbook    0.5.0a1
```
But this bug has nothing to do with plugins

* Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

This also doesn't matter.

### What went wrong?

<img width="1073" alt="ape-bug" src="https://user-images.githubusercontent.com/13678461/200179295-8b9acb9c-4409-41c1-b91f-d900d72739c0.png">

And discord [link](https://discord.com/channels/922917176040640612/922928308109344778/1036431871665524817) to the above.

Essentially, this problem boils down to the fact that `parse_type` has a bug. I debugged it for a little while, but then realised that there is an easier way to solve the problem. There is no need to go from `canonical` type string back to Python representation. There is an `ABIType` that can be easily utilised to produce what `parse_type` was producing, but bug free.

### How can it be fixed?

Going to link the fix here shortly.

unparalleled-js:
Transaction hash that fails: 0x998f1810cc06f7c0aa3a6094082278e882a53a35a57a4f1094213ec78ca45931

PR #1123, opened 2022-11-06
fix: parse_type bug

nazariyv:
### What I did
Removed the previous `parse_type` and implemented one that takes the `dict` form of `ABIType` to produce the same format output that previous `parse_type` was producing, but bug free and more reliably.

fixes: #1122 

### How I did it
Will make a twitter thread and link here later.

### How to verify it
Ape didn't have a proper test for this. There is `tests/functional/utils/test_abi.py` but it doesn't really test the validity. It more so tests that `parse_type` simply parses the regex generated type strings. I would be happy to write the tests for this, but my environment is throwing me an error when I try to run `python -m poetry .`:

```
  File "/home/shredder/git/w/ape/venv/bin/ape", line 5, in <module>
    from ape._cli import cli
  File "/home/shredder/git/w/ape/src/ape/__init__.py", line 7, in <module>
  File "/home/shredder/git/w/ape/src/ape/managers/__init__.py", line 3, in <module>
  File "/home/shredder/git/w/ape/src/ape/plugins/__init__.py", line 10, in <module>
    from .account import AccountPlugin
  File "/home/shredder/git/w/ape/src/ape/plugins/account.py", line 3, in <module>
    from ape.api.accounts import AccountAPI, AccountContainerAPI
  File "/home/shredder/git/w/ape/src/ape/api/__init__.py", line 1, in <module>
    from .accounts import (
  File "/home/shredder/git/w/ape/src/ape/api/accounts.py", line 8, in <module>
    from ape.api.address import BaseAddress
  File "/home/shredder/git/w/ape/src/ape/api/address.py", line 4, in <module>
    from ape.types import AddressType
  File "/home/shredder/git/w/ape/src/ape/types/__init__.py", line 24, in <module>
    from ape.utils.misc import to_int
  File "/home/shredder/git/w/ape/src/ape/utils/__init__.py", line 20, in <module>
    from ape.utils.github import GithubClient, github_client
  File "/home/shredder/git/w/ape/src/ape/utils/github.py", line 10, in <module>
    import pygit2  # type: ignore
  File "/home/shredder/git/w/ape/venv/lib/python3.10/site-packages/pygit2/__init__.py", line 30, in <module>
    from ._pygit2 import *
ImportError: libssl-9ad06800.so.1.1.1k: cannot open shared object file: No such file or directory
```
However, when I run this modified `ape` on the seaport transaction that was reverting. Everything is fine and it works. Here is the reproducible example of the bug before this fix. Feel free to use it to verify that this PR works.

```
#!/usr/bin/env python
import os
from ape import networks, Contract

etherscan_api_key = ''
if not etherscan_api_key:
    raise Exception("you must set ETHERSCAN_API_KEY")

context = networks.parse_network_choice('ethereum:mainnet:geth')
context.__enter__()
receipt = networks.provider.get_receipt('0x998f1810cc06f7c0aa3a6094082278e882a53a35a57a4f1094213ec78ca45931')
receipt.show_trace()
```

I have not verified this solution on outputs.

### Checklist

### DIFF
--- a/src/ape/utils/abi.py
+++ b/src/ape/utils/abi.py
@@ -285,45 +285,12 @@ class LogInputABICollection:
 
         return value
 
-
-def parse_type(output_type: str) -> Union[str, Tuple, List]:
-    if not output_type.startswith("("):
-        return output_type
-
-    # Strip off first opening parens
-    output_type = output_type[1:]
-    found_types: List[Union[str, Tuple, List]] = []
-
-    while output_type:
-        if output_type.startswith(")"):
-            result = tuple(found_types)
-            if "[" in output_type:
-                return [result]
-
-            return result
-
-        elif output_type[0] == "(" and ")" in output_type:
-            # A tuple within the tuple
-            end_index = output_type.index(")") + 1
-            found_type = parse_type(output_type[:end_index])
-            output_type = output_type[end_index:]
-
-            if output_type.startswith("[") and "]" in output_type:
-                end_array_index = output_type.index("]") + 1
-                found_type = [found_type]
-                output_type = output_type[end_array_index:].lstrip(",")
-
-        else:
-            found_type = output_type.split(",")[0].rstrip(")")
-            end_index = len(found_type) + 1
-            output_type = output_type[end_index:]
-
-        if isinstance(found_type, str) and "[" in found_type and ")" in found_type:
-            parts = found_type.split(")")
-            found_type = parts[0]
-            output_type = f"){parts[1]}"
-
-        if found_type:
-            found_types.append(found_type)
-
-    return tuple(found_types)
+def parse_type(t):
+    if t["type"] == "tuple[]":
+        return [tuple([parse_type(c) for c in t["components"]])]
+    elif t["type"] == "tuple":
+        return tuple([parse_type(c) for c in t["components"]])
+    elif t["type"].endswith("]"):
+        return [t["type"][:-2]]
+    else:
+        return t["type"]
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -318,10 +318,11 @@ class Ethereum(EcosystemAPI):
         return HexBytes(b"")
 
     def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
-        output_types = [o.canonical_type for o in abi.outputs]
+        output_types_str = [o.canonical_type for o in abi.outputs]
+        output_types = [parse_type(t.dict()) for t in abi.outputs]
 
         try:
-            vm_return_values = decode(output_types, raw_data)
+            vm_return_values = decode(output_types_str, raw_data)
         except InsufficientDataBytes as err:
             raise DecodingError() from err
 
@@ -332,7 +333,7 @@ class Ethereum(EcosystemAPI):
             vm_return_values = (vm_return_values,)
 
         output_values = [
-            self.decode_primitive_value(v, parse_type(t))
+            self.decode_primitive_value(v, t)
             for v, t in zip(vm_return_values, output_types)
         ]
 
--- a/tests/functional/utils/test_abi.py
+++ b/tests/functional/utils/test_abi.py
@@ -44,5 +44,4 @@ def test_parse_type(s):
     #   - ,int
     #   - int,
     # See tests in `tests_contracts` for specific ABI parsing tests.
-
-    assert parse_type(s)
+    ...
--- a/src/ape/utils/trace.py
+++ b/src/ape/utils/trace.py
@@ -215,13 +215,14 @@ class CallTraceParser(ManagerAccessMixin):
         return parent
 
     def decode_calldata(self, method: MethodABI, raw_data: bytes) -> Dict:
-        input_types = [i.canonical_type for i in method.inputs]
+        input_types_str = [i.canonical_type for i in method.inputs]
+        input_types = [parse_type(i.dict()) for i in method.inputs]
 
         try:
-            raw_input_values = decode(input_types, raw_data)
+            raw_input_values = decode(input_types_str, raw_data)
             input_values = [
                 self.decode_value(
-                    self._ecosystem.decode_primitive_value(v, parse_type(t)),
+                    self._ecosystem.decode_primitive_value(v, t),
                 )
                 for v, t in zip(raw_input_values, input_types)
             ]


src/ape_ethereum/ecosystem.py
```diff
@@ -318,10 +318,11 @@ def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> bytes
         return HexBytes(b"")
 
     def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
-        output_types = [o.canonical_type for o in abi.outputs]
+        output_types_str = [o.canonical_type for o in abi.outputs]
+        output_types = [parse_type(t.dict()) for t in abi.outputs]
```
nazariyv, line 322: can use `o` instead of `t` here for consistency, if you prefer

unparalleled-js:
@nazariyv Next time, can you create a separate branch than `main`? It makes it easier for me to test out the changes locally and manage all of that.

src/ape/utils/abi.py
```diff
@@ -285,45 +285,12 @@ def decode_value(self, abi_type, value):
...
-        if isinstance(found_type, str) and "[" in found_type and ")" in found_type:
-            parts = found_type.split(")")
-            found_type = parts[0]
-            output_type = f"){parts[1]}"
-
-        if found_type:
-            found_types.append(found_type)
-
-    return tuple(found_types)
+def parse_type(t):
```
unparalleled-js, line 288: Can we avoid single-character variable names, to abide by PEP8 style guidelines?

unparalleled-js, line 288: The argument is missing a type annotation and the method is missing a return type annotation.

src/ape_ethereum/ecosystem.py
unparalleled-js: consistency is always good.. you can make them all `t` if you prefer.

tests/functional/utils/test_abi.py
```diff
@@ -44,5 +44,4 @@ def test_parse_type(s):
     #   - ,int
     #   - int,
     # See tests in `tests_contracts` for specific ABI parsing tests.
-
-    assert parse_type(s)
```
unparalleled-js, line 48: Why was this removed? Does it fail now? Why does it fail?

src/ape/utils/abi.py
nazariyv: will address

nazariyv: will address

tests/functional/utils/test_abi.py
nazariyv: because:

- `parse_type` has a different input type
- could not set up a local test environment (precise issue in the description of this PR)

If someone can help me out with the exact steps to set up the test environment, I will address this as well.

nazariyv:
> @nazariyv Next time, can you create a separate branch than `main`? It makes it easier for me to test out the changes locally and manage all of that.

mb. Rushed it a little.

tests/functional/utils/test_abi.py
unparalleled-js: I am not sure about your environment issue, I have not seen that one before.
Seems like `pygit2` is expecting some `libssl` to be installed but it can't find it.
I'd like for us to not rely on `pygit2` in the future... but for now I would figure out how to get that working and document when you had to do.

nazariyv: I tried solving it by installing a specific libssl version on my machine. That didn't help. So I will create a blank repo and write the tests in there and just move it (copy/paste) to this repo (since those tests do not rely on any lib code other than `parse_type`.

nazariyv:
ready for review

src/ape/utils/abi.py
```diff
@@ -285,45 +285,43 @@ def decode_value(self, abi_type, value):
 
         return value
 
+def parse_type(type: Dict[str, Any]) -> Union[str, Tuple, List]:
```
unparalleled-js, line 288: It looks like, based on some failing tests, that this is not handling everything. For example, Solidity structs with nested arrays.

See some of the tests in `tests/functional/test_contract_instance()`!

nazariyv: Managed to make the testing suite work on my Mac (I had problems on Linux Ubuntu).

But now when running the testing suite, it is using `ape` that sits in site-packages (in dependencies). So basically, it's not testing the `ape` in the repo but the one in deps. Why does `ape` install itself as a dep?

How are you running the tests? I go `python -m pytest` in root dir (after activating the `venv` and installing the deps with `pip instsall -e .'[dev]'`)

nazariyv: the very first run of `python -m pytest` after installing a fresh `venv` does use the repo `ape`. However, subsequent uses start using `ape` from deps.

tests/functional/utils/test_abi.py
nazariyv: Ok, the only places it is failing now is these three:

```
FAILED tests/functional/test_contract_instance.py::test_solidity_structs_with_array - assert False
FAILED tests/functional/test_contract_instance.py::test_vyper_struct_arrays - AssertionError: assert False
FAILED tests/functional/test_contract_instance.py::test_solidity_static_struct_arrays - assert False
```

and the changes in this PR certainly do not cause addresses to not be returned in checksum format. I bet the current main has these tests failing too.

nazariyv:
Ok, the only places it is failing now is these three:

```
FAILED tests/functional/test_contract_instance.py::test_solidity_structs_with_array - assert False
FAILED tests/functional/test_contract_instance.py::test_vyper_struct_arrays - AssertionError: assert False
FAILED tests/functional/test_contract_instance.py::test_solidity_static_struct_arrays - assert False
```

and the changes in this PR certainly do not cause addresses to not be returned in checksum format. I bet the current main has these tests failing too.

unparalleled-js:
> I bet the current main has these tests failing too.

The tests pass on main! But maybe we need to use the to_checksum_address / is_checksum_address methods here?

src/ape/utils/abi.py
unparalleled-js: Are you installing in editable mode?

`pip install -e path/to/ape`

nazariyv:
> > I bet the current main has these tests failing too.
> 
> The tests pass on main! But maybe we need to use the to_checksum_address / is_checksum_address methods here?

this is very interesting because this code doesn't touch values at all.

unparalleled-js:
> this is very interesting because this code doesn't touch values at all.

I am going to pull it down and see what I can find out

src/ape/utils/abi.py
nazariyv: `pip install -e .'[dev]'` like it says in `CONTRIBUTING.md`

src/ape/utils/abi.py
```diff
@@ -285,45 +285,47 @@ def decode_value(self, abi_type, value):
...
-    return tuple(found_types)
+    Returns:
+        Union[str, Tuple, List]: Python representation of the type.
+        For example, `uint256[]` becomes ['uint256'].
+    """
+    if type["type"] == "tuple[]":
+        return [tuple([parse_type(c) for c in type["components"]])]
+    elif type["type"] == "tuple":
+        return tuple([parse_type(c) for c in type["components"]])
+    elif type["type"].endswith("[][]"):
```
unparalleled-js, line 326: What about `[][][]` or `[][][][]`, does those still work?
Not that are common data structures in contracts, but programmatically the design should handle it

nazariyv:
> > this is very interesting because this code doesn't touch values at all.
> 
> I am going to pull it down and see what I can find out

Will have time EOD (like in 8 hours) to have a look at this, as well.

src/ape/utils/abi.py
```diff
@@ -285,45 +285,47 @@ def decode_value(self, abi_type, value):
...
+        that we are interested in are `type` and `components`.
+        `components` produces the same two keys (if present),
+        namely `type` and `components`.
 
-    return tuple(found_types)
+    Returns:
+        Union[str, Tuple, List]: Python representation of the type.
+        For example, `uint256[]` becomes ['uint256'].
+    """
+    if type["type"] == "tuple[]":
```
unparalleled-js, line 322: I get tests to pass when this also handles fixed length arrays like `tuple[2]` etc.
Maybe we can swap this check with a regex-like check instead?

`decode_primitive_value()` expects these solidity types to have the component-canonical types in the type-strings but they are missing by the time it gets to the part where it decodes the values. That is how the type + value relate in the context of these failing tests.

nazariyv: What's the definition of `component-canonical types`?

nazariyv: what should parse type produce when we get `uint256[2][3]`?

```python
def parse_type(type):
    match = re.search(r"(^.+?(?=\[|$))(.*)", type["type"])
    if match is None:
        raise ValueError(f"Invalid type: {type['type']}")
    primitive_type, nesting = match.groups()
    num_brackets = nesting.count("[")

    if primitive_type == "tuple" and not nesting:
        return tuple([parse_type(c) for c in type["components"]])
    elif primitive_type == "tuple" and nesting:
        result = tuple([parse_type(c) for c in type["components"]])
        for _ in range(num_brackets):
            result = [result]
        return result
    if nesting:
        result = primitive_type
        for _ in range(num_brackets):
            result = [result]
        return result
    else:
        return primitive_type 
```

returns `[['uint256']]`, but then somewhere down the line this is turned into tuples and the actual output has tuples in there, and the expected one does not.

nazariyv: going to answer this myself. Just ran doctests on the current main, and in fact for `uint256[2][3]` it just returns that.

```
Failed example:
    parse_type('uint256[2][3]')
Expected:
    [['uint256']]
Got:
    'uint256[2][3]'
```

nazariyv: this is very weird, because a `uint256[][]` would be `[['uint256']]`

src/ape/utils/abi.py
nazariyv: I just commented out github test as well as GitHubClient import somewhere, to just silence the error. That made it work for my Ubuntu machine, which is consistently using the correct ape for tests.

nazariyv:
this part in `decode_primitive_value` is a little sus. It expects things like `uint256[2][3]`. But this is a little inconsistent with dynamically sized arrays, which `parse_type` returns as `[['uint256']]`, for example.

nazariyv:
fun fact, if we make `parse_type` to be:

```python
def parse_type(type: Dict[str, Any]) -> Union[str, Tuple, List]:

    if type["type"].startswith("tuple"):
        return tuple([parse_type(c) for c in type["components"]])

    else:
        return type["type"]
```

then pretty much all the tests pass. This is because `decode_primitive_value` is repeating this logic in a sense lol

nazariyv:
Ok, all tests pass. However, I feel like `parse_type` should be nuked in the near future in favour of supporting all this logic in `decode_primitive_value`, since it already handles it anyway. But for now, this should hopefully fix the bug with my transaction. Need to double-check, actually.

unparalleled-js:
> However, I feel like `parse_type` should be nuked in the near future in favour of supporting all this logic in `decode_primitive_value`

 I agree with your assessment and I will add it to our TODOs for Ape 0.6 release.

src/ape/utils/abi.py
```diff
@@ -285,45 +285,28 @@ def decode_value(self, abi_type, value):
 
         return value
 
+def parse_type(type: Dict[str, Any]) -> Union[str, Tuple, List]:
+    """Parses ABIType.dict() into Python types.
```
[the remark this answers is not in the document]
nazariyv: done

unparalleled-js (dismissed):

unparalleled-js (approved):

PR #1124, opened 2022-11-06
fix: pass default contract type to proxy target

jmonteer:
### What I did
Getting a contract type for a proxy pointing to a non-verified contract as target fails and there is no way to specify the default contract type for the target. 

I changed the call to `get(proxy_info.target)` so that it passes the default argument that was specified when calling `get` for the proxy contract. Resulting in `get(proxy_info.target, default)`

### How I did it
the code now passes the optional argument

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -720,7 +720,7 @@ class ContractCache(BaseManager):
 
             if proxy_info:
                 self._local_proxies[address_key] = proxy_info
-                return self.get(proxy_info.target, default)
+                return self.get(proxy_info.target, default=default)
 
             if not self.provider.get_code(address_key):
                 if default:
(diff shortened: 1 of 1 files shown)


src/ape/managers/chain.py
```diff
@@ -720,7 +720,7 @@ def get(
 
             if proxy_info:
                 self._local_proxies[address_key] = proxy_info
-                return self.get(proxy_info.target)
+                return self.get(proxy_info.target, default)
```
unparalleled-js, line 723: We pass optional arguments as kwargs:

```suggestion
                return self.get(proxy_info.target, default=default)
```

unparalleled-js (approved):
Big bonus if you can add a test for this.

unparalleled-js (approved):

PR #1125, opened 2022-11-07
feat: add dev message support in ape.reverts

helloibis:
### What I did

Add support for supplying `dev_message` in `ape.reverts`

### How I did it

1. Added `TransactionError.receipt` which gets supplied by `ReceiptAPI.raise_for_status()`. Because we're expecting to trace failed _committed_ transactions, we should only have to provide the receipt to that single point of failure.
2. Add `ContractType.dev_messages` via https://github.com/ApeWorX/ethpm-types/pull/51
3. Utilize `ReceiptAPI.trace`, `ContractType.pcmap`, and `ContractType.dev_messages` to try and find a suitable message for a given contract inside `RevertsContextManager`

### How to verify it

Test project available at https://github.com/helloibis/ape-revert-test-project

### Checklist


helloibis:
Still needed:

- [ ] `ape-vyper` support for filling in `ContractType.dev_messages`
- [ ] `ape-solidity` support for filling in `ContractType.dev_messages` (tbd via APE-298)
- [ ] `ape-cairo` support for filling in `ContractType.dev_messages` (tbd via APE-299)

Once we have `ape-vyper` at least I think this will be ready for an alpha version. Going to be working on that this week.

unparalleled-js (approved):
Looks good here (1 comment on doc), but let's get a working example (ape-vyper) befor merging just in case?

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,91 @@
...
+
+        contract_dev_message = contract.contract_type.dev_messages[offending_source.line_start]
+
+        if contract_dev_message != self.dev_message:
+            raise AssertionError(f"{assertion_error_prefix} but got '{contract_dev_message}'.")
+
+    def _check_expected_message(self, exception: ContractLogicError):
+        """
+        Compares the revert message given by the exception to the expected message.
+        Raises:
```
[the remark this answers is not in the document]
helloibis: Whoops, remembered to do it on `_check_dev_message()` but not here 😂  

Updated in b7ad17ef

unparalleled-js:
To test this, we may need to pre-compile or adjust some of the already pre-compiled contracts to include the dev messages, and then we can use the `geth` provider for tracing support.

helloibis:
> To test this, we may need to pre-compile or adjust some of the already pre-compiled contracts to include the dev messages, and then we can use the `geth` provider for tracing support.

Good call, I'll be updating my test project with all the test cases and then see about porting that over into here

helloibis:
@unparalleled-js @NotPeopling2day could you take a look over commit 94326c98 specifically? With the new additions to `send_transaction()` we don't call `raise_for_status()` anymore which is where I was attaching the receipt information to the transaction. I updated the `send_transaction()` logic but it's feeling a little clunky - thoughts?

src/ape/api/providers.py
```diff
@@ -960,18 +966,15 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
-            else:
-                raise vm_err from err
+                )
 
-        required_confirmations = (
-            txn.required_confirmations
-            if txn.required_confirmations is not None
-            else self.network.required_confirmations
-        )
+            vm_err.receipt = self.get_receipt(
```
unparalleled-js, line 973: I want to make sure that if for any reason the receipt is non-fetchable, we can carry on.
Like I was going to suggest maybe putting it in a blank `Exception` `try/catch`.

Tests seem to be working though!
But I'd like to make sure we aren't futzing with anything on live network usages.

(task for anyone)
- [ ] See how reverts and other failures work on live networks with these changes

src/ape/api/providers.py
```diff
@@ -960,18 +966,15 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
             if "nonce too low" in str(vm_err):
                 # Add additional nonce information
                 new_err_msg = f"Nonce '{txn.nonce}' is too low"
-                raise VirtualMachineError(
+                vm_err = VirtualMachineError(
```
unparalleled-js, line 969: I think if the `Nonce` is too low, there won't be a receipt, so we maybe could raise here as normal

src/ape/pytest/contextmanagers.py
```diff
@@ -1,26 +1,94 @@
...
+        instruction(s) led to a transaction revert.
 
-    def __exit__(self, exc_type: Type, exc_value: Exception, traceback) -> bool:
-        if exc_type is None:
-            raise AssertionError("Transaction did not revert.")
+        Raises:
+            AssertionError: When the trace or source can not be retrieved, the dev message cannot
+            be found, or the found dev message does not match the expected dev message.
+        """
+        receipt = exception.receipt
```
unparalleled-js, line 22: Instead of storing the receipt on the `exception`, we could potentially look it up now and only store the `txn_hash` on the receipt.

We could make `RevertsContextManager` use the `ManagerAccessMixin` and do

```python
receipt = self.chain_manager.accounts_history[exception.txn_hash]
```

That may help with the refactor you were asking about, during the parts where we get the receipt during `get_vm_err()`

It may also help with other comments I am adding.

Thoughts?

src/ape/api/providers.py
helloibis: Updated in cf803b8b

src/ape/api/providers.py
```diff
@@ -960,18 +966,15 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
 
-        required_confirmations = (
-            txn.required_confirmations
-            if txn.required_confirmations is not None
-            else self.network.required_confirmations
-        )
+            vm_err.receipt = self.get_receipt(
+                txn.txn_hash.hex(),
+                required_confirmations=required_confirmations,
+            )
```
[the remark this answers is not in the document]
helloibis: Updated in f89c5a92

src/ape/api/providers.py
helloibis: Looking through the call chain for `get_receipt()` it seems it can raise `TransactionNotFoundError` or `TransactionError`

Updated in 84b6a53e but will leave this open because there will probably be some more changes around this

src/ape/pytest/contextmanagers.py
helloibis: Updated in 5d4a7c24 but to pass `TransactionAPI` instead, because we basically can get everything we need (`receiver`, `txn_hash`, `provider`, etc) from there

So `TransactionError` now has an optional `txn` parameter for its constructor - I only supply `txn` during failed `Web3Provider.send_transaction()` calls but I'm not sure if we need it elsewhere. For now it basically only facilitates this dev-message assertion feature

helloibis:
Added tests in 02575e8

These include tests for `RevertsContextManager` itself, which we seemed to be missing, as well as tests around the new functionality.

I opted for mocking here so we can just get some functional tests in, I imagine when we get to full integration testing we'd use something like https://github.com/helloibis/ape-revert-test-project. I'm not sure if we would incorporate all those tests into the provider plugins (if the tests differ per provider, then we definitely would) or just move that project into the org and use it as a CI/CD test for providers.

unparalleled-js:
I'd like to add tests next to these ones that include dev message checks: https://github.com/ApeWorX/ape/blob/main/tests/functional/test_geth.py#L67-L80

I can help with this!

unparalleled-js (approved):
Could it be! We finally have dev string assertions? This looks really good. 

I am going to test it some, but the tests look good.

fubuloubu (approved):

fubuloubu:
Yeah agree, some more QA needed, but this looks great!

NotPeopling2day (approved):

helloibis:
@fubuloubu @unparalleled-js here's the link to the more comprehensive test project as well https://github.com/helloibis/ape-revert-test-project

Right now it's primarily situations, but I plan on adding larger, more comprehensive contracts that are more "real world" scenarios too. For now it's mostly like "what situations can Vyper revert" as generally one-liner functions

PR #1128, opened 2022-11-09
perf: only check trace in tests if has a reason to

unparalleled-js:
### What I did

Change it to not check `call_tree` unnecessarily (if no `--gas` flag given).
This makes tests run faster when not using `--gas`

**NOTE**: This PR also fixes many race conditions and flakinesses in the tests that came up when adding a test as well as turning off geth snapshotting for now because of this issue: https://github.com/ethereum/go-ethereum/issues/26154

### How I did it

check if needing gas report _before_ checking the grabbing the call tree.

### How to verify it

Can work with Hardhat project with tracing issues and things still work as far as testing goes (obviously sans all tracing features)

### Checklist


unparalleled-js:
Hmm, thinking I may have a better solution, BRB.

unparalleled-js:
- [ ] Fix issue with snapshotting (noticed while testing)

src/ape_geth/provider.py
```diff
@@ -303,8 +304,21 @@ def _stream_request(self, method: str, params: List, iter_path="result.item"):
...
     name: str = "geth"
 
+    @property
+    def chain_id(self) -> int:
+        if self._chain_id is not None:
+            return self._chain_id
+        elif hasattr(self.web3, "eth"):
+            chain_id = self.web3.eth.chain_id
+        else:
+            return GETH_DEV_CHAIN_ID
```
fubuloubu, line 317: This line won't cache

helloibis (commented):
Is this a configuration that should also apply to the dev-message assertion or do we only mean tracing here as in gas tracing / profiling?

unparalleled-js:
> Is this a configuration that should also apply to the dev-message assertion or do we only mean tracing here as in gas tracing / profiling?

I removed the config because I found out it wasn't needed! So no worries there :)

src/ape_geth/provider.py
```diff
@@ -349,18 +367,41 @@ def disconnect(self):
 
         super().disconnect()
 
+    @raises_not_implemented
+    def snapshot(self) -> SnapshotID:
+        # TODO: Replace with impl below after
+        #  https://github.com/ethereum/go-ethereum/issues/26154 resolved
```
NotPeopling2day, line 373: Do we have a ticket to track this?

NotPeopling2day (approved):

helloibis (approved):

src/ape_geth/provider.py
unparalleled-js: APE-319

Issue #1091, opened 2022-10-06
Collect gas reporting information for view method calls


fubuloubu:
### Overview

Currently, Ape's gas report feature is collecting gas information only for mutable transactions that are executing (due to how tracing works). However, it would be really useful to also profile read-only functions, which will not be executing this way (but will be called)

NOTE: This will also be helpful for test coverage

https://linear.app/apeworx/issue/APE-287/collect-gas-reporting-information-for-view-method-calls

### Specification

Intercept any read-only method calls and send them as a transaction (reverting back scope after the call)

### Dependencies
No dependencies

unparalleled-js:
Question, what do we use as the sender? I guess it does not really matter since it gets reverted, but I am curious.

I wish there was a better way to do this. It feels hacky to use the EVM this way.

fubuloubu:
> Question, what do we use as the sender? I guess it does not really matter since it gets reverted, but I am curious.

Many view methods it doesn't really matter. Could have it default to whatever the first account is if not specified. View methods can be contextual in this way, so it's probably best to have some sort of sender account baked in, even if it doesn't sign.

> I wish there was a better way to do this. It feels hacky to use the EVM this way.

Agreed. Titanoboa will let you perform aribtrary tracing using py-evm opcode extensions

PR #1131, opened 2022-11-11
feat: include view calls in gas report

unparalleled-js:
### What I did

fixes: #1091 

### How I did it

1. For local geth dev, if the user is requesting trace-related info, use `debug_traceCall` in place of `send_call`
(The user can request this either via the `--gas` flag in `ape test` or via kwargs such as `show_trace` or `show_gas_report` while making a call.
3. Adds a protected (`_reports`) manager to `chain_manager`. This is NOT a public API (for now).
4. In `ReceiptCapture` (part of `pytest` plugin code), tell the report manager to track gas. Also, make the `gas_report` from `ReceiptCapture` to the `ReportManager`.
5. Add default implementation in `TestProviderAPI` that sends the call as a transaction, captures its gas, and then reverts it.

**WARN**: The `--gas` flag is really slow for Hardhat (the default implementation)... But using `geth`, it is really fast  I am not sure why that is... Hardhat, when using `--gas` and the changes in this PR, takes twice as long... I think the implementation to send all calls as transaction surrounded by snapshotting is just slow.. There are plans, however, to get `debug_traceCall` in Hardhat _and_ Foundry, so that will help a lot.

### How to verify it

* Use `--gas` flag in a project with Geth provider
* Use `--gas` flag in a project with Hardhat provider

You should now see a list full of the view methods called as well as the mutable ones.

### Checklist


unparalleled-js:
- [ ] add generic impl to `TestProviderAPI` that sends calls as a txn and then reverts it

unparalleled-js:
- [ ] Look into supporting adding flag to log gas report during a call adhocly (out of tests), such as:

```python
contract.balanceOf(address, show_gas_report=True)
```

unparalleled-js:
- [ ] Look into using `CallTraceParser` as more of an injected dependency

NotPeopling2day:
Thanks for the discussion and walk through :)

src/ape/managers/chain.py
```diff
@@ -1064,6 +1105,38 @@ def restore(self, snapshot_id: Optional[SnapshotID] = None):
...
+            end_ecosystem_name = self.provider.network.ecosystem.name
+            end_network_name = self.provider.network.name
+            end_provider_name = self.provider.name
+
+            if (
+                start_ecosystem_name != end_ecosystem_name
+                or start_network_name != end_network_name
+                or start_provider_name != end_provider_name
+            ):
+                logger.warning("Provider changed before could complete isolation.")
```
helloibis, line 1135: Should this maybe read `Provider changed before isolation completed`?

src/ape/api/providers.py
```diff
@@ -850,6 +852,37 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:
...
+                else:
+                    for handler in self.debug_trace_handlers.values():
+                        handler(
+                            call_tree, txn.sender, txn.receiver, transaction_hash=receipt.txn_hash
+                        )
+
+                    return call_tree.returndata
+
+        except APINotImplementedError:
+            return self._send_call(txn, **kwargs)
```
helloibis, line 883: What can raise `APINotImplementedError` in the block above? We might wanna wrap the try/except around only the parts that can raise if possible

helloibis (dismissed):

src/ape/api/providers.py
unparalleled-js: it is the `__enter__` method of the `with` statement, so I think it has include everything in the with part too

unparalleled-js: Also could happen in the `__exit__` maybe

src/ape/api/providers.py
```diff
@@ -850,6 +850,32 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:
...
+            or not txn.receiver
+        ):
+            return self._send_call(txn, **kwargs)
+
+        # Handle call tree, if trace handlers set.
+        # For example, this is used for capturing gas usage in view calls
+        # during tests when the `--gas` flag was set.
+
+        try:
+            with self.chain_manager.isolate():
```
unparalleled-js, line 865: - [ ] Figure out why this part is really really really slow and see if can do anything about it.

src/ape/api/providers.py
unparalleled-js: I made the stuff in the middle into a helper method so it is more obvious that the try /except is about the `with isolate` line.

unparalleled-js:
- [ ] Ensure can ignore methods that are calls

NotPeopling2day (approved):
LGTM

PR #1141, opened 2022-11-19
refactor: replace pygit2 with subprocess calls to `git`

unparalleled-js:
### What I did

Curently, we use `pygit2` when dependencies are specified via branch names instead of release versions because the GitHub release API does not support branches as arguments for obtaining releases; only release version IDs.

There are several reasons to not use `pygit2` as a core dependency for ape though:

* Installation issues on some operating systems
* It is not used in branch; just in a single use-case.

And more that I won't get into right now.

### How I did it

Replace usage of `pygit2` with `subprocess.run()` calls. Now, the expectation of certain git environments are not enforced. Users can avoid branch-based dependencies (as they should anyway) when they don't have the ability to use `git`, and it does not affect the core ape installation.

### How to verify it

Use github dependencies that are branch based:

```
dependencies:
  - name: ds-test
    github: dapphub/ds-test
    branch: master
```

^ Pick a different one than me though.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,7 +91,6 @@ setup(
         "pandas>=1.3.0,<2",
         "pluggy>=1.0.0,<2",
         "pydantic>=1.9.2,<2",
-        "pygit2>=1.7.2,<2",
         "PyGithub>=1.54,<2",
         "pytest>=6.0,<8.0",
         "python-dateutil>=2.8.2,<3",
--- a/src/ape/utils/github.py
+++ b/src/ape/utils/github.py
@@ -1,23 +1,44 @@
 import os
-import re
 import shutil
+import subprocess
 import tempfile
 import zipfile
 from io import BytesIO
 from pathlib import Path
 from typing import Dict, Optional, Set
 
-import pygit2  # type: ignore
 from github import Github, UnknownObjectException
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -91,7 +91,6 @@
         "pandas>=1.3.0,<2",
         "pluggy>=1.0.0,<2",
         "pydantic>=1.9.2,<2",
-        "pygit2>=1.7.2,<2",
         "PyGithub>=1.54,<2",
```
poolpitako, line 94: this doesn't depend on pygit2 as well?

fubuloubu (approved):

fubuloubu:
Just to double check, `pip freeze | grep pygit2` is empty right?

unparalleled-js:
> pip freeze | grep pygit2

It empty!

setup.py
unparalleled-js: This PR replaces the usage of pygit2 with `subprocess.run(["path/to/git"])`.
What do you mean by `as well`?

poolpitako: @unparalleled-js I was wondering if PyGithub had pygit2 as a dep as well but if `pip freeze | grep pygit2` was empty. This is sorted out! :+1:

PR #1143, opened 2022-11-21
test: add reverts test documentation

helloibis:
Adding in documentation for `ape.reverts()` as well as the new `dev_message` feature (and its current caveats)

### DIFF
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -250,6 +250,7 @@ If the message in the `ContractLogicError` raised by the transaction failure is 
 ### `dev_message`
 
 This is the expected dev message corresponding to the line in the contract's source code where the error occurred.
+These can be helpful in optimizing for gas usage and keeping revert reason strings shorter.
 
 Dev messages take the form of a comment in Vyper, and should be placed on the line that may cause a transaction revert:
 
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -254,7 +254,7 @@ This is the expected dev message corresponding to the line in the contract's sou
 Dev messages take the form of a comment in Vyper, and should be placed on the line that may cause a transaction revert:
 
 ```python
-assert x != 0 # dev: invalid value
+assert x != 0  # dev: invalid value
 ```
 
 Take for example:
@@ -264,7 +264,7 @@ Take for example:
 
 @external
 def check_value(_value: uint256) -> bool:
-    assert _value != 0 # dev: invalid value
+    assert _value != 0  # dev: invalid value
     return True
 ```
 
@@ -299,8 +299,8 @@ Due to function inlining, the position of the `# dev: ...` message may sometimes
 
 ```python
 @external
-def foo(_x: decimal) -> decimal: # dev: correct location
-    return sqrt(_x) # dev: incorrect location
+def foo(_x: decimal) -> decimal:  # dev: correct location
+    return sqrt(_x)  # dev: incorrect location
 ```
 
 This typically only applies when trying to add dev messages to statements containing built-in function calls.
@@ -312,13 +312,13 @@ Similarly, if you require dev assertions for non-reentrant functions you must be
 ```python
 @internal
 @nonreentrant('lock')
-def _foo_internal(): # dev: correct location
+def _foo_internal():  # dev: correct location
     pass
 
 @external
 @nonreentrant('lock')
 def foo():
-    self._foo_internal() # dev: incorrect location
+    self._foo_internal()  # dev: incorrect location
 ```
 
 ## Multi-chain Testing
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
     assert actual == expect
 ```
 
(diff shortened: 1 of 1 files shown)


docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+### `expected_message`
+
+This is the expected revert reason given when the transaction fails.
+If the message in the `ContractLogicError` raised by the transaction failure is empty or does not match the `expected_message`, then `ape.reverts()` will raise an `AssertionError`.
+
+### `dev_message`
+
+This is the expected dev message corresponding to the line in the contract's source code where the error occurred.
+
+Dev messages take the form of a comment in Vyper, and should be placed on the line that may cause a transaction revert:
```
unparalleled-js, line 254: Should we include a note about how this is used for gas optimization and keeping revert reasons shorter?

docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+```
+
+Take for example:
+
+```python
+# @version 0.3.7
+
+@external
+def check_value(_value: uint256) -> bool:
+    assert _value != 0 # dev: invalid value
```
unparalleled-js, line 267: To match default python lint rules, should we put 2 spaces between comment and code?

Does this matter in terms of detecting dev reverts?

docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+```
+
+When the transaction reverts and `ContractLogicError` is raised, `ape.reverts()` will check the source contract and see if the line that failed contains the message.
+
+There are a few scenarios where `AssertionError` will be raised when using `dev_message`:
+- If the line in the source contract has a different dev message or no dev message
+- If the contract source cannot be obtained
+- If the transaction trace cannot be obtained
+
+Because `dev_message` relies on transaction tracing to function, you must use a provider like `ape-hardhat` when testing with `dev_message`.
```
unparalleled-js, line 286: We can link to ape-hardhat here

docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+
+### Caveats
+
+#### Language Support
+
+As of `ape` version `0.5.6`, `dev_messages` assertions are only available with Vyper and `ape-vyper`.
+
+You must use `ape-vyper` to compile your contracts for the relevant source information to be present for `ape.reverts()`.
```
unparalleled-js, line 294: I think we should rephrase this a bit. Shortly after we release, some dedicated dev could add dev-str detection to their super secret language plugin, in which case this statement becomes false.

unparalleled-js (commented):
This is concise and helpful. A few notes!

docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+```python
+# @version 0.3.7
+
+@external
+def check_value(_value: uint256) -> bool:
+    assert _value != 0 # dev: invalid value
+    return True
+```
+
+We can explicitly cause a transaction revert and check that it was that assertion line that failed by supplying an expected `dev_message`:
```
NotPeopling2day, line 271:
```suggestion
We can explicitly cause a transaction revert and check the failed line by supplying an expected `dev_message`:
```

docs/userguides/testing.md
```diff
@@ -225,6 +225,102 @@ def test_account_balance(project, owner, receiver, nft):
...
+
+```python
+def test_authorization(my_contract, owner):
+    with ape.reverts(dev_message="dev: invalid value"):
+        my_contract.check_value(sender=owner)
+```
+
+When the transaction reverts and `ContractLogicError` is raised, `ape.reverts()` will check the source contract and see if the line that failed contains the message.
```
NotPeopling2day, line 279:
```suggestion
When the transaction reverts and `ContractLogicError` is raised, `ape.reverts()` will check the source contract to see if the failed line contains a message.
```

docs/userguides/testing.md
helloibis: Updated in 101f372a

> Does this matter in terms of detecting dev reverts?

No, we basically just check for `#\s*(dev:.*)` regardless of what precedes it

docs/userguides/testing.md
helloibis: Updated in 28015695

I basically rephrased it to say "as of 0.5.6 this works with ape-vyper, but not ape-solidity or ape-cairo"

That way it's specific to which of our plugins don't have support yet, rather than saying only `ape-vyper` can ever be used

docs/userguides/testing.md
helloibis: Updated in 68372465 including links for `ape-vyper`, `ape-solidity`, and `ape-cairo` in my other paragraph

docs/userguides/testing.md
helloibis: Updated in 71d5cc10

unparalleled-js (approved):

Issue #1145, opened 2022-11-22
`block_identifier` is not passed correctly to rpc endpoint in v0.5.5


gosuto-inzasheru:
### Environment information

* OS: macOS
* Python Version: 3.10.8
* `ape` and plugin versions:

```sh
$ ape --version
0.5.5

$ ape plugins list
Installed Plugins:
  etherscan    0.5.4
  arbitrum     0.5.1
```

### What went wrong?
```
In [1]: apecoin = Contract('0x4d224452801ACEd8B2F0aebE155379bb5D594381')

In [2]: apecoin.balanceOf(apecoin)
Out[2]: 65352087960278510417976

In [3]: apecoin.balanceOf(apecoin, block_identifier=14404163)
[...]
VirtualMachineError: 400 Client Error: Bad Request for url:

In [4]: apecoin.balanceOf(apecoin, block_identifier=hex(14404163))
Out[4]: 18112000000000000000000
```
for completeness, this is the raw response for rpc (not sure why the code and message do not make their way to the `VirtualMachineError`?):
```
{"jsonrpc":"2.0","id":14,"error":{"code":-32602,"message":"invalid 2nd argument: block_parameter value was not valid, neither block nor hash specified"}}
```

note that this problem does not occur in v0.5.4!

### How can it be fixed?

 i believe this stems from the fact that since commit https://github.com/ApeWorX/ape/commit/a3cde612a2ae7e2b7ebcd06a173ef80c4d49a57f, eth_calls are made using `self.web3.provider.make_request(RPCEndpoint(endpoint), parameters)` and not `self.web3.eth.call(txn.dict(), block_id, state)` as was the case previously. somewhere the conversion of `hex(block_identifier)` does not occur in `provider.make_request`, whereas in `eth.call` it does.

PR #1146, opened 2022-11-22
fix: convert `block_identifier` to hex

gosuto-inzasheru:
### What I did
fixes: #1145

### How I did it
this might be a bit of naive approach, a more thorough solution might lay in understanding why `web3.provider.make_request` does not properly handle `block_identifier` as an int (it should convert to hex for the final raw rpc call).

### How to verify it
see #1145 for a example call to run

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -926,10 +926,9 @@ class Web3Provider(ProviderAPI, ABC):
             if value is not None and not isinstance(value, str):
                 txn_dict[field] = to_hex(value)
 
-        if "block_identifier" in kwargs:
-            block_identifier = hex(kwargs["block_identifier"])
-        else:
-            block_identifier = "latest"
+        block_identifier = kwargs.pop("block_identifier", "latest")
+        if isinstance(block_identifier, int):
+            block_identifier = to_hex(block_identifier)
         arguments = [txn_dict, block_identifier]
         if "state_override" in kwargs:
             arguments.append(kwargs["state_override"])
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -926,7 +926,10 @@ class Web3Provider(ProviderAPI, ABC):
             if value is not None and not isinstance(value, str):
                 txn_dict[field] = to_hex(value)
 
-        block_identifier = kwargs.pop("block_identifier", "latest")
+        if "block_identifier" in kwargs:
+            block_identifier = hex(kwargs["block_identifier"])
+        else:
+            block_identifier = "latest"
         arguments = [txn_dict, block_identifier]
         if "state_override" in kwargs:
             arguments.append(kwargs["state_override"])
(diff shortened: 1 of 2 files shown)


gosuto-inzasheru:
not sure about the relevancy of test cases and the docs here, would like some feedback in general before continuing.

src/ape/api/providers.py
```diff
@@ -926,7 +926,10 @@ def _prepare_call(self, txn: TransactionAPI, **kwargs) -> List:
             if value is not None and not isinstance(value, str):
                 txn_dict[field] = to_hex(value)
 
-        block_identifier = kwargs.pop("block_identifier", "latest")
+        if "block_identifier" in kwargs:
+            block_identifier = hex(kwargs["block_identifier"])
```
unparalleled-js, line 930: What if the user already passed a hex str?
Also, if the user explicitly passes `"latest"`, this will fail and try to convert `"latest"` to a hex str.

Maybe it should be like this:

```python
from eth_utils import to_hex, is_0x_prefixed
...

block_identifier = kwargs.get("block_identifier", "latest")
if isinstance(block_identifier, int):
    block_identifier = to_hex(block_identifier])

# else: assume the user specified a correct HexStr or special-value
```

src/ape/api/providers.py
```diff
@@ -926,7 +926,10 @@ def _prepare_call(self, txn: TransactionAPI, **kwargs) -> List:
             if value is not None and not isinstance(value, str):
                 txn_dict[field] = to_hex(value)
 
-        block_identifier = kwargs.pop("block_identifier", "latest")
+        if "block_identifier" in kwargs:
```
unparalleled-js, line 929: For improving / adding a test, see this test as an example: https://github.com/ApeWorX/ape/blob/main/tests/functional/test_contract_instance.py#L53-L54

unparalleled-js:
I will open the test on a separate PR since it was not included here and I can't push to `patch-1`-type branches for some reason.

unparalleled-js (dismissed):

unparalleled-js:
> not sure about the relevancy of test cases and the docs here, would like some feedback in general before continuing.

For test, this is what I have got. Feel free to add / adjust:

```python
def test_call_use_block_identifier(contract_instance, owner, chain):
    expected = 2
    contract_instance.setNumber(expected, sender=owner)
    block_id = chain.blocks.height  # int
    contract_instance.setNumber(3, sender=owner)  # latest
    actual = contract_instance.myNumber(block_identifier=block_id)
    assert actual == expected

    # Ensure works with hex
    block_id = to_hex(block_id)
    actual = contract_instance.myNumber(block_identifier=block_id)
    assert actual == expected

    # Ensure works keywords like "latest"
    actual = contract_instance.myNumber(block_identifier="latest")
    assert actual == 3
```

gosuto-inzasheru:
awesome, added your suggestions :)

unparalleled-js (approved):

PR #1148, opened 2022-11-22
feat!: move `parse_type` method to ethereum plugin

Ninjagod1251:
### What I did

I am moving parse_type method from ape_utils to ape_ethereum

fixes: #

### How I did it
Move the abi section of code to a new file of abi.py under ape_ethereum

### How to verify it
run pytest and make sure the old test still work


src/ape_ethereum/abi.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any, Dict, List, Tuple, Union
+from ape.utils import is_array
+
+def parse_type(type: Dict[str, Any]) -> Union[str, Tuple, List]:
```
unparalleled-js, line 4: I would put this in `ecosystem.py` somewhere, maybe as `_parse_type()` or inline the functionality into the spot where it's used, no need to export it.. the goal is make it not public anymore

NotPeopling2day: One issue is that it's still used in tracing, `ape/utils/trace.py` in the `decode_calldata()` method.

src/ape/utils/trace.py
```diff
@@ -21,9 +21,11 @@
 
 from ape.exceptions import ContractError, DecodingError
 from ape.utils import ManagerAccessMixin
-from ape.utils.abi import Struct, parse_type
+from ape.utils.abi import Struct
 from ape.utils.misc import ZERO_ADDRESS
 
+from ape_ethereum.ecosystem import _parse_type
```
NotPeopling2day, line 27: This line is creating the circular import error you're getting. While we're still discussing the refactor, let's move this so that it's not erroring out. See the comment below.

src/ape/utils/trace.py
```diff
@@ -223,7 +225,7 @@ def _dim_default_gas(call_sig: str) -> str:
 
     def decode_calldata(self, method: MethodABI, raw_data: bytes) -> Dict:
```
NotPeopling2day, line 226:
```suggestion
    def decode_calldata(self, method: MethodABI, raw_data: bytes) -> Dict:
        from ape_ethereum.ecosystem import _parse_type
               
```

src/ape/utils/trace.py
unparalleled-js: the idea of this ticket is to move the logic from parse_type into `ape_ethereum` and have it happen when calling `decode_primitive_value()` and then it does not need to be imported here.

NotPeopling2day: Yeah, this was just a temporary step to get his tests passing so he could research what else needed to change. 👍

src/ape/utils/trace.py
NotPeopling2day: Comment above, temporary fix :)

unparalleled-js:
See https://github.com/ApeWorX/ape/pull/1160

PR #1150, opened 2022-11-23
feat: improve error when contract type collision

unparalleled-js:
### What I did

Currently, when you have a contract type collision, it looks like this:

```
ERROR: (CompilerError) ContractType collision across compiler plugins with contract: abi/v0.6/Ownable.json
```

but now with these changes, it looks like this:

```
ERROR: (CompilerError) ContractType collision between sources 'abi/v0.6/ENSInterface.json' and 'abi/v0.7/ENSInterface.json'.
```

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -82,7 +82,7 @@ class CompilerManager(BaseManager):
 
         extensions = self._get_contract_extensions(contract_filepaths)
         contracts_folder = self.config_manager.contracts_folder
-        contract_types_dict = {}
+        contract_types_dict: Dict[str, ContractType] = {}
         for extension in extensions:
             path_patterns_to_ignore = self.config_manager.compiler.ignore_files
             ignore_path_lists = [contracts_folder.rglob(p) for p in path_patterns_to_ignore]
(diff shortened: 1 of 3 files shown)


src/ape_pm/compiler.py
```diff
@@ -1,12 +1,21 @@
...
+from ape.exceptions import CompilerError
+from ape.logging import logger
 from ape.utils import get_relative_path
 
+KNOWN_JSON_IGNORE_LIST = (
+    "package.json",
+    "package-lock.json",
+    "tsconfig.json",
+)
+
```
fubuloubu, line 18: Hmm, would this make sense as a global ignore via config file?

e.g,

```yaml
compiler:
    ignore-files:
    - package.json
    - package-lock.json
    - tsconfig.json
```

src/ape_pm/compiler.py
```diff
@@ -21,25 +30,35 @@ def get_versions(self, all_paths: List[Path]) -> Set[str]:
...
         self, filepaths: List[Path], base_path: Optional[Path] = None
     ) -> List[ContractType]:
+        filepaths.sort()  # Sort to assist in reproducing consistent results.
         contract_types: List[ContractType] = []
         for path in filepaths:
-            data = json.loads(path.read_text())
+            try:
+                data = json.loads(path.read_text())
+            except JSONDecodeError as err:
+                logger.error(f"({CompilerError}) {err}")
```
fubuloubu, line 39: what does putting `CompilerError` in the f-string do?

src/ape_pm/compiler.py
unparalleled-js: Done. I made it a separate PR and will strip it out of this one, thanks!
https://github.com/ApeWorX/ape/pull/1151

src/ape/managers/compilers.py
```diff
@@ -99,16 +99,27 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
...
-                if contract_type.name in contract_types_dict:
-                    raise CompilerError(
-                        "ContractType collision across compiler plugins "
-                        f"with contract: {contract_type.source_id}"
+                contract_name = contract_type.name
+                if not contract_name:
+                    if contract_type.source_id:
+                        logger.warning(f"Contract '{contract_type.source_id}' missing name.")
+                    else:
+                        logger.warning(f"Empty contract type found in compiler '{extension}'.")
```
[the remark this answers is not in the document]
unparalleled-js: We had some issues in `ape-pm` but I fixed them there, so these warnings don't happen anymore. **I agree these should be error states!** So I changed them to raise errors here.

This would still happen for custom compiler plugins that are poorly implemented, so I adjusted the warning to describe that:

```python
  raise CompilerError(
      f"Contract '{contract_type.source_id}' missing name. "
      f"Was compiler plugin for '{extension} implemented correctly?"
  )
```

src/ape_pm/compiler.py
unparalleled-js: Meant to do `CompilerError.__name__` but this should actually raise an error so going to **not** do this.. Thanks.

tests/integration/cli/test_compile.py
```diff
@@ -42,13 +43,6 @@ def test_missing_extensions(ape_cli, runner, project):
     assert ".foobar" in result.output
 
 
-@skip_projects_except("unregistered-contracts")
```
unparalleled-js, line 45: duplicate test

fubuloubu (approved):

Issue #799, opened 2022-06-10
brownie style dependencies when compiling a `BrownieProject`


unparalleled-js:
### Overview

New issue coming out of: https://github.com/ApeWorX/ape/issues/749

Dependencies in ape work differently than dependencies in Brownie.
In ape, the dependencies have to be manifest-able. In brownie, they just get pulled in.

If we allow dependencies for Brownie projects to work _brownie-style_, we can unblock a lot of user migrating over. And it makes sense, technically speaking!

**SIZE NOTE**: I am putting a 5 because of potential refactor complications but feel free to adjust 😅 

### Specification

* Don't make manifest from dependencies download when processing brownie dependencies.

NOTE: May require factoring config processing for brownie dependencies. Unknowns alert!

https://linear.app/apeworx/issue/APE-206/brownie-style-dependencies-when-compiling-a-brownieproject

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PatrickAlphaC:
Note: `ContractType collision across compiler plugins ` is what you'll get right now if you try to compile a project where contracts have the same name but different versions of solidity.

hhamud:
Any plans on completing this issue?

unparalleled-js:
> Any plans on completing this issue?

I am working on this now.

PR #1152, opened 2022-11-26
feat: don't compile dependencies

unparalleled-js:
### What I did

fixes: #799 

Goes with https://github.com/ApeWorX/ape-solidity/pull/84

### How I did it

Dependency's no longer need to compile. Instead, we create a contract-type-less `PackageManifest` that contains the sources to be imported in a project. This makes compiling works a lot better and certain dependencies don't compile if they don't ever need to.

### How to verify it

Use dependencies.

### Checklist

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -237,16 +237,8 @@ class DependencyAPI(BaseInterfaceModel):
 
         return None
 
-    def _extract_local_manifest(self, project_path: Path):
-        project_path = project_path.resolve()
-        contracts_folder = project_path / self.contracts_folder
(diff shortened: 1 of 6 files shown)


fubuloubu (approved):
Oh nice, this is clever. Basically, if it fails to compile, there won't be contract types, but there will be sources available for import?

unparalleled-js:
> Oh nice, this is clever. Basically, if it fails to compile, there won't be contract types, but there will be sources available for import?

Yes! As our per talk on Telegram, it seems like maybe we never need to compile dependencies (until publishing) since this is the case.

fubuloubu (approved):

PR #1153, opened 2022-11-26
fix: change to http default in github clone for dependencies

unparalleled-js:
### What I did

This works in CI better than SSH.

### Checklist

### DIFF
--- a/src/ape/utils/github.py
+++ b/src/ape/utils/github.py
@@ -141,7 +141,7 @@ class GithubClient:
         repo_path: str,
         target_path: Path,
         branch: Optional[str] = None,
-        scheme: str = "ssh",
+        scheme: str = "http",
     ):
         """
         Clone a repository from Github.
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
Test locally to make sure it doesn't prompt for username and pasaword

unparalleled-js:
It does not prompt me, but I don't remember if I configured it as such.  You can configure HTTPS to not prompt you. Maybe someone else can test as well? I am curious how CI will handle it.

unparalleled-js:
you know - I believe it was using HTTPS before. When I removed pygit2, I accidentally switched the default... for what it's worth.

fubuloubu:
> you know - I believe it was using HTTPS before. When I removed pygit2, I accidentally switched the default... for what it's worth.

That makes so much more sense lol

fubuloubu:
Verified this works! okay to merge

NotPeopling2day (approved):

fubuloubu:
I believe this still is not working for our CI action
https://github.com/fubuloubu/WETH8/actions/runs/3585793308/jobs/6034209359

PR #1157, opened 2022-11-28
feat: add default yaml config

wakamex:
### What I did
attempt to pull out all the defaults from code into a single yaml

### Why I did it
this can be used to transparently add more detailed parameters, like in https://github.com/ApeWorX/ape-alchemy/pull/35. also allows non-coder users to understand what the default settings are, and to easily override them.

### How I did it
removed parameters and hard-coded them into the standard yaml. maybe a separate `default-config.yaml` would be better than `ape-config.yaml`.
replaced the context-based test_config with an overload to the default config that executes additional overrides (I bet this is causing lots of test failures):
` = 195 failed, 512 passed, 3 skipped, 7 xfailed, 15 warnings, 31 errors in 1125.89s (0:18:45) =`

### How to verify it
possibly additional tests should be needed, though i got it working with test_config.py

### Checklist
- [ ] finish testing (possibly go back to context-based test_config)
- [ ] possibly rename to `default-config.yaml`, distinct from `ape-config.yaml`

### DIFF
--- a/ape-config.yaml
+++ b/ape-config.yaml
@@ -1,0 +1,37 @@
+default_required_confirmations: &default_required_confirmations 2
+default_local_required_confirmations: &default_local_required_confirmations 0
+default_local_block_time: &default_local_block_time 0
+default_transaction_acceptance_timeout: &default_transaction_acceptance_timeout 120
+default_local_transaction_acceptance_timeout: &default_local_transaction_acceptance_timeout 20
+default_gas_limit: &default_gas_limit "auto"
+default_local_gas_limit: &default_local_gas_limit "max"
+ethereum:
+  mainnet:
+    required_confirmations: *default_required_confirmations
+    block_time: 13
+    transaction_acceptance_timeout: *default_transaction_acceptance_timeout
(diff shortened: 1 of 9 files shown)


unparalleled-js:
`ape --config` was supposed to have similar intent here, and not require any source code.
You should be able to do `ape --config` and see the default config

ape-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    required_confirmations: *default_required_confirmations
+    block_time: 13
+    transaction_acceptance_timeout: *default_transaction_acceptance_timeout
+    gas_limit: *default_gas_limit
+    # alchemy:
+    #   min_retry_delay: 1_000 # 1 second delay on fail
+    #   retry_backoff_factor: 2 # explonential backoff
+    #   max_retry_delay: 30_000 # 30 seconds
+    #   max_retries: 3
+    #   retry_jitter: 250 # 250 milliseconds
```
wakamex, line 19: like this @fubuloubu? 😉 (re: getting provider settings from Config)

Issue #947, opened 2022-08-05
Decode calldata using ContractInstance


bout3fiddy:
### Overview

Provide a simple overview of what you wish to see added. Please include:

* What you are trying to do
* Why Ape's current functionality is inadequate to address your goal

In CurveDAO, DAO-controlled voting entities (AragonDAO contracts) can execute random calldata if governance meets quorum. It's important hence to be able to decode the calldata and verify if it isn't malicious. In brownie, this can be achieved by:

```
decoded_calldata = contract.decode_input(calldata)
```
but this is currently not available on the contract instance. 

### Specification

Describe the syntax and semantics of how you would like to see this feature implemented. The more detailed the better!

Remember, your feature is much more likely to be included if it does not involve any breaking changes.

Ape already has this implementation in [ape > utils > trace.py](https://github.com/ApeWorX/ape/blob/main/src/ape/utils/trace.py#L215-L236 ) : 

```
def decode_calldata(self, method: MethodABI, raw_data: bytes) -> Dict:
    input_types = [i.canonical_type for i in method.inputs]  # type: ignore

    try:
        raw_input_values = decode_abi(input_types, raw_data)
        input_values = [
            self.decode_value(
                self._ecosystem.decode_primitive_value(v, parse_type(t)),
            )
            for v, t in zip(raw_input_values, input_types)
        ]
    except (DecodingError, InsufficientDataBytes):
        input_values = ["<?>" for _ in input_types]

    arguments = {}
    index = 0
    for i, v in zip(method.inputs, input_values):
        name = i.name or f"{index}"
        arguments[name] = v
        index += 1

    return arguments
```
The goal would be to allow the contractinstance to be able to do this by:

```
decoded_calldata: Dict = contract.decode_calldata(raw_data: bytes)
```
the decode_calldata would need to figure out the method automagically.

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

fubuloubu:
We have something similar
https://docs.apeworx.io/ape/stable/methoddocs/api.html#ape.api.networks.EcosystemAPI.decode_returndata

But obviously this doesn't work quite as nicely

unparalleled-js:
> We have something similar https://docs.apeworx.io/ape/stable/methoddocs/api.html#ape.api.networks.EcosystemAPI.decode_returndata
> 
> But obviously this doesn't work quite as nicely

That is only the outputs though and not the inputs right?
[...]

PR #1160, opened 2022-11-29
feat!: encode and decode calldata from contracts

unparalleled-js:
### What I did

fixes: #947 
Fixes APE-317
Fixes APE-242

**NOTE: See `.md` changes for more thorough usage example and notes about the feature to help with Testing and review!**

### How I did it

Create a mixin `ContractMethodHandler` that both `ContractCallHandler` and `ContractTransactionHandler` inherit from.
This mixin class contains two new methods in addition to other already-shared coded:

* `encode_calldata()` looks at the args, select the ABI, and encode the calldata.
* `decode_calldata()` takes bytes, checks the selector, and gives you a tuple with both the string selector and a dict mapping input names to values. If you don't give it a selector prefix, it will do its best (copying from Brownie).

Then, creates another mixin `ContractTypeWrapper` that both `ContractInstance` and `ContractContainer` inherit from. It contains already-existing shared code as well as two new methods:

* `encode_calldata()` that looks at the args count and finds the method with the same number of inputs, if it exists. Otherwise, it fails. Note: Not sure about this approach, copied it from brownie.. seems like I would not recommend banking on succeed here.
* `decode_calldata()` that checks for a prefix to find the right ABI, if it exists. If it didn't find anything, it will attempt a brute-force check. Note, it can only return in any case if there is only 1 result, else it is too ambivalent to know I think.

Also had to refactor `ape.utils.trace` to move some of that logic to the ecosystem, and have it be share-able with other parts of the code-base and override-able at the same time. Thanks @Ninjagod1251 for beginning that work.

### How to verify it

```python
# contract is a ContactInstance or ContractContainer
contract.decode_calldata(bytes)

# contract is only a ContractInstance
contract.method.decode_calldata(bytes)
contract.method.encode_calldata(*args)
```

### Checklist


src/ape/api/networks.py
```diff
@@ -439,6 +469,20 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
         """
         return None
 
+    def get_method_selector(self, abi: MethodABI) -> HexBytes:
+        """
+        Get a contract method selector, typically via hashing such as ``keccak``.
+        Defaults to using ``keccak`` but can be overridden in different ecosystems.
```
fubuloubu, line 475: How do you override?

src/ape/contracts/base.py
```diff
@@ -48,6 +48,14 @@ def __init__(
     def __repr__(self) -> str:
         return self.abi.signature if self.abi else "constructor()"
 
+    def encode_calldata(self, *args) -> HexBytes:
+        ecosystem = self.provider.network.ecosystem
+        encoded_calldata = ecosystem.encode_calldata(self.abi, *args)
+        return HexBytes(encoded_calldata)
+
+    def decode_calldata(self, calldata: bytes) -> Dict[str, Any]:
```
unparalleled-js, line 56: - [ ] Change return type to be like `brownie`:

```
(selector, [args])
```

src/ape/api/networks.py
unparalleled-js: I just added an example to the doc!

src/ape/contracts/base.py
```diff
@@ -559,7 +611,88 @@ def poll_logs(
             yield from self.range(new_block.number, stop=new_block.number + 1)
 
 
-class ContractInstance(BaseAddress):
+class ContractTypeWrapper(ManagerAccessMixin):
+    contract_type: ContractType
+
+    def encode_input(self, *args: Any) -> HexBytes:
```
unparalleled-js, line 617: I am wondering if we should drop this method. It seems weird to rely on number of args. Brownie had this so I included it but I can't imagine wanting to rely on this. Thoughts? Maybe I am misunderstanding something as well.

Better if you know at least the method name so you can do `contract.theMethod.encode_input(arg0, arg1, ...)`

delaaxe:
would this work with `ape-starknet` out of the box?

unparalleled-js:
> would this work with `ape-starknet` out of the box?

No, `ape-starknet` would have to implement a few methods in its `EcosystemAPI` implementation. I don't think it would be too difficult though!

delaaxe:
> > would this work with `ape-starknet` out of the box?
> 
> No, `ape-starknet` would have to implement a few methods in its `EcosystemAPI` implementation. I don't think it would be too difficult though!

which ones?

unparalleled-js:
> which ones?

See the changes in `src/ape/api/networks.py::EcosystemAPI`. There are 3 new abstract methos:

```python
def decode_calldata():
def encode_calldata():
def get_method_selector
```

**NOTE**: These are changes are still in draft + code review so this could change!
**Second NOTE**: These changes wouldn't be released until the Ape 0.6 range because of its breaking changes.

src/ape/contracts/base.py
fubuloubu: yeah, I do agree this would be weird to use if there were matching input lengths and types e.g. `ERC20.transfer` and `ERC20.approve`

fubuloubu (dismissed):

src/ape/contracts/base.py
unparalleled-js: - [ ] Delete this method.

src/ape/contracts/base.py
```diff
@@ -48,6 +48,17 @@ def __init__(
...
         return self.abi.signature if self.abi else "constructor()"
 
+    def encode_input(self, *args) -> HexBytes:
+        ecosystem = self.provider.network.ecosystem
+        encoded_calldata = ecosystem.encode_calldata(self.abi, *args)
+        return HexBytes(encoded_calldata)
+
+    def decode_input(self, calldata: bytes) -> Tuple[str, Dict[str, Any]]:
+        input_str = ",".join([str(i.internalType or i.type) for i in self.abi.inputs])
+        selector = f"constructor({input_str})"
```
fubuloubu, line 58: Should this be `self.abi.selector`?

unparalleled-js: `selector` is missing from `ConstructorABI` currently

unparalleled-js: https://github.com/ApeWorX/ethpm-types/pull/54

If that releases, I can use it and bump the ethpm-types

src/ape/contracts/base.py
```diff
@@ -559,7 +611,47 @@ def poll_logs(
...
 
-class ContractInstance(BaseAddress):
+class ContractTypeWrapper(ManagerAccessMixin):
+    contract_type: ContractType
+
+    def decode_input(self, calldata: bytes) -> Tuple[str, Dict[str, Any]]:
+        """
+        Decode the given calldata using this contract.
+        If the calldata has a method ID prefix, Ape with detect it and find
+        the corresponding method. Else, it not be able to succeed.
```
NotPeopling2day, line 621:
```suggestion
    def decode_input(self, calldata: bytes) -> Tuple[str, Dict[str, Any]]:
        """
        Decode the given calldata using this contract.
        If the calldata has a method ID prefix, Ape will detect it and find
        the corresponding method, else it will error.
```

src/ape_ethereum/ecosystem.py
```diff
@@ -314,16 +313,38 @@ def decode_block(self, data: Dict) -> BlockAPI:
 
         return Block.parse_obj(data)
 
-    def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> bytes:
-        if abi.inputs:
-            input_types = [i.canonical_type for i in abi.inputs]
-            return encode(input_types, args)
+    def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> HexBytes:
+        if not abi.inputs:
+            return HexBytes(b"")
```
NotPeopling2day, line 318:
```suggestion
            return HexBytes("")
```

src/ape/contracts/base.py
fubuloubu: released https://github.com/ApeWorX/ethpm-types/releases/tag/v0.3.13

unparalleled-js: UGH!!!
https://github.com/ApeWorX/ethpm-types/pull/55

fubuloubu (commented):
Why is `src/ape/types/trace.py` empty?

fubuloubu (dismissed):

src/ape/contracts/base.py
```diff
@@ -113,9 +122,62 @@ def __repr__(self) -> str:
         abis = sorted(self.abis, key=lambda abi: len(abi.inputs or []))
         return abis[-1].signature
 
+    def encode_input(self, *args) -> HexBytes:
+        args = self._convert_tuple(args)
+        selected_abi = _select_method_abi(self.abis, args)
```
fubuloubu, line 127: Can this be made into a separate method?

src/ape/contracts/base.py
```diff
@@ -48,6 +48,15 @@ def __init__(
     def __repr__(self) -> str:
         return self.abi.signature if self.abi else "constructor()"
 
+    def encode_input(self, *args) -> HexBytes:
```
fubuloubu, line 51: Curious if these all might be better named `encode/decode_calldata` instead of `*_input`

src/ape/contracts/base.py
```diff
@@ -113,9 +122,62 @@ def __repr__(self) -> str:
...
+        args = self._convert_tuple(args)
+        selected_abi = _select_method_abi(self.abis, args)
+        ecosystem = self.provider.network.ecosystem
+        encoded_calldata = ecosystem.encode_calldata(selected_abi, *args)
+        method_id = ecosystem.get_method_selector(selected_abi)
+        return HexBytes(method_id + encoded_calldata)
+
+    def decode_input(self, calldata: bytes) -> Tuple[str, Dict[str, Any]]:
```
fubuloubu, line 133: Does this do similar logic to what Brownie does?

fubuloubu (approved):

fubuloubu:
CI failure is something new, seems like `actions/checkout@v2` is unable to fetch https repos. Try upgrading to v4

src/ape/contracts/base.py
unparalleled-js: I did that first, but I found that `input` / `output` describes the high-level nature of these methods for aptly whereas `calldata` and `returndata` describe the low level data nicely. In `EcoystemAPI`, we use `calldata` and `returndata` this way.

Also, this way it is the same as Brownie.

But I don't feel super strongly about it, just was explaining my thought process for why I thought `input` made more sense.

src/ape/contracts/base.py
unparalleled-js: What do you mean? I am interpreting that you mean make `select_method_abi()` a public method here for users but am not 100% sure!

src/ape/contracts/base.py
unparalleled-js: Yes but the second part of the return tuple's format is a little different.
Brownie returns a tuple yes, but the second part is a list of decoded values.
Here, it is a dict of input names (or stringified index if not found) to decoded values.

src/ape/contracts/base.py
fubuloubu: Okay that's fair

NotPeopling2day (approved):

Issue #873, opened 2022-07-12
Allow regex patterns in revert assertions


helloibis:
### Overview

Adding regex support for `ape.reverts()` messages, allowing users to pass either `str` or `re.Pattern`.

### Specification

The type for the messages will effectively be `Optional[str or re.Pattern]`:

```python
# No message checking
with ape.reverts():
    foo()

with ape.reverts():
    bar()

# Normal usage
with ape.reverts("A transaction error message"):
    foo()

with ape.reverts("Another transaction error message"):
    bar()

# Pattern matching usage
pattern = re.compile(r"error message$")

with ape.reverts(pattern):
    foo()

with ape.reverts(pattern):
    bar()
```

### Dependencies

Should begin after https://github.com/ApeWorX/ape/issues/251 adds the new `dev_message` parameter

PR #1162, opened 2022-11-29
feat: support regex in ape.reverts()

helloibis:
### What I did
- Added regex support in `ape.reverts()`
- Cleared up `RevertsContextManager` tests and test docstrings

fixes: #873 
fixes: APE-311

### How I did it
Allow `expected_message` and `dev_message` in `RevertsContextManager` to accept `re.Pattern` objects

### How to verify it
```python
# Matches explictly "dev: foo" or "dev: bar"
with ape.reverts(re.compile(r"^dev: (foo|bar)$")):
    ...

# Matches explictly "dev: foo" or "dev: bar"
with ape.reverts(dev_message=re.compile(r"^dev: (foo|bar)$")):
    ...
```

### Checklist

### DIFF
--- a/src/ape/pytest/contextmanagers.py
+++ b/src/ape/pytest/contextmanagers.py
@@ -80,11 +80,13 @@ class RevertsContextManager:
         if offending_source is None or offending_source.line_start is None:
             raise AssertionError("Could not find line that caused revert.")
 
-        assertion_error_prefix = "Expected dev revert message '{}.'".format(
+        assertion_error_message = (
             self.dev_message.pattern
             if isinstance(self.dev_message, re.Pattern)
             else self.dev_message
         )
+
+        assertion_error_prefix = f"Expected dev revert message '{assertion_error_message}'"
 
         dev_messages = contract.contract_type.dev_messages or {}
 
@@ -112,11 +114,13 @@ class RevertsContextManager:
         """
         actual = exception.revert_message
 
-        assertion_error_prefix = "Expected revert message '{}.'".format(
+        assertion_error_message = (
             self.expected_message.pattern
             if isinstance(self.expected_message, re.Pattern)
             else self.expected_message
         )
+
+        assertion_error_prefix = f"Expected revert message '{assertion_error_message}'"
 
         message_matches = (
             (self.expected_message.match(actual) is not None)
--- a/src/ape/pytest/contextmanagers.py
+++ b/src/ape/pytest/contextmanagers.py
@@ -1,5 +1,6 @@
+import re
 from collections import deque
-from typing import Optional, Type
+from typing import Optional, Type, Union
 
 import ape
 from ape.exceptions import (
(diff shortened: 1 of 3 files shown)


src/ape/pytest/contextmanagers.py
```diff
@@ -75,7 +80,11 @@ def _check_dev_message(self, exception: ContractLogicError):
         if offending_source is None or offending_source.line_start is None:
             raise AssertionError("Could not find line that caused revert.")
 
-        assertion_error_prefix = f"Expected dev revert message '{self.dev_message}'."
+        assertion_error_prefix = "Expected dev revert message '{}.'".format(
```
unparalleled-js, line 83: This is probably just a personal preference, but I find f-strs more readable than `.format()` strings.

```python
message = self.dev_message.pattern if isinstance(self.dev_message, re.Pattern) else self.dev_message
assertion_error_prefix = f"Expected dev revert message '{message}.'"
```

tests/functional/test_revert_context_manager.py
```diff
@@ -43,32 +63,59 @@ def test_revert_msg_raises_partial(owner, reverts_contract_instance):
...
+    Test that ``AssertionError`` is raised if the contract dev message does not match the supplied
+    dev revert pattern.
     """
+    with pytest.raises(AssertionError):
+        with reverts(dev_message=re.compile(r"dev: [^ero]+")):
+            reverts_contract_instance.revertStrings(2, sender=owner)
+
 
+@geth_process_test
+def test_both(owner, reverts_contract_instance, geth_provider):
```
unparalleled-js, line 115:
```suggestion
def test_both_message_and_dev_str(owner, reverts_contract_instance, geth_provider):
```

or something a ta more

unparalleled-js (approved):
Looks good! A couple small suggestions but nothing blocking.

unparalleled-js (approved):

unparalleled-js:
Just had this thought: If the user provides a normal message but the assertion fails, I wonder if we should go ahead and check if it is a regex string and re-try the assertion. Thoughts?

I am worried users may get confused when regex-strings are not working.

NotPeopling2day (approved):

helloibis:
> Just had this thought: If the user provides a normal message but the assertion fails, I wonder if we should go ahead and check if it is a regex string and re-try the assertion. Thoughts?
> 
> I am worried users may get confused when regex-strings are not working.

@unparalleled-js The issue is we can't really check a regex without either having them pass in `re.Pattern` or using a different kwarg name (which is what Brownie does, e.g. `dev_msg_pattern` or similar)

If someone were to pass in a string that happened to contain regex characters like `?` or `.`, we can't really tell if they were trying to use regex or trying to actually use a string

For example `"test?"` could be passed in as a string by accident, but let's say the dev comment in the contract is `"test"`. If we treated that as regex then, it would actually pass the assertions even though it shouldn't. It seems like that could create confusion about what should or shouldn't be matched, since we're not using Brownie's method of using differently named kwargs

PR #1170, opened 2022-12-01
feat: `ape accounts export <ALIAS>` exports private key

sabotagebeats:
### What I did
created a click command in ape accounts cli to export private keys from keyfiles

fixes: APE-345
fixes: APE-328

### How I did it
created a click account that loads the alias keyfile and then decrypts the private key with a password prompt

### How to verify it
tests coming soon

### Checklist

### DIFF
--- a/src/ape_accounts/_cli.py
+++ b/src/ape_accounts/_cli.py
@@ -64,7 +64,7 @@ def _list(cli_ctx, show_all_plugins):
             click.echo()
 
 
-@cli.command(short_help="Create a new keyfile account with a random mnemonic seed phrase")
+@cli.command(short_help="Create an account with a random mnemonic seed phrase")
 @click.option(
     "--hide-mnemonic",
     help="Hide the newly generated mnemonic from the terminal",
@@ -113,9 +113,7 @@ def generate(cli_ctx, alias, hide_mnemonic, word_count, custom_hd_path):
 
 
 # Different name because `import` is a keyword
-@cli.command(
-    name="import", short_help="Add a new keyfile account by entering a private key or seed phrase"
-)
+@cli.command(name="import", short_help="Import an account by private key or seed phrase")
 @click.option(
     "--use-mnemonic", "import_from_mnemonic", help="Import a key from a mnemonic", is_flag=True
 )
--- a/src/ape_accounts/_cli.py
+++ b/src/ape_accounts/_cli.py
@@ -157,7 +157,7 @@ def _import(cli_ctx, alias, import_from_mnemonic, custom_hd_path):
     )
 
 
-@cli.command(short_help="Unlock an account and export the private key")
+@cli.command(short_help="Export an account private key")
 @ape_cli_context()
 @existing_alias_argument()
 def export(cli_ctx, alias):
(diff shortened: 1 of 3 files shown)


src/ape_accounts/_cli.py
```diff
@@ -157,6 +157,17 @@ def _import(cli_ctx, alias, import_from_mnemonic, custom_hd_path):
     )
 
 
+@cli.command(name="export", short_help="Unlock an account and export the private key")
```
fubuloubu, line 160: Don't need to do this because `export` is not a python reserved keyword
```suggestion
@cli.command(short_help="Unlock an account and export the private key")
```

src/ape_accounts/_cli.py
```diff
@@ -157,6 +157,17 @@ def _import(cli_ctx, alias, import_from_mnemonic, custom_hd_path):
...
+@ape_cli_context()
+@existing_alias_argument()
+def _export(cli_ctx, alias):
```
fubuloubu, line 163: Click will generate command name from the function name here, there's no need for the underscore (unlike `_import`) 
```suggestion
def export(cli_ctx, alias):
```

sabotagebeats: ok perfect thanks!

sabotagebeats:
todo: documentation

fubuloubu (dismissed):

src/ape_accounts/_cli.py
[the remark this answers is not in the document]
sabotagebeats: how about `Export the private key of an existing account`?

fubuloubu (dismissed):

fubuloubu:
@NotPeopling2day have someone do a little QA and can merge

src/ape_accounts/_cli.py
```diff
@@ -157,6 +157,17 @@ def _import(cli_ctx, alias, import_from_mnemonic, custom_hd_path):
     )
 
 
+@cli.command(short_help="Export an account private key")
+@ape_cli_context()
+@existing_alias_argument()
```
unparalleled-js, line 162: So this `existing_alias_argument` decorator takes an `account_type` kwarg. I think we should set that to the `KeyfileAccount`. Else, it lets me give it a ledger / trezor account or even a starknet one it fails when trying to open the file because it doesn't exist. Those accounts should not even be options.

unparalleled-js (commented):
Another feedback, but more so to the last couple PRs:
The short-help text is kind of long now for the `import` and `generate` command and causes some unnecessary newlines.

src/ape_accounts/_cli.py
```diff
@@ -157,6 +157,17 @@ def _import(cli_ctx, alias, import_from_mnemonic, custom_hd_path):
...
+def export(cli_ctx, alias):
+    path = _get_container().data_folder.joinpath(f"{alias}.json")
+    account = json.loads(path.read_text())
+    password = click.prompt("Enter password to decrypt account", hide_input=True)
+    private_key = EthAccount.decrypt(account, password)
+    cli_ctx.logger.success(f"Account 0x{account['address']} private key: {private_key.hex()}")
```
unparalleled-js, line 168: Optional thing, just for awareness: There is a way to bold things, if you want to bold the key or address or anything.
I think it is like `click.style` or something and it takes some markup args.

unparalleled-js (dismissed):
I found a couple small things while testing, but overall looks good

src/ape_accounts/_cli.py
sabotagebeats: ok I got this working :)

src/ape_accounts/_cli.py
sabotagebeats: I am unsure what to do to fix this, but it could alternatively try and then fail with the error that it's not an exportable key?

fubuloubu (approved):

PR #1171, opened 2022-12-01
chore: prevent multiple CI runs when using pushed branched

unparalleled-js:
### What I did

Currently, when pushing a branch and opening a PR, you get multiple runs.
This makes it so only one happens.
https://stackoverflow.com/questions/58895283/stop-already-running-workflow-job-in-github-actions

### Checklist


fubuloubu:
This seems to work great! closing #1004

fubuloubu (approved):

.github/workflows/test.yaml
```diff
@@ -2,6 +2,10 @@ on: ["push", "pull_request"]
 
 name: Test
 
+concurrency:
+  group: ${{ github.head_ref }}
```
fubuloubu, line 6: someone gave the suggestion to use `github.ref` instead

fubuloubu:
Okay I guess `github.ref` doesn't work

unparalleled-js:
> Okay I guess `github.ref` doesn't work

Reverted!

Another nice thing about this that I noticed is that if you happen to push again before the workflow finishes, it will cancel it and only focus on the new one.

unparalleled-js:
So it seems like when merging to main, the tests don't run again. But is that actually a problem? Because we require all branches to be caught up with main before they can merge anyway, so maybe its seen as a duplicate.

unparalleled-js:
Ah... Found the problem:

```
Invalid workflow file
The workflow is not valid. .github/workflows/test.yaml (Line: 7, Col: 10): Unexpected type of value '', expected type: String.
```

fubuloubu:
> Ah... Found the problem:
> 
> ```
> Invalid workflow file
> The workflow is not valid. .github/workflows/test.yaml (Line: 7, Col: 10): Unexpected type of value '', expected type: String.
> ```
> 
> The `.ref` thing does not exist on that event perhaps.

Ah yes, so maybe it needs to detect if it is a pull request or it is a branch push

.github/workflows/test.yaml
fubuloubu, line 6: Vyper is using `${{ github.workflow }}-{{ github.head_ref }}` here:

https://github.com/vyperlang/vyper/pull/3179

unparalleled-js:
It seems like we are getting duplicate runs again, after copying Vyper's approach.

NotPeopling2day (approved):

PR #1173, opened 2022-12-01
feat: allow setting balance if current provider supports it

fubuloubu:
### What I did
Allow setting an account's balance directly

### How I did it
Tried to add a setter to the `balance` property, but that didn't work, so I had to override Pydantic behavior a bit

### How to verify it
Load an `Address`, any `AccountAPI` subclass, and then any `ContractInstance` and try `account.balance += "1 ether"`

### Checklist
Needs some tests

### DIFF
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -43,7 +43,7 @@ The accounts generated from this seed are solely for testing and debugging purpo
 
 Learn more about test accounts from the [testing guide](./testing.html#accounts-fixture).
 
-If your testing provider supports this feature, it is possible to directly set the balanaces of any address, by performing the following action:
+If your testing provider supports this feature, it is possible to directly set the balances of any address by performing the following action:
 
 ```python
 account.balance += int(1e18)  # Gives `account` 1 Ether
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -43,6 +43,12 @@ The accounts generated from this seed are solely for testing and debugging purpo
 
 Learn more about test accounts from the [testing guide](./testing.html#accounts-fixture).
 
+If your testing provider supports this feature, it is possible to directly set the balanaces of any address, by performing the following action:
+
+```python
+account.balance += int(1e18)  # Gives `account` 1 Ether
+```
+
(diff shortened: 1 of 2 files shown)


src/ape/api/address.py
```diff
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, List, Union
```
fubuloubu, line 1:
```suggestion
from typing import List, Union
```

src/ape/api/address.py
```diff
@@ -90,6 +90,22 @@ def balance(self) -> int:
...
 
+    # @balance.setter
+    # NOTE: commented out because of failure noted within `__setattr__`
+    def _set_balance_(self, value: Union[str, int]):
+        self.provider.set_balance(
+            self.address,
+            self.conversion_manager.convert(value, int) if isinstance(value, str) else value,
+        )
+
+    def __setattr__(self, attr: str, value: Any) -> None:
```
fubuloubu, line 101: Think it's best not even to bother with typing on this temporary code
```suggestion
    def __setattr__(self, attr, value):
```

fubuloubu:
Tested with Foundry plugin, `Address`, `ContractInstance` and `AccountAPI` all work properly

Note that `account.balance += "1 ether"` doesn't work, need to use something like `int(1e18)` instead

docs/userguides/accounts.md
```diff
@@ -43,6 +43,12 @@ The accounts generated from this seed are solely for testing and debugging purpo
 
 Learn more about test accounts from the [testing guide](./testing.html#accounts-fixture).
 
+If your testing provider supports this feature, it is possible to directly set the balanaces of any address, by performing the following action:
```
unparalleled-js, line 46:
```suggestion
If your testing provider supports this feature, it is possible to directly set the balances of any address by performing the following action:
```

unparalleled-js (dismissed):

unparalleled-js (approved):

src/ape/api/address.py
banteg: possible now with pydantic 2
https://github.com/pydantic/pydantic/pull/5502

Issue #1176, opened 2022-12-07
gas_limit param causes error in 0.5.7


r4VP4:
### Environment information

* OS: ubuntu 22.04
* Python Version: 3.9.15

```sh
$ ape --version
# 0.5.7

$ ape plugins list
#Installed Plugins:
  polygon      0.5.1a1
  hardhat      0.5.3
  foundry      0.5.2
  vyper        0.5.2
  alchemy      0.5.3
  solidity     0.5.2
  etherscan    0.5.4
```

* Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
hardhat:
  fork:
    ethereum:
      mainnet:
        upstream_provider: alchemy  

foundry:
  fork:
    ethereum:
      mainnet:
        upstream_provider: alchemy       
        
ethereum:
  mainnet:
    gas_limit: "max"
```

### What went wrong?

calling a contract in a script with gas_limit parameter 
```
mycontract.mymethod(
   myparam,
   sender=myaccount,
   gas_limit='auto' # or gas_limit=750000 makes no difference
)
```
```
ape run my_script --network ethereum:mainnet-fork:foundry
# or
ape run my_script --network ethereum:mainnet-fork:hardhat
```

leads to the following error

```
ERROR: (ConversionError) Type 'typing.Union[typing.Literal['auto', 'max'], int, str, NoneType]' must be one of [ChecksumAddress, bytes, int, Decimal, list, tuple].
```

fubuloubu:
Taking a stab at this, but think it's originating from here: https://github.com/ApeWorX/ape/blob/main/src/ape/contracts/base.py#L25

r4VP4:
i've seen your commits in #1180 but i'm not sure if this is the solution, because i have the problem always if i use the gas_limit parameter. it is not just for 'auto'. sorry for the misleading example.

fubuloubu:
related: #1220

PR #1180, opened 2022-12-07
fix: gas_limit of "auto" causes failure with unsupported providers

fubuloubu:
### What I did

Some providers don't support "auto", so just avoid getting gas estimation for those types by removing the argument

fixes: #1176

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -730,6 +730,14 @@ class Web3Provider(ProviderAPI, ABC):
         """
 
         txn_dict = txn.dict()
+
+        # "auto" does not usually work
+        if "gas" in txn_dict and txn_dict["gas"] == "auto":
+            txn_dict.pop("gas")
+            # Also pop these, they are overriden by "auto"
+            txn_dict.pop("maxFeePerGas", None)
+            txn_dict.pop("maxPriorityFeePerGas", None)
+
         try:
             block_id = kwargs.pop("block_identifier", None)
             txn_params = cast(TxParams, txn_dict)
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -934,6 +934,12 @@ class Web3Provider(ProviderAPI, ABC):
             if value is not None and not isinstance(value, str):
                 txn_dict[field] = to_hex(value)
 
+        # Remove unneeded properties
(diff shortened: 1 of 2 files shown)


src/ape/api/providers.py
```diff
@@ -730,6 +730,14 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
         """
 
         txn_dict = txn.dict()
+
+        # "auto" does not usually work
```
unparalleled-js, line 734:
```suggestion
```

`auto` only means get into the method we are in now.

I don't see a problem with popping it out at this point! That makes sense.

src/ape/api/providers.py
```diff
@@ -730,6 +730,14 @@ def estimate_gas_cost(self, txn: TransactionAPI, **kwargs) -> int:
...
+        if "gas" in txn_dict and txn_dict["gas"] == "auto":
+            txn_dict.pop("gas")
+            # Also pop these, they are overriden by "auto"
+            txn_dict.pop("maxFeePerGas", None)
+            txn_dict.pop("maxPriorityFeePerGas", None)
```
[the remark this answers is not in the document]
unparalleled-js: It should be fine. We are not modifying anything used out of this method. So this is only preparing the dict needed for the estimate call.

unparalleled-js:
I tested on several all our providers and did not notice any issues!

unparalleled-js (approved):

unparalleled-js (approved):

src/ape_test/provider.py
```diff
@@ -117,13 +117,17 @@ def priority_fee(self) -> int:
...
-        if "gas" not in data or data["gas"] == 0:
-            data["gas"] = int(1e12)
-
         block_id = kwargs.pop("block_identifier", None)
         state = kwargs.pop("state_override", None)
         call_kwargs = {"block_identifier": block_id, "state_override": state}
-        tx_params = cast(TxParams, txn.dict())
+
+        # Remove unneeded properties
+        data.pop("gas", None)
```
unparalleled-js, line 125: I noticed this while testing... Is this ok? @fubuloubu 
I don't see a point in having `gas` values in calls and it sometimes causes issues when it is

Issue #804, opened 2022-06-14
Allow folders inside the scripts folder


PatrickAlphaC:
### Overview

Provide a simple overview of what you wish to see added. Please include:

* What you are trying to do
* Why Ape's current functionality is inadequate to address your goal

### Specification

I'd like to have a folder inside the scripts folder and be able to run it as a script.

Let's say I have a file like so:

*scripts/deploy/deploy_me.py*

```python
from ape import project
from ape import accounts

def main():
  print("hi")
  contract = accounts[-1].deploy(project.MyContract)
```

I can't run it with:

```
ape run scripts/deploy/deploy_me.py
```

banteg:
`ape run deploy.deploy_me` also sounds like a good convention, similar to how you'd invoke python modules with `python -m scripts.deploy.deploy_me`

fubuloubu:
> `ape run deploy.deploy_me` also sounds like a good convention, similar to how you'd invoke python modules with `python -m scripts.deploy.deploy_me`

Supporting both should be fine

PR #1185, opened 2022-12-12
fix: run scripts from subfolders

sabotagebeats:
### What I did
Attempting to get scripts subfolders to work

fixes: APE-234
fixes: #804 

### Checklist


sabotagebeats:
`ImportError: Error while finding module specification for 'scripts.hello' (ModuleNotFoundError: No module named 'scripts')`

src/ape_run/_cli.py
```diff
@@ -115,14 +115,16 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
...
         commands = {}
-        for filepath in self._project.scripts_folder.glob("*.py"):
-            if filepath.stem.startswith("_"):
-                continue  # Ignore any "private" files
-
-            cmd = self._get_command(filepath)
-            if cmd:  # NOTE: Don't allow calling commands that failed to load
-                commands[filepath.stem] = cmd
-
+        scripts_folders = [x for x in self._project.scripts_folder.iterdir() if x.is_dir()]
```
unparalleled-js, line 118: So we don't need to gather the dirs here. We need to parse the user input and find the script. For example, if the users does:

```
ape run sequencer.middleware.build
```

, we would parse that out and figure out that they are trying to run the script `sequencer/middleware/build.py`
So we don't need to gather all the directories, we need to figure out the to the script they want.

**Also, this allows us to have scripts with the same name across different sub-directories as well as n-level deep directories.**

unparalleled-js: Correction:

```
ape run sequencer middleware build
```

as click helps us a lot; the subdirs can be Groups.

fubuloubu: > Correction:
> 
> ```
> ape run sequencer middleware build
> ```
> 
> as click helps us a lot; the subdirs can be Groups.

Yoooo, this is sick

I love this optimization of turning them into groups

sabotagebeats:
getting this error after merging main: 
```
  File "/home/doge/.pyenv/versions/apeworx/bin/ape", line 5, in <module>
    from ape._cli import cli
  File "/home/doge/ape/ape/src/ape/__init__.py", line 7, in <module>
  File "/home/doge/ape/ape/src/ape/managers/__init__.py", line 3, in <module>
  File "/home/doge/ape/ape/src/ape/plugins/__init__.py", line 10, in <module>
    from .account import AccountPlugin
  File "/home/doge/ape/ape/src/ape/plugins/account.py", line 3, in <module>
    from ape.api.accounts import AccountAPI, AccountContainerAPI
  File "/home/doge/ape/ape/src/ape/api/__init__.py", line 14, in <module>
    from .projects import DependencyAPI, ProjectAPI
  File "/home/doge/ape/ape/src/ape/api/projects.py", line 9, in <module>
    from ethpm_types.utils import AnyUrl, compute_checksum
ImportError: cannot import name 'AnyUrl' from 'ethpm_types.utils' 
```

sabotagebeats:
the scripts module is not being loaded properly because of f string on line 78

src/ape_run/_cli.py
```diff
@@ -66,7 +66,11 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
         elif "main" in code.co_names:
             logger.debug(f"Found 'main' method in script: {relative_filepath}")
 
-            @click.command(cls=NetworkBoundCommand, short_help=f"Run '{relative_filepath}:main'")
+            @click.command(
+                cls=NetworkBoundCommand,
+                short_help=f"Run '{relative_filepath}:main'",
+                name=relative_filepath.stem,
```
unparalleled-js, line 72: - [ ] Make this same change in the else statement below

src/ape_run/_cli.py
```diff
@@ -114,14 +118,33 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
         if not self._project.scripts_folder.exists():
             return {}
 
+        return self._get_cli_commands(self._project.scripts_folder)
+        # TODO: loop through and find subdirectories
```
unparalleled-js, line 122: - [ ] Delete this TODO comment

src/ape_run/_cli.py
```diff
@@ -114,14 +118,33 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
...
-            cmd = self._get_command(filepath)
-            if cmd:  # NOTE: Don't allow calling commands that failed to load
-                commands[filepath.stem] = cmd
+            elif filepath.is_dir():
+                group = click.Group(name=filepath.stem)
+                subcommands = self._get_cli_commands(filepath)
+                for subcommand in subcommands.values():
+                    group.add_command(subcommand)
+                commands[filepath.stem] = group
+                # TODO: commands is a group, subcommands is a networkboundcommand.
```
unparalleled-js, line 141: This is not true, groups are commands (but we already discussed this :) )

- [ ] Delete this TODO comment

src/ape_run/_cli.py
```diff
@@ -114,14 +118,33 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
...
+
+        for filepath in base_path.iterdir():
             if filepath.stem.startswith("_"):
                 continue  # Ignore any "private" files
 
-            cmd = self._get_command(filepath)
-            if cmd:  # NOTE: Don't allow calling commands that failed to load
-                commands[filepath.stem] = cmd
+            elif filepath.is_dir():
+                group = click.Group(name=filepath.stem)
```
unparalleled-js, line 136:
```suggestion
                group = click.Group(name=filepath.stem, short_help=f"Run a script from 'scripts/{filepath.stem}'")
```

unparalleled-js (commented):
I am helping out with the exec issue now!

unparalleled-js:
Also I mentioned elsewhere if you do:

```python
path = str(filepath).replace(os.path.sep, ".").split("scripts.")[-1].replace(filepath.suffix, "")
ns = run_module(f"scripts.{path}")
```

it works..

although that is quick n dirty approach and could maybe be done better!

sabotagebeats:
> ```python
> path = str(filepath).replace(os.path.sep, ".").split("scripts.")[-1].replace(filepath.suffix, "")
> ns = run_module(f"scripts.{path}")
> ```

do I put this in main code.co_names or in cli code.co_names or both?

unparalleled-js:
> > ```python
> > path = str(filepath).replace(os.path.sep, ".").split("scripts.")[-1].replace(filepath.suffix, "")
> > ns = run_module(f"scripts.{path}")
> > ```
> 
> do I put this in main code.co_names or in cli code.co_names or both?

I believe both, because we want to be able to run `main` or `cli` scripts from sub directories and not just one of them.

src/ape_run/_cli.py
```diff
@@ -133,13 +132,13 @@ def _get_cli_commands(self, base_path: Path) -> Dict:
                 continue  # Ignore any "private" files
 
             elif filepath.is_dir():
-                group = click.Group(name=filepath.stem)
+                group = click.Group(
+                    name=filepath.stem, short_help=f"Run a script from 'scripts/{filepath.stem}'"
```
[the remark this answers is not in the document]
sabotagebeats: I think this works with the base level but still doesn't work with my subdirectory scripts.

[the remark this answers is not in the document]
sabotagebeats: I think I got it working by removing that in a couple other places as well.

src/ape_run/_cli.py
```diff
@@ -75,7 +82,13 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
             def call(network):
                 _ = network  # Downstream might use this
                 with use_temp_sys_path(filepath.parent.parent):
-                    ns = run_module(f"scripts.{filepath.stem}")
+                    path = (
+                        str(filepath)
+                        .replace(os.path.sep, ".")
+                        .split("scripts.")[-1]
+                        .replace(filepath.suffix, "")
```
unparalleled-js, line 89: I think this could potentially cause issues if the suffix appears in another part of the path like `scripts.pyfiles.script.py`

unparalleled-js:
```python
                    path = ".".join(
                        (
                            str(filepath)
                            .replace(os.path.sep, ".")
                            .split("scripts.")[-1]
                            .split(".")[:-1]
                        )
                    )
```

src/ape_run/_cli.py
```diff
@@ -66,12 +73,22 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
...
+                    )
+                    ns = run_module(f"{path}")
```
unparalleled-js, line 91:
```suggestion
                    ns = run_module(f"scripts.{path}")
```

src/ape_run/_cli.py
```diff
@@ -53,7 +54,13 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
...
+                    path = (
+                        str(filepath)
+                        .replace(os.path.sep, ".")
+                        .split("scripts.")[-1]
+                        .replace(filepath.suffix, "")
+                    )
+                    ns = run_module(f"scripts.{path}")
                 except Exception as e:
                     logger.error_from_exception(
                         e, f"Exception while parsing script: {relative_filepath}"
```
unparalleled-js, line 66: To fix the middle-dir click scripts, checnage the part a couple lines below from:

```python
return ns["cli"]
```

to 

```python
cli_obj = ns["cli"]
cli_obj.name = filepath.stem
return cli_obj
```

unparalleled-js: perhaps we should only set the name if it not equal to `"cli"` though.

sabotagebeats: ok awesome this got it working

src/ape_run/_cli.py
sabotagebeats: made this into helper method

src/ape_run/_cli.py
```diff
@@ -46,32 +47,46 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
             logger.error_from_exception(e, f"Exception while parsing script: {relative_filepath}")
             return None  # Prevents stalling scripts
 
+        def _get_filepath(filepath):
```
unparalleled-js, line 50: missing `self` here.
Perhaps this is best as a module-level method

unparalleled-js: oops nevermind

unparalleled-js, line 50:
```suggestion
        def _get_module(filepath: Path):
```

sabotagebeats:
passing run test, requires additional testing for subdirectory scripts and addition of script/scripts/subdirectories directory for subdirectories testing

src/ape_run/_cli.py
```diff
@@ -46,32 +47,46 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
             logger.error_from_exception(e, f"Exception while parsing script: {relative_filepath}")
             return None  # Prevents stalling scripts
 
+        def _get_module(filepath: Path):
+            path = ".".join(
```
unparalleled-js, line 51: Might be better to call this like `module_str` or something

src/ape_run/_cli.py
```diff
@@ -46,32 +47,46 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
...
+                (str(filepath).replace(os.path.sep, ".").split("scripts.")[-1].split(".")[:-1])
+            )
+            return path
```
unparalleled-js, line 54: This should include the `scripts.` prefix and then you can remove it everywhere below.

src/ape_run/_cli.py
```diff
@@ -46,32 +47,46 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
...
                 except Exception as e:
                     logger.error_from_exception(
                         e, f"Exception while parsing script: {relative_filepath}"
                     )
                     return None  # Prevents stalling scripts
 
             self._namespace[filepath.stem] = ns
-            return ns["cli"]
+            cli_obj = ns["cli"]
+            cli_obj.name = filepath.stem
```
unparalleled-js, line 73: Let's only change this if it was not already changed by the user normally.
So let's change it if it is the default `cli`:

```suggestion
            cli_obj.name = filepath.stem if cli_obj.name == "cli" else cli_obj.name
```

tests/integration/cli/projects/script/scripts/subdirectory/subdirectory_deploy.py
```diff
@@ -0,0 +1,2 @@
+def main():
```
unparalleled-js, line 1: Include the other kinds of scripts as well, like we did in our demo project.
The `cli` one and the one that does not use anything.

tests/integration/cli/test_run.py
```diff
@@ -31,6 +31,17 @@ def test_run(ape_cli, runner, project):
             assert "Super secret script output" in result.output
 
 
+@skip_projects_except("script")
+def test_run_subdirectories(ape_cli, runner):
+    result = runner.invoke(ape_cli, ["run"])
+    assert result.exit_code == 0, result.output
+    # By default, no commands are run
+    assert "Super secret script output" not in result.output
+    result = runner.invoke(ape_cli, ["run", "subdirectory", "subdirectory_deploy"])
```
unparalleled-js, line 40: After adding the other kinds, we will want to run them here as well.

unparalleled-js:
- [ ] Update the scripts.md userguide

tests/integration/cli/test_run.py
sabotagebeats: I added some more to the subdirectory folder and then am running them with the list comprehension. Does this meet that requirement or do I need to add more subdirectories as well?

src/ape_run/_cli.py
```diff
@@ -46,32 +47,46 @@ def _get_command(self, filepath: Path) -> Union[click.Command, click.Group, None
...
                 except Exception as e:
                     logger.error_from_exception(
                         e, f"Exception while parsing script: {relative_filepath}"
                     )
                     return None  # Prevents stalling scripts
 
             self._namespace[filepath.stem] = ns
-            return ns["cli"]
+            cli_obj = ns["cli"]
+            cli_obj.name = filepath.stem if cli_obj.name == "cli" else cli_obj.name
```
unparalleled-js, line 73:
```suggestion
            cli_obj.name = filepath.stem if cli_obj.name in ("cli", "", None) else cli_obj.name
```

tests/integration/cli/test_run.py
```diff
@@ -31,6 +31,23 @@ def test_run(ape_cli, runner, project):
...
+    # By default, no commands are run
+    assert "Super secret script output" not in result.output
+    subdirectory_scripts = [
+        s
+        for s in (project.scripts_folder / "subdirectory").rglob("*.py")
+        if not s.name.startswith("error")
+    ]
+    for each in subdirectory_scripts:
+        result = runner.invoke(ape_cli, ["run", "subdirectory", each.stem])
+    assert result.exit_code == 0
```
unparalleled-js, line 47: Make these assertions for each script

```suggestion
        assert result.exit_code == 0
```

tests/integration/cli/test_run.py
```diff
@@ -31,6 +31,23 @@ def test_run(ape_cli, runner, project):
...
+    assert "Super secret script output" in result.output
```
unparalleled-js, line 48:
```suggestion
        assert "Super secret script output" in result.output
```

tests/integration/cli/test_run.py
unparalleled-js: This suffices! But make sure to indent the assertions so that we properly test each script:
https://github.com/ApeWorX/ape/pull/1185#discussion_r1072837966
https://github.com/ApeWorX/ape/pull/1185#discussion_r1072838146

src/ape_run/_cli.py
sabotagebeats: I was trying to do this but it seemed to break `Unable to load CLI endpoint for plugin 'ape_run'`

src/ape_run/_cli.py
```diff
@@ -114,14 +134,28 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
...
+                )
+                subcommands = self._get_cli_commands(filepath)
+                for subcommand in subcommands.values():
+                    group.add_command(subcommand)
+                commands[filepath.stem] = group
+
+            if filepath.suffix == ".py":
+                cmd = self._get_command(filepath)
+                if cmd:  # NOTE: Don't allow calling commands that failed to load
+                    commands[filepath.stem] = cmd
```
NotPeopling2day, line 158: Screenshot 2023-01-17 at 7 54 41 PM

Looks like mypy is unhappy in the test because of the types here, you're setting the value in the dictionary as a click group in line 153 but as an individual command in line 158

unparalleled-js: A click group is a subtype of Command I believe, so maybe there is a baseclass we can use

src/ape_run/_cli.py
unparalleled-js: this got resolved by I don't see the changes?

src/ape_run/_cli.py
```diff
@@ -114,14 +134,28 @@ def commands(self) -> Dict[str, Union[click.Command, click.Group]]:
         if not self._project.scripts_folder.exists():
             return {}
 
+        return self._get_cli_commands(self._project.scripts_folder)
+
+    def _get_cli_commands(self, base_path: Path) -> Dict:
         commands = {}
```
unparalleled-js, line 140: I think this would fix the mypy issue.

(may have to import `Command`)

```suggestion
        commands: Dict[str, Command] = {} 
```

src/ape_run/_cli.py
unparalleled-js: That happens because of a syntax error or you typed something wrong.
You can use `-v debug` to see more stack trace.

unparalleled-js (approved):

PR #1189, opened 2022-12-15
feat: add EVM support for Multicall3

fubuloubu:
### What I did

Adds support for [Multicall3](https://github.com/mds1/multicall) on a bunch of different chains directly to `ape-ethereum`

fixes: APE-391
fixes: APE-394

### How I did it
Created `ape_ethereum.multicall` module

### Checklist
based on: ~~#1160~~ #1230


fubuloubu:
- [ ] Needs tests

fubuloubu:
Best test I could find (but only works with Foundry or Hardhat plugins):

```py
def test_multicall(multicall, chain, accounts):
    assert chain.blocks.head.number == 0

    call = multicall.Call()
    call.add(call.contract.aggregate, [])
    call.add(call.contract.aggregate, [])
    call.add(call.contract.aggregate, [])
    call.add(call.contract.aggregate, [])
    (a, _), (b, _), (c, _), (d, _) = call()
    assert a == b == c == d == 0
    assert chain.blocks.head.number == 0

    txn = multicall.Transaction()
    txn.add(txn.contract.aggregate, [])
    txn.add(txn.contract.aggregate, [])
    txn.add(txn.contract.aggregate, [])
    txn.add(txn.contract.aggregate, [])                                                                                                                                               
    txn(sender=accounts[0])                                                                                                                                                           
    assert chain.blocks.head.number == 1                                                                                                                                              
                                                                                                                                                                                      
                                                                                                                                                                                      
def test_multicall_again(multicall, chain):                                                                                                                                           
    # Test isolation                                                                                                                                                                  
    assert chain.blocks.head.number == 0                                                                                                                                              
                                                                                                                                                                                      
    call = multicall.Call()                                                                                                                                                           
    call.add(call.contract.aggregate, [])                                                                                                                                             
    call.add(call.contract.aggregate, [])                                                                                                                                             
    call.add(call.contract.aggregate, [])                                                                                                                                             
    call.add(call.contract.aggregate, [])                                                                                                                                             
    (a, _), (b, _), (c, _), (d, _) = call()                                                                                                                                           
    assert a == b == c == d == 0
```

unparalleled-js:
Can we rebase this one now that the the PRs it depended on are merged in?

fubuloubu:
> Can we rebase this one now that the the PRs it depended on are merged in?

Done!

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+                    requireSuccess if requireSuccess is not None else True,
+                    value or 0,
+                    call.encode_input(*args),
+                )
+            )
+
+
+class Call(BaseMulticall):
+    """
+    Create a sequence of calls to execute at once using `eth_call` via the Multicall3 contract.
```
unparalleled-js, line 121:
```suggestion
    Create a sequence of calls to execute at once using ``eth_call`` via the Multicall3 contract.
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def add(self, call: ContractMethodHandler, *args, **kwargs):
+        if "value" in kwargs:
+            raise InvalidOption("value")
+
+        super().add(call, *args, **kwargs)
+        self.abis.append(_select_method_abi(call.abis, args))
+
+    @property
+    def returnData(self) -> List[HexBytes]:
```
unparalleled-js, line 149: Or is this supposed to be camelCase?

```suggestion
    def return_data(self) -> List[HexBytes]:
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Returns:
+            :class:`~ape.api.transactions.TransactionAPI`
+        """
+        return self.handler.as_transaction(self.calls, **txn_kwargs)  # type: ignore[attr-defined]
+
+
+class Transaction(BaseMulticall):
+    """
+    Create a sequence of calls to execute at once using `eth_sendTransaction`
```
unparalleled-js, line 202:
```suggestion
    Create a sequence of calls to execute at once using ``eth_sendTransaction``
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+              on the current chain at the expected address.
+
+        Args:
+            **txn_kwargs: the kwargs to pass through to the transaction handler.
+
+        Returns:
+            :class:`~ape.api.transactions.ReceiptAPI`
+        """
+        self._validate_calls(**txn_kwargs)
+        return self.handler(self.calls, **txn_kwargs)  # type: ignore [operator]
```
unparalleled-js, line 249:
```suggestion
        return self.handler(self.calls, **txn_kwargs)  # type: ignore[operator]
```

Although, is this type ignore actually needed?

src/ape_ethereum/multicall/exceptions.py
```diff
@@ -0,0 +1,25 @@
+from ape.exceptions import ApeException
+
+
+class MulticallException(ApeException):
```
unparalleled-js, line 4: I wonder if the base exception would make sense at the root exceptions levels so Starknet could also use it.

unparalleled-js (commented):
I am testing now. The PR looks good for Ethereum but multicall is easy in Starknet so I am hoping we can make room for it. Also, there are lots of type-ignores happening, most are related to the call-type differences. I would love a better way, but its not super important.

src/ape_ethereum/multicall/handlers.py
fubuloubu: Yeah this one is camelCase to match the method's return value

src/ape/managers/chain.py
```diff
@@ -950,24 +950,24 @@ def get_deployments(self, contract_container: ContractContainer) -> List[Contrac
 
         return instances
 
-    def lookup_method(
-        self, contract: Union[AddressType, ContractType], selector: HexBytes
-    ) -> MethodABI:
+    def lookup_method(self, address_or_contract_type: Any, selector: HexBytes) -> MethodABI:
```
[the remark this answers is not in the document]
fubuloubu: tracking this in #1217

src/ape/pytest/fixtures.py
```diff
@@ -68,6 +68,19 @@ def project(self) -> ProjectManager:
 
         return self.project_manager
 
+    @pytest.fixture()
+    def multicall(self):
+        """
+        Activate and access the multicall module. Must use a provider that supports `debug_setCode`.
+        """
+        from ape_ethereum import multicall
```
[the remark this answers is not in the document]
fubuloubu: Maybe out of scope for this PR? Come back to it in another one?

src/ape_ethereum/multicall/handlers.py
fubuloubu: Yes

src/ape_ethereum/multicall/exceptions.py
fubuloubu: Added to #1217

fubuloubu: Also add to future refactoring work into EcosystemAPI

fubuloubu:
Note: Can't test this feature because `eth-tester` provider doesn't not support `debug_setCode`

src/ape/pytest/fixtures.py
```diff
@@ -68,6 +68,20 @@ def project(self) -> ProjectManager:
...
 
+    @pytest.fixture
+    def multicall(self):
+        """
+        Activate and access the multicall module.
+        Must use a provider that supports ``debug_setCode``.
+        """
+        from ape_ethereum import multicall
+
+        self.provider.set_code(
```
unparalleled-js, line 79: In a multi-provider / chain context, this fixture could get weird. It uses the provider connected at the time of the start up of the test. However, if you switch mid test, it won't change the chain:

For example:

```python
def test_multichain(multicall, networks):
    # Do stuff with multi-call and top-level provider.
    assert True

    with networks.parse_network_choice("evm-eosystem:network:provider") as provider:
        address = multicall.constants.MULTICALL3_ADDRESS
        assert provider.get_code(address)  # fails because code was not set on this chain.
```

I am thinking multi-call makes the most sense of the ecosystem and can be used in the active provider context continuously as such. This would also address https://github.com/ApeWorX/ape/pull/1189/files#r1068171185

Unless there is another way around this?

fubuloubu: Yeah, I could see that being the case

fubuloubu:
Edit: not a breaking change, no longer requires an un-merged PR

src/ape/pytest/fixtures.py
fubuloubu: somewhat hacky solution, but I basically just made a classmethod to deploy the contract if you needed it (via `debug_setCode`): 715c322

unparalleled-js:
I found a bug while testing this. For some reason, when using Multicall on a call with no arguments, I get the `ArgumentsLengthError()`. The same call works without the multicall. Doing some investigating, I found that receives this as the args:

```
([],)
```

which triggers as 1 arg.

unparalleled-js:
OK, no running into this:

```
eth_abi.exceptions.EncodingTypeError: Value `603786816` of type <class 'int'> cannot be encoded by ByteStringEncoder
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,288 @@
...
+              on the current chain at the expected address.
+
+        Args:
+            **call_kwargs: the kwargs to pass through to the call handler.
+
+        Returns:
+            Iterator[Any]: the sequence of values produced by performing each call stored
+              by this instance.
+        """
+        self._result = self.handler(self.calls, **call_kwargs)  # type: ignore [operator]
```
unparalleled-js, line 213: this space causes issues

```suggestion
        self._result = self.handler(self.calls, **call_kwargs)  # type: ignore[operator]
```

but would love to mypy working better in this module, there are many type ignores.

fubuloubu: yeah I agree

fubuloubu:
@unparalleled-js :eyes: [2bc4c1c](https://github.com/ApeWorX/ape/pull/1189/commits/2bc4c1cc16c314410d22b2fd5885d8ebf45e2064)

fubuloubu:
> OK, now running into this:
> 
> ```
> eth_abi.exceptions.EncodingTypeError: Value `603786816` of type <class 'int'> cannot be encoded by ByteStringEncoder
> ```

This is the integer encoding of the 4bytes calldata (just the signature, no additional args)

Investigating...

src/ape/api/providers.py
```diff
@@ -884,6 +884,9 @@ def _eth_call(self, arguments: List) -> bytes:
         except Exception as err:
             raise self.get_virtual_machine_error(err) from err
 
+        if "error" in result:
```
unparalleled-js, line 887: I think this already happens in `_make_request()`

src/ape/pytest/fixtures.py
unparalleled-js: Ya that works for now I suppose, but I may try it out to see what it feels like

src/ape_ethereum/multicall/handlers.py
fubuloubu: I _think_ #1230 might actually make this a lot better, since I can just use a dict with different members, and it will only `getattr` ones that matter when making the appropiate calls

src/ape/api/providers.py
fubuloubu: It has some sort of bug, because I was reaching the point where it tried to convert an error message dict `result` from RPC to `HexBytes` below

src/ape/pytest/fixtures.py
fubuloubu: For sure, can't really test it w/o `debug_setCode` in foundry/hardhat

fubuloubu:
[64c7424](https://github.com/ApeWorX/ape/pull/1189/commits/64c7424d894276d8e22062175faf5135e65dac8b) doesn't resolve all the mypy issues, but some of them (and it's cleaner to understand)

src/ape/api/providers.py
unparalleled-js: hmm, weird, if you set a breakpoint, does it hit this part? https://github.com/ApeWorX/ape/blob/main/src/ape/api/providers.py#L1114-L1116

fubuloubu:
Dropped the gnosis safe call refactor because it doesn't support `eth_getStorageAt`

fubuloubu:
> > OK, now running into this:
> > ```
> > eth_abi.exceptions.EncodingTypeError: Value `603786816` of type <class 'int'> cannot be encoded by ByteStringEncoder
> > ```
> 
> This is the integer encoding of the 4bytes calldata (just the signature, no additional args)
> 
> Investigating...

this happens because `self.conversion_manager.convert(v, tuple)` is converting the `HexBytes` arg into an integer from within `ContractCallHandler.__call__`, when it hits this line:

https://github.com/ApeWorX/ape/blob/e0f7b57de50b0dd155e071e2ccffd454f28f661a/src/ape/contracts/base.py#L188

inside of the `multicall.Call.__call__` method being triggered

fubuloubu:
> > > OK, now running into this:
> > > ```
> > > eth_abi.exceptions.EncodingTypeError: Value `603786816` of type <class 'int'> cannot be encoded by ByteStringEncoder
> > > ```
> > 
> > 
> > This is the integer encoding of the 4bytes calldata (just the signature, no additional args)
> > Investigating...
> 
> this happens because `self.conversion_manager.convert(v, tuple)` is converting the `HexBytes` arg into an integer from within `ContractCallHandler.__call__`, when it hits this line:
> 
> https://github.com/ApeWorX/ape/blob/e0f7b57de50b0dd155e071e2ccffd454f28f661a/src/ape/contracts/base.py#L188
> 
> inside of the `multicall.Call.__call__` method being triggered
> 
> EDIT: I believe #1247 added this bug when the `HexIntConverter` was created

This bug is the only blocker to this PR being ready for QA and merge

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,290 @@
...
+
+class BaseMulticall(ManagerAccessMixin):
+    def __init__(self) -> None:
+        """
+        Initialize a new Multicall session object. By default, there are no calls to make.
+        """
+        self.calls: List[dict] = []
+
+    @classmethod
+    def deploy(cls):
```
fubuloubu, line 33: Thinking of renaming this to `inject`

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,290 @@
...
+
+        else:
+            return self.contract.aggregate
+
+    def add(
+        self,
+        call,
+        *args,
+        allowFailure=None,
+        value=None,
```
fubuloubu, line 92:
```suggestion
        allowFailure: bool=False,
        value: int=0,
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+
+        # Append call dict to the list
+        # NOTE: Depending upon `_handler_method_abi` at time when `__call__` is triggered,
+        #       some of these properties will be unused
+        self.calls.append(
+            {
+                "target": call.contract.address,
+                "allowFailure": allowFailure if allowFailure is not None else False,
+                "value": value or 0,
```
fubuloubu, line 114:
```suggestion
                "allowFailure": allowFailure,
                "value": self.conversion_manager.convert(value, int),
```

fubuloubu (commented):
Some more simplications

unparalleled-js:
> EDIT: I believe #1247 added this bug when the `HexIntConverter` was created

I fixed the issues on this PR: https://github.com/ApeWorX/ape/pull/1230 (last few commits), as it was affecting that as well!
So after we take in those changes, we should be good with this part.

fubuloubu:
> > EDIT: I believe #1247 added this bug when the `HexIntConverter` was created
> 
> I fixed the issues on this PR: #1230 (last few commits), as it was affecting that as well! So after we take in those changes, we should be good with this part.

This looks good!

fubuloubu:
#1230 not working entirely like I expected:

```
EncodingTypeError: Value `{'target': '0xcA11bde05977b3631167028862bE2a173976CA11', 'allowFailure': Fal...` of type <class 'dict'> cannot be encoded by TupleEncoder: must be list-like object such as array or tuple
```

unparalleled-js:
> #1230 not working entirely like I expected:

Give it a spin now! The problem was happening only for list-of-struct inputs.

fubuloubu:
Sick it works w/ #1230 

```py
In [1]: from ape_ethereum import multicall
In [2]: call = multicall.Call()
In [3]: call.add(call.contract.getLastBlockHash)
In [4]: call.add(call.contract.aggregate, [])
In [5]: call.add(call.contract.getBasefee)
In [7]: list(call())
Out[7]: 
[HexBytes('0x70cd88dbc792de30c4df33934feeaa8e0a203d9a5eee7eed02a2edb06fc0fac4'),
 aggregate_return(blockNumber=16522283, returnData=[]),
 21543193782]
```

src/ape_ethereum/multicall/handlers.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        Adds a call to the Multicall session object.
+
+        Raises:
+            :class:`InvalidOption`: If one of the kwarg modifiers is not able to be used.
+
+        Args:
+            call: :class:`ContractMethodHandler` The method to call.
+            *args: The arguments to invoke the method with.
+            requireSuccess: bool Whether the call must be successful.
```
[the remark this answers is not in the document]
fubuloubu: [d00958c](https://github.com/ApeWorX/ape/pull/1189/commits/d00958cfa15fb76e2d9ea515a1b6e570219c2f6d)

src/ape_ethereum/multicall/handlers.py
fubuloubu: [d00958c](https://github.com/ApeWorX/ape/pull/1189/commits/d00958cfa15fb76e2d9ea515a1b6e570219c2f6d)

src/ape_ethereum/multicall/handlers.py
fubuloubu: [d00958c](https://github.com/ApeWorX/ape/pull/1189/commits/d00958cfa15fb76e2d9ea515a1b6e570219c2f6d)

src/ape_ethereum/multicall/handlers.py
fubuloubu: [d00958c](https://github.com/ApeWorX/ape/pull/1189/commits/d00958cfa15fb76e2d9ea515a1b6e570219c2f6d)

src/ape/api/providers.py
fubuloubu: Should probably figure out if this is a good thing to merge?

unparalleled-js (approved):

PR #1192, opened 2022-12-15
fix: issue where `poll_blocks()` did not detect chain re-org

unparalleled-js:
### What I did

Fixes an issue where a re-org was never detected.
Additionally, when a chain reorganization occurs, `poll_blocks()` yields the missed blocks even if they were previously yielded with different block numbers.
There are also spelling fixes and other small fixes in the `chain.py` module.

### How I did it

Noticed reorg detection wasn't working.
Realized `chain.py` had several warnings (thanks Pycharm)
Fixed everything and added a test.

### How to verify it

Poll blocks on mainnet; you should get a sequential amount of blocks (nothing changed).
Poll blocks with unreasonable small required confs to test reorg works.

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -233,6 +233,11 @@ class BlockContainer(BaseManager):
     ) -> Iterator[BlockAPI]:
         """
         Poll new blocks. Optionally set a start block to include historical blocks.
+
+        **NOTE**: When a chain reorganization occurs, this method logs an error and
+        yields the missed blocks even if they were previously yielded with different
+        block numbers.
+
         **NOTE**: This is a daemon method; it does not terminate unless an exception occurs
         or a ``stop`` is given.
 
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -917,7 +917,7 @@ class ContractCache(BaseManager):
         Find the method ABI from the given contract address and selector.
 
         Args:
-            contract_address (``AddressType``): The address of the contract.
+            contract (``AddressType``): The address of the contract.
             selector (``HexBytes``): The method ID.
 
         Returns:
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -282,8 +282,6 @@ class BlockContainer(BaseManager):
             raise ValueError("'stop' argument must be in the future.")
 
         # Get number of last block with the necessary amount of confirmations.
-        last_yielded_hash = None
-        last_yielded_number = None
         block = None
 
         if start_block is not None:
@@ -294,9 +292,17 @@ class BlockContainer(BaseManager):
         if block:
             last_yielded_hash = block.hash
             last_yielded_number = block.number
-            time_since_last = time.time()
 
-        time.sleep(block_time)
+            # Only sleep if pre-yielded a block
+            time.sleep(block_time)
+
+        else:
+            last_yielded_hash = None
+            last_yielded_number = None
+
+        # Set `time_since_last` even if haven't yield yet.
+        # This helps with timing out the first time.
+        time_since_last = time.time()
 
         def _try_timeout():
             if time.time() - time_since_last > timeout:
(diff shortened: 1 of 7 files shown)


src/ape/managers/chain.py
```diff
@@ -276,13 +277,16 @@ def poll_blocks(
...
         if start_block is not None:
-            # Front-load historically confirmed blocks.
-            yield from self.range(start_block, latest_confirmed_block_number + 1)
-            has_yielded_before_reorg = True
+            # Front-load historical blocks.
+            for block in self.range(start_block, self.height - required_confirmations + 1):
+                last_yielded_hash = block.hash
+                last_yielded_number = block.number
+                time_since_last = time.time()
+                yield block
```
fubuloubu, line 294: Maybe a little optimization trick here:

```suggestion
                yield block

            # NOTE: using last `block` from historical iterator
            last_yielded_hash = block.hash
            last_yielded_number = block.number
            time_since_last = time.time()
```

In python, `block` exists past the scope of the iterator here, and you only need the last one, so leveraging that will reduce a bunch of small computation if you're doing a large number of blocks

src/ape/managers/chain.py
```diff
@@ -276,13 +282,16 @@ def poll_blocks(
...
-        latest_confirmed_block_number = self.height - required_confirmations
-        has_yielded_before_reorg = False
+        last_yielded_hash = None
+        last_yielded_number = None
 
         if start_block is not None:
-            # Front-load historically confirmed blocks.
-            yield from self.range(start_block, latest_confirmed_block_number + 1)
-            has_yielded_before_reorg = True
+            # Front-load historical blocks.
```
fubuloubu, line 289: Or another option
```suggestion
            last_yielded_number = self.height - required_confirmations
            yield from self.range(start_block, last_yielded_number + 1)
            block = self[last_yielded_number + 1]
            last_yielded_hash = block.hash
            last_yielded_number = block.number
            time_since_last = time.time()
            yield block
```

unparalleled-js: I think there could be a race condition in this one.
I like the idea of using `block` from the outer scope because than it for sure is the block that got yielded

src/ape/managers/chain.py
```diff
@@ -284,17 +284,19 @@ def poll_blocks(
         # Get number of last block with the necessary amount of confirmations.
         last_yielded_hash = None
         last_yielded_number = None
+        block = None
```
unparalleled-js, line 287: have to start as `None` in case the range is empty

src/ape/managers/chain.py
```diff
@@ -276,16 +282,21 @@ def poll_blocks(
             raise ValueError("'stop' argument must be in the future.")
 
         # Get number of last block with the necessary amount of confirmations.
-        latest_confirmed_block_number = self.height - required_confirmations
-        has_yielded_before_reorg = False
+        last_yielded_hash = None
+        last_yielded_number = None
```
fubuloubu, line 286: Might be better to move these as the `else` case in `if block` below

src/ape/managers/chain.py
```diff
@@ -276,16 +282,21 @@ def poll_blocks(
...
-            yield from self.range(start_block, latest_confirmed_block_number + 1)
-            has_yielded_before_reorg = True
+            # Front-load historical blocks.
+            for block in self.range(start_block, self.height - required_confirmations + 1):
+                yield block
+
+        if block:
+            last_yielded_hash = block.hash
+            last_yielded_number = block.number
+            time_since_last = time.time()
```
fubuloubu, line 298:
```suggestion
else:
    last_yielded_hash = None
    last_yielded_number = None

```

Might also need `time_since_last` here?

fubuloubu (dismissed):

src/ape/managers/chain.py
```diff
@@ -232,7 +233,12 @@ def poll_blocks(
     ) -> Iterator[BlockAPI]:
         """
         Poll new blocks. Optionally set a start block to include historical blocks.
-        **NOTE**: This is a daemon method; it does not terminate unless an exception occurrs
+
+        **NOTE**: When a chain reorganization occurs, this method logs an error and
+        yields the missed blocks even if they were previously yielded with different
+        block numbers.
```
NotPeopling2day, line 239: nitpick:
```suggestion
        **NOTE**: When a chain reorganization occurs, this method logs an error and
        yields the missed blocks, even if they were previously yielded with different
        block numbers.
```

NotPeopling2day (dismissed):

PR #1196, opened 2022-12-19
fix: use cached path

unparalleled-js:
### What I did

Another failed test in `ape-solidity` caught this issue where the path was being mistakenly reset, causing issues with calculated contracts folder

### Checklist

### DIFF
--- a/src/ape/managers/project/manager.py
+++ b/src/ape/managers/project/manager.py
@@ -246,13 +246,11 @@ class ProjectManager(BaseManager):
                 self.path, self.contracts_folder
             )
 
-        project = self._cached_projects[self.path.name]
-        project.path = self.path
-        return project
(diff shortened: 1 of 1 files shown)


src/ape/managers/project/manager.py
```diff
@@ -246,13 +246,11 @@ def _project(self) -> ProjectAPI:
                 self.path, self.contracts_folder
             )
 
-        project = self._cached_projects[self.path.name]
-        project.path = self.path
```
fubuloubu, line 250: Is this still getting set somewhere else? Is this an unnecessary update?

unparalleled-js: It gets set the first, I don't know why it was getting overridden here, but it causes issues.

NotPeopling2day (approved):

PR #1199, opened 2022-12-20
feat: add support to create test accounts runtime

bilbeyt:
### How I did it

Changed TestAccountContainer and TestAccountManager to implement.

### How to verify it

Run tests.

### Checklist

### DIFF
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -41,6 +41,14 @@ test:
 **WARN**: NEVER put a seed phrase with real funds here.
 The accounts generated from this seed are solely for testing and debugging purposes.
 
+You can create a new test account by doing the following:
+
+```python
+account = accounts.test_accounts.generate_test_account()
+```
+
+**NOTE**: Creating a new test account means it will be unfunded by default.
+
 Learn more about test accounts from the [testing guide](./testing.html#accounts-fixture).
 
 If your testing provider supports this feature, it is possible to directly set the balances of any address by performing the following action:
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -432,6 +432,12 @@ class TestAccountContainerAPI(AccountContainerAPI):
     ``AccountContainerAPI`` directly. This is how they show up in the ``accounts`` test fixture.
     """
 
+    @abstractmethod
+    def generate_account(self) -> "TestAccountAPI":
+        """
+        Generate a new test account
+        """
+
 
 class TestAccountAPI(AccountAPI):
     """
--- a/src/ape/managers/accounts.py
+++ b/src/ape/managers/accounts.py
@@ -20,7 +20,7 @@ class TestAccountManager(list, ManagerAccessMixin):
         accounts_str = ", ".join([a.address for a in self.accounts])
         return f"[{accounts_str}]"
 
-    @property
+    @cached_property
     def containers(self) -> Dict[str, TestAccountContainerAPI]:
         containers = {}
         account_types = [
@@ -96,6 +96,9 @@ class TestAccountManager(list, ManagerAccessMixin):
 
     def __contains__(self, address: AddressType) -> bool:  # type: ignore
         return any(address in container for container in self.containers.values())
+
+    def generate_test_account(self, container_name: str = "test") -> TestAccountAPI:
+        return self.containers[container_name].generate_account()
 
 
 class AccountManager(BaseManager):
--- a/src/ape_test/accounts.py
+++ b/src/ape_test/accounts.py
@@ -6,15 +6,17 @@ from eth_utils import to_bytes
 
 from ape.api import TestAccountAPI, TestAccountContainerAPI, TransactionAPI
 from ape.types import AddressType, MessageSignature, TransactionSignature
-from ape.utils import GeneratedDevAccount, cached_property, generate_dev_accounts
+from ape.utils import GeneratedDevAccount, generate_dev_accounts
 
 
 class TestAccountContainer(TestAccountContainerAPI):
+    _num_generated = 0
+
     @property
     def config(self):
         return self.config_manager.get_config("test")
 
-    @cached_property
+    @property
     def _dev_accounts(self) -> List[GeneratedDevAccount]:
         mnemonic = self.config["mnemonic"]
         return generate_dev_accounts(mnemonic, number_of_accounts=self.config["number_of_accounts"])
@@ -36,6 +38,19 @@ class TestAccountContainer(TestAccountContainerAPI):
 
     def __len__(self) -> int:
         return len(self._dev_accounts)
+
+    def generate_account(self) -> "TestAccountAPI":
+        new_index = len(self) + self._num_generated
+        self._num_generated += 1
+        generated_account = generate_dev_accounts(
+            self.config["mnemonic"], 1, start_index=new_index
+        )[0]
+        acc = TestAccount(
+            index=new_index,
+            address_str=generated_account.address,
+            private_key=generated_account.private_key,
+        )
+        return acc
 
 
 class TestAccount(TestAccountAPI):
--- a/tests/functional/test_accounts.py
+++ b/tests/functional/test_accounts.py
@@ -8,6 +8,7 @@ from ape.exceptions import AccountsError, NetworkError, ProjectError, SignatureE
 from ape.types.signatures import recover_signer
 from ape.utils.testing import DEFAULT_NUMBER_OF_TEST_ACCOUNTS
 from ape_ethereum.ecosystem import ProxyType
+from ape_test.accounts import TestAccount
 
 MISSING_VALUE_TRANSFER_ERR_MSG = "Must provide 'VALUE' or use 'send_everything=True"
 
@@ -362,6 +363,19 @@ def test_test_accounts_repr(test_accounts):
 def test_account_comparison_to_non_account(core_account):
     # Before, would get a ConversionError.
     assert core_account != "foo"
+
+
+def test_create_account(test_accounts):
+    created_acc = test_accounts.generate_test_account()
+
+    assert isinstance(created_acc, TestAccount)
+    assert created_acc.index == len(test_accounts)
+
+    second_created_acc = test_accounts.generate_test_account()
+
+    assert created_acc.address != second_created_acc.address
+
+    assert second_created_acc.index == created_acc.index + 1
 
 
 def test_dir(core_account):
--- a/src/ape/utils/testing.py
+++ b/src/ape/utils/testing.py
@@ -25,7 +25,8 @@ Config example::
 def generate_dev_accounts(
     mnemonic: str = DEFAULT_TEST_MNEMONIC,
     number_of_accounts: int = DEFAULT_NUMBER_OF_TEST_ACCOUNTS,
-    hd_path_format="m/44'/60'/0'/{}",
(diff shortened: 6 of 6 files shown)


docs/userguides/accounts.md
```diff
@@ -41,6 +41,12 @@ test:
 **WARN**: NEVER put a seed phrase with real funds here.
 The accounts generated from this seed are solely for testing and debugging purposes.
 
+You can create random test account by:
```
fubuloubu, line 44:
```suggestion
You can create a new test account by doing the following:
```

docs/userguides/accounts.md
```diff
@@ -41,6 +41,12 @@ test:
...
+
+```python
+accounts.test_accounts.containers['test'].create_account()
+```
+
```
fubuloubu, line 49:
```suggestion

**NOTE**: Creating a new test account means it will be unfunded by default.
```

src/ape/managers/accounts.py
```diff
@@ -20,17 +21,39 @@ def __repr__(self) -> str:
...
         ]
         for plugin_name, (container_type, account_type) in account_types:
             # Pydantic validation won't allow passing None for data_folder/required attr
             containers[plugin_name] = container_type(data_folder="", account_type=account_type)
+            configs[plugin_name] = self.config_manager.get_config(plugin_name)
+        self._containers = containers
+        self._configs = configs
 
-        return containers
+    def _get_configs(self) -> Dict[str, PluginConfig]:
```
fubuloubu, line 40: Why is this method needed?

tests/functional/test_accounts.py
```diff
@@ -364,6 +364,16 @@ def test_account_comparison_to_non_account(core_account):
     assert core_account != "foo"
 
 
+def test_add_new_account(test_accounts):
```
unparalleled-js, line 367:
```suggestion
def test_create_account(test_accounts):
```

src/ape_test/accounts.py
```diff
@@ -6,36 +6,51 @@
 
 from ape.api import TestAccountAPI, TestAccountContainerAPI, TransactionAPI
 from ape.types import AddressType, MessageSignature, TransactionSignature
-from ape.utils import GeneratedDevAccount, cached_property, generate_dev_accounts
+from ape.utils import generate_dev_accounts
 
 
 class TestAccountContainer(TestAccountContainerAPI):
+    _accounts: List
```
unparalleled-js, line 13:
```suggestion
    _accounts: List["TestAccount"]
```

src/ape/managers/accounts.py
bilbeyt: If we dont have this method to check if config is changed or not then one of the tests were failing. Basically there was something like:
`
with temp_config(config):
`
and it was checking for number of test accounts

bilbeyt: test_custom_num_of_test_accounts_config this test.

fubuloubu: Hmm, I wonder if we can take an alternative approach here and just not add the new account to `test_accounts`? Basically, those accounts are only the pre-configured genesis accounts, and if you ad-hoc create a new one it is not apart of that set. I think this is how it works if you fetch an account via address e.g. `acct = accounts["myname.eth"]`

bilbeyt: those accounts are coming from the same approach using generated_dev_accounts and creating TestAccount which has capabilities such as signing messages or transactions. the name test_accounts is also giving people the hint hey dont use these with production accounts. In my opinion, the last approach was just limiting the capabilities of the devs who are using them in tests, so it is perfectly fine for me to use it like this. actually to me, it is better than using context manager approach written in specified test.

tests/functional/test_accounts.py
```diff
@@ -364,6 +364,16 @@ def test_account_comparison_to_non_account(core_account):
     assert core_account != "foo"
 
 
+def test_create_account(test_accounts):
+    assert len(test_accounts) == DEFAULT_NUMBER_OF_TEST_ACCOUNTS
+
+    created_acc = test_accounts.containers["test"].create_account()
+
+    assert len(test_accounts) == DEFAULT_NUMBER_OF_TEST_ACCOUNTS + 1
```
unparalleled-js, line 372: Do we need to add this to the account list?
I am thinking it should work more like how impersonated accounts work when using `ape-hardhat` or `ape-foundry`.

fubuloubu: Yeah, this is my opinion too. I don't think it should be added to the list, instead there are "genesis accounts" in `TestAccountContainer` and impersonated accounts that come from `TestAccountContainer.create_account()` or `TestAccountContainer.__getitem__(addr)`

src/ape/managers/accounts.py
```diff
@@ -15,6 +15,7 @@
 
 class TestAccountManager(list, ManagerAccessMixin):
     __test__ = False
+    num_generated = 0
```
istankovic, line 18: This should probably be private.

src/ape_test/accounts.py
```diff
@@ -37,6 +37,17 @@ def accounts(self) -> Iterator["TestAccount"]:
...
+        generated_account = generate_dev_accounts(
+            self.config["mnemonic"], 1, start_index=len(self) + generation_index
+        )[0]
+        acc = TestAccount(
+            index=len(self) + generation_index,
+            address_str=generated_account.address,
+            private_key=generated_account.private_key,
+        )
+        return acc
+
```
istankovic, line 50: Now I'm wondering whether this even needs to be on the container -- this methods just uses `self.config` and `len(self)`, which I assume can be accessed externally. So it would make sense to also move this method to the manager, no? Essentially, this just calls `generate_dev_accounts`.

src/ape/managers/accounts.py
bilbeyt: done

src/ape_test/accounts.py
bilbeyt: changed this

src/ape/managers/accounts.py
```diff
@@ -97,6 +104,20 @@ def __getitem_str(self, account_str: str):
     def __contains__(self, address: AddressType) -> bool:  # type: ignore
         return any(address in container for container in self.containers.values())
 
+    def generate_test_account(self) -> TestAccount:
```
unparalleled-js, line 107: Here are my thoughts here:

1. Make the return type a `TestAccountAPI` (more generic)
2. Add abstract method `generate_account` to the `TestAccountContainerAPI` class
3. Move bulk of this logic to the `TestAccountContainer` class in `ape_test`
4. Change this implementation to:

```python
container = self.containers["test"]
return container.generate_account()
```

It is not perfect and does not 100% mitigate the concerns I have here, but it is at least closer in the right direction.

unparalleled-js (approved):
Approving because I think this is good enough for now, but please consider my refactor comment!

src/ape/managers/accounts.py
fubuloubu: Could even make this a bit more flexible via 
```suggestion
    def generate_test_account(self, container_name: str = "test") -> TestAccountAPI:
        return self.containers[container_name].generate_account()
```

with the above-mentioned changes to `TestAccountContainerAPI`

bilbeyt:
@fubuloubu @unparalleled-js i have changed my implementation according to your suggestions. you can check it.

docs/userguides/accounts.md
```diff
@@ -41,6 +41,14 @@ test:
 **WARN**: NEVER put a seed phrase with real funds here.
 The accounts generated from this seed are solely for testing and debugging purposes.
 
+You can create a new test account by doing the following:
+
+```python
+accounnt = accounts.test_accounts.generate_test_account()
```
fubuloubu, line 47:
```suggestion
account = accounts.test_accounts.generate_test_account()
```

src/ape/api/accounts.py
```diff
@@ -421,6 +421,12 @@ class TestAccountContainerAPI(AccountContainerAPI):
     ``AccountContainerAPI`` directly. This is how they show up in the ``accounts`` test fixture.
     """
 
+    @abstractmethod
+    def generate_account(self, generation_index: int) -> "TestAccountAPI":
```
fubuloubu, line 425:
```suggestion
    def generate_account(self) -> "TestAccountAPI":
```

src/ape/managers/accounts.py
```diff
@@ -15,6 +15,7 @@
 
 class TestAccountManager(list, ManagerAccessMixin):
     __test__ = False
+    _num_generated = 0
```
fubuloubu, line 18: Can this be moved to `TestAccountContainer`?

src/ape/managers/accounts.py
```diff
@@ -97,6 +98,12 @@ def __getitem_str(self, account_str: str):
     def __contains__(self, address: AddressType) -> bool:  # type: ignore
         return any(address in container for container in self.containers.values())
 
+    def generate_test_account(self, container_name: str = "test") -> TestAccountAPI:
+        container = self.containers[container_name]
+        acc = container.generate_account(self._num_generated)
+        self._num_generated += 1
+        return acc
```
fubuloubu, line 105:
```suggestion
        return self.containers[container_name].generate_account()
```

src/ape_test/accounts.py
```diff
@@ -37,6 +37,18 @@ def accounts(self) -> Iterator["TestAccount"]:
     def __len__(self) -> int:
         return len(self._dev_accounts)
 
+    def generate_account(self, generation_index: int) -> "TestAccountAPI":
+        new_index = len(self) + generation_index
```
fubuloubu, line 41:
```suggestion
    def generate_account(self) -> "TestAccount":
        new_index = len(self) + self._generation_index
        self._generation_index += 1
```

fubuloubu (approved):
As a suggestion, would this work? Removes the implemention-specific use of `generation_index`

src/ape/managers/accounts.py
bilbeyt: @fubuloubu this is the thing i mentioned in discord also. if i move this to TestAccountContainer, then this num_generated is not changing because containers is a property that is reinitializing all containers everytime you hit and meaning num_generated will be 0 all the time, so in tests, if i want to have 3 different accounts i cant have it.

fubuloubu: Good catch. As mentioned in discord, I believe this can be resolved by using `cached_property` for `.containers` property in `TestAccountManager`. I believe this to be an oversight from a previous feature implementation that is causing the issue you've encountered

src/ape_test/accounts.py
```diff
@@ -6,15 +6,17 @@
...
 
 class TestAccountContainer(TestAccountContainerAPI):
+    _num_generated = 0
+
     @property
     def config(self):
         return self.config_manager.get_config("test")
 
-    @cached_property
+    @property
```
fubuloubu, line 19: This could probably stay cached, since config shouldn't be changing

fubuloubu (approved):

src/ape_test/accounts.py
fubuloubu: Okay, following up from discord, this causes a failure when it's cached due to how we are testing ape

Oppurtunity for future optimization

NotPeopling2day (approved):

unparalleled-js (approved):

Issue #1198, opened 2022-12-19
Compiling an Ape project with a Brownie Project dependency with its own other dependencies doesn't work


fubuloubu:
### Environment information

- OS: linux
- Python Version: 3.10.7
- `ape` and plugin versions:

```sh
$ ape --version
0.5.8

$ ape plugins list
Installed Plugins:
  vyper        0.5.2
  hardhat      0.5.4
  tokens       0.5.0
  etherscan    0.5.4
  solidity     0.5.3
  infura       0.5.3
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```yaml
name: Test

plugins:
  - name: solidity
  - name: vyper
  - name: hardhat
  - name: infura
  - name: tokens
  - name: etherscan

default_ecosystem: ethereum

dependencies:
  # - name: openzeppelin
  #   github: OpenZeppelin/openzeppelin-contracts
  #   version: 4.7.1
  - name: vault
    github: yearn/yearn-vaults
    version: 0.4.5

# solidity:
#   import_remapping:
#   - "@openzeppelin/contracts=openzeppelin"
#   - "@yearnvaults/contracts=vault/0.4.5

hardhat:
  port: auto
  fork:
    ethereum:
      mainnet:
        upstream_provider: alchemy

ethereum:
  default_network: local
  local:
    default_provider: hardhatape
```

### What went wrong?

Attempted to compile a Brownie project (https://yearn/yearn-vaults) that defined additional dependencies:

```sh
$ ape compile
INFO: Compiling 'CommonHealthCheck.sol'.                     
INFO: Compiling 'test/TestStrategy.sol'.
INFO: Compiling 'BaseFeeOracle.sol'.
INFO: Compiling 'test/Token.sol'.
INFO: Compiling 'BaseStrategy.sol'.
Traceback (most recent call last):
  File "~/.pyenv/versions/test/bin/ape", line 8, in <module>
    sys.exit(cli())
  File "~/.pyenv/versions/3.10.7/envs/test/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "~/.pyenv/versions/3.10.7/envs/test/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "~/.pyenv/versions/3.10.7/envs/test/lib/python3.10/site-packages/ape/_cli.py", line 36, in invoke
[...]

PR #1200, opened 2022-12-20
fix: dependencies of dependencies 

unparalleled-js:
### What I did

fixes: #1198 

Goes with https://github.com/ApeWorX/ape-solidity/pull/86
Requires https://github.com/ApeWorX/ethpm-types/pull/56

### Checklist


src/ape/api/projects.py
```diff
@@ -190,6 +190,13 @@ def version_id(self) -> str:
         Most often, this is either a version number or a branch name.
         """
 
+    @property
+    @abstractmethod
+    def url(self) -> AnyUrl:
```
fubuloubu, line 195: nit, but shouldn't this be `uri`?

src/ape/managers/project/dependency.py
```diff
@@ -77,6 +77,16 @@ def version_id(self) -> str:
         latest_release = github_client.get_release(self.github, "latest")
         return latest_release.tag_name
 
+    @property
+    def url(self) -> AnyUrl:
```
fubuloubu, line 81: should be `uri`

src/ape/api/projects.py
unparalleled-js: The type is `AnyUrl`, not `AnyUri`! JK, I am actually sure on the technically correct answer. Is there a difference asking for a friend?

I am happy to switch to `uri` though!

src/ape/api/projects.py
```diff
@@ -422,4 +422,5 @@ def _load_manifest_from_file(file_path: Path) -> Optional[PackageManifest]:
     except ValidationError as err:
         logger.warning(f"Existing manifest file '{file_path}' corrupted. Re-building.")
         logger.debug(str(err))
+        breakpoint()
```
fubuloubu:
```suggestion
```

fubuloubu (approved):
Seems like this impacts some of the testing workflow. May need some additional QA on usage within projects with tests in them to make sure everything still works.

unparalleled-js:
> Seems like this impacts some of the testing workflow. May need some additional QA on usage within projects with tests in them to make sure everything still works.

I tested against like 4 of Yearn repos and I suggest others do the same.

fubuloubu:
> > Seems like this impacts some of the testing workflow. May need some additional QA on usage within projects with tests in them to make sure everything still works.
> 
> I tested against like 4 of Yearn's repos and I suggest others do the same.

Ah, this is good then. Would appreciate one other person doing QA, then I think this can merge

NotPeopling2day (approved):
Looks good on my end

PR #1201, opened 2022-12-21
refactor!: modify `AccountAPI.sign_transaction` to return `TransactionAPI`

fubuloubu:
BREAKING CHANGE: This is done to ensure account plugins have the ability to modify the transaction they are signing, i.e. for Gnosis Safe plugin.

### What I did

fixes: APE-388

### How I did it

When working on `ape-safe`, it became apparent that there needs to be a way for an account plugin that wraps some more complex processes with other accounts to be able to mutate the transaction to a different type.

This is also handy if the account plugin wants to make adjustments to things like gas, etc. of the transaction it is giving over to the signing process.

### How to verify it

All tests pass, check with other account plugins (Trezor, Ledger)

### Checklist

- [x] All 2nd party plugins updated

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -9,8 +9,7 @@ from ape.api.address import BaseAddress
 from ape.api.transactions import ReceiptAPI, TransactionAPI
 from ape.exceptions import AccountsError, AliasAlreadyInUseError, SignatureError, TransactionError
 from ape.logging import logger
-from ape.types import AddressType, MessageSignature, SignableMessage, TransactionSignature
-from ape.types.signatures import _Signature
+from ape.types import AddressType, MessageSignature, SignableMessage
 from ape.utils import BaseInterfaceModel, abstractmethod
 
 if TYPE_CHECKING:
@@ -64,7 +63,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         """
 
     @abstractmethod
-    def sign_transaction(self, txn: TransactionAPI) -> Optional[TransactionSignature]:
+    def sign_transaction(self, txn: TransactionAPI) -> Optional[TransactionAPI]:
         """
         Sign a transaction.
 
@@ -72,7 +71,11 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
           txn (:class:`~ape.api.transactions.TransactionAPI`): The transaction to sign.
 
         Returns:
-          :class:`~ape.types.signatures.TransactionSignature` (optional): The signed transaction.
+          :class:`~ape.api.transactions.TransactionAPI` (optional): A signed transaction.
+            The `TransactionAPI` returned by this method may not correspond to `txn` given as input,
+            however returning a properly-formatted transaction here is meant to be executed.
+            Returns `None` if the account does not have a transaction it wishes to execute.
+
         """
 
     def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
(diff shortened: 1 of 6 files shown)


fubuloubu:
Note: considering adding `**signer_options` to `sign_transaction`

src/ape/api/accounts.py
```diff
@@ -64,18 +63,28 @@ def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
...
+            Returns `None` if the account does not have a transaction it wishes to execute.
+
         """
 
-    def call(self, txn: TransactionAPI, send_everything: bool = False) -> ReceiptAPI:
+    def call(
+        self,
+        txn: TransactionAPI,
+        send_everything: bool = False,
+        **signer_options,
```
unparalleled-js, line 86: Thoughts on using `**kwargs` here? Or maybe `signer_options: SignerOptions`?

Reason: It could be risky assigning a specific meaning here to the `**` case since `call()` does a lot.
Note: It makes sense in `sign_transaction()` but in `call()` is where I think it gets gray.

src/ape/api/accounts.py
```diff
@@ -156,21 +166,19 @@ def transfer(
...
-            txn.value = self.conversion_manager.convert(value, int)
-            return self.call(txn)
-
-        elif not kwargs.get("send_everything"):
+        if not value and not kwargs.get("send_everything"):
             raise AccountsError("Must provide 'VALUE' or use 'send_everything=True'")
 
-        else:
-            return self.call(txn, send_everything=True)
+        elif value and "send_everything" in kwargs and kwargs["send_everything"]:
```
unparalleled-js, line 175: I am pretty sure this works but I could be missing something dumb:

```suggestion
        elif value and kwargs.get("send_everything"):
```

src/ape/api/accounts.py
```diff
@@ -215,29 +223,31 @@ def deploy(
...
     ) -> bool:
         """
         Verify a message or transaction was signed by this account.
 
         Args:
             data (Union[:class:`~ape.types.signatures.SignableMessage`, :class:`~ape.api.transactions.TransactionAPI`]):  # noqa: E501
               The message or transaction to verify.
-            signature (Optional[``_Signature``]): The :class:`~ape.types.signatures.MessageSignature` or the
-              :class:`~ape.types.signatures.TransactionSignature`.
+            signature (Optional[``_Signature``]): The :class:`~ape.types.signatures.MessageSignature`.
```
unparalleled-js, line 234: Mostly fixing the type here but also making the link happen at in the type-doc instead of the description.

```suggestion
            signature (Optional[:class:`~ape.types.signatures.MessageSignature`]): The signature.
```

src/ape_accounts/accounts.py
```diff
@@ -118,20 +118,22 @@ def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
...
 
-        signed_txn = EthAccount.sign_transaction(
+        signature = EthAccount.sign_transaction(
             txn.dict(exclude_none=True, by_alias=True), self.__key
         )
-        return TransactionSignature(  # type: ignore
-            v=signed_txn.v,
-            r=to_bytes(signed_txn.r),
-            s=to_bytes(signed_txn.s),
+        txn.signature = TransactionSignature(  # type: ignore
```
unparalleled-js, line 129: This should not longer be needed now that we have Pydantic mypy extension working!

```suggestion
        txn.signature = TransactionSignature(
```

unparalleled-js (approved):
Added some small thoughts but this one seems pretty dang close to merge

src/ape/api/accounts.py
fubuloubu: Nice, thanks!

src/ape_accounts/accounts.py
fubuloubu: I tried directly and via `pre-commit`, and seems like this is still a problem (perhaps because this is a `NamedTuple` subclass?)

fubuloubu: Ah, noticed this was resolved in https://github.com/ApeWorX/ape/pull/1212

unparalleled-js (approved):
ship it!

PR #1202, opened 2022-12-22
refactor!: `prepare_transaction()` logic to `Web3Provider`

NotPeopling2day:
### What I did

Current implementation on `ProviderAPI` is too specific to EVM, making method abstract and moving implementation to `Web3Provider`

fixes: #

### Checklist


fubuloubu (dismissed):
Nice. Test that this works with alchemy and hardhat plugins, as well as Starknet and then can merge

unparalleled-js (approved):

PR #1205, opened 2023-01-04
docs: update PyPI Links

fubuloubu:
### What I did

Noticed that you can add additional links to a PyPI page (including funding) so I added some

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,7 @@ setup(
         "Funding": "https://gitcoin.co/grants/5958/ape-maintenance-fund",
         "Source": "https://github.com/ApeWorX/ape",
         "Tracker": "https://github.com/ApeWorX/ape/issues",
+        "Twitter": "https://twitter.com/ApeFramework",
     },
     include_package_data=True,
     install_requires=[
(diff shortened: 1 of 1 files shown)


NotPeopling2day (dismissed):
🔥

setup.py
```diff
@@ -80,7 +80,13 @@
     long_description_content_type="text/markdown",
     author="ApeWorX Ltd.",
     author_email="admin@apeworx.io",
-    url="https://github.com/ApeWorX/ape",
+    url="https://apeworx.io",
+    project_urls={
+        "Documentation": "https://docs.apeworx.io/ape/",
+        "Funding": "https://gitcoin.co/grants/5958/ape-maintenance-fund",
+        "Source": "https://github.com/ApeWorX/ape",
+        "Tracker": "https://github.com/ApeWorX/ape/issues",
```
unparalleled-js, line 88: If you add `Twitter` here, I think it will show up with the little bird! Wondering if there are any more?

unparalleled-js (dismissed):

setup.py
fubuloubu: Twitter is a good one!

fubuloubu:
```suggestion
        "Tracker": "https://github.com/ApeWorX/ape/issues",
        "Twitter": "https://twitter.com/ApeFramework",
```

unparalleled-js (approved):

PR #1210, opened 2023-01-06
feat!: refactor account history APIs

Ninjagod1251:
### What I did

changed all account_history to transaction history and passed the OG test

fixes: APE-304

Change descriptions:

* `chain.history` -> `chain.account_history`
* `chain.history[txn_hash]` ->`chain.account_history.get_receipt(txn_hash)`
* `chain.history[account_api]` now returns an `AccountHistory` that we can fill in more later.
* `chain.history[address]` does the same as above
* To get sessional receipts for an account, you can do `chain.history[account].sessional`.

I think that is everything?

### Checklist


sabotagebeats:
tests are passing for me

sabotagebeats (dismissed):
looks good to me

src/ape/managers/chain.py
```diff
@@ -1157,15 +1157,15 @@ def blocks(self) -> BlockContainer:
         return self._block_container_map[self.chain_id]
 
     @property
-    def account_history(self) -> AccountHistory:
+    def transaction_history(self) -> TransactionHistory:
```
fubuloubu, line 1160: Can this just be `history`?

fubuloubu (commented):
Had a thought originally of making this work like the following for a higher-level API:

```py
chain.history # all txns in a chain
chain.history[txn_hash] # a specific txn
chain.history[acct] # all txns from account, ordered by nonce
# maybe `chain.history[acct].incoming`, `.outgoing`, `.all`, etc.
chain.history[acct][nonce] # a specific txn from acct at given nonce
```

unparalleled-js:
> Had a thought originally of making this work like the following for a higher-level API:
> 
> ```python
> chain.history # all txns in a chain
> chain.history[txn_hash] # a specific txn
> chain.history[acct] # all txns from account, ordered by nonce
> # maybe `chain.history[acct].incoming`, `.outgoing`, `.all`, etc.
> chain.history[acct][nonce] # a specific txn from acct at given nonce
> ```

I like this! We can at least do the fist two (`txn_hash` and `acct`) and add the nonce stuff later if we want.

I vote we do this now for 0.6!

fubuloubu:
> > Had a thought originally of making this work like the following for a higher-level API:
> > ```python
> > chain.history # all txns in a chain
> > chain.history[txn_hash] # a specific txn
> > chain.history[acct] # all txns from account, ordered by nonce
> > # maybe `chain.history[acct].incoming`, `.outgoing`, `.all`, etc.
> > chain.history[acct][nonce] # a specific txn from acct at given nonce
> > ```
> 
> I like this! We can at least do the fist two (`txn_hash` and `acct`) and add the nonce stuff later if we want.
> 
> I vote we do this now for 0.6!

Okay great! Yeah, we can implement a piece at the time. The caching system takes care of querying by account and nonce for transactions, so easy to add that eventually too

unparalleled-js:
I updated the linear ticket to include the remaining parts.

Also, I think the property style naming is okay in the `AccountHistory` because they are iterators so that's fine. We can include it now too since it is an easy but maybe that's an extra credit item.

Sorry for all the back and forths and everyhing, but I think we have a path forward on this one.

@Ninjagod1251 let me know how I can help tomorrow or whenever, let's finish this.

fubuloubu:
> I updated the linear ticket to include the remaining parts.

I added some additional color to Linear, but agree that this ticket is pretty defined in scope and should be able to be completed just fine, pushing some of the scope creep that was discussed into future tickets

src/ape/managers/chain.py
```diff
@@ -377,89 +383,109 @@ def _get_block(self, block_id: BlockID) -> BlockAPI:
...
-            sender: AddressType = receipt.sender
-            if receipt.txn_hash not in [r.txn_hash for r in self._address_list_map.get(sender, [])]:
-                self.append(receipt)
+    @property
+    def all(self) -> Iterator[ReceiptAPI]:
+        """
+        All account transactions.
+        """
+        yield from self.outgoing
+        yield from self._transaction_list
```
fubuloubu, line 414: Might be good to yield "in order" meaning pull from each iterator whatever the oldest timestamp is (or newest, just document what direction it goes in and be consistent)

src/ape/pytest/fixtures.py
```diff
@@ -146,14 +146,14 @@ def capture_range(self, start_block: int, stop_block: int):
...
-            if t.receiver and t.sender and t.sender in self.chain_manager.account_history
+            if t.receiver and t.sender and t.sender in self.chain_manager.history
         ]
 
         for txn in transactions:
             self.capture(txn.txn_hash.hex())
 
     def capture(self, transaction_hash: str):
-        receipt = self.chain_manager.account_history.get_receipt(transaction_hash)
+        receipt = self.chain_manager.transaction_history.get_receipt(transaction_hash)
```
fubuloubu, line 156: Should be?
```suggestion
        receipt = self.chain_manager.history.get_receipt(transaction_hash)
```

fubuloubu (dismissed):
Looking great!

fubuloubu:
Seems like test error is that `get_receipt` was not kept in `TransactionHistory`

src/ape/managers/chain.py
```diff
@@ -389,49 +389,77 @@ class AccountHistory(BaseInterfaceModel):
...
+                # Likely ``self.address`` is a contract.
+                # Cache the receipts by their sender instead and skip them here.
+                self.chain_manager.history.append(receipt)
+                continue
+
+            elif receipt.txn_hash not in [t.txn_hash for t in self._session_receipts]:
                 yield receipt
 
     @property
     def all(self) -> Iterator[ReceiptAPI]:
```
fubuloubu, line 414: One thought I was having is that we probably shouldn't have this property until we legitimately have *all* of the transactions indexed, otherwise it's semantics are gonna change when adding `.incoming`, `.internal`, etc.

src/ape/managers/chain.py
```diff
@@ -389,49 +389,77 @@ class AccountHistory(BaseInterfaceModel):
...
+            else:
+                yield sessional.pop(0)
 
+                # Catch up before yielding outgoing receipt.
+                if sessional:
+                    for i in range(len(sessional)):
+                        if sessional[i].block_number <= receipt.block_number:
+                            yield sessional.pop(i)
+                        else:
+                            break
```
fubuloubu, line 431: This seems like a good candidate for something in `ape.utils`

src/ape/managers/chain.py
```diff
@@ -389,49 +389,77 @@ class AccountHistory(BaseInterfaceModel):
...
 
-    _address_list_map: Dict[AddressType, AccountHistory] = {}
+        # Yield remaining sessional.
+        for sessional_receipt in sessional:
+            yield sessional_receipt
 
-    # Facilitates faster look-ups.
-    _hash_to_receipt_map: Dict[str, ReceiptAPI] = {}
+    def __iter__(self) -> Iterator[ReceiptAPI]:  # type: ignore[override]
+        yield from self.all
```
fubuloubu, line 441: This is better as `self.outgoing` because the idea is that `__getitem__` would be iterable by an account's `nonce` (which is the same semantically speaking)

fubuloubu (approved):
Some good progress!

Love that this can facilitate queries such as `total_gas_used = sum(txn.gas_used for txn in chain.history[account])` and `ether_sent = sum(txn.value for txn in chain.history[account][10:90]`

Eventually could event facilitate a query kind of like `2022_txns = chain.history[account].query("*", start_date="1/1/2022", end_date="12/31/2022")

src/ape/managers/chain.py
unparalleled-js: It would be great if Ethereum created an RPC for this. Etherscan saves the day at least, but it only works of the Etherscan plugin is installed.

I found this gem: https://github.com/ethereum/go-ethereum/issues/1897

src/ape/managers/chain.py
unparalleled-js: Removed it for now, but we can consider that next time! We can find it again on this PR.

unparalleled-js:
I am guessing this PR is pretty close, at least for a first pass? It sounds like we have some good plans to fill in account history more and solve the problem of Ethereum to be able to see txns of an account.

@Ninjagod1251 Please review changes more and let me know if you have any questions.
Thank you for your hard work on this!

Thank @fubuloubu for helping with the design and everything else.

src/ape/managers/chain.py
fubuloubu: Nice! Long term, my idea was to use this query type:

https://github.com/ApeWorX/ape/blob/0e31f26578aa41d9015489de5ff1dd6f9ee725da/src/ape/api/query.py#L97

So, Etherscan could implement this functionality as a query plugin to source this info

But this is an implementation detail, we can refactor how this works after 0.6.0 and when the query system works a little better

fubuloubu: Perhaps a nice side effect of rebasing to use this though is that we can make this type of query unhandled by default unless you have a query plugin installed that handles it

src/ape/managers/chain.py
```diff
@@ -377,89 +383,113 @@ def _get_block(self, block_id: BlockID) -> BlockAPI:
...
 
-        return self._address_list_map.get(address_key, [])
+    def __iter__(self) -> Iterator[ReceiptAPI]:  # type: ignore[override]
+        yield from self.outgoing
 
-    def __iter__(self) -> Iterator[AddressType]:
-        """
-        Iterate through the accounts listed in the history map.
+    def __len__(self) -> int:
+        return len(list(self.outgoing))
```
fubuloubu, line 416: This can just be `self.provider.get_nonce(self.address)`

src/ape/managers/chain.py
```diff
@@ -377,89 +383,113 @@ def _get_block(self, block_id: BlockID) -> BlockAPI:
...
+                # TODO: Replace with query manager once supports receipts
+                #  instead of transactions.
+                # TODO: Add timeout = 0 once in API method to not wait for txns
+                receipt = self.provider.get_receipt(account_or_hash)
+                self.append(receipt)
+
+            return receipt
+
+    @singledispatchmethod
+    def __contains__(self, key):
```
fubuloubu, line 472: Is there an implementation for this?

fubuloubu (approved):
Two small comments, but good to go I think

src/ape/managers/chain.py
fubuloubu: or `nonce + 1`, not sure if it's the next nonce or the last one

src/ape/managers/chain.py
fubuloubu: If not, might as well leave out for now (unless it needed to have `__contains__` for some reason)

src/ape/managers/chain.py
unparalleled-js: I love this!
But one thing to note is that it could be ecosystem dependent regarding whether it starts at 0 or 1, lol. I think Starknet was different at one point.

unparalleled-js: But for Ethereum it would be `account.nonce` I think because it starts as `0` and you haven't sent any txns yet.
I am not sure about contracts though.

fubuloubu: Worth following up on

Wonder if starknet started at 1 because accounts needed to be deployed first

fubuloubu: > I am not sure about contracts though.

contracts have no ability to create outgoing transactions (defined as transactions where they are original source)
https://etherscan.io/txs?a=0xfeb4acf3df3cdea7399794d0869ef76a6efaff52&f=2

src/ape/managers/chain.py
unparalleled-js: removed!

src/ape/managers/chain.py
fubuloubu: So for contracts, you won't be able to search their history using `.outgoing`, only by using `.incoming` or `.internal` (which are much heavier queries to make)

fubuloubu (approved):

PR #1213, opened 2023-01-09
refactor!: replace evm-types in our APIs with custom generic types

unparalleled-js:
### What I did

fixes: #1040
fixes: #1217
Fixes: APE-62

### How I did it

* Create out own `CallTreeNode` class that is more generic and have provider create those. This class also takes some of the logic and responsibilities from the `CallTreeParser` and now `CallTreeParser` is not needed.
* Refactor more of a data-enrichment pattern. Now, the data is enriched as it goes along. The logic that did not move to `CallTreeNode` is now part of `ape-ethereum.Ethereum` which implements the new method `enrich_calltree()`.

### Checklist


unparalleled-js:
There might be some merge conflicts with https://github.com/ApeWorX/ape/pull/1160 so I want to merge that one first before taking this out of draft

src/ape_ethereum/ecosystem.py
```diff
@@ -379,6 +396,37 @@ def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
 
         return tuple(output_values)
 
+    def _enrich_value(self, value: Any) -> Any:
```
fubuloubu, line 399: Some food for thought: IPython supports rich display through `_repr_pretty_`

src/ape/types/trace.py
```diff
@@ -0,0 +1,113 @@
...
+            reports.append(report)
+
+        return _merge_gas_reports(*reports)
+
+
+def _merge_gas_reports(*reports: GasReport) -> GasReport:
+    if len(reports) == 1:
+        return reports[0]
+    elif len(reports) > 1:
+        return merge_reports(*reports)
```
fubuloubu, line 87: Maybe helpful to upstream this fix?

fubuloubu (commented):
Nice amount of code savings, I didn't review too deeply but it looks good

spinoch:
Does this mean we can remove `ethpm-types` as a dependency?

fubuloubu:
> Does this mean we can remove `ethpm-types` as a dependency?

No, this is to reduce the reliance of ape core on `evm-trace` (which was maybe a typo)

We will still be using that package however for `ape-ethereum` (which is bundled with this package)

src/ape/api/transactions.py
```diff
@@ -318,22 +317,17 @@ def await_confirmations(self) -> "ReceiptAPI":
...
         The method ABI of the method called to produce this receipt.
-        Requires both that the user is using a provider that supports traces
-        as well as for this to have been a contract invocation.
         """
 
-        call_tree = self.call_tree
-        if not call_tree:
-            return None
-
-        return self.chain_manager.contracts.lookup_method(call_tree.address, call_tree.calldata)
```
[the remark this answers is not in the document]
unparalleled-js: Ya I removed it on this PR, it is no longer needed.

We could deprecate it but I think since it is a breaking change and probably not a common method, it is probably ok to delete for 0.6.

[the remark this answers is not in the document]
fubuloubu: Sounds good to me

src/ape_ethereum/ecosystem.py
unparalleled-js: I implemented it and did some refactoring that was necessary that

src/ape/types/trace.py
unparalleled-js: Ya, good call!

https://github.com/ApeWorX/evm-trace/pull/32

src/ape_ethereum/transactions.py
```diff
@@ -190,14 +203,27 @@ def show_trace(self, verbose: bool = False, file: IO[str] = sys.stdout):
             sender=self.sender,
             transaction_hash=self.txn_hash,
             revert_message=revert_message,
-            verbose=verbose,
```
unparalleled-js, line 193: - [ ] Re-handle `verbose` somehow.

fubuloubu (dismissed):

fubuloubu (dismissed):

fubuloubu (approved):

NotPeopling2day (dismissed):
Thanks for answering my questions, LGTM!

unparalleled-js:
- [ ] Ensure is good based on migration results from `ape-hardhat`

Let me do this final step. I have a hunch that something is iffy with transfer txns now and there are some good tests in `ape-hardhat`. Also, `ape-hardhat` relies on the `TraceFrame` approach so it is a good test in that regard.

fubuloubu (dismissed):
Needs QA on the report printing to screen, make sure all is well

src/ape_ethereum/ecosystem.py
```diff
@@ -224,7 +242,7 @@ def str_to_slot(text):
             outputs=[ABIType(type="string")],
         )
         try:
-            name = ContractCall(abi, address)()
+            name = ContractCall(abi, address)(skip_trace=True)
```
fubuloubu, line 245: Hmm, I missed where `skip_trace` kwarg was introduced. In this PR or another one?

Its just an optimization for internal uses? Why do traces happen normally, and can this be opt in behavior instead of opt out?

fubuloubu (commented):
Does this PR include events or is that another backlog PR?

unparalleled-js:
> Does this PR include events or is that another backlog PR? 

That is separate work; it is being tracked already. I can do that next if we want!

unparalleled-js:
More testing to do on my end before merging, just realized:

- [ ] Ensure brownie dependency stuff all still works
- [ ] Migrate `ape-solidity`
- [ ] Migrate `ape-etherscan`

fubuloubu:
> > Does this PR include events or is that another backlog PR?
> 
> That is separate work; it is being tracked already. I can do that next if we want!

Would be nice to have eventually, especially in regard to the first comment here: https://github.com/ApeWorX/ape/pull/1239#issuecomment-1400663680

src/ape_ethereum/ecosystem.py
unparalleled-js: It was added on a separate PR but missed here.

If you are running tests and need traces (such as running with `--gas`), it would attempt to gather trace data from every single proxy-check call we do here, which made debugging trace-related stuff challenging and also made things slow.

It is opt-in in the sense that it only gathers trace data on calls if it needs it (such as with `--gas`), but if the kwarg were reversed, every call in a project's tests would need the kwarg in order for it to show up in gas report.

fubuloubu: Gotcha, so it's sort of an internal kwarg that ape injects during moments it needs it

fubuloubu (approved):

PR #1214, opened 2023-01-11
feat!: conform ProviderAPI.get_receipt()

dtdang:
### What I did

Add kwargs to ProviderAPI.get_receipt() to match what is present in the Web3Provider.get_receipt()

fixes: APE-302

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -326,12 +326,18 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def get_receipt(self, txn_hash: str) -> ReceiptAPI:
+    def get_receipt(
+        self, txn_hash: str,  required_confirmations: int = 0, timeout: Optional[int] = None
+    ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
 
         Args:
             txn_hash (str): The hash of the transaction to retrieve.
+            required_confirmations (int): The amount of block confirmations
+              to wait before returning the receipt. Defaults to ``0``.
+            timeout (Optional[int]): The amount of time to wait for a receipt
+              before timing out. Defaults ``None``.
 
         Returns:
             :class:`~api.providers.ReceiptAPI`:
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -327,7 +327,7 @@ class ProviderAPI(BaseInterfaceModel):
 
     @abstractmethod
     def get_receipt(
-        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None
+        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None, **kwargs
     ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -326,23 +326,12 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @abstractmethod
-    def get_receipt(
-        self,
-        txn_hash: str,
-        required_confirmations: int = 0,
-        timeout: Optional[int] = None,
-        **kwargs,
-    ) -> ReceiptAPI:
+    def get_receipt(self, txn_hash: str, **kwargs) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
 
         Args:
             txn_hash (str): The hash of the transaction to retrieve.
-            required_confirmations (int): The amount of block confirmations
-              to wait before returning the receipt. Defaults to ``0``.
-            timeout (Optional[int]): The amount of time to wait for a receipt
-              before timing out. Defaults to ``None``. When ``None``, the default
-              value comes from the network config in ``ape-config.yaml``.
             kwargs: Any other kwargs that other providers might allow when fetching a receipt.
 
         Returns:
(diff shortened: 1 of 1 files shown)


src/ape/api/providers.py
```diff
@@ -316,12 +316,18 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
...
+        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None
+    ) -> ReceiptAPI:
         """
         Get the information about a transaction from a transaction hash.
 
         Args:
             txn_hash (str): The hash of the transaction to retrieve.
+            required_confirmations (int): The amount of block confirmations
+              to wait before returning the receipt. Defaults to ``0``.
+            timeout (Optional[int]): The amount of time to wait for a receipt
```
unparalleled-js, line 329: We should what it means to be `None`.
I think `None` means use the settings block time out thing? Or does it mean to never timeout?

fubuloubu: Is there a default timeout?"

NotPeopling2day: When `None` the default value returns from the `transaction_acceptance_timeout` setting.

fubuloubu (approved):

NotPeopling2day (dismissed):

src/ape/api/providers.py
```diff
@@ -316,12 +316,19 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
...
+              before timing out. Defaults ``None``. When ``None``, the default
```
unparalleled-js, line 330:
```suggestion
              before timing out. Defaults to ``None``. When ``None``, the default
```

src/ape/api/providers.py
```diff
@@ -316,12 +316,19 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
...
+              value returns from the network value in ``ape-config.yaml``.
```
unparalleled-js, line 331:
```suggestion
              value comes from the network config in ``ape-config.yaml``.
```

unparalleled-js (dismissed):

unparalleled-js (approved):
If you want a green check mark, you have to commit using the cz guidelines.
Looks good though!

fubuloubu:
> If you want a green check mark, you have to commit using the cz guidelines. Looks good though!

Actually, @dtdang our best practice here is to submit PRs from a fork repo instead of against the main one, that's why this CI was executed

src/ape/api/providers.py
```diff
@@ -316,12 +316,19 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
         """
 
     @abstractmethod
-    def get_receipt(self, txn_hash: str) -> ReceiptAPI:
+    def get_receipt(
+        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None
```
fubuloubu, line 320: Do we picture having more kwargs here in the future? do we want to add `**kwargs` so it's not breaking?

fubuloubu (dismissed):

src/ape/api/providers.py
unparalleled-js: I think that is a good thing to do. This may save us an breaking change item in the future. Are there any other methods where a safety kwargs could be added perhaps? `send_transaction()` ?

fubuloubu (approved):

src/ape/api/providers.py
NotPeopling2day: If we do that, I assume the other updated methods will be on separate PRs/tickets?

To recap, we want @dtdang to update this PR with **kwargs?
```suggestion
        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None, **kwargs
```

src/ape/api/providers.py
```diff
@@ -316,12 +316,19 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
         """
 
     @abstractmethod
-    def get_receipt(self, txn_hash: str) -> ReceiptAPI:
+    def get_receipt(
+        self, txn_hash: str, required_confirmations: int = 0, timeout: Optional[int] = None, **kwargs
```
fubuloubu, line 320:
```suggestion
        self,
        txn_hash: str,
        required_confirmations: int = 0,
        timeout: Optional[int] = None,
        **kwargs,
```

src/ape/api/providers.py
```diff
@@ -316,12 +316,19 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:  # Return value of
...
         """
         Get the information about a transaction from a transaction hash.
 
         Args:
             txn_hash (str): The hash of the transaction to retrieve.
+            required_confirmations (int): The amount of block confirmations
+              to wait before returning the receipt. Defaults to ``0``.
+            timeout (Optional[int]): The amount of time to wait for a receipt
+              before timing out. Defaults to ``None``. When ``None``, the default
+              value comes from the network config in ``ape-config.yaml``.
```
fubuloubu, line 332:
```suggestion
            kwargs: Any other kwargs that other providers might allow when fetching a receipt.

```

fubuloubu (approved):

unparalleled-js (dismissed):

fubuloubu:
Should make `ProviderAPI` just kwargs and Web3Provider have the additional kwargs

fubuloubu (approved):

fubuloubu (approved):

PR #1218, opened 2023-01-13
docs: add anonymized analytics

fubuloubu:
### What I did

It would be really helpful for us to monitor growth by seeing our analytics from the docs page

fixes: APE-453

### How I did it

I chose Plausible because they are privacy-preserving and open source

### How to verify it

Once docs are released (on main) should be able to see the page

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,6 +29,7 @@ extensions = [
     "sphinx.ext.autosummary",
     "sphinx.ext.napoleon",
     "sphinx_rtd_theme",
+    "sphinx_plausible",
 ]
 autosummary_generate = True
 
@@ -48,6 +49,8 @@ source_suffix = [".rst", ".md"]
 # The master toctree document.
 master_doc = "index"
 
+# Configure Pluasible
+plausible_domain = "docs.apeworx.io"
 
 # -- Options for HTML output -------------------------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,7 +20,7 @@ sys.path.insert(0, os.path.abspath(".."))
 # -- Project information -----------------------------------------------------
 
 project = "ape"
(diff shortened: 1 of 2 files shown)


fubuloubu:
Also found this: https://www.goatcounter.com/

fubuloubu:
Another: https://github.com/plausible/analytics

unparalleled-js:
Note, I had to switch to Chrome (and not Brave) to get the script to not be blocked.

unparalleled-js (dismissed):
I did get the script to load on chrome. Does anything show up on the account?

fubuloubu:
Hmm, do all scripts get blocked by brave by default?

unparalleled-js:
There is also the chance the privacy-browser purposely misleads the analytics tools, in the case to avoid penalty.

https://brave.com/privacy-updates/1-web-resource-replacements/#solution-dont-just-block-replace

fubuloubu:
I wish github pages just had analytics so it could tell us how many downloads we were getting lol

unparalleled-js:
Maybe we should try the other 2 that you linked and see if they are considered "trackers" by Brave?

The GOAT one seems like it may be cognizant of this.

fubuloubu:
> Maybe we should try the other 2 that you linked and see if they are considered "trackers" by Brave?
> 
> The GOAT one seems like it may be cognizant of this.
> 

I'm into this. I like that the others are open source

fubuloubu:
> I did get the script to load on chrome. Does anything show up on the account?

Try again, it wasn't watching

fubuloubu:
> It works on Chrome but not Brave! I don't expect it to work on Brave though.

There's a way to set it up with a proxy method which may bypass this, but not screwing around with this now

> I like how easy it integrates with our docs stack too.

Yeah, it's a 3rd party plugin, but seems relatively active so that's cool.

Issue #1220, opened 2023-01-14
Can't use Union type for converter


fubuloubu:
### Environment information

- OS: linux
- Python Version: 3.9.14
- `ape` and plugin versions:

```sh
$ ape --version
0.5.10.dev9+gb57d4a9b

$ ape plugins list
Installed Plugins:
  infura       0.5.3
  ens          0.5.1
  etherscan    0.5.4
  foundry      0.5.2
  ledger       0.5.0
```

### What went wrong?

When making a call with `my_contract.myMethod(..., gas_limit=<some-integer-value>), it fails to process because it is asking the conversion plugin system to convert to a Union type, which is not possible:

```py
File ~/.pyenv/versions/3.9.14/envs/ape-safe/lib/python3.9/site-packages/ape/contracts/base.py:336, in ContractTransactionHandler.__call__(self, *args, **kwargs)
    334 function_arguments = self._convert_tuple(args)
    335 contract_transaction = self._as_transaction(*function_arguments)
--> 336 return contract_transaction(*function_arguments, **kwargs)

File ~/.pyenv/versions/3.9.14/envs/ape-safe/lib/python3.9/site-packages/ape/contracts/base.py:274, in ContractTransaction.__call__(self, *args, **kwargs)
    273 def __call__(self, *args, **kwargs) -> ReceiptAPI:
--> 274     txn = self.serialize_transaction(*args, **kwargs)
    276     if "sender" in kwargs and isinstance(kwargs["sender"], AccountAPI):
    277         return kwargs["sender"].call(txn, **kwargs)

File ~/.pyenv/versions/3.9.14/envs/ape-safe/lib/python3.9/site-packages/ape/contracts/base.py:268, in ContractTransaction.serialize_transaction(self, *args, **kwargs)
    264 if "sender" in kwargs and isinstance(kwargs["sender"], ContractInstance):
    265     # Automatically impersonate contracts (if API available) when sender
    266     kwargs["sender"] = self.account_manager.test_accounts[kwargs["sender"].address]
--> 268 kwargs = _convert_kwargs(kwargs, self.conversion_manager.convert)
    269 return self.provider.network.ecosystem.encode_transaction(
    270     self.address, self.abi, *args, **kwargs
    271 )

File ~/.pyenv/versions/3.9.14/envs/ape-safe/lib/python3.9/site-packages/ape/contracts/base.py:24, in _convert_kwargs(kwargs, converter)
     22 fields = TransactionAPI.__fields__
[...]

PR #1224, opened 2023-01-16
fix: don't use Union type in BaseInterfaceModel

fubuloubu:
### What I did

Move some of the transaction validation logic for `gas_limit` from `prepare_transaction` to transaction validator on `TransactionAPI` class

fixes: #1220

### Checklist


src/ape/api/transactions.py
```diff
@@ -46,6 +47,31 @@ class TransactionAPI(BaseInterfaceModel):
     class Config:
         allow_population_by_field_name = True
 
+    @validator("gas_limit", pre=True)
+    def validate_gas_limit(cls, value):
+        if value is None:
+            if not cls.network_manager.active_provider:
+                raise NetworkError("Must be connected to use default gas config")
```
NotPeopling2day, line 54:
```suggestion
                raise NetworkError("Must be connected to use default gas config.")
```

src/ape/api/transactions.py
```diff
@@ -46,6 +47,31 @@ class TransactionAPI(BaseInterfaceModel):
...
+
+            value = cls.network_manager.active_provider.network.gas_limit
+
+        if value == "auto":
+            return None  # Delegate to `ProviderAPI.estimate_gas_cost`
+
+        elif value == "max":
+            if not cls.network_manager.active_provider:
+                raise NetworkError("Must be connected to use 'max'")
```
NotPeopling2day, line 63:
```suggestion
                raise NetworkError("Must be connected to use 'max'.")
```

fubuloubu:
Don't quite understand how the test failure could arise

src/ape/api/transactions.py
```diff
@@ -30,7 +31,7 @@ class TransactionAPI(BaseInterfaceModel):
     chain_id: int = Field(0, alias="chainId")
     receiver: Optional[AddressType] = Field(None, alias="to")
     sender: Optional[AddressType] = Field(None, alias="from")
-    gas_limit: Optional[GasLimit] = Field(None, alias="gas")
```
unparalleled-js, line 33: Should we delete this type from `ape.types`? I think this is was the only reason it was created.

unparalleled-js:
> Don't quite understand how the test failure could arise

It seems like the test is failing best the test provider is using estimated gas again instead of MAX_GAS.
When you use estimate gas, you get a contact logic error during gas estimation.

src/ape/api/transactions.py
fubuloubu:
```
src/ape_ethereum/ecosystem.py:    gas_limit: GasLimit = "auto"
src/ape_ethereum/ecosystem.py:    def validate_gas_limit(cls, value: GasLimit) -> GasLimit:
src/ape/api/networks.py:    def gas_limit(self) -> GasLimit:
```

fubuloubu: I wanted to move this Union to the outer limit of validating when `kwargs` first come in so that `TransactionAPI` can be simple, might need to do some work though

fubuloubu:
> > Don't quite understand how the test failure could arise
> 
> It seems like the test is failing best the test provider is using estimated gas again instead of MAX_GAS. When you use estimate gas, you get a contact logic error during gas estimation.

Should we catch and re-raise a GasEstimationError for these situations?

src/ape/api/transactions.py
```diff
@@ -46,6 +47,31 @@ class TransactionAPI(BaseInterfaceModel):
     class Config:
         allow_population_by_field_name = True
 
+    @validator("gas_limit", pre=True)
+    def validate_gas_limit(cls, value):
+        if value is None:
+            if not cls.network_manager.active_provider:
+                raise NetworkError("Must be connected to use default gas config.")
+
+            value = cls.network_manager.active_provider.network.gas_limit
```
unparalleled-js, line 56: OK - I figured out why things are not working on this PR. Thins are happening in the wrong order.

The default network config for `gas_limit` is **not** honored here (only in this PR).
The gas estimation logic in `prepare_transaction()` triggers _before_ this validation logic here in the txn model.
Right now, it will always use `auto-estimate`, regardless of what is specified in the config file, and that is why there are failing tests.

unparalleled-js: hmm, maybe it is actually that there are conflicting validators in `ape_ethereum`. So maybe not a timing issue but a conflict.

unparalleled-js:
@fubuloubu if you don't mind, I am going to complete this one (unless you are working on it still).
It is causing other issues and I would like to make sure this gets in.

fubuloubu:
> @fubuloubu if you don't mind, I am going to complete this one (unless you are working on it still). It is causing other issues and I would like to make sure this gets in.

Thanks, yeah this one sort of slipped through my ability to complete it

src/ape_ethereum/ecosystem.py
```diff
@@ -101,23 +103,24 @@ class NetworkConfig(PluginConfig):
...
-
-            # Value could be an integer string
-            if value.isdigit():
-                return int(value)
-            # Enforce "0x" prefix on base 16 integer strings
-            elif value.lower().startswith("0x"):
-                return int(value, 16)
-            else:
-                raise ValueError("Invalid gas_limit, must be 'auto', 'max', or a number")
+    @validator("gas_limit", pre=True)
```
fubuloubu, line 106: I think `allow_reuse` is necessary so that subclasses which use this class as a base inherit this validator also

unparalleled-js: ya good catch, a lot of ecosystems subclass `Ethereum`.

unparalleled-js: oh but this is only the config class, not sure if that gets subclassed ever, but it doesn't hurt I suppose

unparalleled-js: I put it in, just in case

fubuloubu:
Also realized this is breaking, so definitely need for 0.6.0

unparalleled-js (dismissed):
@fubuloubu let me know when this looks good to you, I don't think it will let you approve

src/ape/api/transactions.py
```diff
@@ -45,6 +46,31 @@ class TransactionAPI(BaseInterfaceModel):
     class Config:
         allow_population_by_field_name = True
 
+    @validator("gas_limit", pre=True)
```
fubuloubu, line 49:
```suggestion
    @validator("gas_limit", pre=True, allow_reuse=True)
```

fubuloubu:
> @fubuloubu let me know when this looks good to you, I don't think it will let you approve

Looks good with this one change

unparalleled-js (approved):

fubuloubu (commented):
Approved

Issue #1227, opened 2023-01-17
Allow using bytecode from JSON artifacts.


sajal:
### Overview

I am trying to use a JSON artifact created using hardhat(from a much larger project), as my test dependency.

The current JSON compiler only allows using JSON artifacts as interfaces, and ignores the bytecode.

Here is an example JSON artifact : https://gist.github.com/sajal/8c793c5066c3684a6178906aa248f347

When deploying using `deployer.deploy(project.MockLinearPool, ..... args ...)` the following warning is emitted `WARNING  ape:base.py:45 Deploying an empty contract (no bytecode)` and when trying to call the contract the following error happens

`ape.exceptions.ContractError: Unable to make contract call. '0x6Af9BB9Cf7307AC439cc7E37859bdD844874ebc1' is not a contract on network 'local'.`

### Specification

I propose adapting the [InterfaceCompiler](https://github.com/ApeWorX/ape/blob/0e31f26578aa41d9015489de5ff1dd6f9ee725da/src/ape_pm/compiler.py#L12-L50) so that `bytecode` and `deployedBytecode` fields (if present in JSON) are included in the final compiled output.

I do not think this change breaks anything.

### Dependencies

I do not think there are any dependencies for this.

unparalleled-js:
Fixed in https://github.com/ApeWorX/ape/pull/1261
Please let us know if you have any more issues! Thank you for reporting and contributing to the fix.

PR #1228, opened 2023-01-17
feat: use bytecode if available when using JSON compiler

sajal:
### What I did

Include bytecode if present when using JSON compiler.

fixes: #1227

fixes: APE-462

### How I did it

I wanted to use contracts coming from much larger repo with their own toolchain. I figured the easiest way would be to compile it as an artifact and use the resulting JSON as test dependency within ape. I looked at how ape_vyper provides the bytecode and replicated that in JSON compiler.

### How to verify it

Use [any JSON artifact](https://gist.github.com/sajal/8c793c5066c3684a6178906aa248f347), put it in your contracts directory and use in test cases/console.

### Checklist

### DIFF
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -36,6 +36,9 @@ contract.my_method()
 2. **Pre-existing Contract Types**: If you have a contract type JSON that was compiled elsewhere, you can include it in your project.
    This is useful if you are unable or unwilling to install a compiler.
 
+3. **Pre-existing Contract Binary**: If you have an artifact with binary compiled elsewhere, you can include it in your project.
+  This is useful if you want to use contracts from much larger projects as dependency for your test cases.
+
 **WARN**: You may have to adjust the name and source ID of the contract type in the JSON to match the new file name in your project.
 
 ## Other Compiler Plugins
--- a/src/ape_pm/compiler.py
+++ b/src/ape_pm/compiler.py
@@ -39,6 +39,10 @@ class InterfaceCompiler(CompilerAPI):
             ):
                 # Raw contract type JSON
                 contract_type_data = data
+                if "deployedBytecode" in data:
+                    contract_type_data["runtimeBytecode"] = {"bytecode": data["deployedBytecode"]}
+                if "bytecode" in data:
+                    contract_type_data["deploymentBytecode"] = {"bytecode": data["bytecode"]}
 
             else:
                 logger.warning(f"Unable to parse {ContractType.__name__} from '{source_id}'.")


docs/userguides/compile.md
```diff
@@ -36,6 +36,9 @@ contract.my_method()
 2. **Pre-existing Contract Types**: If you have a contract type JSON that was compiled elsewhere, you can include it in your project.
    This is useful if you are unable or unwilling to install a compiler.
 
+3. **Pre-existing Contract Binary**: If you have an artifact with binary compiled elsewhere, you can include it in your project.
```
unparalleled-js, line 39: Can you show how to do this?

src/ape_pm/compiler.py
```diff
@@ -39,6 +39,10 @@ def compile(
             ):
                 # Raw contract type JSON
                 contract_type_data = data
+                if "deployedBytecode" in data:
+                    contract_type_data["runtimeBytecode"] = {"bytecode": data["deployedBytecode"]}
```
unparalleled-js, line 43: If there is also `runtimeBytecode` already, do we still wish to override it?

src/ape_pm/compiler.py
```diff
@@ -39,6 +39,10 @@ def compile(
...
+                if "bytecode" in data:
+                    contract_type_data["deploymentBytecode"] = {"bytecode": data["bytecode"]}
```
unparalleled-js, line 45: How do your contract type data end up in this state?
Is it possible you can migrate it downstream somehow before compiling?

docs/userguides/compile.md
sajal: In any brownie project, if you run `brownie compile` then the resulting .json files in `build/contracts/` have the bytecodes included. the process would be similar for hardhat, and i suspect every other tooling around.

src/ape_pm/compiler.py
sajal: I would say so, just for the sake of completeness (as the vyper compiler also adds that)

src/ape_pm/compiler.py
sajal: It is possible, but I think its better if ape accepts the commonly used format by all other tools rather than adding another step to edit the artifact to make it "ape-compatible"

src/ape_pm/compiler.py
unparalleled-js: Ohhh I get it now!

unparalleled-js: The issue is that these are different JSONs. This `if` block is assuming `ContractType.json()`-like structure, which is like step above the raw compiler output.

src/ape_pm/compiler.py
unparalleled-js: Disregard this!

docs/userguides/compile.md
unparalleled-js: I realize now you are looking to add this from raw compilation output. The reason I got confused is because the logic is happening in the `ContractType.json()`-like handler

unparalleled-js:
Superseded by: https://github.com/ApeWorX/ape/pull/1261

I added you as a coauthor there.

PR #1229, opened 2023-01-17
feat: add default sender manager

bilbeyt:
### What I did

Added the support for default_sender in contract transactions

fixes: #

### How I did it

Implemented a new context manager to use in tests.

### How to verify it

Run the tests.

### Checklist

### DIFF
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
 account.balance += int(1e18)  # Gives `account` 1 Ether
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
+
+```python
+with accounts.use_sender(0): # Use first account from test mnemonic
+  contract.myFunction(1)
+
+with accounts.use_sender("<address>"): # Impersonate an account
+  contract.myFunction(1)
+
+with accounts.use_sender(a): # a is a `TestAccountAPI` object
+  contract.myFunction(1)
+```
+
 ## Live Network Accounts
 
 When using live networks, you need to get your accounts into Ape.
@@ -157,6 +172,24 @@ Then, in your scripts, you can [load](../methoddocs/managers.html#ape.managers.a
 from ape import accounts
 
 account = accounts.load("<ALIAS>")
+```
+
+### Default Sender Support
+
+In order to reduce repetition of adding `sender` in your contract calls, you can use `use_sender` context manager.
+
+```python
+with accounts.use_sender(0):
+  contract.myFunction(1)
+
+with accounts.use_sender("<address>"):
+  contract.myFunction(1)
+
+with accounts.use_sender("<alias>"):
+  contract.myFunction(1)
+
+with accounts.use_sender(a): # a is a `AccountAPI` object
+  contract.myFunction(1)
 ```
 
 ## Automation
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -70,6 +70,8 @@ class ContractConstructor(ManagerAccessMixin):
         if "sender" in kwargs and isinstance(kwargs["sender"], AccountAPI):
             sender = kwargs["sender"]
             return sender.call(txn, **kwargs)
+        elif "sender" not in kwargs and self.account_manager.default_sender is not None:
+            return self.account_manager.default_sender.call(txn, **kwargs)
 
         return self.provider.send_transaction(txn)
 
@@ -333,6 +335,8 @@ class ContractTransactionHandler(ContractMethodHandler):
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
         function_arguments = self._convert_tuple(args)
         contract_transaction = self._as_transaction(*function_arguments)
+        if "sender" not in kwargs and self.account_manager.default_sender is not None:
+            kwargs["sender"] = self.account_manager.default_sender
         return contract_transaction(*function_arguments, **kwargs)
 
     def _as_transaction(self, *args) -> ContractTransaction:
--- a/src/ape/managers/accounts.py
+++ b/src/ape/managers/accounts.py
@@ -1,4 +1,7 @@
-from typing import Dict, Iterator, List, Type
+import contextlib
+from typing import ContextManager, Dict, Generator, Iterator, List, Optional, Type, Union
+
+from eth_utils import is_hex
 
 from ape.api.accounts import (
     AccountAPI,
@@ -11,6 +14,17 @@ from ape.types import AddressType
 from ape.utils import ManagerAccessMixin, cached_property, singledispatchmethod
 
 from .base import BaseManager
+
+_DEFAULT_SENDERS: List[AccountAPI] = []
+
+
+@contextlib.contextmanager
+def _use_sender(
+    account: Union[AccountAPI, TestAccountAPI]
+) -> Generator[AccountAPI, TestAccountAPI, None]:
+    _DEFAULT_SENDERS.append(account)
+    yield account
+    _DEFAULT_SENDERS.pop()
 
 
 class TestAccountManager(list, ManagerAccessMixin):
@@ -100,6 +114,13 @@ class TestAccountManager(list, ManagerAccessMixin):
     def generate_test_account(self, container_name: str = "test") -> TestAccountAPI:
         return self.containers[container_name].generate_account()
 
+    def use_sender(self, account_id: Union[TestAccountAPI, AddressType, int]) -> ContextManager:
+        if not isinstance(account_id, TestAccountAPI):
+            account = self[account_id]
+        else:
+            account = account_id
+        return _use_sender(account)
+
 
 class AccountManager(BaseManager):
     """
@@ -116,6 +137,10 @@ class AccountManager(BaseManager):
 
         my_accounts = accounts.load("dev")
     """
+
+    @property
+    def default_sender(self) -> Optional[AccountAPI]:
+        return _DEFAULT_SENDERS[-1] if _DEFAULT_SENDERS else None
 
     @cached_property
     def containers(self) -> Dict[str, AccountContainerAPI]:
@@ -317,3 +342,16 @@ class AccountManager(BaseManager):
             any(address in container for container in self.containers.values())
             or address in self.test_accounts
         )
+
+    def use_sender(
+        self,
+        account_id: Union[AccountAPI, AddressType, str, int],
+    ) -> ContextManager:
+        if not isinstance(account_id, AccountAPI):
+            if isinstance(account_id, int) or is_hex(account_id):
+                account = self[account_id]
+            elif isinstance(account_id, str):  # alias
+                account = self.load(account_id)
+        else:
+            account = account_id
+        return _use_sender(account)
--- a/tests/functional/test_default_sender_context_manager.py
+++ b/tests/functional/test_default_sender_context_manager.py
@@ -1,0 +1,65 @@
+import pytest
+
+from ape.exceptions import SignatureError
+from ape.pytest.contextmanagers import RevertsContextManager as reverts
+
+
+def test_default_sender_test_account(solidity_contract_instance, owner, test_accounts):
+    with test_accounts.use_sender(owner):
+        tx = solidity_contract_instance.setNumber(1)
+        assert tx.transaction.sender == owner.address
+    with pytest.raises(SignatureError):
+        solidity_contract_instance.setNumber(2)
+
+    with test_accounts.use_sender(owner.address):
+        tx = solidity_contract_instance.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with test_accounts.use_sender(owner.index):
+        tx = solidity_contract_instance.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+
+def test_default_sender_account(
+    solidity_contract_container, networks_connected_to_tester, accounts, keyfile_account
+):
+    owner = accounts[0]
+    passphrase = "a"
+
+    with accounts.use_sender(owner) as acct:
+        acct.set_autosign(True, passphrase)
+        contract = owner.deploy(solidity_contract_container, 0)
+
+    with accounts.use_sender(owner) as acct:
+        acct.set_autosign(True, passphrase)
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with pytest.raises(SignatureError):
+        contract.setNumber(2)
+
+    with accounts.use_sender(owner.address) as acct:
+        acct.set_autosign(True, passphrase)
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with accounts.use_sender(owner.alias) as acct:
+        acct.set_autosign(True, passphrase)
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with accounts.use_sender(0) as acct:
+        acct.set_autosign(True, passphrase)
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+
+def test_nested_default_sender(solidity_contract_instance, owner, test_accounts):
+    with test_accounts.use_sender(owner):
+        tx = solidity_contract_instance.setNumber(1)
+        assert tx.transaction.sender == owner.address
+        with test_accounts.use_sender(test_accounts[0]):
+            with reverts():
+                solidity_contract_instance.setNumber(2)
+        solidity_contract_instance.setNumber(3)
+        assert tx.transaction.sender == owner.address


docs/userguides/testing.md
```diff
@@ -304,6 +304,17 @@ with ape.reverts(dev_message=re.compile(r"^dev: (foo|bar)$")):
     ...
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `default_sender` context manager.
+
+```python
+def test_default_sender(solidity_contract_instance, owner):
+    with ape.default_sender(owner):
```
fubuloubu, line 313: `use_sender` might be clearer
```suggestion
    with ape.use_sender(owner):
```

src/ape/pytest/runners.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         self.receipt_capture = receipt_capture
         self._provider_is_connected = False
         ape.reverts = RevertsContextManager  # type: ignore
+        ape.default_sender = default_sender  # type: ignore
```
fubuloubu, line 26: Actually think this might be good in a general sense, not specific to pytest. just add to `ape/__init__.py`

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,19 @@
+from ape.pytest.contextmanagers import RevertsContextManager as reverts
+from ape.pytest.contextmanagers import default_sender
+
+
+def test_default_sender(solidity_contract_instance, owner):
```
fubuloubu, line 5: Have a test outside of the context manager that shows it fails "has no sender="

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,24 @@
...
+    with use_sender(owner):
+        tx = solidity_contract_instance.setNumber(1)
+        assert tx.transaction.sender == owner.address
+    try:
+        solidity_contract_instance.setNumber(2)
+    except Exception as e:
+        assert isinstance(e, SignatureError)  # No sender for transaction to sign
```
istankovic, line 13: Think `pytest.raises` would be more appropriate here.

src/ape/contextmanagers.py
```diff
@@ -0,0 +1,17 @@
+import contextlib
+from typing import List, Optional
+
+from ape_accounts.accounts import AccountAPI
+
+_DEFAULT_SENDERS: List[Optional[AccountAPI]] = []
+
+
+@contextlib.contextmanager
+def use_sender(address):
```
unparalleled-js, line 10: Thoughts on making this off the `AccountManager` and `TestAccountManager` such that the syntax could be like:

```python
# myscript.py
from ape import accounts

from .contract import my_contract  # Made up

def main():
    with account.use_sender("joebob"):  # Also allow address?
        my_contract.mutableMethod(123)
```

and:

```python
# test_mycontract.py
def test_mymutablemethod(accounts, my_contract):
    with account.use_sender(0):  # where 0 is the first test account
        my_contract.mutableMethod(123)
```

bilbeyt: that. address is actually directly an account. AccountAPI

unparalleled-js:
```python
with accounts.use_sender(account_api)
```

unparalleled-js (changes requested):
I think we need to work out the design.
I am still for the idea of not putting this in the global ape namespace and instead placing in the various account managers.

src/ape/contextmanagers.py
fubuloubu: I like using `accounts` as the way to use this context manager, we can make it be abstract such that you can use `Union[AccountAPI, AddressType, ENS, Alias]`:
1. use `AccountAPI` directly (like it does now), for both `AccountManager` and `TestAccountManager`
2. `accounts.load` if `AccountManager` and `Alias` (scripting usage)
3. `convert(address_or_ens, AddressType` then `accounts[address]` if anything else provided, for both `AccountManager` and `TestAccountManager`

Also could expand scripting usage with `@account_option` click decorator setting this up too, so that scripts don't even have to use `sender=` at all (not sure about this, I sort of like the explicitness a bit)

fubuloubu:
> I am still for the idea of not putting this in the global ape namespace and instead placing in the various account managers.

This makes a tons of sense

istankovic:
> I think we need to work out the design. I am still for the idea of not putting this in the global ape namespace and instead placing in the various account managers.

With the context manager residing on various account managers, would they share the same state or would each account manager maintain their own state?

src/ape/contextmanagers.py
bilbeyt: changed my implementation a bit @fubuloubu @unparalleled-js

unparalleled-js:
> With the context manager residing on various account managers, would they share the same state or would each account manager maintain their own state?

The normal account manager would only work with regular accounts and the test account manager would only work with test accounts, so the state would be different, at least that is how I was thinking of it.

docs/userguides/accounts.md
```diff
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
 account.balance += int(1e18)  # Gives `account` 1 Ether
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
+
+```python
+with test_accounts.use_sender(0):
```
unparalleled-js, line 65: `test_accounts` is the `accounts` fixture in the context of running tests.
It make more sense to show that:

```python
def test_example(accounts):
    with accounts.use_sender(0):  # First account from test mnemonic
        ...
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
 account = accounts.load("<ALIAS>")
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
+
+```python
```
unparalleled-js, line 181: I don't think we should include numeric example here.

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
...
+with accounts.use_sender(0, 'pass'):
```
unparalleled-js, line 182: password should be optional and should use a kwarg in the example

```python
with accounts.use_sender('address', passphrase="pass"):
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
...
+  contract.myFunction(1)
+
+with accounts.use_sender('address', 'pass'):
+  contract.myFunction(1)
+
+with accounts.use_sender('alias', 'pass'):
+  contract.myFunction(1)
+
+with accounts.use_sender(KeyfileAccount, 'pass'):
```
unparalleled-js, line 191: For consistency's sake, can we use double quoted strings on all the examples?

src/ape/contracts/base.py
```diff
@@ -14,6 +14,7 @@
 from ape.api.query import ContractEventQuery, extract_fields
 from ape.exceptions import ArgumentsLengthError, ChainError, ContractError, TransactionNotFoundError
 from ape.logging import logger
+from ape.managers.accounts import get_default_sender
```
unparalleled-js, line 17: I think the default sender should be accessible on the class.

```python
class AccountManager:
    @property
    def default_sender(self) -> AccountAPI:
        return self._default_senders[-1] if self._default_senders else None
```

src/ape/managers/accounts.py
```diff
@@ -100,6 +114,11 @@ def __contains__(self, address: AddressType) -> bool:  # type: ignore
     def generate_test_account(self, container_name: str = "test") -> TestAccountAPI:
         return self.containers[container_name].generate_account()
 
+    def use_sender(self, account):
```
unparalleled-js, line 117: - [ ] Missing types on inputs and outputs

src/ape/managers/accounts.py
```diff
@@ -317,3 +336,12 @@ def __contains__(self, address: AddressType) -> bool:
             any(address in container for container in self.containers.values())
             or address in self.test_accounts
         )
+
+    def use_sender(self, account, passphrase: Optional[str] = None):
+        if not isinstance(account, AccountAPI):
+            try:
+                account = self.load(account)  # if account is alias
+            except (ValueError, IndexError):
```
unparalleled-js, line 344: Instead of `try` / `except`, check if it is a hex via the `is_hex` method from `eth_utils`

src/ape/managers/accounts.py
```diff
@@ -317,3 +336,12 @@ def __contains__(self, address: AddressType) -> bool:
...
+                account = self[account]
+        account.set_autosign(True, passphrase)
```
unparalleled-js, line 346: I don't think autosign should automatically be enabled behind the user's back.
We could have a kwarg `autosign: bool` though.
Otherwise, it makes this method a lot more dangerous if the user is not aware of this.

unparalleled-js (changes requested):

src/ape/managers/accounts.py
unparalleled-js: Note: this is the main blocker for me right now!

docs/userguides/accounts.md
bilbeyt: but tests are not failing and if accounts is really accounts then __getitem_int wont also fallback to the test_accounts and this would even raise an error. couldnt get really

docs/userguides/accounts.md
unparalleled-js: The password is not being stored anywhere here; it is being passed in. In some situations this is required, like scripting with keyfile accounts.

That being said, setting the default sender and setting an account to `autosign=True` are two different things and maybe it does not make sense to combine them here.

src/ape/managers/accounts.py
bilbeyt: in this scenario account is alias so just string, i would do that if account was just address, on the other hand, account could be even int

docs/userguides/accounts.md
bilbeyt: so we dont use passphrase here? keyfile_account needs passphrase

bilbeyt:
@unparalleled-js @fubuloubu changed my implementation according to your comments, can you check again ?

docs/userguides/accounts.md
```diff
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
...
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
+
+```python
+with test_accounts.use_sender(0):
+  contract.myFunction(1)
+
+with test_accounts.use_sender("address"):
+  contract.myFunction(1)
+
+with test_accounts.use_sender(TestAccount):
```
fubuloubu, line 71: What does this one do?

docs/userguides/accounts.md
fubuloubu: we don't test the docs right now, so they wouldn't be failing

docs/userguides/accounts.md
```diff
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
 account.balance += int(1e18)  # Gives `account` 1 Ether
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
+
+```python
+with test_accounts.use_sender(0):
+  contract.myFunction(1)
```
fubuloubu, line 66: But yeah, I agree
```suggestion
def test_example(accounts):
    with accounts.use_sender(0):  # Use first account from test mnemonic
        contract.myFunction(1)
```

docs/userguides/accounts.md
```diff
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
...
+
+with test_accounts.use_sender("address"):
+  contract.myFunction(1)
```
fubuloubu, line 69:
```suggestion
def test_example(accounts):
    with accounts.use_sender("<address>"):  # Impersonate an account
        contract.myFunction(1)
```

docs/userguides/accounts.md
fubuloubu: Or do you mean `<TestAccount>`?

docs/userguides/accounts.md
```diff
@@ -57,6 +57,21 @@ If your testing provider supports this feature, it is possible to directly set t
...
+
+with test_accounts.use_sender(TestAccount):
+  contract.myFunction(1)
```
fubuloubu, line 72: This might be clearer:
```suggestion
def test_example(accounts):
    with accounts.use_sender(a):  # a is a `TestAccount` class
        contract.myFunction(1)
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
 account = accounts.load("<ALIAS>")
 ```
 
+### Default Sender Support
+
+In order to eliminate the usage of sender in contract calls, you can use `use_sender` context manager.
```
fubuloubu, line 179:
```suggestion
In order to reduce repetition of adding `sender` in your contract calls, you can use `use_sender` context manager.
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
...
+  contract.myFunction(1)
+
+with accounts.use_sender("address", autosign=True, passphrase="pass"):
+  contract.myFunction(1)
+
+with accounts.use_sender("alias", autosign=True, passphrase="pass"):
+  contract.myFunction(1)
+
+with accounts.use_sender(KeyfileAccount, autosign=True, passphrase="pass"):
+  contract.myFunction(1)
```
fubuloubu, line 192: The reason I don't like `passphrase=` kwarg here is that it encourages people to just write their passphrase into a script file they very likely will upload to a repo

If they also upload their keyfile somehow, the funds are as good as gone

I like `unlock=True` better because it just uses the native unlock facility of `KeyfileAccount`

fubuloubu: Also, how would this work for other account types? (like ledger/trezor)

src/ape/contracts/base.py
```diff
@@ -333,6 +334,8 @@ def call(self) -> ContractCallHandler:
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
         function_arguments = self._convert_tuple(args)
         contract_transaction = self._as_transaction(*function_arguments)
+        if "sender" not in kwargs and ape.accounts.default_sender is not None:
```
fubuloubu, line 337: Because this class subclasses `ManagerAccessMixin`, we don't need to import `ape` (which causes a cycle)
```suggestion
        if "sender" not in kwargs and self.account_manager.default_sender is not None:
```

src/ape/contracts/base.py
```diff
@@ -333,6 +334,8 @@ def call(self) -> ContractCallHandler:
...
+            kwargs["sender"] = ape.accounts.default_sender
```
fubuloubu, line 338:
```suggestion
            kwargs["sender"] = self.account_manager.accounts.default_sender
```

src/ape/managers/accounts.py
```diff
@@ -9,9 +12,19 @@
 )
 from ape.types import AddressType
 from ape.utils import ManagerAccessMixin, cached_property, singledispatchmethod
+from ape_accounts.accounts import KeyfileAccount
```
fubuloubu, line 15: This might cause a cycle

src/ape/managers/accounts.py
```diff
@@ -317,3 +341,20 @@ def __contains__(self, address: AddressType) -> bool:
             any(address in container for container in self.containers.values())
             or address in self.test_accounts
         )
+
+    def use_sender(
+        self,
+        account_id: Union[AccountAPI, AddressType, str, int],
+        autosign: bool,
+        passphrase: Optional[str] = None,
```
fubuloubu, line 349: This might be a little devious, but hear me out:
```suggestion
       **kwargs,
```

src/ape/managers/accounts.py
```diff
@@ -317,3 +341,20 @@ def __contains__(self, address: AddressType) -> bool:
...
+    ) -> ContextManager:
+        if not isinstance(account_id, AccountAPI):
+            if isinstance(account_id, int) or is_hex(account_id):
+                account = self[account_id]
+            elif isinstance(account_id, str):  # alias
+                account = self.load(account_id)
+        else:
+            account = account_id
+        if isinstance(account, KeyfileAccount):
+            account.set_autosign(autosign, passphrase)
```
fubuloubu, line 359: And then something like this:
```suggestion
        for fn_name, args in kwargs.items():
            getattr(account, fn_name)(*args)
```

fubuloubu: This will make it more extensible over any account type! (also we won't need to import from `ape_accounts` anymore)

fubuloubu: Could also do it this way:
```suggestion
       for fn_name, fn_args in kwargs.items():
            if isinstance(fn_args, (tuple, list)):
                getattr(account, fn_name)(*fn_args)
            elif isinstance(fn_args, dict):
                getattr(account, fn_name)(**fn_args)
            else:
                raise  # can't handle `type(fn_args)`
```

istankovic: Yeah, from the point of view of the user of the API, this is strictly worse. And being that ape documentation is nothing to write home about, losing the well-named arguments is just ugh. Not to mention the type information that is lost.

bilbeyt: I agree with @istankovic, clear kwargs are better than unknown things.

docs/userguides/accounts.md
bilbeyt: they dont even have unlock. i just looked at TrezorAccount and LedgerAccount implementations, so this is a problem like i wrote in discord

bilbeyt: i also dont think it would be good idea to eliminate passphrase here and just allow autosign without it, then if you get hacked then it is not fun :D

docs/userguides/accounts.md
unparalleled-js: Right, but for highlighting best practice, all I meant was this:

```python
import keyring

passphrase = keyring.get_password("ape-scripts", "default-user")
with accounts.use_sender('address', passphrase=password):
```

or even

```python
with accounts.use_sender('address', passphrase=os.environ.get("KEYFILE_PASSPHRASE")):
```

because hardcoded passphrases looks bad from a security perspective, even in documentation.

This is not a blocker for me though, I am only explaining the thought process.

bilbeyt: ok, got. it

src/ape/managers/accounts.py
bilbeyt: I think if that is the case, then the implementer of the plugin should handle the change in this part adding another kwarg if it is needed. As i am not using in the code right now, if i add the kwargs, i need to suppress it also as it wont be used.

fubuloubu: I am okay with just sticking to `autosign=True` and `unlock=True`, I am still very much against `passphrase="my super secure passphrase that I accidentally commited to github then tried to scrub but it was far too late and now I've lost all my monies"`

fubuloubu: btw, I think `autosign` sets `unlock=True`, so you might only need to do `set_autosign` if both are present

bilbeyt: yes, that is why i used just autosign

src/ape/managers/accounts.py
```diff
@@ -317,3 +342,20 @@ def __contains__(self, address: AddressType) -> bool:
             any(address in container for container in self.containers.values())
             or address in self.test_accounts
         )
+
+    def use_sender(
+        self,
+        account_id: Union[AccountAPI, AddressType, str, int],
+        autosign: bool,
+        passphrase: Optional[str] = None,
```
fubuloubu, line 350:
```suggestion
```

src/ape/managers/accounts.py
```diff
@@ -317,3 +342,20 @@ def __contains__(self, address: AddressType) -> bool:
...
+    ) -> ContextManager:
+        if not isinstance(account_id, AccountAPI):
+            if isinstance(account_id, int) or is_hex(account_id):
+                account = self[account_id]
+            elif isinstance(account_id, str):  # alias
+                account = self.load(account_id)
+        else:
+            account = account_id
+        if type(account).__name__ == "KeyfileAccount":
+            account.set_autosign(autosign, passphrase)  # type: ignore
```
fubuloubu, line 360:
```suggestion
```

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_default_sender_account(
+    solidity_contract_container, networks_connected_to_tester, accounts, keyfile_account
+):
+
+    owner = accounts[0]
+    passphrase = "a"
+    owner.set_autosign(True, passphrase)
+    contract = owner.deploy(solidity_contract_container, 0)
+    with accounts.use_sender(owner, autosign=True, passphrase=passphrase):
```
fubuloubu, line 31:
```suggestion
    with accounts.use_sender(owner):
```

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+    with pytest.raises(SignatureError):
+        contract.setNumber(2)
+
+    with accounts.use_sender(owner.address, autosign=True, passphrase=passphrase):
```
fubuloubu, line 37:
```suggestion
    with accounts.use_sender(owner.address):
```

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with accounts.use_sender(owner.alias, autosign=True, passphrase=passphrase):
```
fubuloubu, line 41:
```suggestion
    with accounts.use_sender(owner.alias):
```

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        tx = contract.setNumber(1)
+        assert tx.transaction.sender == owner.address
+
+    with accounts.use_sender(0, autosign=True, passphrase=passphrase):
```
fubuloubu, line 45:
```suggestion
    with accounts.use_sender(0):
```

fubuloubu (changes requested):

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
 account = accounts.load("<ALIAS>")
 ```
 
+### Default Sender Support
+
+In order to reduce repetition of adding `sender` in your contract calls, you can use `use_sender` context manager.
+
+```python
+with accounts.use_sender(0, autosign=True, passphrase="<PASSPHRASE>"):
```
fubuloubu, line 182:
```suggestion
with accounts.use_sender(0):
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
...
+  contract.myFunction(1)
+
+with accounts.use_sender("<address>", autosign=True, passphrase="<PASSPHRASE>"):
```
fubuloubu, line 185:
```suggestion
with accounts.use_sender("<address>"):
```

docs/userguides/accounts.md
```diff
@@ -159,6 +174,24 @@ from ape import accounts
...
+  contract.myFunction(1)
+
+with accounts.use_sender("<alias>", autosign=True, passphrase="<PASSPHRASE>"):
```
fubuloubu, line 188:
```suggestion
with accounts.use_sender("<alias>"):
```

fubuloubu (changes requested):

docs/userguides/accounts.md
fubuloubu: > i also dont think it would be good idea to eliminate passphrase here and just allow autosign without it, then if you get hacked then it is not fun :D

Due to how `KeyfileAccount` works, it can't sign anything without unlocking first because you need the passphrase to decrypt the key

fubuloubu (dismissed):

tests/functional/test_default_sender_context_manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_default_sender_account(
+    solidity_contract_container, networks_connected_to_tester, accounts, keyfile_account
+):
+
+    owner = accounts[0]
+    passphrase = "a"
+    owner.set_autosign(True, passphrase)
+    contract = owner.deploy(solidity_contract_container, 0)
+    with accounts.use_sender(owner) as acct:
```
fubuloubu, line 31:
```suggestion
    with accounts.use_sender(owner) as acct:
        contract = solidity_contract_container.deploy(0)  # Deploys using acct
```

fubuloubu (dismissed):

fubuloubu (approved):

Issue #661, opened 2022-04-16
Dict Converter


fubuloubu:
### Overview

Sometimes contract functions accept structs as inputs, however it is non-trivial to work with structs as they are basically ABI-encoded to tuples. It would be nicer to work with dict-like types (maybe `typing.Mapping`?) and have the ABI conversion happen under the hood

### Specification

Example Vyper code:
```vyper
struct Params:
    a: bool
    b: bool 

@external
def f(p: Params): 
    pass
```

produces this ABI:
```json
{
  "stateMutability": "nonpayable",
  "type": "function",
  "name": "f",
  "inputs": [
    {
      "name": "p",
      "type": "tuple",
      "components": [
        {"name": "a", "type": "bool"},
        {"name": "b", "type": "bool"}
      ]
    }
  ],
  "outputs": []
}
```

If we call in ape like this:

```py
c.f({"a": True, "b": False}, sender=a)
```

under the hood we should be able to do a lookup of the ABI to see that it is a "named tuple" (aka a struct), and convert the dictionary input into the corresponding tuple type *in the correct order* that the ABI tuple requests it.

Could perhaps add this as a built-in conversion function via the conversion manager

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #1230, opened 2023-01-17
feat: encode struct-like arguments as tuples for ape-ethereum

fubuloubu:
### What I did

Allows any struct-like object (either dict or accessor via getattr) to be used whenever a struct input is required, instead of requiring a conversion to a tuple

fixes: #661

fixes: APE-464
### How I did it

I put it into ape-ethereum only because this is a bit more specific to how Ethereum's ABI encoding works. Wasn't sure if Starknet would allow the same

### How to verify it

Use it in a project with eip712 structs passed as an arg maybe

### Checklist


tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
+from ethpm_types.abi import MethodABI
+from hexbytes import HexBytes
+from pydantic import BaseModel
+
+from ape import networks
+
+ABI = MethodABI.parse_obj(
```
unparalleled-js, line 7: Does this align with both vyper and solidity ABI outputs?

tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
...
+    "0000000000000000000000000000000000000000000000000000000000000001"
+    "3200000000000000000000000000000000000000000000000000000000000000"
+    "0000000000000000000000000000000000000000000000000000000000000001"
+)
+
+
+def test_encode_structs():
+    encode_calldata = networks.ethereum.encode_calldata
+
+    assert EXPECTED == encode_calldata(ABI, (1, b"2", True))
```
unparalleled-js, line 42: the "actual" goes first I believe, for some pytest output to make sense.

```suggestion
    assert encode_calldata(ABI, (1, b"2", True)) == EXPECTED
```

tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
...
+    assert EXPECTED == encode_calldata(ABI, {"a": 1, "b": b"2", "c": True})
```
unparalleled-js, line 43:
```suggestion
    assert encode_calldata(ABI, {"a": 1, "b": b"2", "c": True}) == EXPECTED
```

tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
...
+    assert EXPECTED == encode_calldata(ABI, Struct(a=1, b=b"2", c=True))
```
unparalleled-js, line 44:
```suggestion
    assert encode_calldata(ABI, Struct(a=1, b=b"2", c=True)) == EXPECTED
```

tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
...
+                    {"name": "b", "type": "bytes32"},
+                    {"name": "c", "type": "bool"},
+                ],
+            }
+        ],
+    }
+)
+
+
+class Struct(BaseModel):
```
unparalleled-js, line 26: Ethereum does not use `https://github.com/ApeWorX/ethpm-types/blob/main/ethpm_types/abi.py#L394` does it?

fubuloubu: No, not yet

unparalleled-js (approved):

unparalleled-js (changes requested):
One thing is missing!

During testing, I noticed the conversion is failing. I am unable to use `AccountAPI` as an input in a dictionary for an input.

tests/functional/conversion/test_encode_structs.py
```diff
@@ -0,0 +1,44 @@
...
+    a: int
```
unparalleled-js, line 27: can we make one of these an `AddressType` and include a test that passes a test account an input to ensure conversion works?

src/ape_ethereum/ecosystem.py
```diff
@@ -320,6 +320,23 @@ def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> HexBy
...
+                and not isinstance(arg, tuple)
+            ):
+                if isinstance(arg, dict):
+                    args[idx] = tuple([arg[m.name] for m in ipt.components])  # type: ignore[index]
+
+                else:
+                    arg = [getattr(arg, m.name) for m in ipt.components]  # type: ignore[arg-type]
+                    args[idx] = tuple(arg)  # type: ignore[index]
+
         input_types = [i.canonical_type for i in abi.inputs]
```
unparalleled-js, line 340: do we also need to handle nested / recursive structures?

fubuloubu: I think `.canonical_type` is recursive

fubuloubu: https://github.com/ApeWorX/ethpm-types/blob/main/ethpm_types/abi.py#L58

unparalleled-js: I moreso meant struture values as inputs, like nested dicts with account objects in the middle:

```
account = accounts[0]
argument = {"foo": {"bar": account}}
contract.my_method(arguments, sender=account)
```

src/ape_ethereum/ecosystem.py
```diff
@@ -337,8 +338,39 @@ def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> HexBy
...
+                    arguments[idx] = tuple([arg[m.name] for m in ipt.components])
+
+                else:
+                    arg = [getattr(arg, m.name) for m in ipt.components if m.name]
+                    arguments[idx] = tuple(arg)
+
         input_types = [i.canonical_type for i in abi.inputs]
-        encoded_calldata = encode(input_types, args)
+
+        def convert(ls):
```
unparalleled-js, line 360: idk, is there a better way to do this @fubuloubu ?
I tried using the tuple converter but it attempts to convert too much.

unparalleled-js (approved):
Approving, but waiting until at least @fubuloubu looks at the changes I added

fubuloubu (commented):
Looking good!

PR #1232, opened 2023-01-18
docs: fix `gas_limit` example in config guide

milkyklim:
Update docs. Current setup throws:

```sh
pydantic.error_wrappers.ValidationError: 1 validation error for EthereumConfig
gas_limit
  extra fields not permitted (type=value_error.extra)
```


unparalleled-js (approved):
yep, good catch thank you.

PR #1238, opened 2023-01-20
refactor: return default config when network not found

fubuloubu:
### What I did

Not sure if this is the right solution, but it matches what happens in the other case when the ecosystem is not found

Currently gives a rather nasty looking AttributeError that doesn't tell you anything useful

fixes: #

### Checklist

### DIFF
--- a/src/ape_geth/provider.py
+++ b/src/ape_geth/provider.py
@@ -205,6 +205,9 @@ class BaseGethProvider(Web3Provider, ABC):
 
         # Use value from config file
         network_config = config.get(self.network.name)
+        if not network_config:
+            return DEFAULT_SETTINGS["uri"]
+            
         return network_config.get("uri", DEFAULT_SETTINGS["uri"])
 
     @property


src/ape_geth/provider.py
```diff
@@ -205,6 +205,9 @@ def uri(self) -> str:
 
         # Use value from config file
         network_config = config.get(self.network.name)
+        if not network_config:
```
unparalleled-js, line 208: You can combine all of this like this:

```python
network_config = config.get(self.network.name) or DEFAULT_SETTINGS["uri"]
```

unparalleled-js (approved):
Good catch.
One optional comment!

unparalleled-js:
Ha I just ran into this while working with Optimism.

src/ape_geth/provider.py
```diff
@@ -205,6 +205,9 @@ def uri(self) -> str:
...
+            return DEFAULT_SETTINGS["uri"]
+            
         return network_config.get("uri", DEFAULT_SETTINGS["uri"])
```
fubuloubu, line 211:
```suggestion
        network_config = config.get(self.network.name) or DEFAULT_SETTINGS
        return network_config.get("uri", DEFAULT_SETTINGS["uri"])
```

unparalleled-js (approved):

PR #1239, opened 2023-01-20
feat: create abi files per contract

bilbeyt:
### What I did

Creating multiple abi files for compile

fixes: APE-479

### How I did it

Changed ape.managers.project.types.create_manifest to have seperate files.

### How to verify it

Just run compile command and /abis folder should have compiled contracts as seperate.

### Checklist

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -91,7 +91,23 @@ class ProjectAPI(BaseInterfaceModel):
         if it exists and is valid.
         """
 
-        return _load_manifest_from_file(self.manifest_cachefile)
+        manifest = _load_manifest_from_file(self.manifest_cachefile)
+        if manifest is not None:
+            manifest.contract_types = self.contracts
+        return manifest
+
+    @property
+    def contracts(self) -> Dict[str, ContractType]:
+        contracts = {}
+        for p in self._cache_folder.glob("*.json"):
+            if p == self.manifest_cachefile:
+                continue
+            contract_name = p.name.replace(".json", "")
+            contract_type = ContractType().parse_file(p)
+            if contract_type.name is None:
+                contract_type.name = contract_name
+            contracts[contract_type.name] = contract_type
+        return contracts
 
     @property
     def _cache_folder(self) -> Path:
--- a/src/ape/managers/project/types.py
+++ b/src/ape/managers/project/types.py
@@ -196,6 +196,11 @@ class BaseProject(ProjectAPI):
             )
             # Cache the updated manifest so `self.cached_manifest` reads it next time
             self.manifest_cachefile.write_text(manifest.json())
+            if compiled_contract_types:
+                for name, contract_type in compiled_contract_types.items():
+                    (self.project_manager.local_project._cache_folder / f"{name}.json").write_text(
+                        contract_type.json()
+                    )
             return manifest
 
     def _compile(self, project_sources: _ProjectSources) -> Dict[str, ContractType]:


fubuloubu (commented):
In general, this was something we were looking to do in a much more comprehensive way

unparalleled-js (changes requested):
As @fubuloubu mentioned, we have discussed doing this in a way such that we are not double-caching here. Instead, the local project manifest would get composed from the contract type artifacts during runtime as needed.

Can you describe your use-case more? Maybe there is a short-term solution we can offer. Else, we can try to work out the design now and help you implement the rest of it.

I still left other review comments anyway!

fubuloubu:
> As @fubuloubu mentioned, we have discussed doing this in a way such that we are not double-caching here. Instead, the local project manifest would get composed from the contract type artifacts during runtime as needed.
> 
> Can you describe your use-case more? Maybe there is a short-term solution we can offer. Else, we can try to work out the design now and help you implement the rest of it.
> 
> I still left other review comments anyway!

I think a lot of alternative tools to Ape (some in other languages) cache build files by contract type, instead of by project. So to integrate with them it would be easier if we did the same (which isn't really required for us until publishing time for the local package)

bilbeyt:
> As @fubuloubu mentioned, we have discussed doing this in a way such that we are not double-caching here. Instead, the local project manifest would get composed from the contract type artifacts during runtime as needed.
> 
> Can you describe your use-case more? Maybe there is a short-term solution we can offer. Else, we can try to work out the design now and help you implement the rest of it.
> 
> I still left other review comments anyway!

So, we are using brownie and web3 heavily in our code. Normally, brownie is providing contract based compile files when you simply run `brownie compile` and it is in the form that you can easily crease web3.Contract instances etc so list of dicts directly. We would like to use this in ape also, those compiled files also used in many other areas like in frontend and other parts of the project.

fubuloubu:
> So, we are using brownie and web3 heavily in our code.

What Brownie features is Ape missing that we can include in our roadmap for Brownie equivalency (besides coverage)

bilbeyt:
> > So, we are using brownie and web3 heavily in our code.
> 
> What Brownie features is Ape missing that we can include in our roadmap for Brownie equivalency (besides coverage)

In tests, we cant use easy usage of 'EventName' in tx.events kind of syntax with ape it is kind of 

```
assert next(contract.Event.from_receipt(tx), None) is not None
```

Gas calculation problem also, all tests with eth.get_balance failing now.

Transactions has no timestamp, and i need to use below to create web3py Contract Instance with address and abi

```
[abi.dict() for abi in contract.contract_type.abi]
```

and for tx timestamp max i can use is

```
ape.chain.blocks[tx.block_number]
```

fubuloubu:
> In tests, we cant use easy usage of 'EventName' in tx.events kind of syntax with ape it is kind of
> 
> ```
> assert next(contract.Event.from_receipt(tx), None) is not None
> ```

I am in favor of refactoring `from_receipt` to return a list here, since it should be a defined list (less than 100s of events) and it doesn't need additional data to fetch (the receipt contains all the raw event data)

Maybe can add this to 0.6.0 breaking changes?

> Gas calculation problem also, all tests with eth.get_balance failing now.

Is there an issue open for this? We should solve any issues with this

> Transactions has no timestamp, and i need to use below to create web3py Contract Instance with address and abi
> 
> ```
> [abi.dict() for abi in contract.contract_type.abi]
> ```

You want to be able to pull out a `Web3py` contract instance from an Ape contract?

> and for tx timestamp max i can use is
> 
> ```
> ape.chain.blocks[tx.block_number]
> ```

Yeah, I've wanted to add `ReceiptAPI.timestamp` that's a computed property to fetch the timestamp from the corresponding block, I don't think we have tracking for this, but this is basically what you need?

unparalleled-js:
> > In tests, we cant use easy usage of 'EventName' in tx.events kind of syntax with ape it is kind of
> > ```
> > assert next(contract.Event.from_receipt(tx), None) is not None
> > ```
> 
> I am in favor of refactoring `from_receipt` to return a list here, since it should be a defined list (less than 100s of events) and it doesn't need additional data to fetch (the receipt contains all the raw event data)
> 
> Maybe can add this to 0.6.0 breaking changes?

Yes, this is a grand idea. It should have been a list all along on the Receipt.
Everything else too, but right now focused on this first bit as we are moving towards 0.6.

bilbeyt:
> > In tests, we cant use easy usage of 'EventName' in tx.events kind of syntax with ape it is kind of
> > ```
> > assert next(contract.Event.from_receipt(tx), None) is not None
> > ```
> 
> I am in favor of refactoring `from_receipt` to return a list here, since it should be a defined list (less than 100s of events) and it doesn't need additional data to fetch (the receipt contains all the raw event data)
> 
> Maybe can add this to 0.6.0 breaking changes?
> 
> > Gas calculation problem also, all tests with eth.get_balance failing now.
> 
> Is there an issue open for this? We should solve any issues with this
> 
> > Transactions has no timestamp, and i need to use below to create web3py Contract Instance with address and abi
> > ```
> > [abi.dict() for abi in contract.contract_type.abi]
> > ```
> 
> You want to be able to pull out a `Web3py` contract instance from an Ape contract?
> 
> > and for tx timestamp max i can use is
> > ```
> > ape.chain.blocks[tx.block_number]
> > ```
> 
> Yeah, I've wanted to add `ReceiptAPI.timestamp` that's a computed property to fetch the timestamp from the corresponding block, I don't think we have tracking for this, but this is basically what you need?

There is no issue regarding gas calculation or so only what i wrote to discord. And web3py contract part seems good to me.

fubuloubu:
> > > In tests, we cant use easy usage of 'EventName' in tx.events kind of syntax with ape it is kind of
> > > ```
> > > assert next(contract.Event.from_receipt(tx), None) is not None
> > > ```
> > 
> > 
> > I am in favor of refactoring `from_receipt` to return a list here, since it should be a defined list (less than 100s of events) and it doesn't need additional data to fetch (the receipt contains all the raw event data)
> > Maybe can add this to 0.6.0 breaking changes?
> 
> Yes, this is a grand idea. It should have been a list all along on the Receipt. Everything else too, but right now focused on this first bit as we are moving towards 0.6.

Yeah, a lot to do for 0.6.0 but this refactor to a list can be something we can do for 0.6.0 pretty easily, turns this:
```py
assert next(contract.Event.from_receipt(tx), None) is not None
```
into:
```py
assert len(contract.Event.from_receipt(tx)) == 0
```
which is definitely more intuitive, and can also maybe do stuff like this (eventually):
```py
assert contract.Event.from_receipt(tx) == [contract.Event(1, 2, 3)]
```

unparalleled-js:
https://github.com/ApeWorX/ape/pull/1244

bilbeyt:
@fubuloubu @unparalleled-js are we merging this? there is no comment left so this is ready imo.

fubuloubu:
> @fubuloubu @unparalleled-js are we merging this? there is no comment left so this is ready imo.

I think this is something we should work on post-0.6.0 release

fubuloubu:
> > Yeah, I've wanted to add `ReceiptAPI.timestamp` that's a computed property to fetch the timestamp from the corresponding block, I don't think we have tracking for this, but this is basically what you need?
> 
> Yes, that would be a nice improvement in API ergonomics.

Tracking this here: https://github.com/ApeWorX/ape/issues/1257

unparalleled-js:
> @fubuloubu @unparalleled-js are we merging this? there is no comment left so this is ready imo.

Yes, some refactoring is needed! And I think there are some unknowns yet as well, regarding what exactly we want. We won't forget it, and I am thinking the design now.

bilbeyt:
> > @fubuloubu @unparalleled-js are we merging this? there is no comment left so this is ready imo.
> 
> Yes, some refactoring is needed! And I think there are some unknowns yet as well, regarding what exactly we want. We won't forget it, and I am thinking the design now.

Whenever it is ready i can also work on it as we need this in shorter time span.

unparalleled-js:
Having discussed some with @fubuloubu , @bilbeyt this is the spec we would like to go with:

1. Create each `ContractType` in the project's `.build/` folder like: `.build/{ContractType.name}.json`.
2. Don't include the contact types in the local `.build/manifest.json` - that will be just for the sources.
Step 2 will require altering how `__getattr__` works in `ProjectManager` so that `project.MyContract` still works, which may be the challenging part here.

src/ape/managers/project/types.py
```diff
@@ -170,7 +173,7 @@ def _as_ape_project(self, **kwargs):
 
     def create_manifest(
         self, file_paths: Optional[List[Path]] = None, use_cache: bool = True
-    ) -> PackageManifest:
+    ) -> Tuple[PackageManifest, Dict[str, ContractType]]:
```
unparalleled-js, line 176: `contract_types` is a property on the `PackageManifest`, so could we set the contract types property on the manifest and only return the manifest?

src/ape/api/projects.py
```diff
@@ -84,6 +84,16 @@ def manifest_cachefile(self) -> Path:
         file_name = self.name or "__local__"
         return self._cache_folder / f"{file_name}.json"
 
+    @property
+    def contract_types(self) -> Dict[str, ContractType]:
```
unparalleled-js, line 88: Maybe we should call this `contracts()` to echo the same in `ProjectManager`

src/ape/managers/project/types.py
unparalleled-js: Can we make the cached manifest also grab the type JSONs? Like can we have it be excluded from the cached file but not excluding from the Python object?

Thinking out-loud here. I really appreciate you working on this.

fubuloubu: This might look like a local project implementing this method only after all the other files are created and are consistent with each other (as a result of an intermediate pass we use for compiling a local project, likely many times in a row), compiling the manifest to extract only when necessary for the publishing step

src/ape/api/projects.py
```diff
@@ -91,7 +91,20 @@ def cached_manifest(self) -> Optional[PackageManifest]:
...
+
+    @property
+    def contracts(self) -> Dict[str, ContractType]:
+        contracts = {}
+        for p in self._cache_folder.glob("*.json"):
+            if p == self.manifest_cachefile:
+                continue
+            contract_name = p.name.replace(".json", "")
+            contracts[contract_name] = ContractType().parse_file(p)
+        return contracts
```
fubuloubu, line 107: This part looks solid

src/ape/managers/project/types.py
```diff
@@ -194,8 +194,15 @@ def create_manifest(
                 name=self.name,
                 version=self.version,
             )
+            del manifest.contract_types
```
fubuloubu, line 197: I think [this function](https://github.com/ApeWorX/ape/pull/1239/files#diff-e56910138250dd5c83fc082fb33e1acd230a5f00b386aab6405099d768b1c09bR171-R173) likely needs to be rewritten 

Ultimately, what we want is for dependencies to be fetched/compiled as manifests and disassembled as needed for individual files (for imports), and the local project to compile contract types as individual files under the contract types folder, then assemble the manifest if/when you publish the manifest (which the framework doesn't support... _yet_)

bilbeyt: will you do the rewrite and accept this one ?

fubuloubu: `ApeProject.create_manifest` should basically do this (for a local dev project):
1. fetch sources from `contracts/`
2. fetch checksum information, etc. from `.build/__local__.json`
3. fetch contract types from `.build/<ContractTypeName>.json`
4. fetch metadata from `ape-config.yaml`
5. combine into `PackageManifest` type

bilbeyt: thanks for steps @fubuloubu but to me when i look into the code, this is already there, so i am more confused right now.

istankovic: @fubuloubu are the above changes really necessary for @bilbeyt's PR to be merged?

From what I can tell, his PR fixes a real issue for us and would allow us to proceed with our move to ape.
While we definitely want to help improve ape and have already demonstrated that, we do not have unlimited time to do so and have to balance our priorities.

Apart from this issue, there's still a big blocker for our move, namely the pretty bad performance compared to
what we have with brownie. In that light, I would like to ask you to reconsider whether requiring additional
work here is really reasonable.

src/ape/managers/project/types.py
```diff
@@ -194,8 +194,15 @@ def create_manifest(
...
             # Cache the updated manifest so `self.cached_manifest` reads it next time
             self.manifest_cachefile.write_text(manifest.json())
+            manifest.contract_types = contract_types
+            if manifest.contract_types:
+                for name, contract_type in manifest.contract_types.items():
+                    (self.project_manager.local_project._cache_folder / f"{name}.json").write_text(
```
unparalleled-js, line 203: What happens if the file already exists here?

src/ape/managers/project/types.py
unparalleled-js, line 197: Do we need this `del` statement? Does it hurt if there are contracts?
The point is that we don't want to compile at this stage, but they are already compiled, it should be harmless to include them.

unparalleled-js: If you address my small comments below, we can merge this and I can handle refactors later.

src/ape/managers/project/types.py
bilbeyt: now only newly. compiled files would be written

unparalleled-js (dismissed):

unparalleled-js (approved):

unparalleled-js:
- [ ] Approved but let me finish testing with `ape-cairo` to make sure it is not broken

src/ape/managers/project/types.py
fubuloubu: Updates the cached copy, which is what we want (driving updates based on compilation changes)

src/ape/api/projects.py
```diff
@@ -91,7 +91,23 @@ def cached_manifest(self) -> Optional[PackageManifest]:
         if it exists and is valid.
         """
 
-        return _load_manifest_from_file(self.manifest_cachefile)
+        manifest = _load_manifest_from_file(self.manifest_cachefile)
+        if manifest is not None:
```
fubuloubu, line 95: Hmm, what is the implications of this?

src/ape/api/projects.py
```diff
@@ -91,7 +91,23 @@ def cached_manifest(self) -> Optional[PackageManifest]:
...
+            manifest.contract_types = self.contracts
+        return manifest
+
+    @property
+    def contracts(self) -> Dict[str, ContractType]:
+        contracts = {}
+        for p in self._cache_folder.glob("*.json"):
+            if p == self.manifest_cachefile:
+                continue
+            contract_name = p.name.replace(".json", "")
```
fubuloubu, line 105:
```suggestion
            contract_name = p.stem
```

fubuloubu (approved):
What's the worst case scenario here? Projects do an extra full recompile, even if they had the old caching style? I think that's acceptable, just want to make sure people thought how this impacts previous projects using the old setup when they upgrade

src/ape/managers/project/types.py
fubuloubu: @istankovic I was talking a bit more out loud to describe the overall vision, we have this tracked elsewhere. @bilbeyt's change is basically good, just wanted to make sure he can see what the design was supposed to fit in to

unparalleled-js (changes requested):
When I delete the JSON files manually, they don't come back when I run `ape compile` and the `__getattr__` fails. 

Currently, re-compiling is only triggered when the `__local__.json` file is missing.

unparalleled-js (approved):
Willing to ignore my last review and handle it later myself.

src/ape/api/projects.py
unparalleled-js: What exactly do you mean here? What are you thinking?

The cached manifest is optional because it does not always exist.

src/ape/api/projects.py
unparalleled-js: https://github.com/ApeWorX/ape/pull/1291

src/ape/api/projects.py
unparalleled-js: Add thoughts here: https://github.com/ApeWorX/ape/pull/1291

PR #1242, opened 2023-01-23
feat: utilize ImpersonatedAccount for contract transactions

bilbeyt:
### What I did

Implemented support for ImpersonatedAccount to send transaction if account is actually a contract.

fixes: APE-488

### Checklist


src/ape/contracts/base.py
```diff
@@ -263,16 +263,16 @@ def __repr__(self) -> str:
...
-            # Automatically impersonate contracts (if API available) when sender
-            kwargs["sender"] = self.account_manager.test_accounts[kwargs["sender"].address]
-
         kwargs = _convert_kwargs(kwargs, self.conversion_manager.convert)
         return self.provider.network.ecosystem.encode_transaction(
             self.address, self.abi, *args, **kwargs
         )
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        if "sender" in kwargs and isinstance(kwargs["sender"], ContractInstance):
```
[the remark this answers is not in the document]
bilbeyt: Can i connect to mainnet then randomly select one account that at the end account_manager returns ImpersonatedAccount then i interact with a contract lets say open contract without owner protection can i just contract.myMethod(x=1) with this then? so i dont have any idea about tolga.eth and dont have any private key or so but i am trying to send a transaction like i am tolga.eth, kind of security problem to me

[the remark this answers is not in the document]
bilbeyt: Got it now what you mean. It is ok.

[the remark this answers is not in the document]
unparalleled-js: `BaseAddress` is a shared base class between `ContractInstance` and `Address`, so you could do

```python
isinstance(kwargs["sender"], BaseAddress)
```

src/ape/api/providers.py
```diff
@@ -1046,7 +1047,13 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
 
     def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         try:
-            txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            account = self.account_manager[txn.sender]
+            is_contract = self.web3.eth.get_code(txn.sender) != "0x"
```
unparalleled-js, line 1051: `web3.eth.get_code()` returns `HexBytes`, and therefore it will never be equal to `"0x"` (the str).
Using `is_contract` as @fubuloubu said would fix that thought!

src/ape/contracts/base.py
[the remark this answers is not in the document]
fubuloubu: But so is `AccountAPI`, which is why I wanted to not use `BaseAddress`

src/ape/api/providers.py
```diff
@@ -1046,7 +1047,12 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
 
     def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         try:
-            txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            account = self.account_manager[txn.sender]
+            if isinstance(account, ImpersonatedAccount) and account.is_contract:
```
fubuloubu, line 1051: Broader question, why should we only take this path if the impersonated account is a contract?

unparalleled-js, line 1051: If it is helper, we did this already in `ape-hardhat`: https://github.com/ApeWorX/ape-hardhat/blob/main/ape_hardhat/provider.py#L398-L415

Notice something that is missing the exception handling for revert-detection.

Perhaps we could remove the `ape-hardhat` custom implementation after this.

fubuloubu: Yeah, seems like a better way to handle it

bilbeyt: @unparalleled-js there is no problem regarding revert-detection here, if you see if-else part is wrapped with try except and in except part VmError is already handled. I am only removing is_contract part to apply the ape-hardhat changes as logically if sender is self.unlocked_accounts that means the sender is ImpersonatedAccount

bilbeyt: Also in ape-hardhat, this send_transaction only used for unlocked_accounts and for others super call will make it work with send_raw_transaction. which is already here now. To me, they are equal.

unparalleled-js (dismissed):
I think this is good, it is a nice default handler. I don't think we are able to test this in Core since we don't have impersonate account for either Geth or EthTester.

but if this is working for you all, then let's ship it.

src/ape/contracts/base.py
```diff
@@ -263,16 +263,16 @@ def __repr__(self) -> str:
...
         kwargs = _convert_kwargs(kwargs, self.conversion_manager.convert)
         return self.provider.network.ecosystem.encode_transaction(
             self.address, self.abi, *args, **kwargs
         )
 
     def __call__(self, *args, **kwargs) -> ReceiptAPI:
+        if "sender" in kwargs and isinstance(kwargs["sender"], (ContractInstance, Address)):
+            # Automatically impersonate contracts (if API available) when sender
+            kwargs["sender"] = self.account_manager.test_accounts[kwargs["sender"].address]
+
```
fubuloubu, line 275: What's the reasoning for moving this? `serialize_transaction` gets called below

bilbeyt: to set kwargs["sender"] here to be ImpersonatedAccount so later if condition will work and use ImpersonatedAccount.call function

fubuloubu: So the flow here is that this same convert step was happening inside of `serialize_transaction`:

https://github.com/ApeWorX/ape/blob/2a82957dcbf97471fe5d2c6dc0b8359676aaaee9/src/ape/contracts/base.py#L276

Which is before the use of it to do `kwargs["sender"].call`:

https://github.com/ApeWorX/ape/blob/2a82957dcbf97471fe5d2c6dc0b8359676aaaee9/src/ape/contracts/base.py#L278-L279

Meaning they are functionally equivalent, except we loose the ability to call `serialize_transaction` directly and take advantage of this same feature

fubuloubu (approved):

fubuloubu:
@bilbeyt just update the branch

Issue #1231, opened 2023-01-18
Type error in empty DynArray of struct


milkyklim:
### Environment information

- OS: macOS 
- Python Version: Python 3.8.6
- `ape` and plugin versions:

```sh
$ ape --version
0.5.9

$ ape plugins list
Installed Plugins:
  etherscan    0.1.0b3
  tokens       0.1.0b5.dev2+g9dc883e
  vyper        0.5.2
  solidity     0.5.4
  foundry      0.5.2
```

- Contents of your `ape-config.yaml`:

```yaml
name: test
plugins:
  - name: vyper
  - name: solidity
  - name: tokens
  - name: ethereum
  - name: foundry

ethereum:
  default_network: mainnet-fork
  mainnet_fork:
    default_provider: foundry

foundry:
  port: auto
  fork:
    ethereum:
      mainnet:
        upstream_provider: geth

geth:
  ethereum:
    mainnet:
      uri: ${WEB3_PROVIDER_URI}
```

### What went wrong?

This gets compiled:

```vyper
# @version 0.3.7
MAX_NUM: public(constant(uint256)) = 10

struct MyStruct:
    one: uint256
    two: uint256
    there: uint256

@view
@external
def get_values() -> (DynArray[uint256, MAX_NUM], DynArray[MyStruct, MAX_NUM]):
    _ids: DynArray[uint256, MAX_NUM] = []
    _my_structs: DynArray[MyStruct, MAX_NUM] = []

    return _ids, _my_structs
```

When I try to call `.get_values()` in ape I get an error: 

`TypeError: __init__() missing 3 required positional arguments: 'one', 'two', and 'there'`

### How can it be fixed?

--

fubuloubu:
Diving into this, it seems like for the second return type it is returning an empty tuple instead of an empty list from `ape.utils.abi.StructParser.parse`

PR #1243, opened 2023-01-23
fix: parse empty array of structs into empty list

fubuloubu:
### What I did

Avoid trying to create a struct value when the returndata is empty if the return data is expeted to be a dynamic array of structs

fixes: #1231

fixes: APE-490

### How I did it

Found two conditions that needed to be fixed:
1. If there is no data, you cannot parse out any components for a struct, so return `None` early (avoids `TypeError` per #1231)
2. If the type is an array of structs, and you get a `None` value, then it should be an empty list instead.

### How to verify it

Not sure if this was the right approach to solving the problem, it seems weird to me that `_create_struct` would not be handled better by `parse` upstream of the second change I had to make to handle that condition

### Checklist

- ~~[ ] Documentation has been updated~~


fubuloubu:
Needs tests for both tuple of dynarray of structs and dynarray of structs

src/ape/utils/abi.py
```diff
@@ -111,6 +111,10 @@ def parse(self, output_types: List[ABIType], values: Union[List, Tuple]) -> Any:
                         }
                         item_type = ABIType.parse_obj(item_type_data)
                         parsed_item = self.parse([item_type], [value])
+                        # If it's an empty array of structs, replace `None` with empty list
+                        if output_type.type == "tuple[]" and parsed_item is None:
```
unparalleled-js, line 115: What about fixed size, like `tuple[2]`?

unparalleled-js (commented):
We could use a test for this. Let me know if need help or would like me to do it.

src/ape/utils/abi.py
fubuloubu: A lot more permutations to test, some examples:
```vyper
struct Foo:
    a: address
    b: bytes32
    c: uint256

...
def bar1() -> Foo:
...
def bar2() -> DynArray[Foo, N]:
...
def bar3() -> (DynArray[Foo, N], DynArray[Foo, N]):
...
def bar4() -> (Foo[N], DynArray[Foo, N]):
...
def bar5() -> DynArray[Foo[M], N]:
```

src/ape/utils/abi.py
```diff
@@ -132,6 +136,9 @@ def _create_struct(self, out_abi: ABIType, out_value) -> Optional[Any]:
             name = out_abi.name or self.default_name
 
         components = self._parse_components(out_abi.components, out_value[0])
+        if not components:
+            return None  # Empty struct
```
unparalleled-js, line 140: When return `empty(MyStruct)` in Vyper, it gives me a components like this: 
```
['0x0000000000000000000000000000000000000000', HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000')]
```

Is `empty(MyStruct)` not what you mean by `# Empty struct` here?

unparalleled-js: Were you able to hit this return statement in your example?

fubuloubu: `empty(MyStruct)` means "a `MyStruct` with all default (empty) values", so it actually does produce a struct (which is perhaps confusing)

fubuloubu: An "empty struct" is a struct w/o any components, e.g. `struct MyStruct: pass`, which I don't think is valid in Vyper (and maybe Solidity)

fubuloubu: Perhaps this should raise instead

unparalleled-js: It is needed for empty tuples actually, but I adjusted it a little.

src/ape/utils/abi.py
unparalleled-js: I am adding these now and adjusting

unparalleled-js: I added a bunch more tests cases that I think align with these.
Let me know if you think anything is missing yet!

tests/functional/test_contract_instance.py
```diff
@@ -194,7 +195,35 @@ def test_nested_structs_in_tuples(contract_instance, sender, chain):
     assert is_checksum_address(struct_2.t.a)
 
 
-def test_vyper_structs_with_array(vyper_contract_instance, sender):
+def test_get_empty_dyn_array_of_structs(contract_instance):
+    actual = contract_instance.getEmptyDynArrayOfStructs()
+    expected = None
```
unparalleled-js, line 200: I think this should actually be an empty list.

unparalleled-js: k it list now

fubuloubu (commented):
Approved

unparalleled-js (approved):

PR #1244, opened 2023-01-24
refactor!: use lists for logs on receipt instead of iterator

unparalleled-js:
### What I did

Since we know the number of logs should be small enough on receipts, if we decoding logs for a specific receipt, uses lists instead of iterators.

This simplifies tests for everyone.

### Checklist

### DIFF
--- a/docs/userguides/transactions.md
+++ b/docs/userguides/transactions.md
@@ -108,21 +108,27 @@ When declaring `type="0x0"` and _not_ specifying a `gas_price`, the `gas_price` 
 
 ## Transaction Logs
 
-To get logs that occurred during a transaction, you can use the [ContractEvent.from_receipt(receipt)](../methoddocs/contracts.html?highlight=contractevent#ape.contracts.base.ContractEvent.from_receipt) and access your data from the [ContractLog](../methoddocs/types.html#ape.types.ContractLog) objects that it returns.
-
-The following is an example demonstrating how to access logs from an instance of a contract:
+In Ape, you can easily get all the events on a receipt.
+Use the `.events` property to access the ([ContractLog](../methoddocs/types.html#ape.types.ContractLog)) objects.
+Each object represents an event emitted from the call.
 
 ```python
 receipt = contract.fundMyContract(value="1 gwei", type="0x0", sender=sender)
-for log in contract.MyFundEvent.from_receipt(receipt):
+print(receipt.events)
+```
+
+To only get specific log types, use the `decode_logs()` method and pass the event ABIs as arguments:
+
+```python
+for log in receipt.decode_logs(contract.FooEvent.abi, contract.BarEvent.abi):
     print(log.amount)  # Assuming 'amount' is a property on the event.
 ```
 
-You can also access the logs from the receipt itself if you know the ABI:
+You can also use the [ContractEvent.from_receipt(receipt)](../methoddocs/contracts.html?highlight=contractevent#ape.contracts.base.ContractEvent.from_receipt) method:
 
 ```python
-event_type = contract.MyEvent
-for log in receipt.decode_logs(event_type.abi):
+receipt = contract.fooMethod(value="1 gwei", type="0x0", sender=sender)
+for log in contract.FooEvent.from_receipt(receipt):
     print(log.amount)  # Assuming 'amount' is a property on the event.
 ```
 
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -236,7 +236,7 @@ class ReceiptAPI(BaseInterfaceModel):
         abi: Optional[
             Union[List[Union[EventABI, "ContractEvent"]], Union[EventABI, "ContractEvent"]]
         ] = None,
-    ) -> Iterator[ContractLog]:
+    ) -> List[ContractLog]:
         """
         Decode the logs on the receipt.
 
@@ -244,7 +244,7 @@ class ReceiptAPI(BaseInterfaceModel):
             abi (``EventABI``): The ABI of the event to decode into logs.
 
         Returns:
-            Iterator[:class:`~ape.types.ContractLog`]
+            List[:class:`~ape.types.ContractLog`]
         """
 
     def raise_for_status(self):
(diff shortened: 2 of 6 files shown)


docs/userguides/transactions.md
```diff
@@ -108,21 +108,27 @@ When declaring `type="0x0"` and _not_ specifying a `gas_price`, the `gas_price`
...
-
-The following is an example demonstrating how to access logs from an instance of a contract:
+In Ape, you can easily get all the events on a receipt.
+Use the `.events` property to access the ([ContractLog](../methoddocs/types.html#ape.types.ContractLog)) objects.
+Each object represents an event emitted from the call.
 
 ```python
 receipt = contract.fundMyContract(value="1 gwei", type="0x0", sender=sender)
-for log in contract.MyFundEvent.from_receipt(receipt):
+print(receipt.events)
```
fubuloubu, line 117: Much nicer API

docs/userguides/transactions.md
```diff
@@ -108,21 +108,27 @@ When declaring `type="0x0"` and _not_ specifying a `gas_price`, the `gas_price`
...
-You can also access the logs from the receipt itself if you know the ABI:
+You can also use the [ContractEvent.from_receipt(receipt)](../methoddocs/contracts.html?highlight=contractevent#ape.contracts.base.ContractEvent.from_receipt) method:
 
 ```python
-event_type = contract.MyEvent
-for log in receipt.decode_logs(event_type.abi):
+receipt = contract.fooMethod(value="1 gwei", type="0x0", sender=sender)
+for log in contract.FooEvent.from_receipt(receipt):
     print(log.amount)  # Assuming 'amount' is a property on the event.
 ```
```
fubuloubu, line 134: Eventually would be nice to have something like:
```py
tx = contract.fundMyContract(value="1 gwei", type="0x0", sender=sender)
# Validate at least one instance of `MyFundEvent` in `tx.events`
assert contract.MyFundEvent in tx.events
# Validate at least one instance of `MyFundEvent` in `tx.events` with argument `amount` set to 1
assert contract.MyFundEvent(amount=1) in tx.events
# Validate at least one instance of `MyFundEvent` in `tx.events` with argument `amount` set to 1 and `receiver` set to `acct`
assert contract.MyFundEvent(amount=1, receiver=acct) in tx.events
# Validate exactly one event in `tx.events`, which is a `MyFundEvent` with argument `amount` set to 1 and `receiver` set to `acct`
assert tx.events == [contract.MyFundEvent(amount=1, receiver=acct)]
```

Compared to the [current ways](https://github.com/fubuloubu/WETH8/blob/622ef7e29d260ff055e2eb8c4fdc07a4411447e7/tests/test_weth.py#L112-L117) of doing event checking in tests, this is much nicer API

fubuloubu (approved):

docs/userguides/transactions.md
fubuloubu: Note: the "fuzzy matching" (example number 2) is probably quite a bit more complicated than the other 3 and can get backlogged for now

src/ape/api/transactions.py
```diff
@@ -230,21 +230,29 @@ def _confirmations_occurred(self) -> int:
 
         return latest_block.number - self.block_number
 
+    @cached_property
+    def events(self) -> List[ContractLog]:
```
istankovic, line 234: What is the reason this has been made a list (a mutable object)? Do we really require mutability here?
My feeling is this should have been a tuple, an immutable object.

unparalleled-js: Oh ya, I guess I meant sorted, not sort.

istankovic: > sorted_logs = list(tx.events).sort(key=foo)

This wouldn't work, but this would:
```python
sorted_logs = sorted(tx.events, key=foo)
```

unparalleled-js: However! I just noticed that `sorted()` can take a tuple, but it returns a list.
Still I am leaning `list` though.

istankovic: > Still I am leaning list though.

Let me elaborate on why I think it is a good idea to go with an immutable type in cases like this.
If it is not immediately obvious, perhaps by looking at the function's name, that the function returns a new object,
without references being held by the original object (e.g . the instance in case of instance methods), then one
may think that either (1) the return value is not referenced by the original object and is therefore safe to mutate, or (2) the return value _is_ referenced by the original object and is therefore not safe to mutate. Either of those would be valid. This situation is even more pronounced in cases like this where we have an attribute access.

By choosing immutable types over mutable ones, the API authors signal the correct assumption to the user of the API.
For that reason it also a fairly wide practice in the Python community.

unparalleled-js: Since it is a cached_property, it is technically is referenced by the original object, but I see what you mean.

istankovic: > Since it is a cached_property, it is technically is referenced by the original object

... which means that repeated attribute accesses of `.events` on the same object are not guaranteed to return the same set of event, right?

fubuloubu: I think this is a good catch, and it should have been a normal `property` (consistent with how `ContractEvent.from_receipt` is implemented) and produce a new list object every time it's called.

It adds a small amount of overhead, but there are no RPC calls done in response to triggering this property, just a call to `EcosystemAPI` for decoding the log data with the given ABIs.

Not fundamentally against tuples, as it's only slightly more awkward to remember the comma when comparing against expected logs in a test, but I think a list is pretty natural to work with, and you can rearrange it like @unparalleled-js mentioned to have an easier time comparing e.g. `assert sorted(tx.events, key=event_sorting_fn)[:2] == [EventA(...), EventB(...)] ` (has at least two events in any order with those exact comparisons)

unparalleled-js: @cached_property has been removed. Thanks for patience and detailed explanation @istankovic.

unparalleled-js: > ... which means that repeated attribute accesses of `.events` on the same object are not guaranteed to return the same set of events, right?

`@cached_property` does return the same set each time because it is cached on the object.

But now, we made it a normal calculated property. So you can mutate it without affects the receipt.

PR #1247, opened 2023-01-24
fix: issues with static fee transactions

unparalleled-js:
### What I did

I noticed while debugging https://github.com/ApeWorX/ape-optimism/issues/12 that the issue was because of static fee transactions. I was able to reproduce with Ethereum as well.

### How I did it

* Remove `Union` type from `TransactionAPI` and require the transaction type to be an `int`.
* Make a `HexIntConverter` class
* Fix weird chain ID issues on receipt's transactions that I found during txn-replay for extracting revert messages.

### How to verify it

* `contract.method(arg, sender=me, type=0)`
* Also make sure works in reverts testing

### Checklist

### DIFF
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -3,7 +3,14 @@ from decimal import Decimal
 from typing import Any, Dict, List, Tuple, Type, Union
 
 from dateutil.parser import parse  # type: ignore
-from eth_utils import is_checksum_address, is_hex, is_hex_address, to_checksum_address, to_hex
+from eth_utils import (
+    is_checksum_address,
+    is_hex,
+    is_hex_address,
+    to_checksum_address,
+    to_hex,
+    to_int,
+)
 from hexbytes import HexBytes
 
 from ape.api import ConverterAPI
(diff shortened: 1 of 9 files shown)


src/ape/managers/converters.py
```diff
@@ -38,6 +45,28 @@ def convert(self, value: str) -> bytes:
...
 
 
+class HexIntConverter(ConverterAPI):
+    """
+    Convert hex values to integers.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return (
+            isinstance(value, int)
```
fubuloubu, line 55: Should skip this value tbh, as it's just pass-through since if the value is already the target type, it skips conversion

src/ape/managers/converters.py
```diff
@@ -38,6 +45,28 @@ def convert(self, value: str) -> bytes:
...
+            or (isinstance(value, str) and is_hex(value))
+            or isinstance(value, bytes)
+        )
+
+    def convert(self, value: Any) -> int:
+        if isinstance(value, str) and is_hex(value):
+            return to_int(HexBytes(value))
```
fubuloubu, line 62: Can probably reduce it to just this, since it should be guaranteed to be a hex string or bytes coming in

tests/functional/conversion/test_timestamp.py
```diff
@@ -25,9 +25,8 @@ def test_convert_string_timestamp(args):
     "args",
     (
         [
-            "0000",
-            "2003",
-            "9999999999",
```
fubuloubu, line 30: These became valid?

fubuloubu (approved):

tests/functional/conversion/test_timestamp.py
unparalleled-js: they are valid in the other converter

src/ape/managers/converters.py
unparalleled-js: https://github.com/ApeWorX/ape/pull/1248

src/ape/managers/converters.py
unparalleled-js: https://github.com/ApeWorX/ape/pull/1248

PR #1255, opened 2023-01-26
feat: support selection of default transaction type in config

bilbeyt:
### What I did

Changed Network config and ape-ethereum a little to support

fixes: APE-513

### Checklist


unparalleled-js (commented):
You can test this by using the `temp_config` fixture:

`test_ecosystem.py`

```python
def test_configure_default_txn_type(temp_config, Ethereum):
    config_dict = {"ethereum": {"local": {"default_transaction_type": 0}}}
    with temp_config(config_dict):
        assert ethereum.local.default_transaction_type == TransactionType.STATIC
```

(I didn't run this)

fubuloubu (approved):

fubuloubu:
Just needs tests!

fubuloubu (dismissed):

unparalleled-js (approved):

tests/functional/test_ecosystem.py
```diff
@@ -182,3 +183,10 @@ def test_decode_receipt(eth_tester_provider, ethereum):
...
     )
+
+
+def test_configure_default_txn_type(temp_config, ethereum):
+    config_dict = {"ethereum": {"local": {"default_transaction_type": "0x00"}}}
```
unparalleled-js, line 189: It looks like we need this to go through a pydantic validator so that it is can be recognized as `0` before it creates the enum value.

fubuloubu: Yeah, the work we did on the gas limit PR to refactor to using Validators could come into play here too

But maybe a bit differently, maybe in `TransactionAPI` it is just an integer value, allowing each subclass (e.g. `ape_ethereum.transactions.BaseTransaction`) to take it in as an Enum Value and convert to an integer value when forming the transaction

fubuloubu: Also, refactoring it to be the top item in `TransactionAPI` will allow that value to be used for parsing a raw transaction into the given type respected by our framework. https://docs.pydantic.dev/usage/types/#discriminated-unions-aka-tagged-unions

bilbeyt: i can put pydantic validator to do the job but other refactoring parts, i couldnt get it, but if you explain more i can try refactoring other parts too.

fubuloubu: It's a bit more complex, but in this PR we refactored the gas limit field of `TransactionAPI` to be a bit simpler (just `Optional[int]` instead of `Union[int, str, "max", "auto", None]` which was causing errors downstream:

https://github.com/ApeWorX/ape/pull/1224/files#diff-84df2743691337b305524d23ee375e6f5d476f2e684da37155480eab02d51001R50

A similar approach here might work, basically adding a validator for `type` that converts whatever input is given (probably `Union[str, int, bytes]`) to just `int`

The Ethereum plugin can do the same thing, although now that I think about it, no validator should be needed to downcast the Enum if it subclasses `int`

fubuloubu (dismissed):
I like the changes here, but need to fix the problem with CI

tests/functional/test_ecosystem.py
unparalleled-js: I believe  a temporary fix for the CI problem would be to change this from `"0x00"` to `0`, and then we can handle the conversions on a separate PR, if you want and test the scenarios where the user uses hex values in their config.

bilbeyt: @unparalleled-js the problem is i have python 3.10 and mac setup locally and i cant give anything in test other than "0x00". otherwise, it is failing.

bilbeyt: :D sorry you actually changed these 6 days ago about optimism fix static transactions as i didnt rebase my branch i never noticed :D

unparalleled-js (approved):

PR #1266, opened 2023-01-31
fix: add option to ignore signer matching

bilbeyt:
### What I did

Add option to ignore signer matching when serialize

fixes: #

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -52,7 +52,7 @@ class AccessList(BaseModel):
 
 
 class BaseTransaction(TransactionAPI):
-    def serialize_transaction(self) -> bytes:
+    def serialize_transaction(self, ignore_sign_match: bool = False) -> bytes:
         if not self.signature:
             raise SignatureError("The transaction is not signed.")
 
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/transactions.py
```diff
@@ -61,14 +61,18 @@ def serialize_transaction(self) -> bytes:
         signature = (self.signature.v, to_int(self.signature.r), to_int(self.signature.s))
         signed_txn = encode_transaction(unsigned_txn, signature)
 
-        if self.sender and EthAccount.recover_transaction(signed_txn) != self.sender:
+        if (
+            self.sender
+            and EthAccount.recover_transaction(signed_txn) != self.sender
+            and not ignore_sign_match
```
fubuloubu, line 67: If you put `and txn.signature` prior to the previous leg of the conditional, you can emulate this same thing without the kwarg

bilbeyt: this is another thing, if you use impersonated account then traverse block transactions, you will see there is a signature for the transaction which to me really weird.

bilbeyt: i tried to find the account address who signed this but never found the address coming in with tx.

bilbeyt: so self.signature is not doing that

bilbeyt: it would also fail at line 56 otherwise

bilbeyt:
maybe this is not relevant anymore because of #1274

unparalleled-js:
> maybe this is not relevant anymore because of #1274

Did we ever figure this out?

bilbeyt:
> > maybe this is not relevant anymore because of #1274
> 
> Did we ever figure this out?

closing this as i dont have any problem in my perf PR

PR #1267, opened 2023-01-31
fix: remove direct install from setup file

unparalleled-js:
### What I did

Thanks @fubuloubu for the help with this one and the fast turn-around.
This should fix publishing 🤞🏻 

fixes: APE-550

### How I did it

* Comment out myst parser in setup.py
* Install from github in the docs CI action

### How to verify it

We try releasing again!

### Checklist


NotPeopling2day (dismissed):

fubuloubu (approved):

NotPeopling2day (approved):

.github/workflows/prtitle.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         - name: Install Dependencies
           run: |
             python -m pip install --upgrade pip
-            pip install .[dev]
+            pip install "commitizen>=2.40,<2.41"
```
unparalleled-js, line 25: The action runs much faster this way and saves action minutes.
It only needs the one tiny tool to do this job! It does not need Ape installed.

unparalleled-js:
The Pr-title action somehow only uses either the yaml from when the PR was first opened or the yaml from main, can't tell.
It always fails even though I am 99.9% sure I fixed the issue.

I say we merge this and I can open a dummy PR next to show that it is indeed fixed.

.github/workflows/prtitle.yaml
fubuloubu: Yes this is true! I wanted it to source the same version of commitizen from setup.py, only reason I changed it that way

fubuloubu (dismissed):

.github/workflows/prtitle.yaml
unparalleled-js: Should we go back? I am good with other. Pros and cons!

fubuloubu: so like if it could just read that line directly from `setup.py` that would be best

unparalleled-js: [f35f8eb](https://github.com/ApeWorX/ape/pull/1267/commits/f35f8ebe9d227a906d4b245c151e15d573dcaf3e)

setup.py
```diff
@@ -47,15 +47,16 @@
         "twine==3.8.0",  # Package upload tool
     ],
     "dev": [
-        "commitizen>=2.19,<2.20",  # Manage commits and publishing releases
+        # commitizen: Manage commits and publishing releases
+        (here / "cz-requirement.txt").read_text().strip(),
```
unparalleled-js, line 51: @fubuloubu Now, `setup.py` and the actions that use CZ share the same source of truth.

fubuloubu (approved):

fubuloubu (approved):

NotPeopling2day (approved):

PR #1274, opened 2023-02-01
perf: improve performance of web3 provider send transaction

bilbeyt:
### What I did

Improved performance of transaction handling

fixes: APE-563

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -26,7 +26,6 @@ from web3.exceptions import ContractLogicError as Web3ContractLogicError
 from web3.exceptions import TimeExhausted
 from web3.types import RPCEndpoint
 
-from ape.api.accounts import ImpersonatedAccount
 from ape.api.config import PluginConfig
 from ape.api.networks import LOCAL_NETWORK_NAME, NetworkAPI
 from ape.api.query import BlockTransactionQuery
@@ -1041,30 +1040,22 @@ class Web3Provider(ProviderAPI, ABC):
         return txn
 
     def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
-        if not txn.sender:
-            txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
 
-        else:
-            try:
-                account = self.account_manager[txn.sender]
-                if isinstance(account, ImpersonatedAccount):
-                    txn_dict = txn.dict()
-                    txn_params = cast(TxParams, txn_dict)
-                    txn_hash = self.web3.eth.send_transaction(txn_params)
-                else:
-                    txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
-            except ValueError as err:
-                vm_err = self.get_virtual_machine_error(err, txn=txn)
-
-                if "nonce too low" in str(vm_err):
-                    # Add additional nonce information
-                    new_err_msg = f"Nonce '{txn.nonce}' is too low"
-                    raise VirtualMachineError(
-                        new_err_msg, base_err=vm_err.base_err, code=vm_err.code
-                    )
-
-                vm_err.txn = txn
-                raise vm_err from err
+        try:
+            if txn.signature or not txn.sender:
+                txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            else:
+                if (
+                    txn.sender not in self.chain_manager.provider.web3.eth.accounts  # type: ignore # noqa
+                ):
+                    self.chain_manager.provider.unlock_account(txn.sender)
+                txn_dict = txn.dict()
+                txn_params = cast(TxParams, txn_dict)
+                txn_hash = self.web3.eth.send_transaction(txn_params)
+        except ValueError as err:
+            vm_err = self.get_virtual_machine_error(err, txn=txn)
+            vm_err.txn = txn
+            raise vm_err from err
 
         receipt = self.get_receipt(
             txn_hash.hex(),
@@ -1173,6 +1164,13 @@ class Web3Provider(ProviderAPI, ABC):
         err_msg = err_data.get("message")
         if not err_msg:
             return VirtualMachineError(base_err=exception, txn=txn)
+
+        if txn is not None and "nonce too low" in str(err_msg):
+            txn = cast(TransactionAPI, txn)
+            new_err_msg = f"Nonce '{txn.nonce}' is too low"
+            return VirtualMachineError(
+                new_err_msg, base_err=exception, code=err_data.get("code"), txn=txn
+            )
 
         return VirtualMachineError(str(err_msg), code=err_data.get("code"), txn=txn)
 
--- a/src/ape_test/accounts.py
+++ b/src/ape_test/accounts.py
@@ -10,7 +10,29 @@ from ape.utils import GeneratedDevAccount, generate_dev_accounts
 
 
 class TestAccountContainer(TestAccountContainerAPI):
-    _num_generated = 0
+    _num_generated: int
+    _accounts: List["TestAccount"]
+    _mnemonic: str
+    _num_of_accounts: int
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.init()
+
+    def init(self):
+        self._num_generated = 0
+        self._accounts = []
+        self._mnemonic = self.config["mnemonic"]
+        self._num_of_accounts = self.config["number_of_accounts"]
+        for index, account in enumerate(self._dev_accounts):
+            self._accounts.append(
+                TestAccount(
+                    index=index, address_str=account.address, private_key=account.private_key
+                )
+            )
+
+    def __len__(self) -> int:
+        return self._num_of_accounts
 
     @property
     def config(self):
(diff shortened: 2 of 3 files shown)


src/ape/api/providers.py
```diff
@@ -1042,13 +1041,23 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
 
     def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         try:
-            account = self.account_manager[txn.sender]
-            if isinstance(account, ImpersonatedAccount):
+            # account = self.account_manager[txn.sender]
```
unparalleled-js, line 1044: - [ ] Remove commented code

src/ape/api/providers.py
```diff
@@ -1042,13 +1041,23 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
...
+            #     txn_dict = txn.dict()
+            #     txn_params = cast(TxParams, txn_dict)
+            #     txn_hash = self.web3.eth.send_transaction(txn_params)
+            # else:
+            #     txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            if txn.signature:
+                txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            else:
+                if (
+                    txn.sender not in self.chain_manager.provider.web3.eth.accounts  # type: ignore # noqa
```
unparalleled-js, line 1055: I am not exactly sure why, but when we send calls as txns for the sake of getting gas and trace info on the call, the txn goes through with a sender equal to `None`. That is what led to this PR: https://github.com/ApeWorX/ape/pull/1276
It was causing issues because Ape was trying to impersonate `None`.

Perhaps there is a better way, such as setting sender on `AccountAPI.call()` if it is missing. Not sure.

But `sender` is `Optional` on a txn so we shouldn't assume it is there anyway.

src/ape/contracts/base.py
```diff
@@ -275,9 +275,9 @@ def __repr__(self) -> str:
         return self.abi.signature
 
     def serialize_transaction(self, *args, **kwargs) -> TransactionAPI:
-        if "sender" in kwargs and isinstance(kwargs["sender"], (ContractInstance, Address)):
-            # Automatically impersonate contracts (if API available) when sender
-            kwargs["sender"] = self.account_manager.test_accounts[kwargs["sender"].address]
+        # if "sender" in kwargs and isinstance(kwargs["sender"], (ContractInstance, Address)):
```
unparalleled-js, line 278: - [ ] Remove commented code

src/ape/api/providers.py
bilbeyt: I just tried accounts[None] and test_accounts[None], as getitem has no implementation regarding None it cant impersonate it and should raise an error. How that was possible in your case?

bilbeyt: I introduced a new error type SenderNotFound and edited some parts so technically in send_transaction there shouldnt be any None txn.sender

unparalleled-js: It not about impersonation, it happens when extracting gas / trace info from calls (like when `--gas`) flag is specified.  So there really isn't a sender needed.

If there is no sender, can we just send the txn like we were doing before? Else it breaks the `ape-hardhat` and `ape-foundry` plugins.

unparalleled-js: I am testing on `ape-hardhat` and `ape-foundry`... if it is not broken, then it's fine.

bilbeyt: I will revert my error type as it is also neglecting contract instances and creating issues with tests. I will just send txn if there is no sender anyways and continue with other improvements if i can find other things

unparalleled-js (changes requested):
So I am not hitting the `None` account issue I was before, but now getting this error in `ape-hardhat` from these changes:

`AttributeError: 'TestContractSol' has no attribute 'call'.`

It happens on a test that impersonates a contract.

tests/functional/test_accounts.py
```diff
@@ -291,7 +293,9 @@ def test_impersonate_not_implemented(accounts):
         _ = accounts[test_address]
 
 
-def test_contract_as_sender_non_fork_network(contract_instance):
+def test_contract_as_sender_non_fork_network(mocker, contract_instance):
+    mocked_provider = mocker.patch("ape_test.provider.LocalProvider")
+    mocked_provider.unlock_account.side_effect = NotImplementedError
```
unparalleled-js, line 298: I think by default, `unlock_account` is not implemented in the ethtester provider, no?

src/ape_test/provider.py
```diff
@@ -138,7 +141,16 @@ def send_call(self, txn: TransactionAPI, **kwargs) -> bytes:
 
     def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
         try:
-            txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            if txn.signature:
+                txn_hash = self.web3.eth.send_raw_transaction(txn.serialize_transaction())
+            else:
+                if (
+                    txn.sender not in self.chain_manager.provider.web3.eth.accounts  # type: ignore # noqa
```
unparalleled-js, line 148: Can you place the specific linting rules you need to disable here? Like `noqa[E501]`

src/ape_test/provider.py
```diff
@@ -58,6 +58,9 @@ def disconnect(self):
     def update_settings(self, new_settings: dict):
         pass
 
+    def unlock_account(self, address: AddressType) -> bool:
```
unparalleled-js, line 61: This is pretty cool; I'd like to see some new tests showcasing this.

src/ape/exceptions.py
```diff
@@ -28,6 +28,12 @@ class AccountsError(ApeException):
     """
 
 
+class SenderNotFoundError(ApeException):
```
unparalleled-js, line 31: - [ ] So we can delete this?

bilbeyt:
> So I am not hitting the `None` account issue I was before, but now getting this error in `ape-hardhat` from these changes:
> 
> `AttributeError: 'TestContractSol' has no attribute 'call'.`
> 
> It happens on a test that impersonates a contract.

Yes, it is ignoring contractinstance as sender either i need to also add that or revert my change regarding sender and just add or condition to send send_raw_transaction which should be more performant i guess as there will be less code line executed

src/ape/api/providers.py
unparalleled-js: Another idea I didnt try:  In the method `_send_call_as_txn`, maybe we can include the account as the sender there when it calls `call()` https://github.com/ApeWorX/ape/blob/main/src/ape/api/providers.py#L851

src/ape/api/providers.py
```diff
@@ -1041,30 +1040,26 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
...
+                ):
+                    self.chain_manager.provider.unlock_account(txn.sender)
+                txn_dict = txn.dict()
+                txn_params = cast(TxParams, txn_dict)
+                txn_hash = self.web3.eth.send_transaction(txn_params)
+        except ValueError as err:
+            vm_err = self.get_virtual_machine_error(err, txn=txn)
+            if "nonce too low" in str(vm_err):
```
unparalleled-js, line 1057: Should this part should occur within `def get_virtual_machine_error()`? That way, this custom error happens in all the necessary spots.

tests/functional/test_accounts.py
unparalleled-js: nevermind, I see you implemented it!

src/ape/api/providers.py
bilbeyt: added

src/ape/exceptions.py
bilbeyt: deleted

src/ape/api/providers.py
bilbeyt: It is already using account.call and for those cases i still added sender for txn if there is none.

src/ape_test/provider.py
bilbeyt: will add 1-2 tests

unparalleled-js: Oh wait I just realized this was a thing: https://github.com/ethereum/eth-tester/blob/master/eth_tester/main.py#L205

Can we use their official method?

bilbeyt: :s i just tried this and it didnt workout well, so the problem is you can create a new account using private key and lock/unlock to that account, so you cant random add an account with an address, so i will revert this because eth-tester doesnt support this.

unparalleled-js: Thanks for trying!

src/ape/managers/accounts.py
```diff
@@ -36,7 +36,7 @@ def __repr__(self) -> str:
         accounts_str = ", ".join([a.address for a in self.accounts])
         return f"[{accounts_str}]"
 
-    @cached_property
+    @property
```
fubuloubu, line 39: Containers should only be initialized once

bilbeyt:
How about changing whole architecture to use just simple classes like no cached_property or property, and create just pub-sub system so when config changes, managers would be notified. This is basically already keeping everything in memory and doesnt compute them if it is not required really. Also, is using yield necessary? mostly projects wont have thousands users for example that yield would perform, in other context just using lists are faster. @unparalleled-js  @fubuloubu

src/ape_test/accounts.py
```diff
@@ -10,41 +10,61 @@
...
-        return generate_dev_accounts(mnemonic, number_of_accounts=self.config["number_of_accounts"])
+        return generate_dev_accounts(self._mnemonic, number_of_accounts=self._num_of_accounts)
 
     @property
     def aliases(self) -> Iterator[str]:
-        for index in range(len(self)):
+        for index in range(self._num_of_accounts):
             yield f"dev_{index}"
 
+    def _is_config_changed(self):
```
unparalleled-js, line 50: `ProviderAPI`'s are allowed to implement a method called `update_settings()`.
We should use this method to change the mnemonic and handle regenerating accounts.
All you would do is add a method `update_settings()` to this class that takes a dictionary. And then check for mnemonic / num_of_accounts in there, set them, and do everything you do when you call `_is_config_changed()`

This way, you don't need to run `init` form within the `__len__` method and other stuff like that.
`update_settings()` exists for this reason!

unparalleled-js (commented):
I am getting errors in `ape-foundry` for contract-as-sender based tests when using this PR... I can look into it.

`ape-hardhat` seems to work though.

src/ape_test/accounts.py
bilbeyt: but, currently, temp_config context manager in ape is not using that just creating a temporary file as new config and using config.using_project to do the change, so update_settings wont be called anyway

bilbeyt: I cann change the context manager ofc but we may still get user errors i guess

bilbeyt: > but, currently, temp_config context manager in ape is not using that just creating a temporary file as new config and using config.using_project to do the change, so update_settings wont be called anyway

This was the reason i used this like this

unparalleled-js:
> I am getting errors in `ape-foundry` for contract-as-sender based tests when using this PR... I can look into it.

OK foundry errors may not related to this PR, not sure why they suddenly have come up.. But fixed here: https://github.com/ApeWorX/ape-foundry/pull/32

src/ape_test/accounts.py
unparalleled-js: I suppose the issue is that you want to change the config while in the same Python session?

bilbeyt: yes, so that test wont fail, and i looked at the update_settings part and saw it is only for ProviderAPI and in here we dont need something like this

unparalleled-js (approved):

unparalleled-js:
I am unable to update this branch with `main` for some reason, else I would do that and merge now.

fubuloubu:
> I am unable to update this branch with `main` for some reason, else I would do that and merge now.

@bilbeyt has to enable maintainers to update

bilbeyt:
> > I am unable to update this branch with `main` for some reason, else I would do that and merge now.
> 
> @bilbeyt has to enable maintainers to update

I rebased this one now and as i tried with another MR, i dont see a checkbox to enable maintainers to edit

unparalleled-js:
> I rebased this one now and as i tried with another MR, i dont see a checkbox to enable maintainers to edit

Is it not appearing in the usual spot? Odd, maybe an organization setting?

Here is the guide on it in case its that's helpful: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #1281, opened 2023-02-03
feat: add boto to docker image for aws usage

sabotagebeats:
### What I did

added boto to docker image for better integrations with AWS

### How I did it

add pip install boto to the dockerfile

### How to verify it

run the new dockerfile and then import boto3 in a script or in the python console

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,8 @@ RUN pip install --upgrade pip \
 # Fix RLP installation issue
     && pip uninstall rlp --yes \
     && pip install rlp==3.0.0 \
+# Add boto for aws compatibility
+    && pip install boto3 \
 # Validate installation
     && ape --version
 


Dockerfile
```diff
@@ -30,6 +30,8 @@ RUN pip install --upgrade pip \
 # Fix RLP installation issue
     && pip uninstall rlp --yes \
     && pip install rlp==3.0.0 \
+# Add boto for aws compatibility
```
unparalleled-js, line 33: I have some thought here... Discussing on Telegram / Discord.

sabotagebeats:
After discussion with @unparalleled-js It is possible we should do this in another way possibly creating an alternative docker image that has this tooling since Someone running local docker containers on their computer may not care about AWS and it may bloat the image unnecessarily.

helloibis:
> After discussion with @unparalleled-js It is possible we should do this in another way possibly creating an alternative docker image that has this tooling since Someone running local docker containers on their computer may not care about AWS and it may bloat the image unnecessarily.

Docker has `--build-arg` where you can pass inputs in when building a Docker image, would that be an option maybe that they could toggle whether they want boto?

unparalleled-js:
> Docker has `--build-arg` where you can pass inputs in when building a Docker image, would that be an option maybe that they could toggle whether they want boto?

There is also an ARG on the Dockerfile that you can use `--build-arg` for and point to a custom requirements file.

helloibis:
> There is also an ARG on the Dockerfile that you can use `--build-arg` for and point to a custom requirements file.

That'd be really cool!

sabotagebeats:
Closed because it is much better for image size to put this in a custom docker file using apeworx/ape as a base.

PR #1284, opened 2023-02-04
feat: trace of source code lines ran per receipt

unparalleled-js:
### What I did
lines

(using contracts):

Registry.vy:

```python
# @version 0.3.7

addr: public(address)

@external
def register(addr: address):
    self.addr = addr
```

VyperContract.vy:

```python
# @version 0.3.7

import interfaces.IRegistry as IRegistry

_balance: public(uint256)
registry: public(IRegistry)


@external
def __init__(registry: IRegistry):
    self.registry = registry


@external
def addBalance(  # ::: def?
    num: uint256  # def addBalance(): yop
) -> uint256:  # def
    assert num != self._balance
    self.registry.register(msg.sender)
    self._balance = self._balance + self.addInterest(num)

    # Run some loops.
    for i in [1, 2, 3, 4, 5]:
        if i == num:
            break

    # Comments in the middle (is a test)

    for i in [1, 2, 3, 4, 5]:
        if i != num:
            continue

    return self._balance  # TEST COMMENT 9 def foo2():


@internal
def addInterest(num: uint256) -> uint256:
    return 123 + num
```

Unblocks: APE-577

* Add method `ProjectManager.get_pc_map()` which gets a mapping of PC to a mapping of line number to line content of a given contract.
* Add property `ReceiptAPI.line_trace`  which is a list of trace-nodes representing blocks of text (lines of code)
* Add method `show_line_trace()` to Ethereum Receipts.

The following is some output from `show_line_trace()` at the time of opening the PR.

Matching WIP ape-vyper PR: https://github.com/ApeWorX/ape-vyper/pull/67

https://docs.soliditylang.org/en/v0.8.18/internals/source_mappings.html

### Checklist


fubuloubu (commented):
A lot of deleted files, will give more thoughts when I have a chance to process more

unparalleled-js:
> A lot of deleted files

I split into 2 PRs, there are less deleted files on this one now.
Other PR: https://github.com/ApeWorX/ape/pull/1285

unparalleled-js:
- [ ] CompilerAPI delegation

src/ape/types/trace.py
```diff
@@ -216,3 +222,26 @@ class TraceFrame(BaseInterfaceModel):
...
+    """
+
+
+class LineTraceNode(BaseModel):
+    source_id: str
+    """
+    The local project's contract's relative path str as an ID.
+    """
+
+    method_id: str
```
fubuloubu, line 238: Sometimes there are items that aren't contained within a particular method

For example, when you don't have any payable methods in Vyper, it will inline a nonpayable check (e.g. `assert msg.value == 0`) at the top of the code, and not within any particular method

src/ape/types/trace.py
```diff
@@ -216,3 +222,26 @@ class TraceFrame(BaseInterfaceModel):
...
+    """
+    The method containing the lines.
+    """
+
+    lines: Dict[int, str]
```
fubuloubu, line 243: I need to understand the intent behind this better. I am not sure it's a 1:1 releationship in all cases.

unparalleled-js: Line number -> Line content is 1 to 1

src/ape/types/trace.py
unparalleled-js: But do we want to show these ever?

src/ape/types/trace.py
fubuloubu: Ah, so this is just the model of lines from a `ethpm_types.Source` object? Is there a way to expand more functionality there instead? maybe `Source.__getitem__`

unparalleled-js: Ya there are probably a few things we can move to ethpm types!

I will explain more soon about these line traces soon... I am still kind of in a research phase

src/ape/types/trace.py
fubuloubu: Yes this would be an important programming block to assure we had conditional coverage for

unparalleled-js: It will be part of branch coverage.

PR #1292, opened 2023-02-08
feat: way to get compilers easier [APE-607]

unparalleled-js:
### What I did

Noticed that it is not the easiest to get a hold of one of the compiler API implementation classes, like `ape-solidity`.
Perhaps there are use-cases (like weird Solidity stuff) where one may need the Solidity compiler at their fingertips.

Before, you had to do this:

```python
project.compiler_manager.registered_compilers[".sol"]
```

but now you can do this:

```python
compilers.solidity
```

### How I did it

* Adds method `get_compiler()` to `CompilerManager` that allows getting a compiler by name
* Adds `__getattr__` so you can do the ape-fancy dances that uses `get_compiler()`
* Adds `compilers` pytest fixture to our pytest integration.

### How to verify it

Use solidity in a fixture or in a test. Call the methods.

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Dict, List, Optional, Set
+from typing import Any, Dict, List, Optional, Set
 
 from ethpm_types import ContractType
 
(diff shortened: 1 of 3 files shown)


NotPeopling2day (approved):

src/ape/managers/compilers.py
```diff
@@ -30,6 +30,19 @@ def __repr__(self):
         num_compilers = len(self.registered_compilers)
         return f"<{self.__class__.__name__} len(registered_compilers)={num_compilers}>"
 
+    def __getattr__(self, name: str) -> Any:
+        try:
+            return self.__getattribute__(name)
+        except AttributeError:
+            # Check if a contract.
```
fubuloubu, line 37: What's this note for? :eyes:

src/ape/pytest/fixtures.py
```diff
@@ -44,6 +44,14 @@ def accounts(self) -> List[TestAccountAPI]:
 
         return self.account_manager.test_accounts
 
+    @pytest.fixture(scope="session")
+    def compilers(self):
+        """
+        Access compiler classes directly.
```
fubuloubu, line 50:
```suggestion
        Access compiler manager directly.
```

fubuloubu (approved):

fubuloubu (approved):

PR #1298, opened 2023-02-13
perf: improve performance of ape [APE-620]

bilbeyt:
### What I did

Added more caching

fixes: #

### Checklist

### DIFF
--- a/docs/userguides/console.md
+++ b/docs/userguides/console.md
@@ -9,6 +9,9 @@ In [1]: chain.blocks.head.timestamp
 Out[1]: 1647323479
 ```
 
+WARNING: Contract changes are not reflected in the active console session.
+If you need to make changes to your contract, you must re-start your console session for the compiler to handle the changes.
+
 ## Ape Namespace
 
 Your console comes with pre-initialized root ape objects in your namespace.
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -43,6 +43,10 @@ class ProjectAPI(BaseInterfaceModel):
 
     config_file_name: str = "ape-config.yaml"
 
+    _cached_manifest: Optional[PackageManifest] = None
+
+    _contracts: Optional[Dict[str, ContractType]] = None
+
     def __repr__(self):
         return f"<{self.__class__.__name__} {self.path.name}>"
 
@@ -90,28 +94,29 @@ class ProjectAPI(BaseInterfaceModel):
         The ``PackageManifest`` at :py:attr:`~ape.api.projects.ProjectAPI.manifest_cachefile`
         if it exists and is valid.
         """
-
-        manifest = _load_manifest_from_file(self.manifest_cachefile)
-        if manifest is not None:
-            manifest.contract_types = self.contracts
-
-        return manifest
+        if self._cached_manifest is None:
+            manifest = _load_manifest_from_file(self.manifest_cachefile)
+            if manifest is not None:
+                manifest.contract_types = self.contracts
+            self._cached_manifest = manifest
+        return self._cached_manifest
 
     @property
     def contracts(self) -> Dict[str, ContractType]:
-        contracts = {}
-        for p in self._cache_folder.glob("*.json"):
-            if p == self.manifest_cachefile:
-                continue
+        if self._contracts is None:
+            contracts = {}
+            for p in self._cache_folder.glob("*.json"):
+                if p == self.manifest_cachefile:
+                    continue
 
-            contract_name = p.stem
-            contract_type = ContractType().parse_file(p)
-            if contract_type.name is None:
-                contract_type.name = contract_name
+                contract_name = p.stem
+                contract_type = ContractType().parse_file(p)
+                if contract_type.name is None:
+                    contract_type.name = contract_name
 
-            contracts[contract_type.name] = contract_type
-
-        return contracts
+                contracts[contract_type.name] = contract_type
+            self._contracts = contracts
+        return self._contracts
 
     @property
     def _cache_folder(self) -> Path:
@@ -199,6 +204,8 @@ class DependencyAPI(BaseInterfaceModel):
     A list of glob-patterns for excluding files in dependency projects.
     """
 
+    _cached_manifest: Optional[PackageManifest] = None
+
     def __repr__(self):
         return f"<{self.__class__.__name__} name='{self.name}'>"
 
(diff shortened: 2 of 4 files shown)


fubuloubu (commented):
I don't know if this one will work out ultimately, I'll let @unparalleled-js review and provide any additional color

In general, we are relying on the behavior of re-compilng changed dependencies, so we can't lose that

I think with the refactor to using per-contract type based file compiling, we could reduce the processing path here by holding the checksum information in memory, and caching the compiled types based on this checksum. At the end of the session, write out that file (`__local__.json`) back to disk, maybe using a context manager-style approach.

A different approach is that we could _always_ hold the manifest in memory, and only write it back whenever a compilation happens, and only compile once automatically the first time a type is asked for. If a re-compile is needed, then create an API to trigger a re-compile through the `ProjectManager` class. This would be consistent to Python behavior that a file changed after it is loaded is cached in memory, and does not reference the contents again.

This alternative behavior would probably be fine for testing and scripting (maybe even prefferable?) but might be annoying in the console (without the ability to force a recompile)

bilbeyt:
> I don't know if this one will work out ultimately, I'll let @unparalleled-js review and provide any additional color
> 
> In general, we are relying on the behavior of re-compilng changed dependencies, so we can't lose that
> 
> I think with the refactor to using per-contract type based file compiling, we could reduce the processing path here by holding the checksum information in memory, and caching the compiled types based on this checksum. At the end of the session, write out that file (`__local__.json`) back to disk, maybe using a context manager-style approach.
> 
> A different approach is that we could _always_ hold the manifest in memory, and only write it back whenever a compilation happens, and only compile once automatically the first time a type is asked for. If a re-compile is needed, then create an API to trigger a re-compile through the `ProjectManager` class. This would be consistent to Python behavior that a file changed after it is loaded is cached in memory, and does not reference the contents again.
> 
> This alternative behavior would probably be fine for testing and scripting (maybe even prefferable?) but might be annoying in the console (without the ability to force a recompile)

sure @unparalleled-js can have a look, but i guess this is the most important problem that ape is using lots of computed property and also using IO operations, these things are making it slower and in this PR i addressed this and tests are passing. ape should use in-memory operations more and only store when it is really necessary like you suggested @fubuloubu

istankovic:
@bilbeyt we saw some significant perf improvements with this PR, could you please post the before/after timings for our test suite?

unparalleled-js:
`project.MyContract` went from 712 ms to 49.1 µs. 💯 
It is noticeably snappier.

unparalleled-js:
Testing I have done:

- [ ] Made sure when you change an imported file, that the files that import that file still get re-compiled on `ape compile`
- [ ] Tests a project that relies on dependency contracts.
- [ ] Run tests in `ape-vyper` and `ape-solidity` plugins
- [ ] General `ape compile` testing 

This is a nice performance improvement, it feels good to use.

@fubuloubu is there anything else you are concerned about or want me to double check?

fubuloubu:
> @fubuloubu is there anything else you are concerned about or want me to double check?

Is it changing the behavior of working with the manifest where file changes within a session do not force a recompile? e.g. during `ape console` if I make a change to an outside file, will it recompile if I try to do `project.MyContract` or is it stuck with the source from when I entered the session?

If the answer is "no change in behavior" we can go ahead and ship this. If there is a change, not sure if we'd consider that breaking behavior or not, but I would be okay with that behavior change (either way, needs communication about the update in this scenario)

bilbeyt:
> > @fubuloubu is there anything else you are concerned about or want me to double check?
> 
> Is it changing the behavior of working with the manifest where file changes within a session do not force a recompile? e.g. during `ape console` if I make a change to an outside file, will it recompile if I try to do `project.MyContract` or is it stuck with the source from when I entered the session?
> 
> If the answer is "no change in behavior" we can go ahead and ship this. If there is a change, not sure if we'd consider that breaking behavior or not, but I would be okay with that behavior change (either way, needs communication about the update in this scenario)

This behaviour of using ape console and doing a change is kind of weird to me and never heard such a use case before actually. Is there anyone really using it ?

bilbeyt:
> @bilbeyt we saw some significant perf improvements with this PR, could you please post the before/after timings for our test suite?

Before this 56 tests passed 10 mins 47 secs, with this 56 tests passed 2 minutes 38 secs

fubuloubu:
> 

I use it sometimes, but like I said I'm not opposed to breaking that expectation, we just have to communicate it to our users

unparalleled-js:
Ah, you're right @fubuloubu , that is one change between `main` and this branch.

unparalleled-js:
We could disable the manifest's in-memory caching when in a TTY.

fubuloubu:
> We could disable the manifest's in-memory caching when in a TTY.

I don't think we have to, like @bilbeyt said it is a little weird that this is the behavior. However, some people may be using this. So while it is technically "breaking" I don't think this was a defined "feature" prior, more of a side effect, so we just need to make people generally aware that this behavior changes with whatever release we put this in

fubuloubu:
@unparalleled-js please follow up on this one tomorrow

unparalleled-js (approved):
@bilbeyt In `docs/userguides/console.md`, can you put a small note below the first `bash` code block that says something similar to:

```
WARNING: Contract changes are not reflected in the active console session.
If you need to make changes to your contract, you must re-start your console session for the compiler to handle the changes.
```

Then, I will make sure either I or @violetsummerzine can call this change out in a tweet, like:

```
Ape just got a LOT faster at compiling and accessing compiled artifacts! One small caveat is you may need to restart your console session when making contract-source code changes.
```

bilbeyt:
> @bilbeyt In `docs/userguides/console.md`, can you put a small note below the first `bash` code block that says something similar to:
> 
> ```
> WARNING: Contract changes are not reflected in the active console session.
> If you need to make changes to your contract, you must re-start your console session for the compiler to handle the changes.
> ```
> 
> Then, I will make sure either I or @violetsummerzine can call this change out in a tweet, like:
> 
> ```
> Ape just got a LOT faster at compiling and accessing compiled artifacts! One small caveat is you may need to restart your console session when making contract-source code changes.
> ```

Done

fubuloubu (approved):

unparalleled-js (approved):

istankovic:
Amazing improvement, thanks @bilbeyt, @unparalleled-js and @fubuloubu! :purple_heart:

fubuloubu:
Well get this one released formally either today or tomorrow

violetsummerzine:
Great optimization!!!

PR #1299, opened 2023-02-14
feat: Allow any mnemonic to be used with generate_dev_accounts [APE-625]

bilbeyt:
Refactored generate_dev_account to allow any hd path and mnemonic to be used from config

fixes: #

### Checklist

### DIFF
--- a/src/ape_test/__init__.py
+++ b/src/ape_test/__init__.py
@@ -5,7 +5,7 @@ from pydantic import PositiveInt
 from ape import plugins
 from ape.api import PluginConfig
 from ape.api.networks import LOCAL_NETWORK_NAME
-from ape.utils import DEFAULT_NUMBER_OF_TEST_ACCOUNTS, DEFAULT_TEST_MNEMONIC
+from ape.utils import DEFAULT_HD_PATH, DEFAULT_NUMBER_OF_TEST_ACCOUNTS, DEFAULT_TEST_MNEMONIC
 
 from .accounts import TestAccount, TestAccountContainer
 from .provider import LocalProvider
@@ -55,6 +55,11 @@ class Config(PluginConfig):
     disconnect_providers_after: bool = True
     """
     Set to ``False`` to keep providers connected at the end of the test run.
+    """
+
+    hd_path: str = DEFAULT_HD_PATH
+    """
+    The hd_path to use when generating the test accounts.
     """
 
 
--- a/src/ape/utils/__init__.py
+++ b/src/ape/utils/__init__.py
@@ -44,6 +44,7 @@ from ape.utils.os import (
 )
 from ape.utils.process import JoinableQueue, spawn
 from ape.utils.testing import (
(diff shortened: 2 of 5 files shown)


src/ape_test/__init__.py
```diff
@@ -57,6 +57,8 @@ class Config(PluginConfig):
     Set to ``False`` to keep providers connected at the end of the test run.
     """
 
+    hd_path: str = "m/44'/60'/0'"
```
fubuloubu, line 60: default path is `m/44'/60'/0'/0/{}`
https://github.com/ethereum/eth-account/blob/master/eth_account/hdaccount/__init__.py#L12

fubuloubu (commented):
Similar to what I was saying in discord, I think you might get an efficiency boost by refactoring `generate_account` more so that it only produces as many accounts as you need it to and not all at once

src/ape_test/__init__.py
unparalleled-js, line 60: - [ ] Add property doc str

unparalleled-js (commented):
Thanks for another contribution! Can you add a test? I would like to see the use-case in action. It might help me do some further manual testing as well.

src/ape_test/__init__.py
unparalleled-js, line 60: This should be made into a global variable next to `DEFAULT_TEST_MNEMONIC` and `DEFAULT_NUMBER_OF_TEST_ACCOUNTS`

Then, `generate_dev_accounts` should make use of it for its `hd_path_format` default kwarg.

This way, it's all in 1 spot.

unparalleled-js: Yes, same as this one: https://github.com/ApeWorX/ape/blob/main/src/ape/utils/testing.py#L40
It should be driven from 1 spot, as a constant, like we do for the mnemonic and number of accounts (See my other comment!)

bilbeyt: this value was there before me that is why i kept it, now will just change it

bilbeyt:
> Thanks for another contribution! Can you add a test? I would like to see the use-case in action. It might help me do some further manual testing as well.

added some functional tests

bilbeyt:
> Similar to what I was saying in discord, I think you might get an efficiency boost by refactoring `generate_account` more so that it only produces as many accounts as you need it to and not all at once

for now, this is not necessary for us as it can be also controlled by directly num_of_accounts.

bilbeyt:
@unparalleled-js @fubuloubu now i know why the value was different as hd_path, the thing is eth_tester hd_path is like that https://github.com/ethereum/eth-tester/blob/master/eth_tester/backends/pyevm/main.py#L127 and now lots of tests are failing so i will change it back to old value

unparalleled-js (approved):
LGTM!

fubuloubu:
Maybe we can release this one so that https://github.com/ApeWorX/ape-ganache/pull/22 can be made to work?

PR #1300, opened 2023-02-14
feat: sources and source path lookup improvements [APE-628]

unparalleled-js:
### What I did

* Adds `sources` property to `ProjectManager` so I can get source content by source ID easier.
* Improves `lookup_path` math to handle source ID arg types
* Fixes more issues related to directories with matching source file extensions getting sucked into places they shouldn't

e.g.:

```python
# Get line content from a Source in Ape
source = project.sources[ct.source_id]
line = source[line_index]

# Lookup real path to a source using a source ID.
src_path = project.lookup_path(ct.source_id)
```

### How I did it

* use internal project api `classmethod` for building source dict. NOTE: No compilation is necessary!
* wrap arg in `Path()` so it always works.

### Checklist

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -110,6 +110,7 @@ class ProjectAPI(BaseInterfaceModel):
                 contract_type.name = contract_name
 
             contracts[contract_type.name] = contract_type
+
         return contracts
 
     @property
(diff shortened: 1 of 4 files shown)


unparalleled-js:
- [ ] leverage `ethpm_types.Source`

fubuloubu (approved):

sabotagebeats (approved):

PR #1305, opened 2023-02-17
refactor: remove account history method from ExplorerAPI [APE-637]

fubuloubu:
### What I did

Removes a method from ExplorerAPI that is better off routing through QueryAPI instead

depends on: https://github.com/ApeWorX/ape/pull/1277 (rebase after merge)

depends on: https://github.com/ApeWorX/ape-etherscan/pull/71 (must be released first)

### Checklist


fubuloubu:
NOTE: Removing methods from API classes aren't breaking changes because it reduces the set of features that are required by plugins to implement (meaning they will simply be dead code afterwards that can be pruned at their leisure)

unparalleled-js (approved):

PR #1306, opened 2023-02-17
refactor: improve query manager [APE-639]

fubuloubu:
### What I did

A couple of small improvements to the QueryManager / default query provider, and one bug fix to improve estimation of single item queries (it previously was replying 0)

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel, NonNegativeInt, PositiveInt, root_validator
 
 from ape.logging import logger
 from ape.types import AddressType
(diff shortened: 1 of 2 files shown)


unparalleled-js (approved):
Since there is a bug fix, we should highlight that in the PR title and changelog over the refactor.

fubuloubu:
> Since there is a bug fix, we should highlight that in the PR title and changelog over the refactor.

The bug fix doesn't cause any faults, it was just short-circuting the query manager to whatever is the first query engine registered that returns 0 for an estimate

helloibis (approved):

PR #1309, opened 2023-02-19
perf: add fastentrypoints

NotPeopling2day:
### What I did

fixes: #

### Checklist


antazoey (approved):
On my local machine, `ape --help` takes over 8 seconds with or without this PR, and the majority of the time is spent from setting up the `Web3Provider` and `Ethereum` ecosystem, it seems.

We ultimately need a way to not load all plugins at start up, perhaps by utilizing pluggy's features correclty (@fubuloubu 's suggestion).

I don't see a harm in adding this, but it currently doesn't do anything because we stiill end up loading everything because of the plugin system's current approach

fubuloubu (changes requested):
Without a demonstrable improvement in loading time, I'd like to avoid adding dependencies

PR #1314, opened 2023-02-20
fix: handle missing contract type JSONs [APE-650]

unparalleled-js:
### What I did

Follow up to https://github.com/ApeWorX/ape/pull/1298

Right now, if you migrate from 0.6.2 to the soon-to-be 0.6.3, you will get issues where Ape cannot find your contracts because of some recent in-memory caching advancements. This PR does 2 things:

1. Allows migration to 0.6.3 by searching for the missing contract and recompiling it
2. Handles the case where if you manually delete a file JSON from your .build cache, it will re-compile it and give it back to you.

### How to verify it

* Install ape 0.6.2 and compile a project
* Switch to ape env with this branch installed
* Try to run a script or tests that uses one of the compiled scripts
* Before, it would fail saying the contract was not found, which may be confusing. Now, it will seemlessly work like nothing every happened (migration).
Additionally, if you delete a contract type JSON, Ape will now notice that and recompile them and things will still work as well.

**Basically, there is no reason to get a `ContractTypeNotFound` error in a local project for your own contracts when they actually exist.**

### Checklist

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -95,11 +95,13 @@ class ProjectAPI(BaseInterfaceModel):
         if it exists and is valid.
         """
         if self._cached_manifest is None:
-            manifest = _load_manifest_from_file(self.manifest_cachefile)
-            if manifest is not None:
-                manifest.contract_types = self.contracts
-            self._cached_manifest = manifest
-        return self._cached_manifest
+            self._cached_manifest = _load_manifest_from_file(self.manifest_cachefile)
+            if self._cached_manifest is None:
+                return None
+
+        manifest = self._cached_manifest
+        manifest.contract_types = self.contracts
+        return manifest
 
     @property
     def contracts(self) -> Dict[str, ContractType]:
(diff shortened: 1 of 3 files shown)


unparalleled-js:
trying to find a better approach

src/ape/api/projects.py
```diff
@@ -95,11 +95,29 @@ def cached_manifest(self) -> Optional[PackageManifest]:
...
-            self._cached_manifest = manifest
-        return self._cached_manifest
+            self._cached_manifest = _load_manifest_from_file(self.manifest_cachefile)
+            if self._cached_manifest is None:
+                return None
+
+        manifest = self._cached_manifest
+        if manifest.contract_types and not self.contracts:
+            # Extract contract types from cached manifest.
+            # This helps migrate to >= 0.6.3.
```
fubuloubu, line 105: Attitude note to remove in 0.7.0

fubuloubu (dismissed):

src/ape/api/projects.py
unparalleled-js: OMG, "Add a TODO note" = "Attitude note", always now.

fubuloubu: lmfao whoops

fubuloubu (approved):

fubuloubu (approved):

PR #1315, opened 2023-02-21
feat: `run_until_complete()` support multiple coroutines [APE-653]

unparalleled-js:
### What I did

Needing to parallelize something potentially and wanted to use this method. However, it only supports 1 coroutine at a time, which isn't super helpful for parallelization. 
Also, `ape-starknet` already had to implement this, so delete it from there and use this on next release.

### Checklist

### DIFF
--- a/src/ape/utils/misc.py
+++ b/src/ape/utils/misc.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import sys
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
Wasn't aware we have async methods

PR #1316, opened 2023-02-24
feat: implement contract log container [APE-662]

bilbeyt:
### What I did

Added more functionality for tx.events

fixes: #

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
 
 from eth_abi.abi import encode
 from eth_abi.packed import encode_packed
@@ -24,6 +24,9 @@ from web3.types import FilterParams
 from ape.types.signatures import MessageSignature, SignableMessage, TransactionSignature
 from ape.types.trace import CallTreeNode, GasReport, TraceFrame
 from ape.utils.misc import to_int
+
+if TYPE_CHECKING:
+    from ape.contracts import ContractEvent
 
 BlockID = Union[int, HexStr, HexBytes, Literal["earliest", "latest", "pending"]]
 """
@@ -264,6 +267,23 @@ class ContractLog(BaseModel):
         return self.event_arguments.get(item, default)
 
 
+class ContractLogContainer(list):
+    """
+    Container for ContractLogs which is adding capability of filtering logs
+    """
+
+    def filter(self, event: "ContractEvent", **kwargs) -> List[ContractLog]:
+        found_events = []
+        for log in self:
+            if log.event_name == event.name:
+                match = all(
+                    v == log.event_arguments.get(k) and v is not None for k, v in kwargs.items()
+                )
+                if match:
+                    found_events.append(log)
+        return found_events
+
+
 __all__ = [
     "ABI",
     "AddressType",
@@ -273,6 +293,7 @@ __all__ = [
     "Checksum",
     "Compiler",
     "ContractLog",
+    "ContractLogContainer",
     "ContractType",
     "GasReport",
     "MessageSignature",
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -16,7 +16,7 @@ from pydantic import BaseModel, Field, root_validator, validator
 from ape.api import ReceiptAPI, TransactionAPI
 from ape.contracts import ContractEvent
 from ape.exceptions import OutOfGasError, SignatureError, TransactionError
-from ape.types import CallTreeNode, ContractLog
+from ape.types import CallTreeNode, ContractLog, ContractLogContainer
 from ape.utils import cached_property
 
 
@@ -235,13 +235,15 @@ class Receipt(ReceiptAPI):
         abi: Optional[
             Union[List[Union[EventABI, "ContractEvent"]], Union[EventABI, "ContractEvent"]]
         ] = None,
-    ) -> List[ContractLog]:
+    ) -> ContractLogContainer:
         if abi is not None:
             if not isinstance(abi, (list, tuple)):
                 abi = [abi]
 
             event_abis: List[EventABI] = [a.abi if not isinstance(a, EventABI) else a for a in abi]
-            return list(self.provider.network.ecosystem.decode_logs(self.logs, *event_abis))
+            return ContractLogContainer(
+                self.provider.network.ecosystem.decode_logs(self.logs, *event_abis)
+            )
 
         else:
             # If ABI is not provided, decode all events
@@ -253,7 +255,7 @@ class Receipt(ReceiptAPI):
                 for address, contract in contract_types.items()
             }
 
-            decoded_logs: List[ContractLog] = []
+            decoded_logs: ContractLogContainer = ContractLogContainer()
             for log in self.logs:
                 contract_address = log["address"]
                 if contract_address not in selectors:
@@ -268,7 +270,7 @@ class Receipt(ReceiptAPI):
                         decoded_logs.append(library_log)
                 else:
                     decoded_logs.extend(
-                        list(self.provider.network.ecosystem.decode_logs([log], event_abi))
+                        self.provider.network.ecosystem.decode_logs([log], event_abi)
                     )
 
             return decoded_logs
--- a/tests/functional/test_contract_event.py
+++ b/tests/functional/test_contract_event.py
@@ -281,3 +281,9 @@ def test_contract_decode_logs_no_abi(owner, contract_instance):
     assert events[0].event_name == "NumberChange"
     assert events[0].newNum == 1
     assert events[0].transaction_index == 0
+
+
+def test_contract_log_container(owner, contract_instance):
+    receipt = contract_instance.setNumber(1, sender=owner)
+    events = receipt.events.filter(contract_instance.NumberChange, newNum=1)
+    assert len(events) == 1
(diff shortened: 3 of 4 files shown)


tests/functional/test_contract_event.py
```diff
@@ -281,3 +281,9 @@ def test_contract_decode_logs_no_abi(owner, contract_instance):
     assert events[0].event_name == "NumberChange"
     assert events[0].newNum == 1
     assert events[0].transaction_index == 0
+
+
+def test_contract_log_container(owner, contract_instance):
+    receipt = contract_instance.setNumber(1, sender=owner)
+    assert "NumberChange" in receipt.events
+    assert receipt.events["NumberChange"]["newNum"] == 1
```
fubuloubu, line 289: My idea was something more akin to this:
```suggestion
    # at least 1 NumberChange event with arg `newNum` set to 1
    assert contract_instance.NumberChange(newNum=1) in receipt.events
```
Allowing more strict comparisons like this:
```py
    # exactly 1 event (NumberChange) with arg `newNum` set to 1
    assert receipt.events == [contract_instance.NumberChange(newNum=1)]
```

fubuloubu (commented):
I think this change is pointing at a more Brownie-style events assertion system, but I think we can do better than that

tests/functional/test_contract_event.py
bilbeyt: but if i have lots of attributes in the event and i dont care all about them so actually dont want to create the object but only care lets say 1 attribute then i am stuck

fubuloubu: you could also do this:

```py
    # exactly 1 event (NumberChange), don't care what `newNum` is set to
    assert receipt.events == [contract_instance.NumberChange()]
```

fubuloubu: hmm, I do like this approach!

except call it `.filter` instead of `.find`

if it still returns a list we can leverage it's truthiness via:

```py
assert receipt.events.filter(contract.NumberChange, oldNum=1, newNum=2)
```

or if you want an exact check:

```py
assert len(receipt.events.filter(contract.NumberChange, oldNum=1, newNum=2)) == 1
```

fubuloubu: > it requires one to specify all the event attributes in order for the event to match, even in cases where you only care about event type

Just responding to this, the intent was to have un-specified event args work as wildcards. It would be a bit of work to get a proper wildcard system working

bilbeyt: the current implementation should be ok

src/ape/types/__init__.py
```diff
@@ -264,6 +264,23 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
...
+    def filter(self, event, **kwargs) -> List[ContractLog]:
+        found_events = []
+        for log in self:
+            if log.event_name == event.name:
+                found = True
+                for k, v in kwargs.items():
+                    if log.event_arguments.get(k) != v:
+                        found = False
+                        break
+                if found:
```
istankovic, line 279: This has a slight issue in that it would incorrectly match non-existent attributes if the value is `None`, e.g. `.filter(foo=None)`.

A fairly common patter to address this is something like
```python
_NONE = object() # unique object to detect non-existent entries, somewhere on the module level
match = all(v == log.event_arguments.get(k, _NONE) for k, v in kwargs.items())
if match:
  found_events.append(log)
```

bilbeyt: changed.

src/ape/types/__init__.py
```diff
@@ -264,6 +267,19 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
         return self.event_arguments.get(item, default)
 
 
+class ContractLogContainer(list):
+    __test__ = False
```
fubuloubu, line 271: What's this for?

src/ape/types/__init__.py
```diff
@@ -264,6 +267,19 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
...
+
+    def filter(self, event, **kwargs) -> List[ContractLog]:
```
fubuloubu, line 273: Needs a docstrings and typing

src/ape/types/__init__.py
```diff
@@ -264,6 +267,19 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
...
+        found_events = []
+        for log in self:
+            if log.event_name == event.name:
+                match = all(v == log.event_arguments.get(k, _NONE) for k, v in kwargs.items())
```
fubuloubu, line 277: Wouldn't this work the same? 
```suggestion
                match = all(v == log.event_arguments.get(k, None) for k, v in kwargs.items())
```

src/ape_ethereum/transactions.py
```diff
@@ -235,13 +235,15 @@ def decode_logs(
...
-    ) -> List[ContractLog]:
+    ) -> ContractLogContainer:
         if abi is not None:
             if not isinstance(abi, (list, tuple)):
                 abi = [abi]
 
             event_abis: List[EventABI] = [a.abi if not isinstance(a, EventABI) else a for a in abi]
-            return list(self.provider.network.ecosystem.decode_logs(self.logs, *event_abis))
+            return ContractLogContainer(
+                list(self.provider.network.ecosystem.decode_logs(self.logs, *event_abis))
```
fubuloubu, line 245: Since `ContractLogContainer` subclasses `list`, you can just do this now

```suggestion
                self.provider.network.ecosystem.decode_logs(self.logs, *event_abis)
```

src/ape_ethereum/transactions.py
```diff
@@ -271,7 +273,7 @@ def decode_logs(
                         list(self.provider.network.ecosystem.decode_logs([log], event_abi))
```
fubuloubu, line 273:
```suggestion
                         self.provider.network.ecosystem.decode_logs([log], event_abi)
```

fubuloubu (commented):
This is very slick, I like it

Needs a few updates to be production ready

src/ape/types/__init__.py
unparalleled-js: The only time I have used this is when my class name starts with the word `Test` and I don't want `pytest` picking it up.

src/ape/types/__init__.py
unparalleled-js: `None` is the default so you wouldn't need to specify, but ya wondering why we need the `_None` object as well.

bilbeyt: as @istankovic explained, if we have expression like .filter(event, foo=None) which foo doesnt exist in event then log.event_arguments.get("foo", None) == None which would be true and we could see some unwanted things in response. using _None makes sure the log wont be there.

src/ape_ethereum/transactions.py
bilbeyt: somehow this wasnt working actually that is why because it assumes type is just list which has no filter method

src/ape/types/__init__.py
bilbeyt: removed test part.

src/ape/types/__init__.py
bilbeyt: if i give type for event it would be ContractEvent and .contracts.base importing ContractLog then this is a import cycle. and i cant just use "ContractEvent" if i dont import it some point and kwargs i dont have any type basically, but will add docstring

src/ape/types/__init__.py
fubuloubu: > The only time I have used this is when my class name starts with the word `Test` and I don't want `pytest` picking it up, such as `TestAccountContainer`.

Ahhhhh, so that's how you do it

fubuloubu: I think some stuff in `ape_test` needs this

src/ape/types/__init__.py
unparalleled-js: Could you do:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from ape.contracts.base import ContractEvent

...

def filter(self, event: "ContractEvent", **kwargs):

Comment (fubuloubu) [reply src/ape/types/__init__.py:275]: Yeah, you don't need a type for kwargs, although I think `Dict[str, Any] ` might work

Comment (fubuloubu) [reply src/ape/types/__init__.py:275]: BTW for import cycles you can solve that using quotes around your type

File: src/ape/types/__init__.py (hunk 5 above)
Comment (fubuloubu) [reply src/ape/types/__init__.py:279]: Can just assert that the value×in the kwarg is not `None`

File: src/ape/types/__init__.py (hunk 4 above)
Comment (unparalleled-js) [reply src/ape/types/__init__.py:275]: Types on `**kwargs` can be a little unintuitive. I agree, don't worry about them; we don't.

But, just so you know, for `*args` and `**kwargs`, you can only put the type of the value(s), like `*args: Any` or `**kwargs: Any`, you don't do `*args: List[Any]` or `**kwargs: Dict[str, Any]`.

Comment (bilbeyt) [reply src/ape/types/__init__.py]: yes, i found this example in code regarding TYPE_CHECKING so it is ok now.

File: src/ape_ethereum/transactions.py (hunk 6 above)
Comment (bilbeyt) [reply src/ape_ethereum/transactions.py]: now working

File: src/ape/types/__init__.py (hunk 5 above)
Comment (bilbeyt) [reply src/ape/types/__init__.py]: also works so changed

File: src/ape_ethereum/transactions.py [hunk 8]
```diff
@@ -268,7 +270,9 @@ def decode_logs(
                         decoded_logs.append(library_log)
                 else:
                     decoded_logs.extend(
-                        list(self.provider.network.ecosystem.decode_logs([log], event_abi))
+                        ContractLogContainer(
```
unparalleled-js, line 273: Could we override `def extend()` in `ContractLogContainer` to auto-wrap the given iterator so we can do:

```python
decoded_logs.extend(self.provider.network.ecosystem.decode_logs([log], event_abi)
```

bilbeyt: worked directly

unparalleled-js (approved):

unparalleled-js:
All feedback has been addressed! I tested this and things seem good.

Issue #1289, opened 2023-02-07
regression: can no longer convert hex strings to bytes ABI types


fubuloubu:
### Environment information
https://github.com/ApeAcademy/ERC721/actions/runs/4118991819/jobs/7112166914

### What went wrong?
This test used to pass pre-v0.6.0

### How can it be fixed?
Think that the arg converter is no longer picking up this conversion

fubuloubu:
@vany365 get this one in the queue, multiple people reporting being affected by it

PR #1319, opened 2023-02-25
fix: issue with providing str for bytes in calldata [APE-667]

unparalleled-js:
### What I did

Just opening this to show @Ninjagod1251 what to do and get it started.
See comments for more info!

fixes: #1289
fixes: APE-627

### Checklist


src/ape_ethereum/ecosystem.py
```diff
@@ -345,7 +345,11 @@ def encode_calldata(self, abi: Union[ConstructorABI, MethodABI], *args) -> HexBy
         parser = StructParser(abi)
         arguments = parser.encode_input(args)
         input_types = [i.canonical_type for i in abi.inputs]
-        converted_args = self.conversion_manager.convert(arguments, tuple)
+
+        # TODO: Convert input_types to python_types. e.g. uint256 -> int, bytes4 -> HexBytes etc.
```
unparalleled-js, line 349: @Ninjagod1251 I saved the hardest part for you to figure out!
We need to derive the Pythonic types from the ABI input types here so we can pass them into the conversion manager.

Like `bytes4` should be `HexBytes`
`address` -> `AddressType`
etc.

tests/functional/test_ecosystem.py
```diff
@@ -52,6 +53,26 @@ def test_encode_address(ethereum):
...
+            ABIType(name="d", type="bytes4[]"),
+        ],
+    )
+    address = "0xd8da6bf26964af9d7eed9e03e53415d37aa96045"
+    byte_array = ["0x456", "0x678"]
+    values = ("0x123", address, HexBytes(55), byte_array)
+
+    actual = ethereum.encode_calldata(abi, *values)
+    expected = [HexBytes("0x123"), address, 55, [HexBytes(x) for x in byte_array]]
+    assert actual == expected
```
unparalleled-js, line 73: @Ninjagod1251 This test fails now but should pass if we get this working, but also feel free to ad just any of this as needed and we can also chat about this on Monday. Thanks

src/ape/managers/converters.py
```diff
@@ -296,7 +298,22 @@ def convert(self, value: Any, type: Type) -> Any:
             any: The same given value but with the new given type.
         """
 
-        if type not in self._converters:
+        if isinstance(value, (list, tuple)) and isinstance(type,tuple):
+            # We expected to convert a tuple type, so convert each item in the tuple.
+            # NOTE: We allow values to be a list, just incase it is a list
```
unparalleled-js, line 303:
```suggestion
            # NOTE: We allow values to be a list, just in case it is a list
```

src/ape/managers/converters.py
```diff
@@ -27,7 +27,7 @@ class HexConverter(ConverterAPI):
     """
     A converter that converts ``str`` to ``HexBytes``.
     """
-
+
```
unparalleled-js, line 30: some whitespace hanging out here

src/ape/managers/converters.py
```diff
@@ -296,7 +298,22 @@ def convert(self, value: Any, type: Type) -> Any:
             any: The same given value but with the new given type.
         """
 
-        if type not in self._converters:
+        if isinstance(value, (list, tuple)) and isinstance(type,tuple):
```
unparalleled-js, line 301:
```suggestion
        if isinstance(value, (list, tuple)) and isinstance(type, tuple):
```

unparalleled-js (commented):
Looks good! I can't approve it won't let me, just run the linters! thank you :)

fubuloubu:
Note: in debuging the implementation for this with @Ninjagod1251, I figured out that https://github.com/ApeWorX/ethpm-types/issues/63 would be very helpful for some of the code we added

src/ape/managers/converters.py
```diff
@@ -278,7 +280,7 @@ def is_type(self, value: Any, type: Type) -> bool:
         else:
             return isinstance(value, type)
 
-    def convert(self, value: Any, type: Type) -> Any:
+    def convert(self, value: Any, type: Union[Type, Tuple[Type], List[Type]]) -> Any:
```
Ninjagod1251, line 283: error with pre-commit `(expression has type "Tuple[Any, ...]", variable has type "Tuple[Type[Any]]")`

any suggestions? @unparalleled-js

src/ape/managers/converters.py
```diff
@@ -246,6 +246,8 @@ def _converters(self) -> Dict[Type, List[ConverterAPI]]:
             Decimal: [],
             list: [ListTupleConverter()],
             tuple: [ListTupleConverter()],
+            bool: [],
+            str: [],
```
fubuloubu, line 250: Also had to add these so that we could convert these types without issue. There's no converters I can think of for these types

src/ape_ethereum/ecosystem.py
```diff
@@ -338,14 +338,46 @@ def decode_block(self, data: Dict) -> BlockAPI:
 
         return Block.parse_obj(data)
 
+    def _python_type_for_abi_type(self, abi_type: ABIType) -> Any[Type[Any]]:
+        # NOTE: An array can be an array of tuples, so we start with an array check
+        if abi_type.type.endswith("]"):
+            # remove one layer of the potential onion of array
+            new_type = "[".join(str(abi_type.type).split("[")[:-1])
```
fubuloubu, line 345: This is the code that I think would be better to fix in `ethpm_types.abi.ABIType` by adding a method to fetch the inner array type

src/ape/managers/converters.py
unparalleled-js: Try `Union[Type, Tuple[Type, ...], List[Type]]

unparalleled-js: `Tuple` wants to know how many, but if you do `...` it says it can be any length.

src/ape_ethereum/ecosystem.py
unparalleled-js: Do itttttt

src/ape/managers/converters.py
```diff
@@ -246,6 +246,8 @@ def _converters(self) -> Dict[Type, List[ConverterAPI]]:
...
         }
```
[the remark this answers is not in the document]
unparalleled-js: The `ListTupleConverter` is potentially used outside of Core via `convert(thing, tuple)`

src/ape_ethereum/ecosystem.py
```diff
@@ -338,14 +338,49 @@ def decode_block(self, data: Dict) -> BlockAPI:
...
+        elif abi_type.type == "string":
+            return str
+
+        elif "bytes" in abi_type.type:
+            return bytes
+
+        elif "int" in abi_type.type:
+            return int
+
+        raise
```
unparalleled-js, line 370: You can't use an empty raise like this!

```
>>> raise
  File "<stdin>", line 1, in <module>
RuntimeError: No active exception to reraise
>>>
```

Please use an exception class

unparalleled-js (commented):
empty raise is not legal unless an exception is already present on the Exception stack

src/ape_ethereum/ecosystem.py
unparalleled-js: (if there was an exception on the stack, it would work, but in this case, that isn't it)

unparalleled-js (commented):
Approved

src/ape/managers/converters.py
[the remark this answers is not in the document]
fubuloubu: We should make a point to remove this in v0.7

fubuloubu (dismissed):

fubuloubu:
Fix the lining issue and can merge

fubuloubu (approved):
Let's gooooo

Ninjagod1251 (approved):

Issue #867, opened 2022-07-11
Show uncaught VirtualMachineError / TransactionError revert lines


antazoey:
### Overview

Intercept stack trace during unhandled VirtualMachineError or TransactionError and show only the line of revert, 

### Specification

![brvt](https://user-images.githubusercontent.com/19540978/178281328-49cab1d5-a6f6-44d4-b07b-b5bad84b0fd5.jpg)

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

unparalleled-js:
@banteg What is all needed to start development on this?  Do we need the coverage-setup or anything else with tracing support first?

Also, is there anything else we should know about this ticket before beginning development?

fubuloubu:
> @banteg What is all needed to start development on this? Do we need the coverage-setup or anything else with tracing support first?
> 
> Also, is there anything else we should know about this ticket before beginning development?

You need sourcemapping and/or ast layout and/or Vyper's error map to find the PC that a revert corresponds to, and the source code that maps to whatever the program counter of the revert is

@helloibis is investigating some of this

helloibis:
@unparalleled-js @fubuloubu what I've uncovered a bit of as part of the other ticket is that often times we'll have `VirtualMachineError` or `TransactionError` exceptions that arise before the transaction is actually committed. In such cases, it doesn't seem like I can actually get a stack trace (even if I do have a `txn_hash`). One example is the issue I posted in the Discord where I was unable to deploy a contract due to insufficient gas, the trace info just comes up empty.

Should we consider this feature something that is only shown if we can access the trace info (for instance it seems we can always do it for `ContractLogicError`), and otherwise doesn't show additional information?

cc @NotPeopling2day

fubuloubu:
> Should we consider this feature something that is only shown if we can access the trace info (for instance it seems we can always do it for `ContractLogicError`), and otherwise doesn't show additional information?

I think this makes sense, or at least if we can start here and figure out something over time for the other exception types it'll work

helloibis:
Status update: I'm not sure if this ticket is necessarily blocked, but definitely after https://github.com/ApeWorX/ape/issues/937 is completed this should be easier to implement and work in all `TransactionError` cases (other than `OutOfGasError` of course)

PR #1337, opened 2023-03-13
feat: show source code lines / traceback from `ReceiptAPI` [APE-708]

antazoey:
### What I did

fixes: #867 
Fixes: APE-217
Fixes: APE-577

* Adds `traceback` property to `ReceiptAPI` with implementation in `ape_ethereum`
* Adds `show_source_traceback()` method to `ReceiptAPI` the same.

### How I did it

* Using AST, trace execution, and PCMap. Only works with Vyper at first.

### How to verify it

Create `interfaces/IRegistryF.vy`:

```python
# @version 0.3.7

@external
def register_f(addr: address):
    pass
```

Create `RegistryF.vy`:

```python
# @version 0.3.7

addr: public(address)

@external
def register_f(addr2: address):
    assert self.addr != addr2, "doubling."
    self.addr = addr2
```

Create `JustFail.vy`:

```python
# @version 0.3.7

import interfaces.IRegistryF as IRegistryF

_balance: public(uint256)
registry: public(IRegistryF)

@external
def __init__(registry: IRegistryF):
    self.registry = registry

@external
def addBalance_f(
    num: uint256
) -> uint256:
    assert num != self._balance
    self.registry.register(msg.sender)
    self._balance = self._balance + self.addInterest(num)

    # Run some loops.
    for i in [1, 2, 3, 4, 5]:
        if i == num:
            break

    # Fail in the middle (is test)
    # Fails because was already set above.
    self.registry.register_f(msg.sender)

    for i in [1, 2, 3, 4, 5]:
        if i != num:
            continue

    return self._balance

@internal
def addInterest(num: uint256) -> uint256:
    return 123 + num
```

Create a script `tb.py`:

```python
import ape
from ape.exceptions import ContractLogicError


def main():
    account = ape.accounts.test_accounts[0]
    registry = ape.project.RegistryF.deploy(sender=account)
    contract = ape.project.JustFail.deploy(registry, sender=account)
    txn = contract.addBalance_f.as_transaction(123)

    try:
        account.call(txn)
    except ContractLogicError:
        pass

    receipt = txn.provider.get_receipt(txn.txn_hash.hex())
    receipt.show_source_traceback()
```

It should show:

```shell
Traceback (most recent call last)
  File /home/ApeProjects/ape-playground/contracts/JustFail.vy, in addBalance_f(uint256 num) -> uint256
       20     # Run some loops.
       21     for i in [1, 2, 3, 4, 5]:
       22         if i == num:
       23             break
       24
       25     # Fail in the middle (is test)
       26     # Fails because was already set above.
  -->  27     self.registry.register_f(msg.sender)
       28
       29     for i in [1, 2, 3, 4, 5]:
       30         if i != num:
       31             continue

  File /home/ApeProjects/ape-playground/contracts/RegistryF.vy, in register_f(address addr2)
  -->  7     assert self.addr != addr2, "doubling."
       8     self.addr = addr2
```

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -19,7 +19,13 @@ from ape.exceptions import (
     TransactionNotFoundError,
 )
 from ape.logging import logger
-from ape.types import AddressType, ContractLogContainer, TraceFrame, TransactionSignature
+from ape.types import (
+    AddressType,
+    ContractLogContainer,
+    SourceTraceback,
+    TraceFrame,
+    TransactionSignature,
+)
 from ape.utils import BaseInterfaceModel, abstractmethod, cached_property, raises_not_implemented
 
 if TYPE_CHECKING:
@@ -428,6 +434,15 @@ class ReceiptAPI(BaseInterfaceModel):
 
         return output
 
+    @property
+    @raises_not_implemented
+    def source_traceback(self) -> SourceTraceback:  # type: ignore[empty-body]
+        """
+        A pythonic style traceback for both failing and non-failing receipts.
+        Requires a provider that implements
+        :meth:~ape.api.providers.ProviderAPI.get_transaction_trace`.
+        """
+
     @raises_not_implemented
     def show_trace(self, verbose: bool = False, file: IO[str] = sys.stdout):
         """
@@ -443,6 +458,14 @@ class ReceiptAPI(BaseInterfaceModel):
     def show_gas_report(self, file: IO[str] = sys.stdout):
         """
         Display a gas report for the calls made in this transaction.
+        """
+
+    @raises_not_implemented
+    def show_source_traceback(self):
+        """
+        Show a receipt traceback mapping to lines in the source code.
+        Only works when the contract type and source code are both available,
+        like in local projects.
         """
 
     def track_gas(self):
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -5,6 +5,7 @@ from collections import deque
 from functools import cached_property
 from inspect import getframeinfo, stack
 from pathlib import Path
+from types import CodeType, TracebackType
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
 
 import click
@@ -19,7 +20,7 @@ if TYPE_CHECKING:
     from ape.api.networks import NetworkAPI
     from ape.api.providers import SubprocessProvider
     from ape.api.transactions import TransactionAPI
-    from ape.types import AddressType, BlockID, SnapshotID, TraceFrame
+    from ape.types import AddressType, BlockID, SnapshotID, SourceTraceback, TraceFrame
 
 
 class ApeException(Exception):
@@ -112,8 +113,23 @@ class TransactionError(ContractError):
         self.txn = txn
         self.trace = trace
         self.contract_address = contract_address
+        self.source_traceback: Optional["SourceTraceback"] = None
         ex_message = f"({code}) {message}" if code else message
+
+        # Finalizes expected revert message.
         super().__init__(ex_message)
+
+        if not txn:
+            return
+
+        ape_tb = _get_ape_traceback(self, txn)
+        if not ape_tb:
+            return
+
+        self.source_traceback = ape_tb
+        py_tb = _get_custom_python_traceback(self, txn, ape_tb)
+        if py_tb:
+            self.__traceback__ = py_tb
 
 
 class VirtualMachineError(TransactionError):
@@ -532,9 +548,10 @@ def handle_ape_exception(err: ApeException, base_paths: List[Path]) -> bool:
     an exception on the exc-stack.
 
     Args:
-        err (:class:`~ape.exceptions.ApeException`): The transaction error
+        err (:class:`~ape.exceptions.TransactionError`): The transaction error
           being handled.
-        base_paths (List[Path]): Source base paths for allowed frames.
+        base_paths (Optional[List[Path]]): Optionally include additional
+          source-path prefixes to use when finding relevant frames.
 
     Returns:
         bool: ``True`` if outputted something.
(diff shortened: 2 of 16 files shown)


src/ape_ethereum/transactions.py
```diff
@@ -157,17 +188,166 @@ def call_tree(self) -> Optional[CallTreeNode]:
...
         method_id = self.data[:4]
-        if method_id not in contract_type.methods:
+        if method_id not in self.contract_type.methods:
             return None
 
-        return contract_type.methods[method_id]
+        return self.contract_type.methods[method_id]
+
+    @property
+    def traceback(self) -> Traceback:
```
unparalleled-js, line 210: The chaos happens here

src/ape/api/transactions.py
```diff
@@ -402,6 +413,14 @@ def show_gas_report(self, file: IO[str] = sys.stdout):
         Display a gas report for the calls made in this transaction.
         """
 
+    @raises_not_implemented
+    def show_traceback(self):
```
fubuloubu, line 417: I think `show_trace` is a little to close to this name, maybe `show_error_traceback` (and corresponding `error_traceback`)

src/ape/types/trace.py
```diff
@@ -219,3 +221,150 @@ class TraceFrame(BaseInterfaceModel):
     """
     The raw trace frame from the provider.
     """
+
+
+class Traceback(list):
```
fubuloubu, line 226: With pydantic, you can do this:

```py
class Traceback(BaseModel):
    __root__: List[TracebackItem]

    ...
```

unparalleled-js: That isn't quite the same! They are still models wrapped around it and not a subclass directly. Example:

```python
from pydantic import BaseModel
from typing import List

class MyList(BaseModel):
    __root__: List[int] = []

class MyStr(BaseModel):
    __root__: str = ""

my_ls = MyList()
my_str = MyStr()
print(f"has append: {hasattr(my_ls, 'append')}")
print(f"str of custom str: {my_str}")
```

Outputs:

```
has append: False
str of custom str: __root__=''
```

This is why in `ethpm-types`, we had to add the `__str__` and `__repr__` methods still when we did `__root__: str`, 'member?

src/ape/api/transactions.py
unparalleled-js: I agree. However, I was thinking - there is no reason to not have this option for non-failing receipts.. Like it still works, you just make it all the way to the `RETURN` opcode in the traceback. We can make it not be red or something.

unparalleled-js:
- [ ] Main thing this needs is thorough testing, both negative and positive.

unparalleled-js:
Observation: would like to get it like IPython where it selectively shows a few lines before or after:'

```
File ~/PycharmProjects/ape/src/ape/managers/chain.py:540, in AccountHistory.__getitem_slice(self, indices)
    537     stop += len(self)
    539 elif stop > len(self):
--> 540     raise ChainError(
    541         f"'stop={stop}' cannot be greater than account's current nonce ({len(self)})."
    542     )
    544 if stop <= start:
    545     return []  # nothing to query
```

I wonder how it does that...

fubuloubu:
> Observation: would like to get it like IPython where it selectively shows a few lines before or after:'
> 
> ```
> File ~/PycharmProjects/ape/src/ape/managers/chain.py:540, in AccountHistory.__getitem_slice(self, indices)
>     537     stop += len(self)
>     539 elif stop > len(self):
> --> 540     raise ChainError(
>     541         f"'stop={stop}' cannot be greater than account's current nonce ({len(self)})."
>     542     )
>     544 if stop <= start:
>     545     return []  # nothing to query
> ```
> 
> I wonder how it does that...
> 
> * [ ]  Might need to delegate some of this to the `CompilerAPI` but need to figure out what exactly...

Could maybe find the next 1-2 layer higher AST nodes and display the breadth of those

src/ape/api/transactions.py
```diff
@@ -385,6 +385,17 @@ def return_value(self) -> Any:
...
+    @property
+    @raises_not_implemented
+    def traceback(self) -> Traceback:
+        """
+        A pythonic style traceback for both failing and non-failing receipts.
+        Requires a provider that implements
+        :meth:~ape.api.providers.ProviderAPI.get_transaction_trace`.
+        """
+
+        return  # type: ignore[return-value]
```
fubuloubu, line 397: Do you need a `return` statement w/ `raises_not_implemented`?

unparalleled-js: If there is a return it needs something, but the other cases we do the type ignore in the signature without the return, which is better, so I will switch it to that

unparalleled-js: done, but has to squash commits because of hefting rebasing (easier to rebase 1 commit), but the change is in there.

src/ape/types/trace.py
unparalleled-js: I ended up doing this!

unparalleled-js:
- [ ] still trying to find way to exclude irrelevant stack frames when err happens in console

src/ape/_cli.py
```diff
@@ -36,14 +46,17 @@ def invoke(self, ctx) -> Any:
...
-            if logger.level == LogLevel.DEBUG.value:
-                tb = traceback.format_exc()
-                err_message = tb or str(err)
-            else:
-                err_message = str(err)
+        except TransactionError as err:
+            if not err.traceback:
+                raise abort(err) from err
+
+            reporter = ctx.obj.chain_manager._reports
```
unparalleled-js, line 53: - [ ] Either use the `__traceback__` that gets set now, or manually include the frame(s) from the script or plugin making the call that causes the tx err

unparalleled-js:
- [ ] Show in pytest plugin

unparalleled-js:
- [ ] Figure out why isn't working on `ape-foundry` and `ape-hardhat` (only been using `ape-geth` thus far)

unparalleled-js:
Marking this as ready for review... the PR is gotten unwieldly, but this feels ready now.

* Can see awesome stacktraces in scripts
* Can see awesome stacktraces in IPython
* Can see awesome stacktraces in Pytest
* Can see (and interact with) awesome stacktraces when using debuggers.

Future needs:

* Compiler-specific handling pieces (will look into that next)
* Scripting, pytest, and console integration tests

unparalleled-js:
- [ ] Compiler API delegation
- [ ] Handle implicit payable checks in Vyper

src/ape/types/trace.py
```diff
@@ -19,6 +22,9 @@
 """
 
 
+_FUNCTION_AST_TYPES = ("FunctionDef", "Name", "arguments")
```
unparalleled-js, line 25: - [ ] find way to delegate to `CompilerAPI`

src/ape/types/trace.py
```diff
@@ -325,7 +325,7 @@ def get_content_asts(self, location: SourceLocation) -> List[ASTNode]:
         return [
             a
             for a in self.ast.get_nodes_at_line(location)
-            if a.lineno >= location[0] and a.classification == ASTClassification.CONTENT
+            if a.lineno >= location[0] and a.classification != ASTClassification.FUNCTION_DEF
```
fubuloubu, line 328: I think this also works, and maybe be suggested for enums
```suggestion
            if a.lineno >= location[0] and a.classification is ASTClassification.FUNCTION_DEF
```

src/ape/managers/chain.py
```diff
@@ -622,15 +624,24 @@ def __getitem_str(self, account_or_hash: str) -> Union[AccountHistory, ReceiptAP
...
+
+            # If we get here, we failed to get an account or receipt.
+            # Raise top-level exception.
+            raise
 
-            return receipt
+    def _get_receipt(self, txn_hash: str) -> ReceiptAPI:
+        receipt = self._hash_to_receipt_map.get(txn_hash)
+        if not receipt:
+            # TODO: Replace with query manager once supports receipts
```
fubuloubu, line 639: Technically it supports receipts right now, it's just misnamed as transactions

unparalleled-js: what is the best way to query a receipt by its known hash?

src/ape/exceptions.py
```diff
@@ -113,9 +114,82 @@ def __init__(
...
+
+        if not stack:
+            return
+
+        tb_next = None
+        for tb in stack:
+            tb.tb_next = tb_next
+            tb_next = tb
+
+        self.__traceback__ = stack[-1]
```
fubuloubu, line 191: Sorta wish this code was a utility function

antazoey: i can move it to a method in this same module but cant put it in utils because of circular deps

fubuloubu: That's good enough

src/ape/types/trace.py
```diff
@@ -225,3 +229,768 @@ class TraceFrame(BaseInterfaceModel):
...
+        return self.__root__.items()
+
+    def as_list(self) -> List[str]:
+        return list(self.__root__.values())
+
+    def __getitem__(self, lineno: int) -> str:
+        return self.__root__[lineno]
+
+    def __iter__(self):
+        yield from self.__root__
```
fubuloubu, line 267: this would iter the line numbers?

fubuloubu (dismissed):
some of the items under `src/ape/types/trace.py` appear like they could live separately under `ethpm-types`, but not sure I want to advocate for that too much

src/ape/types/trace.py
antazoey: updated this to be cooler and still work

antazoey:
> some of the items under `src/ape/types/trace.py` appear like they could live separately under `ethpm-types`, but not sure I want to advocate for that too much

OK I did manage to share some code there and move some of the basic structures there. they're composed of ethpm types anyway so it makes sense and `Content` can be used on `Source` to share some indexing logic there, so long that the basic spec can still be honored in serialization. but it is kind of a breaking change in `ethpm-types`, so we will have to do an ethpm-types 0.5 to proceed. it does help separate of the concerns and makes this PR a little smaller too

src/ape/types/trace.py
```diff
@@ -225,3 +230,428 @@ class TraceFrame(BaseInterfaceModel):
     """
     The raw trace frame from the provider.
     """
+
+
+class ControlFlow(BaseModel):
```
fubuloubu, line 235: Does this class have ape-specific functionality?

fubuloubu (commented):
looking much cleaner now

src/ape/types/trace.py
antazoey: i thought about moving to `ethpm-types` but decided against it because it really is only about execution. it felt weird putting anything that is solely about execution in ethpm-types for some reason, probably because there wasnt good spot for it on its own.

but the answer is no, it is basically a class that only works with classes from ethpm-types

antazoey: it is kindof like the main helper class to the source traceback too, they go kind of hand in hand and make the most sense together, at least that is what i think currently

PR #1341, opened 2023-03-16
fix: contract caching issue

unparalleled-js:
### What I did

Issue preventing https://github.com/ApeWorX/ape-etherscan/pull/74 this from working

### Checklist

### DIFF
--- a/src/ape/managers/project/types.py
+++ b/src/ape/managers/project/types.py
@@ -202,6 +202,11 @@ class BaseProject(ProjectAPI):
                     (self.project_manager.local_project._cache_folder / f"{name}.json").write_text(
                         contract_type.json()
                     )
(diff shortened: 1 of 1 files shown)


src/ape/managers/project/types.py
```diff
@@ -202,6 +202,11 @@ def create_manifest(
                     (self.project_manager.local_project._cache_folder / f"{name}.json").write_text(
                         contract_type.json()
                     )
+                    if self._contracts is None:
```
unparalleled-js, line 205: this was a regression from the recent caching improvements we made recently, causing us to sometimes have to force re-compile even when we shouldn't, like when Solidity contracts use libraries

fubuloubu (approved):

PR #1343, opened 2023-03-17
feat: two new `ape console` Magics [APE-726]

unparalleled-js:
### What I did

Am looking for more IPython integration opportunities with tracebacks.
Ended up here. But look, this is kinda of neat and sets us up to integrate more with IPython, which ultimately is what we need...

* Creates an `%ape` magic for `ape-console` that allows running CLI commands within the `ape console` session:

```shell
ape console --network ethereum:mainnet:alchemy
In [1]: %ape accounts list
Found 1 account:
  0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C (alias: 'test0')

Out[1]: <Result okay>
```

A `%bal` magic that gets a balance in human-readable form:

```shell
In [2]: account = accounts.load("metamask0")

In [3]: %bal account
Out[3]: '0.00040634 ETH'
```

### How I did it

* `start_ipython` for CLI command but use `embed` everywhere else (`ape test -I` and `ape run -I`)
* Register as an official IPython extension
* Add implementations for the said magics

### How to verify it

See above!

Would like to see much testing though and feedback on even if we want this.

### Checklist

### DIFF
--- a/docs/userguides/console.md
+++ b/docs/userguides/console.md
@@ -122,11 +122,9 @@ Commands:
   compile   Compile select contract source files
   console   Load the console
   init      Initalize an ape project
-  ledger    Manage Ledger accounts
   networks  Manage networks
   plugins   Manage ape plugins
   run       Run scripts from the `scripts/` folder
-  template  Create a project from a Cookiecutter project template (TEMPLATE).
   test      Launches pytest and runs the tests for a project
 
 Out[1]: <Result okay>
--- a/docs/userguides/console.md
+++ b/docs/userguides/console.md
@@ -88,6 +88,17 @@ Out[1]: 1
 
 You can also add an `ape_console_extras.py` file to the global ape data directory (`$HOME/.ape/ape_console_extras.py`) and it will execute regardless of what project context you are in.  This may be useful for variables and utility functions you use across all of your projects.
 
+## Configure
+
+To automatically use other IPython extensions, add them to your `ape-config.yaml` file:
+
+```yaml
+console:
+  plugins:
+    # A plugin that lets you modify Python modules without having close/reopen your console.
+    - autoreload
+```
+
 ## Magic Commands
 
 The `ape-console` plugin ships with custom [magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html#line-magics) that are available when running the `ape console` command or loading the `ape_console.plugin` IPython extension manually.
--- a/docs/userguides/console.md
+++ b/docs/userguides/console.md
@@ -90,7 +90,7 @@ You can also add an `ape_console_extras.py` file to the global ape data director
 
 ## Magic Commands
 
-The `ape-console` plugin ships with some custom [magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html#line-magics).
+The `ape-console` plugin ships with custom [magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html#line-magics).
 
 ### %ape
 
@@ -98,8 +98,6 @@ The `%ape` magic invokes the CLI in your `ape-console` session:
 
 ```shell
 In [1]: %ape
-WARNING: Unable to load CLI endpoint for plugin 'ape_starknet'
-        ModuleNotFoundError: No module named 'starknet_py.hash'
 Usage: cli [OPTIONS] COMMAND [ARGS]...
 
 Options:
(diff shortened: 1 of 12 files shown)


unparalleled-js:
It'd be nice to allow plugins to extend this functionality as well, like have them register their own console plugins...
This way, we could have a `titanaboa` and`tokens` ones:

```
In [2] %boa assert False, "test"

In [2] %erc20 DAI
```

src/ape_console/_cli.py
```diff
@@ -124,11 +125,17 @@ def console(project=None, verbose=None, extra_locals=None):
...
+        _config.HistoryManager.enabled = False
+
+    shared_kwargs = {
+        "colors": "Neutral",
+        "banner1": banner,
+        "user_ns": namespace,
+        "config": _config,
+    }
+    if embed:
+        IPython.embed(**shared_kwargs)
```
unparalleled-js, line 139: One caveat is that you won't get these on consoles started from `-I` for technical reasons beyond our control.
However, you can manually load them via:

```
%load_ext ape_console.plugin
```

unparalleled-js: I just put this in the docs too

unparalleled-js:
the code QL failure is happening on code that I have not touched at all, and the error is simply wrong (it thinks a project name is a password for some reason)

docs/userguides/console.md
```diff
@@ -87,3 +87,66 @@ Out[1]: 1
...
+  --config             Show configuration options (using `ape-config.yaml`)
+  -h, --help           Show this message and exit.
+
+Commands:
+  accounts  Manage local accounts
+  cache     Query from caching database
+  compile   Compile select contract source files
+  console   Load the console
+  init      Initalize an ape project
+  ledger    Manage Ledger accounts
```
fubuloubu, line 125: Maybe remove 2nd party plugins for now

src/ape_console/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+    def ape(self, line: str = ""):
+        """
+        Run Ape CLI commands within an ``ape console`` session.
+
+        Usage example::
+
+            %ape accounts list
+
+        """
+        runner = CliRunner()
```
fubuloubu, line 23: Lol, clever use of the test runner

src/ape_console/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+        elif isinstance(result, str) and result.startswith("0x"):
+            address = result
+        elif isinstance(result, str) and result.isnumeric() or isinstance(result, int):
+            # Happens when excluding quotes from hex str.
+            hex_result = to_hex(int(result))
+            address = ecosystem.decode_address(hex_result)
+        elif isinstance(result, str):
+            address = ape.accounts.load(result).address
+        else:
+            raise ValueError(f"Unable to get account from '{result}'.")
```
fubuloubu, line 57: Can't use the converter?

fubuloubu (approved):
Some examples of some other potentially useful magics

fubuloubu:
> the code QL failure is happening on code that I have not touched at all, and the error is simply wrong (it thinks a project name is a password for some reason)

Seems like the CodeQL had an update?

src/ape_console/plugin.py
unparalleled-js: ya, is this ok, do you think?

src/ape_console/plugin.py
unparalleled-js: changed it to do so! thanks

docs/userguides/console.md
unparalleled-js: done

unparalleled-js:
Another Magic opportunity: `%ape_autoreload` copying what autoreload does except it works on Contracts. So it will re-compile them and put them in the proper caches, mid Console session (since we lost that functionality on some recent caching improvements).

src/ape_console/plugin.py
fubuloubu: Seems fine to me, it's just a wrapper anyways I think

fubuloubu (approved):

fubuloubu:
> Another Magic opportunity: `%ape_autoreload` copying what `%autoreload` does except it works on Contracts. So it will re-compile them and put them in the proper caches, mid Console session (since we lost that functionality on some recent caching improvements).
> 
> (Note this isn't easy but here is an example: https://github.com/ipython/ipython/blob/main/IPython/extensions/autoreload.py)

Wonder if it would be easier to just have a config that avoids caching results in the session?

PR #1353, opened 2023-03-24
chore: upgrade black [APE-745]

unparalleled-js:
### What I did

Upgrade black to next breaking release.
One caveat is also required upgrading `packaging`, hope that is ok.

### How I did it

tried to upgrade black
realized some other incompats
fixed those
upgrades black
ran black

### How to verify it

passing ci
objected conflicts from bumping packaging

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ extras_require = {
         "hypothesis-jsonschema==0.19.0",  # JSON Schema fuzzer extension
     ],
     "lint": [
-        "black>=22.12.0,<23",  # Auto-formatter and linter
+        "black>=23.1.0,<24",  # Auto-formatter and linter
         "mypy>=0.991",  # Static type analyzer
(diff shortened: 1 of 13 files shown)


setup.py
```diff
@@ -97,7 +97,7 @@
         "ijson>=3.1.4,<4",
         "importlib-metadata",
         "ipython>=8.5.0,<9",
-        "packaging>=21,<22",
+        "packaging>=23.0,<24",
```
fubuloubu, line 100: What do we even use this for? Is it pinned because of something else?

fubuloubu (approved):

setup.py
unparalleled-js: let me know - i can create a ticket. else will just leave it for now.

fubuloubu: just leave it for now

PR #1356, opened 2023-03-27
fix: put back console banner

unparalleled-js:
### What I did

Regressions from unreleased feature https://github.com/ApeWorX/ape/pull/1343
Noticed banner was missing.
Noticed we should use the `config` rather than CLI options.
All is better now.

### How I did it

Use the IPython config mechanism to set things, including the banner, which was being silently eaten before.

### How to verify it

Everything in console still works.

### Checklist

### DIFF
--- a/src/ape_console/_cli.py
+++ b/src/ape_console/_cli.py
@@ -126,22 +126,19 @@ def console(project=None, verbose=None, extra_locals=None, embed=False):
     if console_extras:
         namespace.update(console_extras)
 
-    _config = IPythonConfig()
+    ipy_config = IPythonConfig()
     if environ.get("APE_TESTING"):
-        _config.HistoryManager.enabled = False
+        ipy_config.HistoryManager.enabled = False
 
-    ipython_kwargs = {
-        "colors": "Neutral",
-        "banner1": banner,
-        "user_ns": namespace,
-        "config": _config,
-    }
+    ipython_kwargs = {"user_ns": namespace, "config": ipy_config}
     if embed:
-        IPython.embed(**ipython_kwargs)
+        IPython.embed(**ipython_kwargs, colors="Neutral", banner1=banner)
     else:
+        ipy_config.TerminalInteractiveShell.colors = "Neutral"
+        ipy_config.TerminalInteractiveShell.banner1 = banner
         console_config = cast(ConsoleConfig, ape.config.get_config("console"))
-        arguments = ["--ext", "ape_console.plugin"]
+        ipy_config.InteractiveShellApp.extensions.append("ape_console.plugin")
         if console_config.plugins:
-            arguments.extend(["--InteractiveShellApp.extensions", *console_config.plugins])
+            ipy_config.InteractiveShellApp.extensions.extend(console_config.plugins)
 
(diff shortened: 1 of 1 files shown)


fubuloubu (dismissed):
Very nice

How ergonomic is it to use the customized console outside of this module now?

unparalleled-js:
> How ergonomic is it to use the customized console outside of this module now?

when you do `--embe` or use `-I` in `ape run` or `ape test`, it works exactly like it did before, but you can do `%load_ext ape_consoe.plugin` to get added functionality.

when you do `ape console` alone, it is great, it has everything, including custom exception handling on #1337 

if you are using `IPython` from any other tool, such as jupyter or whatever, you can do `%load_ext ape_console.plugin` and it should be like `ape console`, tho haven't tested that too much

fubuloubu:
> if you are using `IPython` from any other tool, such as jupyter or whatever, you can do `%load_ext ape_console.plugin` and it should be like `ape console`, tho haven't tested that too much

Oh sick, this could potentially support `ape-notebook` kernel then

src/ape_console/_cli.py
```diff
@@ -126,22 +127,19 @@ def console(project=None, verbose=None, extra_locals=None, embed=False):
...
+        ipy_config.TerminalInteractiveShell.banner1 = banner
         console_config = cast(ConsoleConfig, ape.config.get_config("console"))
-        arguments = ["--ext", "ape_console.plugin"]
+        ipy_config.InteractiveShellApp.extensions.append("ape_console.plugin")
         if console_config.plugins:
-            arguments.extend(["--InteractiveShellApp.extensions", *console_config.plugins])
+            ipy_config.InteractiveShellApp.extensions.extend(console_config.plugins)
 
-        IPython.start_ipython(**ipython_kwargs, argv=arguments)
+        IPython.start_ipython(**ipython_kwargs, argv=())
```
fubuloubu, line 145: What's the default for `argv`?

src/ape_console/_cli.py
```diff
@@ -28,11 +28,12 @@
...
 )
 @network_option()
+@click.option("--embed", help="Use an embedded IPython", is_flag=True)
 @ape_cli_context()
-def cli(cli_ctx, network):
+def cli(cli_ctx, network, embed):
     """Opens a console for the local project."""
     verbose = cli_ctx.logger.level == logging.DEBUG
-    return console(verbose=verbose)
+    return console(verbose=verbose, embed=embed)
```
fubuloubu, line 36: Should we just always embed?

unparalleled-js: then we can't do cool error handling things or any other fun IPython integration stuff.
we will need an exception handler to properly filter out irrelevant stack frames when showing source code tracebacks

src/ape_console/_cli.py
```diff
@@ -28,11 +28,12 @@
     context_settings=dict(ignore_unknown_options=True),
 )
 @network_option()
+@click.option("--embed", hidden=True, is_flag=True)
```
unparalleled-js, line 31: @fubuloubu `hidden=True` excludes this from the `--help` output / docs

fubuloubu (approved):

fubuloubu (approved):

PR #1359, opened 2023-03-27
fix: send geth process output to devnull when log level above DEBUG [APE-753]

unparalleled-js:
### What I did

Same thing as https://github.com/ApeWorX/ape/pull/1172 but for `ape-geth`

Realized `ape test` output when using `geth` provider is extremely ridden when subprocess logs and i was like "i thought i got rid of these!!!!" but then realized `ape-geth` uses `py-geth` package and not `SubprocessProvider`, then realized how to fix.

### How I did it

overrode the `start()` method to use `DEVNULL` when not `DEBUG`

### Checklist

### DIFF
--- a/src/ape_geth/provider.py
+++ b/src/ape_geth/provider.py
@@ -1,6 +1,7 @@
 import shutil
 from abc import ABC
 from pathlib import Path
+from subprocess import DEVNULL, Popen, PIPE
 from typing import Any, Dict, Iterator, List, Optional, Union
 
 import ijson  # type: ignore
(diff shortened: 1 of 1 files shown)


src/ape_geth/provider.py
```diff
@@ -148,6 +148,19 @@ def connect(self):
         self.start()
         self.wait_for_rpc(timeout=60)
 
+    def start(self):
+        if self.is_running:
```
fubuloubu, line 152: Attr won't exist at first

fubuloubu (approved):

src/ape_geth/provider.py
```diff
@@ -148,6 +150,19 @@ def connect(self):
...
+            return
+
+        self.is_running = True
+        out_file = PIPE if logger.level <= LogLevel.DEBUG else DEVNULL
+        self.proc = Popen(
```
unparalleled-js, line 159: will be easier after this https://github.com/ethereum/py-geth/pull/139

fubuloubu (approved):

PR #1361, opened 2023-03-29
feat: upgrade web3.py to official 6.0 release [APE-759]

unparalleled-js:
### What I did

* requires successful release of eip712 (gh seems down or something)

### How I did it

upgrade and run tests and fix things

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,7 @@ setup(
         "eth-utils>=2.0.0,<3",
         "hexbytes>=0.2.3,<1",
         "py-geth>=3.12.0,<4",
-        "web3[tester]>=6.0.0",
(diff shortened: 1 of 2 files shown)


fubuloubu (commented):
Would this be a breaking change for any plugins?

unparalleled-js:
> Would this be a breaking change for any plugins?
 
It is not supposed to be because plugins avoid upper pins when the dependency is shared by core to avoid it.
However, I did find one thing: https://github.com/ApeWorX/ape-hardhat/pull/130
`ape-hardhat` was improperly importing something that no longer works. We can fix + release this before this core PR to avoid it breaking though!

Issue #271, opened 2021-11-24
Empty plugin configs (but key is still there) causes error


unparalleled-js:
### Environment information

latest ape
macOS
python 3.9.7

### What went wrong?

Had a config file like this:

```yaml
hardhat:
```

Where the plugin hardhat was listed but it had no keys.

Hit this error:

```
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/Users/jules/PycharmProjects/ape/src/ape/__init__.py", line 33, in <module>
    config = _ConfigManager(  # type: ignore
  File "<string>", line 10, in __init__
  File "/Users/jules/PycharmProjects/ape/src/ape/managers/config.py", line 45, in __post_init__
    config = config_class(**user_override)
TypeError: ape_hardhat.providers.HardhatNetworkConfig() argument after ** must be a mapping, not NoneType
```

### How can it be fixed?

Line 45 in `config.py`, check if the obj is not `None` before decomposing

unparalleled-js:
This was fixed sometime ago

unparalleled-js:
I was mistaken, this is not fixed, I was in the wrong directory than the config I tested on

PR #1364, opened 2023-03-30
fix: empty config bug [APE-762]

antazoey:
### What I did

noticed #271 was mistakenly closed early, so i reopened it and fixed the bug

fixes: #271
Fixes: APE-127

### Checklist

### DIFF
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -193,7 +193,7 @@ class ConfigManager(BaseInterfaceModel):
 
         for plugin_name, config_class in self.plugin_manager.config_class:
             # `or {}` to handle the case when the empty config is `None`.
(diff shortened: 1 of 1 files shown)


src/ape/managers/config.py
```diff
@@ -192,7 +192,8 @@ def _plugin_configs(self) -> Dict[str, PluginConfig]:
         )
 
         for plugin_name, config_class in self.plugin_manager.config_class:
-            user_override = user_config.pop(plugin_name, {})
+            # `or {}` to handle the case when the empty config is `None`.
```
unparalleled-js, line 195: for some reason this doesnt happen to all configs, which is maybe why @z80dev couldnt repro, but i get it specifically on the hardat plugin, and i have verified this fixes it.

PR #1360, opened 2023-03-29
feat: filter out irrelevant traceback frames [APE-758]

unparalleled-js:
### What I did

fixes: #442 
Fixes: APE-157

* Removes noisy framework traceback frames in `ape console` during `TransactionError` handling
* Removes noisy framework traceback frames in `ape run` during `TransactionError` handling
* Remove noisy framework traceback frames in `ape test` during `TransactionError` handling.

This makes the PR #1337 a lot simpler to follow and smaller.
There is also a lot of benefit to this alone so wanted to include it incase #1337 does not make it in the next release.

### How I did it

 * filter out frames that didn't occur in the local Ape project during the handling of `TransactionError` in CLI, pytest plugin, and IPython plugin.

### How to verify it

Cause unhandled TransactionError (e.g. ContractLogicError) in all 3 environments.

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #1371, opened 2023-04-03
fix: issue where `ape test --showinternal` would cause crash [APE-782]

unparalleled-js:
### What I did

regression from #1360  causing `--showinternal` to crash pytest terminal output

### How I did it

correct type

### How to verify it

can use `--showinternal` now
added a test

### Checklist


fubuloubu (dismissed):

unparalleled-js:
the 1 test that failed randomly was so flukey, am super unsure how that scenario could have possibly happened. In order to "fix", I renamed a global variable. Really have no idea though.

fubuloubu (approved):

PR #1375, opened 2023-04-04
perf: add cache middleware to `ape-geth` web3 instance [APE-799]

unparalleled-js:
### What I did

fixes: #

### Checklist


src/ape_test/provider.py
```diff
@@ -49,7 +48,6 @@ def connect(self):
             num_accounts=self.config["number_of_accounts"],
         )
         self._web3 = Web3(EthereumTesterProvider(ethereum_tester=self._evm_backend))
-        self._web3.middleware_onion.add(simple_cache_middleware)
```
unparalleled-js, line 52: turns out this is kinda bad... it interferes with all connected web3 instances which can be really bad.

fubuloubu (approved):

PR #1376, opened 2023-04-06
fix: check contract address in log filter

fubuloubu:
### What I did

When you filter the events from `ReceiptAPI.events`, it only paid attention to the name of the event, and not the `ContractInstance` object it came from

fixes: #

### How I did it

Ensure that we check the contract address of the log matches the event definition

### How to verify it

Needs a test

### Checklist

~~- [x] Documentation has been updated~~


src/ape/types/__init__.py
```diff
@@ -284,7 +284,7 @@ class ContractLogContainer(list):
     def filter(self, event: "ContractEvent", **kwargs) -> List[ContractLog]:
         found_events = []
         for log in self:
-            if log.event_name == event.name:
+            if log.event_name == event.name and log.contract_address == event.contract.address:
```
unparalleled-js, line 287: seems good, let me know if need help writing a test

fubuloubu: Yeah, would be appreciated!

tests/functional/test_geth.py
```diff
@@ -22,40 +22,61 @@
 '0x([A-Fa-f0-9]{64})'
 tx\.origin=0x[a-fA-F0-9]{40}
 ContractA\.methodWithoutArguments\(\) -> 0x00..5174 \[\d+ gas\]
-├── SYMBOL\.methodB1\(lolol="ice-cream", dynamo=36\)
+├── SYMBOL\.supercluster\(x=234444\) -> \[
```
unparalleled-js, line 25: srry this changes, futz with the contract an its json was out of date

fubuloubu:
Weird macOS-only failures :thinking:

PR #1377, opened 2023-04-07
feat: show dev message in `ContractLogicError` when no other message provided [APE-807]

unparalleled-js:
### What I did

Currently, if your `ContractLogicError` has no `revert_message`, the default is `Transaction failed.` However, sometimes a dev message is present!! So instead of showing `Transaction failed.`, it will show the dev message in these cases.

### How I did it

Move logic for extracting `dev_message` to `ContractLogicError` itself... now it's a property there.

### How to verify it

Here is what some output of mine looks like for uncaught:

```
SUCCESS: Contract 'MathAbuse' deployed to: 0x274b028b03A250cA03644E6c578D81f019eE1323

  File "/Users/jules/ApeProjects/ape-playground/scripts/math.py", line 7, in main
    contract.num_add(1, sender=account)

ERROR: (ContractLogicError) Integer overflow
```

### Checklist


src/ape/exceptions.py
```diff
@@ -139,6 +153,94 @@ def __init__(
...
+        accessor: Optional["ManagerAccessMixin"] = None
+        if self.txn:
+            accessor = self.txn
+        elif self.trace is not None:
+            self.trace, second_trace = tee(self.trace)
+            if second_trace:
+                accessor = next(second_trace, None)
+
+        if accessor is None:
+            return None, "Missing trace access."
```
fubuloubu, line 167: I think raising an exception may be better than having the dual string return here, it's a little confusing

unparalleled-js: less confusing now [efb90ff](https://github.com/ApeWorX/ape/pull/1377/commits/efb90ff504aa774ea43dcfbc7fd0143200ecd7e3)

fubuloubu (approved):

Issue #690, opened 2022-05-03
Ecosystem.decode_error() (inspired by custom Solidity error types)


unparalleled-js:
### Elevator pitch:

When using `solidity` error definitions, the revert message (the inputs to the custom error) are not decoded (they are just bytes).

e.g.

```
ape.exceptions.ContractLogicError: b'\x9b\x0f\x01\xda\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00;\x9a\xca\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01'
```

**Let's do what we do best and dynamically generate a custom Exception class with the inputs as the properties!**

### Value:

Solidity projects that use custom errors.

### Dependencies:

https://github.com/ApeWorX/ethpm-types/pull/29#pullrequestreview-961074682

### Design approach:

Like structs, events, etc, let's create custom `ContractLogicError` classes that have the inputs as the properties on the class

So in your solidity, if you do:

```solidity
    error InsufficientFund(uint256 given, uint256 required);

    function fund() public payable isOn {
        if (msg.value > 0) {
            revert InsufficientFund(msg.value, 1);
        }
    }
```

, you can access the custom error classes on the `ContractInstance` and they will be the exceptions raised for reverts in that case.

So you can do this in your tests:

```python
with ape.reverts(contract.InsufficientFund) as err:
    ...

assert err.value.foo == 5
assert err.value.bar == 6
```

* Exactly like `pytest.raises()` works.

NOTE: The original still has to work to as not be breaking:

```python
with ape.reverts(message="revert reason"):
```

### Task list:

* [ ] Tasks go here

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

fubuloubu:
Everything looks good, approved


Perhaps can use single dispatch to allow both message and exception class to be used in `ape.reverts`

Issue #1223, opened 2023-01-16
Add error type support to `ape.reverts`


fubuloubu:
### Overview

Solidity v0.8.x series added a new type of item to the abi called custom errors: https://docs.soliditylang.org/en/v0.8.17/structure-of-a-contract.html#errors

We have support for parsing these errors from the ABI as of [ethpm-types v0.2.1](https://github.com/ApeWorX/ethpm-types/issues/22) 

The basic gist is that when a revert happens, Solidity will return abi-encoded dated corresponding to the error type with additional information about the error. Some of the built-in Solidity errors are as follows (it might be good to add an enum with these to `ape-ethereum` or `ape-solidity`):

<img width="552" alt="image" src="https://user-images.githubusercontent.com/3859395/212744399-381987c3-1ff7-4f19-b7dc-7abbc852dcfc.png">

### Specification

It would be nice to use these error types in tests w/ `ape.reverts`. Some examples might include:

```py
# succeds if `CustomError` is raised with any args
with ape.reverts(my_contract.CustomError):
    my_contract.call_that_raises_CustomError(...)

# succeds if `CustomError` is raised with `arg1 = some_value`, fails otherwise
with ape.reverts(my_contract.CustomError, arg1=some_value):
    my_contract.call_that_raises_CustomErrorWithArgs(...)
```

### Dependencies

n/a

pcaversaccio:
To be specific, Solidity [`0.8.4`](https://github.com/ethereum/solidity/releases/tag/v0.8.4) introduced _Custom Errors_.

fubuloubu:
Alternative syntax:

```py
with my_contract.CustomError(_, _, ..., _):
    my_contract.call_that_raises_CustomError(...)

with my_contract.CustomError(some_value, _, ..., _):
    my_contract.call_that_raises_CustomErrorWithArgs(...)
```

But the use of the wildcard `_` might be complex

pcaversaccio:
KISS - I would definitely go with version 1.

fubuloubu:
> KISS - I would definitely go with version 1.

Yeah I agree, but I did like the ability to do `with CustomErrorType: ...` which I think it's a bit nicer looking syntax

pcaversaccio:
I see - maybe worth introducing an alias later?

unparalleled-js:
> @fubuloubu any progress/news on this?

PR open

PR #1378, opened 2023-04-07
feat: compiler API method for enriching error [APE-808]

unparalleled-js:
### What I did

fixes: #690 
Fixes: APE-193

Give compilers the chance to create a custom exception.

Will be used:

* For vyper compiler errors, like math errors
* For solidity custom error abi types

### How I did it

new api method
passthrough for not implemented.

### Checklist


unparalleled-js:
- [ ] Add JSON compiler implementation to make use of error ABIs.

src/ape_pm/compiler.py
```diff
@@ -78,3 +80,62 @@ def compile(
             contract_types.append(contract_type)
 
         return contract_types
+
+    def enrich_error(self, err: ContractLogicError) -> ContractLogicError:
```
unparalleled-js, line 84: one thing is that this is exactly the same implementation `ape-solidity` would have... wondering if we should use a shared method somehow?

unparalleled-js: - [ ] going to make ecosystem decode_error method, that should help

unparalleled-js:
To see it work in Solidity, check out this PR: https://github.com/ApeWorX/ape-solidity/pull/93

Example uncaught exception:

```
  File "/Users/jules/ApeProjects/ape-playground/scripts/haserror.py", line 8, in main
    contract.withdraw(sender=other)

ERROR: (Unauthorized) addr=0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7, counter=123
```

unparalleled-js:
Can do [1223](https://github.com/ApeWorX/ape/issues/1223) right after

fubuloubu (approved):

PR #1380, opened 2023-04-10
feat: `ape.reverts` to work on custom contract error types

unparalleled-js:
### What I did

Fixes: APE-458
fixes: #1223 

### How I did it

Copy stuff for events to get started.
Added new handler in reverts context manager.

### How to verify it

```
with ape.reverts(contract.MyError, arg1=blah, arg2=foo):
    cause_error()
```
### Checklist


src/ape/contracts/base.py
```diff
@@ -628,6 +630,28 @@ def poll_logs(
             yield from self.range(new_block.number, stop=new_block.number + 1)
 
 
+class ContractError(ManagerAccessMixin):
```
fubuloubu, line 633: Hmm, maybe can call this `CustomError` or `ContractErrorType` to avoid having to alias the `ContractError` exception as an import

fubuloubu: Also, don't think it needs `ManagerAccessMixin`

fubuloubu (dismissed):
Asked a few questions about design offline, but otherwise looks great

src/ape/contracts/base.py
```diff
@@ -628,6 +628,28 @@ def poll_logs(
             yield from self.range(new_block.number, stop=new_block.number + 1)
 
 
+class CustomError:
```
unparalleled-js, line 631: Attempting to make this an instance of `ContractLogicError` and have that be what actually gets raised, but the issue there is that this does not know the inputs - only the input types (abi). This class is more like a `CustomErrorType` class rather than an actual exception class.

fubuloubu: actually a good name for the class

src/ape/api/networks.py
```diff
@@ -526,20 +525,8 @@ def decode_error(self, abi: ErrorABI, data: HexBytes, **kwargs) -> ContractLogic
         """
 
         # TODO: Include ErrorABI in official `decode_calldata` API definition for 0.7
```
fubuloubu, line 527: Is this still accurate?

src/ape/contracts/base.py
```diff
@@ -881,6 +881,30 @@ def get_event_by_signature(self, signature: str) -> ContractEvent:
...
+        Args:
+            signature (str): The signature of the error.
+
+        Returns:
+            :class:`~ape.types.CustomErrorType`
+        """
+
+        name_from_sig = signature.split("(")[0].strip()
+        options = self._errors_.get(name_from_sig, [])
+        err = ContractError(f"No error found with signature '{signature}'.")
```
fubuloubu, line 898: Wonder if this exception is too ambigiously named :thinking:

At least something to consider for breaking changes in 0.7.0

src/ape/api/networks.py
unparalleled-js: `decode_calldata` should also permit `ErrorABI` as a valid input type, ya

src/ape/api/networks.py
```diff
@@ -512,35 +510,6 @@ def enrich_calltree(self, call: CallTreeNode, **kwargs) -> CallTreeNode:
         """
         return call
 
-    def decode_error(self, abi: ErrorABI, data: HexBytes, **kwargs) -> ContractLogicError:
```
unparalleled-js, line 515: it is ok to delete because this hasnt been released yet

src/ape/contracts/base.py
unparalleled-js: Add thoughts to APE-822, just made the ticket

src/ape/contracts/base.py
```diff
@@ -900,6 +938,46 @@ def _events_(self) -> Dict[str, List[ContractEvent]]:
...
+
+            errors = {}
+            for abi_name, abi_list in abis.items():
+                errors_to_add = []
+                for abi in abi_list:
+                    error_type = None
+                    for existing_cls in prior_errors:
+                        if existing_cls.abi.signature == abi.signature:
+                            # Error class was previously initialized by contract at same address.
+                            error_type = existing_cls
```
unparalleled-js, line 963: really really important to dynamically-define classes only once if we want to rely on the instances for assertions.... I was seeing weird things in tests when multiple instances of `ContractInstance` with same address / chain would create new classes of custom errors and cause odd test failures like `MyError is not MyError`. So this whole caching system ensures they're only defined one time.

tests/functional/test_contract_instance.py
```diff
@@ -638,3 +640,25 @@ def test_obj_list_as_struct_array_input(contract_instance, owner, data_object):
...
     data = {"a": owner, "b": HexBytes(123), "c": "GETS IGNORED"}
     assert contract_instance.setStructArray([data, data]) is None
+
+
+def test_custom_error(error_contract, not_owner):
+    contract = error_contract
+    unauthorized = contract.Unauthorized
+    assert issubclass(unauthorized, CustomError)
+
+    with pytest.raises(contract.Unauthorized) as err:
```
unparalleled-js, line 650: @fubuloubu Look at this! This means everything you think it means. I think we're good now, at least with this part?

fubuloubu: Yes, this seems great!

src/ape/contracts/base.py
```diff
@@ -668,6 +675,13 @@ def decode_input(self, calldata: bytes) -> Tuple[str, Dict[str, Any]]:
         return method.selector, input_dict
 
 
+def _create_custom_error_type(abi: ErrorABI) -> Type[CustomError]:
+    def exec_body(namespace):
+        namespace["abi"] = abi
```
fubuloubu, line 680: Probably want to add the contract too?

fubuloubu (approved):

src/ape/contracts/base.py
unparalleled-js: Done!

fubuloubu (approved):

istankovic:
Thanks a bunch for this!

Issue #1381, opened 2023-04-11
Uniswap V3 repo version not recognized [APE-820]


MMtis:
### Environment information

- OS: win (but using ubuntu with wsl for this project)
- Python Version: 3.10.6
- `ape` and plugin versions:

```sh
$ ape --version
0.5.9

$ ape plugins list
Installed Plugins:
  chainstack    0.5.0a1
  solidity      0.6.2
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
dependencies:
  - name: Uniswap
    github: Uniswap/v3-core
    version: 1.0.0
  - name: Uniswap
    github: Uniswap/v3-periphery
    version: 1.3.0
```

### What went wrong?

- After compiling the project, the Uniswap V3 repo dependencies versions are not getting recognized. I'm fairly new to Apeworx, so maby I'm missing something.

```sh
$ ape compile
ERROR: (ProjectError) Unknown version '1.0.0' for repo 'v3-core'.
```
Since there is no release for this project, I tried different versions v1.0.0, V1.0.0 but none worked.

### How can it be fixed?

The repo does not have a release. So for it to work I had to fork it and create a release.

unparalleled-js:
Hello @MMtis ! Do this:

```yaml
dependencies:
  - name: Uniswap
    github: Uniswap/v3-core
    branch: v1.0.0
```

Reason:
The `version:` key in the config will use the GitHub release API to try and fetch a published package version. It looks Uniswap is not publishing releases this way. The `branch:` config option will use branches or tags. In this case, the release tag is perfect!

I just tried this and it seems to work.

MMtis:
> Hello @MMtis ! Do this:
> 
> ```yaml
> dependencies:
>   - name: Uniswap
>     github: Uniswap/v3-core
>     branch: v1.0.0
> ```
> 
> Reason: The `version:` key in the config will use the GitHub release API to try and fetch a published package version. It looks Uniswap is not publishing releases this way. The `branch:` config option will use branches or tags. In this case, the release tag is perfect!
> 
> I just tried this and it seems to work.

Hello @unparalleled-js ,

Thank you for the tip ! It's definitely better than forking the repo.
It also worked for me.

unparalleled-js:
> > Hello @MMtis ! Do this:
> 
> > 
> 
> > ```yaml
> 
> > dependencies:
> 
> >   - name: Uniswap
[...]

PR #1385, opened 2023-04-11
feat: smarter handling of references and versions in dependency configs [APE-829]

unparalleled-js:
### What I did

fixes: #1381 
Fixes: APE-820

### How I did it

* document stuff about tags
* if version not found, try as a reference
* if missing `v` prefix, try with `v` prefix.

### How to verify it

this will work now:

```
dependencies:
  - name: Uniswap
    github: Uniswap/v3-core
    version: 1.0.0
```

tho youll get a warning.

best to do this:

```

dependencies:
  - name: Uniswap
    github: Uniswap/v3-core
    ref: v1.0.0
```

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -45,6 +45,17 @@ dependencies:
 
 will download the [Open Zeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts) package with version `4.4.2`.
 
+**WARN**: The `version:` key only works when the version is published to GitHub; use the `branch:` key when the version is only available by tag.
+
+For example, to install a version based on a `git` tag, do the following:
+
+```yaml
+dependencies:
+  - name: Uniswap
+    github: Uniswap/v3-core
+    branch: v1.0.0
+```
+
 To ignore files from a dependency project, use the `exclude` setting to specify glob patterns:
 
 ```yaml
@@ -70,7 +81,7 @@ Now, in your solidity files, import `OpenZeppelin` sources via:
 import "@openzeppelin/token/ERC721/ERC721.sol";
 ```
 
-You can also set the branch and name of the dependency's contracts folder, e.g.:
+You can also set the name of the dependency's contracts folder, e.g.:
 
 ```yaml
 dependencies:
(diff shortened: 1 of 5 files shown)


docs/userguides/config.md
```diff
@@ -45,6 +45,19 @@ dependencies:
...
+For example, to install a version based on a `git` tag, do the following:
+
+```yaml
+dependencies:
+  - name: Uniswap
+    github: Uniswap/v3-core
+    branch: v1.0.0
+```
+
+The `branch:` config installs the code from that branch or tag whereas the `version:` config uses the official GitHub release API.
```
fubuloubu, line 59: Wonder if `branch` is the confusing bit and it should be `ref` or something (breaking change obvs)

fubuloubu (approved):

docs/userguides/config.md
unparalleled-js: We could deprecate `branch:`, add `ref:`, make `branch:` be an alias and undocument it, and then remove `branch:` on 0.7. We could do that now.

unparalleled-js:
another idea that could alongside: if the version release is not found from `version:`, check the tags. maybe warn as well saying to use `branch:` (or `ref:`), but then it at least would still work.

thoughts there?

fubuloubu:
> another idea that could alongside: if the version release is not found from `version:`, check the tags. maybe warn as well saying to use `branch:` (or `ref:`), but then it at least would still work.
> 
> thoughts there?

Both the suggestions are good suggestions, I think `ref:` is clearer to the type of git operation that's being performed (vs. `branch:` which is just one type of `ref:`)

src/ape/managers/project/dependency.py
```diff
@@ -61,19 +62,36 @@ class GithubDependency(DependencyAPI):
     such as ``dapphub/erc20``.
     """
 
+    # TODO: Remove at >= 0.7
```
unparalleled-js, line 65: APE-830, it is in the 0.7 view

fubuloubu (approved):
This seems more robust!

PR #1390, opened 2023-04-18
fix: improve error message when unable to find contract type [APE-813]

NotPeopling2day:
Contributor [@LinuxIsCool](https://github.com/LinuxIsCool) suggested updating the error message to let users know they will need an Explorer plugin to look up contract types by address.

https://github.com/ApeWorX/ape-alchemy/issues/51

Example error on local network, no explorer support:
Screenshot 2023-04-19 at 10 45 54 AM

Example successful path and error path when changing address to unverified contract:
Screenshot 2023-04-19 at 10 51 05 AM


fubuloubu (approved):

fubuloubu (dismissed):

src/ape/managers/chain.py
```diff
@@ -1092,6 +1093,9 @@ def instance_at(
             :class:`~ape.contracts.base.ContractInstance`
         """
 
+        if self.network_manager.active_provider is None:
```
antazoey, line 1096: this happens automatically when doing `self.provider`, so it probably isnt needed here

src/ape/managers/chain.py
```diff
@@ -1111,7 +1115,10 @@ def instance_at(
                 raise  # Current exception
 
         if not contract_type:
-            raise ChainError(f"Failed to get contract type for address '{contract_address}'.")
+            msg = f"Failed to get contract type for address '{contract_address}'."
+            if self.provider.network.explorer is None:
+                msg += " Try installing an explorer plugin, such as Etherscan."
```
antazoey, line 1120: (notes from offline)

- [ ] handle case where using provider without explorer (mention also try using diff network)
- [ ] mention cmd to install plugin
- [ ] if all seems good here, mention that the contract may not be verified

fubuloubu (approved):

src/ape/managers/chain.py
```diff
@@ -1111,7 +1111,17 @@ def instance_at(
                 raise  # Current exception
 
         if not contract_type:
-            raise ChainError(f"Failed to get contract type for address '{contract_address}'.")
+            msg = f"Failed to get contract type for address '{contract_address}'."
+            if self.provider.network.explorer is None:
+                msg += (
+                    f" Current provider '{self.provider.name}' has no associated "
+                    "explorer plugin. Try installing an explorer plugin using "
+                    f"{Colors.GREEN}ape plugins install etherscan{Colors.END}, "
```
antazoey, line 1119: You should `click.style` instead maybe, then we wouldnt need to define all the ansi colors

antazoey: or alternatively, the rich library all over will do stuff like this automatically , or you can control with CLI-markup

antazoey (approved):

PR #1391, opened 2023-04-18
fix: negative quantity transfer value [APE-860]

NotPeopling2day:
Was trying to create a failing receipt and realized we weren't catching negative values quickly and handling them in ape.

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -6,6 +6,7 @@ from functools import partial
 from pathlib import Path
 from typing import IO, Collection, Dict, Iterator, List, Optional, Set, Type, Union, cast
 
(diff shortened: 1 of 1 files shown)


fubuloubu (dismissed):

src/ape/api/accounts.py
```diff
@@ -176,6 +176,8 @@ def transfer(
 
         elif value:
             txn.value = self.conversion_manager.convert(value, int)
+            if txn.value < 0:
```
antazoey, line 179: consider adding a test to ensure this behavior is preserved

antazoey (approved):

fubuloubu (approved):

Issue #1264, opened 2023-01-29
Create event logs for comparison during testing


fubuloubu:
### Overview

Would be super handy to have a way during testing to assert equivalency of a particular set of event logs to a desired outcome.

Thanks to https://github.com/ApeWorX/ape/pull/1244, we now have `ReceiptAPI.events`, which is an explicit list of all the logs that were present during the invocation of a transaction. However, we still have to more explicitly compare particular fields to show that particular fields were set, and particular events were emitted overall. It would be great to have a more succinct ability to do these comparisons.

### Specification

We want the ability to do simple comparisons that all (or a portion) of the events emitted during a transaction are as we expect. Here are some common scenarios:

1. We want to show that only an expected number of events occurred, and that those logs had the expected log arguments.
```py
assert tx.events == [contract.Event(1, 2, 3)]
```

2. We want to show that at least a specific event occurred, and that those logs had the expected log arguments.
```py
assert contract.Event(1, 2, 3) in tx.events
```

3. We might want to show one of the above, but only care about certain arguments being set to specific values, ignoring others for the purpose of a comparison. Not sure how to do this, but perhaps something like `contract.Event(1, _, 3)` or using kwarg syntax instead

4. We might want to show a subset of specific events occurred in the event stream. Not sure how best to provide this, but currently you can just assert multiple times e.g.:
```py
assert contract.EventA(1, 2, 3) in tx.events
assert contract.EventB(4, 5, 6) in tx.events
```

---

A way to do this would be to add a `__call__` to `ContractEvents` that produces the equivalent decoded log to what `ReceiptAPI.events` produces (via `EcosystemAPI.decode_logs`). If those objects have a comparison method (e.g. `ContractLog.__eq__`) then this will work. Also note that `ReceiptAPI.events` is a `cached_property`, so it's not too bad doing multiple comparisons

### Dependencies
n/a

fubuloubu:
From discord:

- maybe just produce a `ContractLog` item directly, instead of encoding and decoding 🤔
- also, specially handle `_` as a variable name that gets skipped for comparison purposes, that might even be the best way to support comparison of the "non-essential" information, if the user decides that it matters in some contexts e.g. `contract.MyEvent(*event_args)` produces a `ContractLog` with `block_number`, etc. set to `_` to skip the comparison, but if you take a specific event log from the live chain and compare it, they have to match more exactly (e.g. same transaction in the same block)

PR #1392, opened 2023-04-19
feat: implementation of event log comparison [APE-864]

z80dev:
### What I did

Implement a way during testing to assert equivalency of a particular set of event logs to a desired outcome 

fixes: #1264 

### How I did it

Implements a MockContractLog class which can be compared to a real ContractLog class. Only fields present in the mock should be compared. Then, also implement a way to construct one of these MockContractLog instances by calling the Event type

### How to verify it

Tests WIP

### Checklist


src/ape/contracts/base.py
```diff
@@ -21,7 +21,7 @@
     TransactionNotFoundError,
 )
 from ape.logging import logger
-from ape.types import AddressType, ContractLog, LogFilter
+from ape.types import AddressType, MockContractLog, BaseContractLog, ContractLog, LogFilter
```
fubuloubu, line 24: Then won't need this
```suggestion
from ape.types import AddressType, MockContractLog, ContractLog, LogFilter
```

src/ape/contracts/base.py
```diff
@@ -463,6 +463,19 @@ def __len__(self):
...
         return sum(1 for _ in logs)
 
+
+    def __call__(self, *args: Any, **kwargs: Any) -> "BaseContractLog":
+        event_args = dict(zip([input.name for input in self.abi.inputs], args))
+        event_args.update(kwargs)
+
+        log_data = {
+            field: event_args.get(field, None) for field in ContractLog.__fields__.keys()
+        }
```
fubuloubu, line 473: I think it looks more like this
```suggestion
        log_data["event_arguments"] = event_args
```

src/ape/contracts/base.py
```diff
@@ -463,6 +463,19 @@ def __len__(self):
...
+        log_data["address"] = self.contract.address
+        log_data["abi"] = self.abi
+
+        return MockContractLog(**log_data)
```
fubuloubu, line 477: Or could do this instead
```suggestion
        return MockContractLog(
            abi=self.abi,
            address=self.contract.address,
            event_arguments=event_args,
        )
```

src/ape/contracts/base.py
```diff
@@ -463,6 +463,19 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+
+    def __call__(self, *args: Any, **kwargs: Any) -> "BaseContractLog":
+        event_args = dict(zip([input.name for input in self.abi.inputs], args))
+        event_args.update(kwargs)
```
fubuloubu, line 469: Maybe add a disjoint set between `event_args.keys()` and `kwargs.keys()` to prevent overwrites

fubuloubu: Also add a check after updating that `event_args.keys()` is a subset of `set(input.name for input in self.abi.inputs)`

src/ape/types/__init__.py
```diff
@@ -275,6 +302,21 @@ def __getitem__(self, item: str) -> Any:
     def get(self, item: str, default: Optional[Any] = None) -> Any:
         return self.event_arguments.get(item, default)
 
+class MockContractLog(BaseContractLog):
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, BaseContractLog):
+            return NotImplemented
```
fubuloubu, line 308: I think we want to be stricter to say we can only compare a Mock to a normal `ContractLog`
```suggestion
        if not isinstance(other, ContractLog):
            return NotImplementedError()
```

src/ape/types/__init__.py
```diff
@@ -275,6 +302,21 @@ def __getitem__(self, item: str) -> Any:
...
+
+        if self.contract_address != other.contract_address or self.event_name != other.event_name:
+            return False
+
+        shared_keys = set(self.event_arguments.keys()) & set(other.event_arguments.keys())
+
+        for k in shared_keys:
+            if self.event_arguments[k] != other.event_arguments[k]:
+                return False
```
fubuloubu, line 317: probably can do it like this
```suggestion
        # NOTE: `self.event_arguments` contains a subset of items from `other.event_arguments`,
        #       but we skip those the user doesn't care to check
        for name, value in self.event_arguments.items():
            # Make sure `value` is not `None` (user explicitly set it `None`)
            # NOTE: `other.event_arguments[name]` will raise `ItemError` only if ABIs don't match
            if value and value != other.event_arguments[name]:
                return False
```

src/ape/types/__init__.py
```diff
@@ -292,6 +334,9 @@ def filter(self, event: "ContractEvent", **kwargs) -> List[ContractLog]:
                     found_events.append(log)
         return found_events
 
+    def __contains__(self, __o: object) -> bool:
+        return any(log == __o for log in self)
```
fubuloubu, line 338: `Any` is a better type, and gave a different name because it's easier to read
```suggestion
    def __contains__(self, val: Any) -> bool:
        return any(log == val for log in self)
```

tests/functional/test_contract_event.py
```diff
@@ -277,6 +277,7 @@ def test_contract_two_events_with_same_name(
 def test_contract_decode_logs_no_abi(owner, contract_instance):
     receipt = contract_instance.setNumber(1, sender=owner)
     events = list(receipt.decode_logs())  # no abi
+    print(contract_instance.NumberChange(1))
```
fubuloubu, line 280: No print statements if not necessary

tests/functional/test_contract_event.py
```diff
@@ -277,6 +277,7 @@ def test_contract_two_events_with_same_name(
 def test_contract_decode_logs_no_abi(owner, contract_instance):
```
fubuloubu, line 277: This might not be the best test case to update

fubuloubu (commented):
Looking good, a couple of change suggestions

Needs a few test cases for:
- [x] Check `c.Event() in tx.events`
- [x] Check `tx.events == [c.Event()]`
- [x] Check `tx.events[0] == c.Event(a)`
- [ ] Check `tx.events[1] == c.Event(a, arg2=b)`
- [x] Check `tx.events[2] == c.Event(arg3=b)`

tests/functional/test_contract_event.py
0xafm: Yeah, I was just trying to test the `__call__` function worked, I picked a random test

tests/functional/test_contract_event.py
z80dev: Yeah this was just my attempt to test out the `__call__` implementation, I only pushed to show what I was trying (and failing)

src/ape/types/__init__.py
```diff
@@ -276,6 +304,25 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
         return self.event_arguments.get(item, default)
 
 
+class MockContractLog(BaseContractLog):
```
antazoey, line 307: - [ ] Add class doc here
- [ ] Include `MockContractLog` in `docs/methoddocs/types.md` members list under the misc header

src/ape/types/__init__.py
```diff
@@ -276,6 +304,25 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
...
+            return NotImplemented
+
+        if self.contract_address != other.contract_address or self.event_name != other.event_name:
+            return False
+
+        # NOTE: `self.event_arguments` contains a subset of items from `other.event_arguments`,
+        #       but we skip those the user doesn't care to check
+        for name, value in self.event_arguments.items():
+            # Make sure `value` is not `None` (user explicitly set it `None`)
+            # NOTE: `other.event_arguments[name]` will raise `ItemError` only if ABIs don't match
```
antazoey, line 319: `ItemError`? Do you mean `IndexError`?

antazoey (commented):
Definitely would like tests!

src/ape/contracts/base.py
```diff
@@ -463,6 +463,32 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def __call__(self, *args: Any, **kwargs: Any) -> "MockContractLog":
```
fubuloubu, line 466: Is the quotes required?
```suggestion
    def __call__(self, *args: Any, **kwargs: Any) -> MockContractLog:
```

src/ape/contracts/base.py
```diff
@@ -463,6 +463,32 @@ def __len__(self):
...
+            raise ValueError(f"Overlapping keys found in arguments: {overlapping_keys}")
+
+        # Update event_args with keyword arguments
+        event_args.update(kwargs)  # type: ignore[arg-type]
+
+        # Check that event_args.keys() is a subset of the expected input names
+        expected_input_names = {input.name for input in self.abi.inputs}
+        if not set(event_args.keys()).issubset(expected_input_names):
+            raise ValueError(
+                f"Invalid argument keys found, expected a subset of {expected_input_names}"
```
fubuloubu, line 482: `setA - setB` gives you items in `setA` that are not in `setB`
```suggestion
        if unknown_input_names := set(event_args.keys()) - {input.name for input in self.abi.inputs}:
            raise ValueError(
                f"Invalid argument keys found, expected a subset of {unknown_input_names}"
```
p.s. I enjoy walrus operators way too much

src/ape/contracts/base.py
```diff
@@ -463,6 +463,32 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def __call__(self, *args: Any, **kwargs: Any) -> "MockContractLog":
+        # Create a dictionary from the positional arguments
+        event_args = dict(zip((input.name for input in self.abi.inputs), args))
+
+        # Check for overwrites by looking for the intersection of the two key sets
+        overlapping_keys = set(event_args.keys()) & set(kwargs.keys())
+        if overlapping_keys:
```
fubuloubu, line 472: moar walruses
```suggestion
        if overlapping_keys := set(event_args.keys()) & set(kwargs.keys())
```

src/ape/types/__init__.py
```diff
@@ -199,6 +199,34 @@ class ContractLog(BaseInterfaceModel):
...
 
+    @validator("contract_address", pre=True)
+    def validate_address(cls, value):
+        from ape import convert
+
+        return convert(value, AddressType)
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, BaseContractLog):
+            return NotImplemented
```
fubuloubu, line 210: Add a note of how this triggers `x.__eq__(y)` to change to `y.__eq__(x)`, and how this fits in with `MockContractLog`

src/ape/types/__init__.py
```diff
@@ -199,6 +199,34 @@ class ContractLog(BaseInterfaceModel):
...
+
+        if self.contract_address != other.contract_address or self.event_name != other.event_name:
+            return False
+
+        for k, v in self.event_arguments.items():
+            other_v = other.event_arguments.get(k)
+            if v is None or other_v is None:
+                continue  # Skip wildcard comparisons
```
fubuloubu, line 218: Shouldn't be needed here, because `MockContractLog.__eq__` has this logic

src/ape/types/__init__.py
fubuloubu: that was my bad

fubuloubu (commented):
Needs:
- [ ] rendering assertion errors in pytest a little nicer in the following circumstances:
  - [ ] when a value in the args mismatches
  - [ ] when there's 1+ more exceptions than expected
  - [ ] when there's 1+ less exceptions than expected
- [ ] convert `event_argument` values (via validator in `MockContractLog`, not applicable to regular `ContractLog`)

fubuloubu:
Also, amend your commit messages to follow https://www.conventionalcommits.org/en/v1.0.0/

p.s. not necessary if you submit PR against your own repo instead of ApeWorX branch

tests/functional/test_contract_event.py
fubuloubu, line 1: Change some of these to use `args`, and a mix of `args` and `kwargs`

src/ape/contracts/base.py
```diff
@@ -463,6 +463,36 @@ def __len__(self):
...
+        if unknown_input_names := set(event_args.keys()) - {input.name for input in self.abi.inputs}:
+            raise ValueError(
+                f"Invalid argument keys found, expected a subset of {unknown_input_names}"
+            )
+
+        # Convert the arguments using the conversion manager
+        converted_args = {}
+        for key, value in event_args.items():
+            input_abi = next(input for input in self.abi.inputs if input.name == key)
+            converted_args[key] = self.conversion_manager.convert(value, input_abi.type)
```
fubuloubu, line 487: Only need to convert if it's not `None`

src/ape/contracts/base.py
```diff
@@ -463,6 +463,36 @@ def __len__(self):
         logs = self.provider.get_contract_logs(self.log_filter)
         return sum(1 for _ in logs)
 
+    def __call__(self, *args: Any, **kwargs: Any) -> MockContractLog:
+        # Create a dictionary from the positional arguments
+        event_args = dict(zip((input.name for input in self.abi.inputs), args))
+
+        # Check for overwrites by looking for the intersection of the two key sets
+        if overlapping_keys := set(event_args.keys()) & set(kwargs.keys()):
+            raise ValueError(f"Overlapping keys found in arguments: {overlapping_keys}")
```
antazoey, line 472: (optional) comma-join them to look nicer
```suggestion
            raise ValueError(f"Overlapping keys found in arguments: '{', '.join(overlapping_keys)}'.")
```

src/ape/types/__init__.py
```diff
@@ -185,9 +185,9 @@ def encode_topic_value(abi_type, value):
         )
 
 
-class ContractLog(BaseInterfaceModel):
+class BaseContractLog(BaseInterfaceModel):
     """
-    An instance of a log from a contract.
+    Base class representing information relevant to an event instance itself.
```
antazoey, line 190:
```suggestion
    Base class representing information relevant to an event instance.
```

src/ape/types/__init__.py
```diff
@@ -199,6 +199,32 @@ class ContractLog(BaseInterfaceModel):
     event_arguments: Dict[str, Any]
     """The arguments to the event, including both indexed and non-indexed data."""
 
+    @validator("contract_address", pre=True)
+    def validate_address(cls, value):
+        from ape import convert
```
antazoey, line 204: you dont need to do this import
you can do:

```python
return cls.conversion_manager.convert(value, AddressType)
```

src/ape/types/__init__.py
```diff
@@ -276,6 +302,48 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
         return self.event_arguments.get(item, default)
 
 
+class MockContractLog(BaseContractLog):
+    """
+    A mock version of the ContractLog class used for testing purposes.
```
antazoey, line 307:
```suggestion
    A mock version of the :class:`~ape.types.ContractLog` class used for testing purposes.
```

src/ape/types/__init__.py
```diff
@@ -276,6 +302,48 @@ def get(self, item: str, default: Optional[Any] = None) -> Any:
...
+    This class is designed to match a subset of event arguments in a ContractLog instance
+    by only comparing those event arguments that the user explicitly provides.
+
+    Inherits from BaseContractLog, and overrides the equality method for custom comparison
```
antazoey, line 311:
```suggestion
    Inherits from :class:`~ape.types.BaseContractLog`, and overrides the equality method for custom comparison
```

docs/methoddocs/types.md
```diff
@@ -4,7 +4,7 @@
 
 ```{eval-rst}
 .. automodule:: ape.types
-    :members: BlockID, AddressType, ContractLog
+    :members: BlockID, AddressType, ContractLog, MockContractLog
```
antazoey, line 7: Do we want to add `BaseContractLog` here as well?

antazoey (commented):
looks like `black` and `mypy` are concerned

fubuloubu (dismissed):

src/ape/contracts/base.py
```diff
@@ -463,6 +463,50 @@ def __len__(self):
...
+            )
+
+        # Convert the arguments using the conversion manager
+        converted_args = {}
+        for key, value in event_args.items():
+            if value is None:
+                continue
+            input_abi = next(input for input in self.abi.inputs if input.name == key)
+            type: Union[str, Type[AddressType], Type[int]] = input_abi.canonical_type
+            if type == "address":
```
antazoey, line 497: this might be getting too ecosystem specific here, since we have this same logic in a helper method in the Ethereum class called `self._python_type_for_abi_type`.

Here are some suggestions:

1. Create another `EcosystemAPI` method called `get_python_types(abi_type) -> List[Type]` or something... and share  the logic between the two cases.
2. Create an internal helper method and share the logic and put a TODO comment here to make this more non-ABI specific (non ecosystem specific), meaning less assumptuous  about type names in the ABI

For example, let's say a new zK ecosystem comes out and their ABIs use `contract_id` instead of `address` for whatever reason. In theory, ape should be able to handle that because of the plugin system.

z80dev: I implemented this, where ethereum's implementation calls `self._python_type_for_abi_type` and it seems to be working as intended

docs/methoddocs/types.md
z80dev: Added

src/ape/types/__init__.py
```diff
@@ -195,6 +195,29 @@ class ContractLog(BaseInterfaceModel):
     event_arguments: Dict[str, Any]
     """The arguments to the event, including both indexed and non-indexed data."""
 
+    @validator("contract_address", pre=True)
+    def validate_address(cls, value):
+        from ape import convert
+
+        return convert(value, AddressType)
```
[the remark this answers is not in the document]
z80dev: Added

src/ape/api/networks.py
```diff
@@ -509,6 +509,18 @@ def enrich_calltree(self, call: CallTreeNode, **kwargs) -> CallTreeNode:
         """
         return call
 
+    @abstractmethod
+    def get_python_types(self, abi_type) -> List[Type]:
```
fubuloubu, line 513:
```suggestion
    def get_python_types(self, abi_type: str) -> List[Type]:
```

tests/functional/test_contract_event.py
```diff
@@ -78,14 +78,14 @@ def test_contract_logs_index_access(contract_instance, owner, assert_log_values)
     contract_instance.setNumber(2, sender=owner)
     contract_instance.setNumber(3, sender=owner)
 
-    assert_log_values(event_type[0], 1)
-    assert_log_values(event_type[1], 2)
-    assert_log_values(event_type[2], 3)
+    assert event_type[0] == contract_instance.NumberChange(None, 0, None, 1)
```
fubuloubu, line 81: I wasn't picturing support `None`, but I could see the benefit of it

@antazoey what do you think?

fubuloubu (dismissed):

tests/functional/test_contract_event.py
antazoey: wouldnt it be needed for full positional support?

src/ape/api/networks.py
```diff
@@ -509,6 +509,18 @@ def enrich_calltree(self, call: CallTreeNode, **kwargs) -> CallTreeNode:
         """
         return call
 
+    @abstractmethod
```
antazoey, line 512: adding an abstract method is technically a breaking chance as it requires plugin to have implemented it.
to get around this, you can use `@raises_not_implemented` or come up with a really generic / basic implementation.

src/ape/contracts/base.py
```diff
@@ -468,6 +468,46 @@ def __len__(self):
...
+            k for k in kwargs.keys() if k is not None
+        )
+
+        if overlapping_keys:
+            raise ValueError(
+                f"Overlapping keys found in arguments: '{', '.join(overlapping_keys)}'."
+            )
+
+        # Update event_args with keyword arguments
+        event_args.update(kwargs)  # type: ignore[arg-type]
```
antazoey, line 485: rather than a type ignore here, you can do `evert_args: Dict` at the top to tell mypy to not care about the types of the elements.

src/ape/contracts/base.py
```diff
@@ -468,6 +468,46 @@ def __len__(self):
...
+
+        # Check that event_args.keys() is a subset of the expected input names
+        if unknown_input_names := set(event_args.keys()) - {
+            input.name for input in self.abi.inputs
+        }:
+            raise ValueError(
+                f"Invalid argument keys found, expected a subset of {unknown_input_names}"
```
antazoey, line 492: m
```suggestion
                f"Invalid argument keys found, expected a subset of {', '.join(unknown_input_names)}"
```

src/ape/contracts/base.py
```diff
@@ -468,6 +468,46 @@ def __len__(self):
...
+            )
+
+        # Convert the arguments using the conversion manager
+        converted_args = {}
+        for key, value in event_args.items():
+            if value is None:
+                continue
+            input_abi = next(input for input in self.abi.inputs if input.name == key)
+            ecosystem = self.provider.network.ecosystem
+            type = ecosystem.get_python_types(input_abi)
```
antazoey, line 502: it is not necessary to overwrite python builtin keywords here

```suggestion
            py_type = ecosystem.get_python_types(input_abi)
```

antazoey (changes requested):
the one blocking issue is that we technically cannot do an abstractmethod at this time, have to do an alternative (raises_not_implemented, generic default, or event make Ethereum approach the default).

src/ape/contracts/base.py
z80dev: I'm still getting the type error unless I use `event_args: Dict[Any, Any]`, just `Dict` still pops the same error

fubuloubu (dismissed):

antazoey (dismissed):
looks like some slight linting concerns but then we good

src/ape/api/networks.py
z80dev: I think the right type here is `ABIType`

antazoey (dismissed):

antazoey (approved):

antazoey (approved):

src/ape/api/networks.py
antazoey: it is potentially `Union[str, ABIType]`

tests/functional/test_contract_event.py
antazoey: `None` makes sense as a skip for EVM languages... but if the language has a concept of `null` (not zero address or zero), we may desire a different behavior here.

🤔

antazoey: would `"*"` be better as the skip value?

```python
 assert event_type[0] == contract_instance.NumberChange("*", 0, "*", 1)
```

otherwise, using kwargs works too.
on the fence with `None`.

antazoey: you are right, good point.. and also `None` makes sense because it is fundamentally the same as a kwarg when not specified, so we have to do it that way

src/ape/api/networks.py
antazoey: we can change `_python_type_for_abi_type` to also be `Union[str, ABIType]`, it should still work

antazoey: ok  just checked, youre right it is just ABIType i think

PR #1412, opened 2023-04-25
test: add tests for return_value [APE-896]

antazoey:
### What I did

I noticed we didnt have any tests for `return_value` in ape core itself.
I think this is because we added geth tracing stuff later on and included that stuff in our tests but didnt copy in any thing for return value

i am trying to debug something for some of our users and i wanted to be sure it worked in ape. seems like it does. let's include these tests so i dont have to do that again next time.

### How I did it

test various return types

**note** i also made sure it still works in foundry and hardhat, but they also already have their own tests.

### How to verify it

n/a or test suite  passes

### Checklist


antazoey:
note, looks like the actual bug id like to fix is related to proxies most likely

fubuloubu (approved):

fubuloubu:
> note, looks like the actual bug id like to fix is related to proxies most likely

yeah, I think we need a lot more proxy testing for the different types of proxies too

Issue #1339, opened 2023-03-15
ApeX Doesn't Work with the Development Version of Web3.py


arjunaskykok:
### Environment information

- OS: linux
- Python version: 3.11
- `ape` and plugin versions:  the latest stable version

### What went wrong?

```bash
$ python3.11 -m venv .venv
$ source .venv/bin/activate
(.venv) $ pip install eth-ape
$ ape init
ImportError: cannot import name 'getargspec' from 'inspect' (/usr/lib/python3.11/inspect.py)
(.venv) $ pip install --upgrade https://github.com/ethereum/web3.py
ERROR: pip\'s dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
eth-ape 0.1.0a2 requires eth-account<0.6.0,>=0.5.4, but you have eth-account 0.8.0 which is incompatible.
(.venv) $ ape init
Traceback (most recent call last):
  File "/tmp/aa/.venv/bin/ape", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/tmp/aa/.venv/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aa/.venv/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/tmp/aa/.venv/lib/python3.11/site-packages/click/core.py", line 1651, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aa/.venv/lib/python3.11/site-packages/click/core.py", line 1698, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aa/.venv/lib/python3.11/site-packages/ape/_cli/__init__.py", line 24, in get_command
    return CLI_PLUGINS[name]()
           ~~~~~~~~~~~^^^^^^
KeyError: 'init'
```

Right now, the fix for Python 3.11 has been put into the development version of web3.py. That's why I installed the web3.py library from GitHub.

fubuloubu:
[...]

PR #1414, opened 2023-04-26
feat: python 11

antazoey:
### What I did

fixes: #1339 
Fixes: APE-715

### How I did it

* Add ci test task
* Change default from 3.8 to 3.9 (to get ready for dropping 3.8)
* update setup.py project tag and supported versions range
* update readme

### How to verify it

everything works the same on 3.11

### Checklist


fubuloubu:
See notes in ethPM-types update, except we actually do use pysha3 here I think

.github/workflows/test.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
         strategy:
             matrix:
                 os: [ubuntu-latest, macos-latest]   # eventually add `windows-latest`
-                python-version: [3.8, 3.9, "3.10"]
```
antazoey, line 66: i removed 3.8 here to save time and because i think we should drop support soon, maybe on our 0.7 release

fubuloubu (approved):

.github/workflows/test.yaml
fubuloubu: Do we want to release this now with 3.8 still supported, and deprecate for 0.7? Not sure if 3.8 can be deprecated yet because of wider OS support (need to check in on this)

fubuloubu: Can we add 3.8 back again?

antazoey: Sure.

PR #1417, opened 2023-04-27
fix: restore pending timestamp at end of isolated context [APE-906]

0xkorin:
### What I did

I noticed that there is a subtle difference in behaviour between Hardhat and Foundry as provider. If you set `chain.pending_timestamp` prior to using an isolated context, Hardhat will re-set the pending timestamp when the context has been exited, while Foundry doesn't. This PR equalizes the behaviour such that the pending timestamp is always restored at the end of the context.

### How I did it

Store the pending timestamp when entering the context, restore it during exit.

### How to verify it

```py
ts = chain.pending_timestamp + 3600
chain.pending_timestamp = ts
with chain.isolate():
  chain.mine()
assert chain.pending_timestamp == ts
```

### Checklist


antazoey:
OK I added some commits, hope that is ok, just because geth and eth tester are both kind of weird compared to ganache, hardhat, and anvil, and they are the ones core uses in its test suite so we had failing tests.

So we have to make sure all the providers work.

I tested on every provider; we are good.

antazoey (approved):

fubuloubu (approved):

0xkorin:
> OK I added some commits, hope that is ok, just because geth and eth tester are both kind of weird compared to ganache, hardhat, and anvil, and they are the ones core uses in its test suite so we had failing tests.
> 
> So we have to make sure all the providers work.
> 
> I tested on every provider; we are good.

Perfect, thanks!

fubuloubu:
@0xkorin please rebase!

PR #1418, opened 2023-04-27
fix: contract fixture update [APE-907]

NotPeopling2day:
Still haven't gotten it to find the fixture, but maybe here's another path to try out.

### DIFF
--- a/src/ape/pytest/fixtures.py
+++ b/src/ape/pytest/fixtures.py
@@ -6,11 +6,11 @@ import pytest
 
 from ape.api import ReceiptAPI, TestAccountAPI
 from ape.logging import logger
-from ape.managers.chain import ChainManager
+from ape.managers.chain import ChainManager, ContractInstance
 from ape.managers.networks import NetworkManager
 from ape.managers.project import ProjectManager
 from ape.pytest.config import ConfigWrapper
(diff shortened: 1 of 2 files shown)


tests/integration/cli/projects/test/tests/test_fixtures.py
```diff
@@ -13,7 +13,7 @@ def test_chain(chain):
 
 
 def test_contract(Contract):
-    assert isinstance(Contract, ContractInstance)
+    assert isinstance(Contract("0x0000000000000000000000000000000000000000"), ContractInstance)
```
fubuloubu, line 16: Technically, this won't be a contract
```suggestion
    assert isinstance(Contract("0x0000000000000000000000000000000000000000"), Address)
```

src/ape/pytest/fixtures.py
```diff
@@ -77,11 +77,13 @@ def project(self) -> ProjectManager:
         return self.project_manager
 
     @pytest.fixture(scope="session")
-    def Contract(self):
+    def Contract(self, address: AddressType) -> ContractInstance:
```
antazoey, line 80: parameters to fixtures are other fixtures, meaning this would be referencing an `address` fixture? i don't think this is the right idea.

Issue #1419, opened 2023-04-28
Control caching behavior of contract_type and proxy_info on live networks


BowTiedDevil:
### Overview
When fetching contract data from a block explorer on a live network, the results are cached in `.ape/{ecosystem_name}/{network_name}/contract_types/{address}.json`, which is very helpful if you intend to use the contract object again. 

However for contracts instantiated with a limited ABI (a minimal ERC-20, for example), the cache remains and must be manually removed and re-fetched to restore full ABI functionality, or to clear storage space.

### Specification
An existing JSON file may be passed as the `contract_type` using `Contract(address,contract_type=ContractType.parse_file(path))` where `ContractType` is imported from `ethpm_types`

The caching behavior could be controlled with an optional argument `cache` that defaults to `True` so existing scripts behave as expected.

Example call that would disable caching to storage:
`Contract(address=POOL_ADDRESS, contract_type=ContractType.parse_file(FILE_PATH), cache=False)`

### Dependencies
None

BowTiedDevil:
I will redo the PR after reviewing CONTRIBUTING.md and reformatting

fubuloubu:
`contract_type` already overrides whatever the cache has (I believe), so we can add a kwarg `ignore_cache` or `update_cache` (boolean defaulted to False) that can be set true to force re-downloading the type via explorer. I think this will improve the UX around purging the cache value in cases where it's not set correctly

Alternatively, there could be some algorithm that merges what's in the cache with whatever the explorer provides (if available), or whatever a user provides via `contract_type` input. This could be used to solve some of the issues with proxies where the proxy has additional methods from the implementation contract (such as upgradeable proxies)

BowTiedDevil:
I had not considered the case of a forced update, I will work on this and wrap into the next PR.

fubuloubu:
> I had not considered the case of a forced update, I will work on this and wrap into the next PR. 

Thanks for taking this on!

PR #1421, opened 2023-04-29
feat: Add argument to control caching, block explorer fetching, forced disk cache updates [APE-913]

BowTiedDevil:
### What I did

Implement booleans with default values to control:
- if retrieved storage types will be cached to disk (useful when working with minimal ABIs) 
- if the `get` method will attempt to retrieve a contract type from the block explorer
- if a forced disk cache update is performed after contract type is retrieved

fixes: #1419 

### How I did it

Add three boolean args (`cache_to_disk: bool = True`, `fetch_from_explorer: bool = True`, `force_disk_cache_update: bool = False`) to two `ContractCache` methods defined in ape / managers / chain.py:
- `get`
- `instance_at`

Add `if` checks before calls to:
- `_cache_contract_to_disk`
- `_cache_proxy_info_to_disk`

Move the caching logic out of `_get_contract_type_from_explorer` into the parent `get` method so it can be managed separately (`_get_contract_type_from_explorer` should be a single-concern method, and it is only called by `get` so this should not break any other functionality).

Refactor the contract type logic inside `get`:
- Track types retrieved from memory, disk cache, and from the block explorer separately
- After retrieving, determine the 'canonical' contract type by comparing against `None` and `default` contract type
- Then manage updates to the cache

Treat calls to `instance_at` with a value for `contract_type` with special behavior:
- If a contract type was provided, assume it was done with the intention of skipping the block explorer fetch, so `fetch_from_explorer` is overridden to `False`.

Wherever contract types are compared to default values, explicitly check against `contract_type is None` instead of `not contract_type`.

### Backwards Compatibility
A default call to `instance_at` (no `contract_type`, `cache_to_disk: bool = True`, `fetch_from_explorer: bool = True`) must preserve the expected behavior of earlier Ape releases (check memory, check the block explorer, cache the result to disk).

### How to verify it
Simple script to verify behavior against a known Uniswap V3 LP address, can observe the caching behavior in a separate terminal:

```
from ethpm_types import ContractType
import os

os.environ['ETHERSCAN_API_KEY'] = 'edit me'

POOL_ADDRESS = "0xCBCDF9626BC03E24F779434178A73A0B4BAD62ED"
V2_FILE_PATH = "./v2lp.json"
V3_FILE_PATH = "./v3lp.json"

with ape.networks.ethereum.mainnet.use_provider("geth"):

    print("Loading contract with UniswapV2 LP ABI (cache disabled)")
    contract_v2_no_cache = ape.Contract(
        address=POOL_ADDRESS,
        contract_type=ContractType.parse_file(V2_FILE_PATH),
        cache_to_disk=False,
    )

    print("Loading contract with UniswapV3 LP ABI (cache disabled)")        
    contract_v3_no_cache = ape.Contract(
        address=POOL_ADDRESS,
        contract_type=ContractType.parse_file(V3_FILE_PATH),
        cache_to_disk=False,        
    )
    
    print("Loading contract with UniswapV2 LP ABI (cache enabled)")        
    contract_v2_cached = ape.Contract(
        address=POOL_ADDRESS,
        contract_type=ContractType.parse_file(V2_FILE_PATH),        
    )
      
    print("Loading contract with UniswapV3 LP ABI (cache enabled)")        
    contract_v3_cached = ape.Contract(
        address=POOL_ADDRESS,
        contract_type=ContractType.parse_file(V3_FILE_PATH),        
    )
    
    print("Loading contract without ABI (cache enabled, block explorer fetch enabled)")
    contract_v3_from_explorer = ape.Contract(
        address=POOL_ADDRESS,        
    )    

    print("Loading contract with UniswapV2 LP ABI (forced disk cache update)")        
    contract_v2_forced = ape.Contract(
        address=POOL_ADDRESS,
        contract_type=ContractType.parse_file(V2_FILE_PATH),        
        force_disk_cache_update=True,
    )
```

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -963,7 +963,11 @@ class ContractCache(BaseManager):
         return contract_types
 
     def get(
-        self, address: AddressType, default: Optional[ContractType] = None
+        self,
+        address: AddressType,
+        default: Optional[ContractType] = None,
+        cache: bool = True,
+        update_cache: bool = True,
     ) -> Optional[ContractType]:
         """
         Get a contract type by address.
@@ -975,6 +979,10 @@ class ContractCache(BaseManager):
             address (AddressType): The address of the contract.
             default (Optional[ContractType]): A default contract when none is found.
               Defaults to ``None``.
+            cache (bool): Cache the result for live networks
+              Defaults to ``True``.
+            update_cache (bool): Update the cache with the new contract
+              Defaults to ``False``.
 
         Returns:
             Optional[ContractType]: The contract type if it was able to get one,
@@ -999,6 +1007,8 @@ class ContractCache(BaseManager):
             return default
 
         contract_type = self._get_contract_type_from_disk(address_key)
+
+        # Handle the case where no contract is cached on disk
         if not contract_type:
             # Contract could be a minimal proxy
             proxy_info = self._local_proxies.get(address_key) or self._get_proxy_info_from_disk(
@@ -1007,22 +1017,26 @@ class ContractCache(BaseManager):
 
             if not proxy_info:
                 proxy_info = self.provider.network.ecosystem.get_proxy_info(address_key)
-                if proxy_info and self._is_live_network:
+                if proxy_info and self._is_live_network and (cache or update_cache):
                     self._cache_proxy_info_to_disk(address_key, proxy_info)
 
             if proxy_info:
                 self._local_proxies[address_key] = proxy_info
-                return self.get(proxy_info.target, default=default)
+                return self.get(proxy_info.target, default=default, cache=cache)
 
             if not self.provider.get_code(address_key):
                 if default:
                     self._local_contract_types[address_key] = default
-                    self._cache_contract_to_disk(address_key, default)
+                    if cache or update_cache:
+                        self._cache_contract_to_disk(address_key, default)
 
                 return default
 
-            # Also gets cached to disk for faster lookup next time.
             contract_type = self._get_contract_type_from_explorer(address_key)
+
+            # Cache to disk for faster lookup next time.
+            if contract_type and (cache or update_cache):
+                self._cache_contract_to_disk(address, contract_type)
 
         # Cache locally for faster in-session look-up.
         if contract_type:
@@ -1031,14 +1045,16 @@ class ContractCache(BaseManager):
         if not contract_type:
             if default:
                 self._local_contract_types[address_key] = default
-                self._cache_contract_to_disk(address_key, default)
+                if cache or update_cache:
+                    self._cache_contract_to_disk(address_key, default)
 
             return default
 
         if default and default != contract_type:
             # Replacing contract type
             self._local_contract_types[address_key] = default
-            self._cache_contract_to_disk(address_key, default)
+            if cache or update_cache:
+                self._cache_contract_to_disk(address_key, default)
             return default
 
         return contract_type
@@ -1062,6 +1078,8 @@ class ContractCache(BaseManager):
         address: Union[str, AddressType],
         contract_type: Optional[ContractType] = None,
         txn_hash: Optional[str] = None,
+        cache: bool = True,
+        update_cache: bool = False,
     ) -> ContractInstance:
         """
         Get a contract at the given address. If the contract type of the contract is known,
@@ -1080,10 +1098,17 @@ class ContractCache(BaseManager):
               in case it is not already known.
             txn_hash (Optional[str]): The hash of the transaction responsible for deploying the
               contract, if known. Useful for publishing. Defaults to ``None``.
+            cache (bool): Cache the result for live networks
+              Defaults to ``True``.
+            update_cache (bool): Re-fetch ContractType from the block explorer and update the cache
+              Defaults to ``False``.
 
         Returns:
             :class:`~ape.contracts.base.ContractInstance`
         """
+
+        if update_cache and not cache:
+            raise ValueError("update_cache=True cannot be set with cache=False")
 
         if self.conversion_manager.is_type(address, AddressType):
             contract_address = cast(AddressType, address)
@@ -1095,7 +1120,9 @@ class ContractCache(BaseManager):
 
         try:
             # Always attempt to get an existing contract type to update caches
-            contract_type = self.get(contract_address, default=contract_type)
+            contract_type = self.get(
+                contract_address, default=contract_type, cache=cache, update_cache=update_cache
+            )
         except Exception as err:
             if contract_type:
                 # If a default contract type was provided, don't error and use it.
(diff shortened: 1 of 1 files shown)


src/ape/managers/chain.py
```diff
@@ -975,71 +980,100 @@ def get(
             address (AddressType): The address of the contract.
             default (Optional[ContractType]): A default contract when none is found.
               Defaults to ``None``.
+            cache_to_disk (bool): Cache the fetched contract type to  (live networks only).
```
fubuloubu, line 983:
```suggestion
            cache_to_disk (bool): Cache the fetched contract type to disk (live networks only).
```

fubuloubu (commented):
Great job! Will need to give it some QA from the team to double check the new features.

Also needs some tests 😁

src/ape/managers/chain.py
BowTiedDevil: Of course! I'll review some old PRs to see how y'all deal with tests and make updates to the PR 👌

BowTiedDevil:
> Great job! Will need to give it some QA from the team to double check the new features.
> 
> Also needs some tests grin


Of course! I'll review some old PRs to see how y'all deal with tests and make updates to mine 👌

src/ape/managers/chain.py
```diff
@@ -975,71 +980,100 @@ def get(
...
+                    if cache_to_disk:
+                        self._cache_contract_to_disk(address_key, default)
 
                 return default
 
-            # Also gets cached to disk for faster lookup next time.
-            contract_type = self._get_contract_type_from_explorer(address_key)
+        # Attempt to get the contract type from block explorer
+        # (unless overridden by `fetch_from_explorer=False`)
+        contract_type_from_explorer = (
```
antazoey, line 1043: this defeats the purpose of caching, doesn't it?
the explorer API method is hit every time with these changes.

should `fetch_from_explorer` default to `False` if the contract type was not found on disk and `True` otherwise? That would safe an API call in most cases I believe.

antazoey (changes requested):
for the caching to make sense, we can only load (by default) if the faster approach is not available

adding the kwargs is a good idea tho! greatly appreciated

src/ape/managers/chain.py
```diff
@@ -975,71 +980,100 @@ def get(
...
+        elif contract_type_from_explorer is not None and contract_type_from_explorer != default:
+            # If a contract type was not provided, and one was found from the block explorer, use it
+            contract_type = contract_type_from_explorer
+        elif contract_type_from_disk is not None and contract_type_from_disk != default:
+            # If a contract type was not provided, and one was found from disk, use it
+            contract_type = contract_type_from_disk
+        elif contract_type_from_memory is not None and contract_type_from_memory != default:
+            contract_type = contract_type_from_memory
+        else:
+            raise ValueError(
```
antazoey, line 1061: In the `get()` method, when the contract type is not found,  it should return `None`.
The `__getitem__` method is the one that raises.
This mimics the Pythonic dictionary.

Adding this raise statement is a breaking change in that sense as well.

BowTiedDevil: Ah, I understand the intent here. I will rework this to eliminate raising the exception.

src/ape/managers/chain.py
```diff
@@ -975,71 +980,100 @@ def get(
...
 
             return default
 
-        contract_type = self._get_contract_type_from_disk(address_key)
-        if not contract_type:
+        # Attempt to get the contract type from memory
+        contract_type_from_memory = self._local_contract_types.get(address_key)
+
+        # Attempt to get the contract type from disk
+        contract_type_from_disk = self._get_contract_type_from_disk(address_key)
```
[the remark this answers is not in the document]
BowTiedDevil: Is the primary purpose of the memory cache to avoid disk I/O? I can rework the logic once I better understand the modality of using one or the other.

src/ape/managers/chain.py
fubuloubu: Wonder if there's some way to validate that an ABI is "fully specified", which would bypass this check because we know we can't expand it anymore to be "more" correct

Would also have to account for  proxymutability in this scenario too

fubuloubu: Gives me ideas for refactoring to the `QueryAPI` system. Basically just fetch the contract type from any available source (cache, etherscan, sourcity, etc.) and merges the results into the best possible ABI

If etherscan is rate limiting, just have it bypass. If the cache was set to a lesser ABI, extend it with info from external sources, etc.

antazoey: if etherscan had a webhook for contract source updates, we could theoretically automatically update caches in the background.

src/ape/managers/chain.py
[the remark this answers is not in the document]
BowTiedDevil: It's simple enough to restore the previous behavior (return the contract type immediately if found in memory). 

Is there a means whereby the memory cache could have some stale state that you'd like to ignore, then continue to the block explorer for a refresh & update? If so, it may be better to ignore or remove the existing entry in the memory cache if some forced update is being performed.

@fubuloubu suggested a kwarg to ignore the cache, which may be more intuitive than interpreting intent based on the set of kwargs I've added thus far.

src/ape/managers/chain.py
BowTiedDevil: > this defeats the purpose of caching, doesn't it? the explorer API method is hit every time with these changes.
The case I'm thinking of is memory or disk has cached some minimal ABI, but you want to update it from the block explorer. `get` is called through `instance_at`, where I've added a ternary statement that disables the block explorer fetch if a contract type was passed via the `default` argument. 
 
> should `fetch_from_explorer` default to `False` if the contract type was not found on disk and `True` otherwise? That would safe an API call in most cases I believe.
If `cache_to_disk` was set to `True` the first time a contract was loaded, the disk cache will be empty on subsequent runs. If we wanted to update the disk cache later, I think we should discount the state of whatever is in the disk cache (missing or outdated).

src/ape/managers/chain.py
BowTiedDevil: I've removed the exception, falling through all the `if`/`elif` checks will preserve the return behavior since `contract_type=None` if not set to a type from explorer, disk cache, or memory

src/ape/managers/chain.py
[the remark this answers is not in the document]
antazoey: ya, i suppose we need to add kwargs for that too, so we can force update from explorer and what not.

src/ape/managers/chain.py
```diff
@@ -975,71 +980,96 @@ def get(
...
             # Don't check disk-cache or explorer when using local
             if default:
                 self._local_contract_types[address_key] = default
 
             return default
 
-        contract_type = self._get_contract_type_from_disk(address_key)
-        if not contract_type:
+        # Attempt to get the contract type from memory
+        contract_type_from_memory = self._local_contract_types.get(address_key)
```
antazoey, line 1005: - [ ] short circuit if found in memory and the user is not updating any caches

antazoey (changes requested):
i think the tests will definitely need to be updated, this logic gets kinda messy and confusing.
also see comments

antazoey (changes requested):
i think the tests will definitely need to be updated, this logic gets kinda messy and confusing.
also see comments

Issue #1112, opened 2022-10-29
Add support for Private Transactions to ProviderAPI


fubuloubu:
### Overview

Some providers have support for [`eth_sendPrivateTransaction`](https://docs.flashbots.net/flashbots-auction/searchers/advanced/rpc-endpoint#eth_sendprivatetransaction) RPC request, such as [Alchemy](https://docs.alchemy.com/reference/eth-sendprivatetransaction).

NOTE: There is a return value for this method that includes an `id` used by the provider to keep track of the request

NOTE: There is also `eth_cancelPrivateTransaction`, but unsure how to support that within the Ape transaction workflow

### Specification

Add a new method to `ProviderAPI` called `send_private_transaction`:

```py
class ProviderAPI(...):
    ...
    @raises_not_implemented
    def send_private_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
        """
        Send a transaction through a private mempool (if supported by the Provider)
        """
```

Providers like Alchemy (and eventually Flashbots) can add this

Add a new keyword arg `private` to anywhere that generates a transaction request (such as `ContractTransaction.__call__` and `AccountAPI.call`) which is a `bool` that when set to `True` will attempt to use `ProviderAPI.send_private_transaction` to publish the signed transaction payload. `private` should be set to `False` by default.

If the Provider does not support `send_private_transaction`, by default it will raise as unimplemented, triggering the user to realize that they must remove the `private` kwarg from their request or choose a different provider.

### Dependencies

N/A

PR #1422, opened 2023-04-30
feat: private transaction support [APE-915]

deepsp94:
### What I did

1. Added send_private_transaction as an abstract method to ProviderAPI
2. Added a `private` bool to `transfer` and `call` methods of AccountAPI to test that a "not implemented" error is raised

fixes: https://github.com/ApeWorX/ape/issues/1112

### How I did it

Pretty much as described in https://github.com/ApeWorX/ape/issues/1112. Will follow up with a PR on ape-alchemy that actually implements the `send_private_transaction` abstract method created here.

### How to verify it

WIP

### Checklist

### DIFF
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -168,6 +168,19 @@ contract = Contract("0x123...")
 receipt = contract(sender=sender, gas=40000, data="0x123")
 ```
 
+### Private Transactions
+
+If you are using a provider that implements `send_private_transaction()`, you are able to use the `private=True` kwarg to privately publish your transaction.
+For example, EVM providers likely will use the `eth_sendPrivateTransaction` RPC to achieve this.
+
+To send a private transaction, do the following:
+
+```python
+receipt = contract.set_number(sender=dev, private=True)
+```
+
+The `private=True` is available on all contract interaction.
+
 ## Decoding and Encoding Inputs
 
 If you want to separately decode and encode inputs without sending a transaction or making a call, you can achieve this with Ape.
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -170,7 +170,7 @@ receipt = contract(sender=sender, gas=40000, data="0x123")
 
 ### Private Transactions
 
-If you are using a provider that implements `send_private_transaction()`, you are able to use the `private=True` kwarg to privately publish your transaction.
+If you are using a provider that allows private mempool transactions, you are able to use the `private=True` kwarg to publish your transaction into a private mempool.
 For example, EVM providers likely will use the `eth_sendPrivateTransaction` RPC to achieve this.
 
 To send a private transaction, do the following:
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -179,7 +179,7 @@ To send a private transaction, do the following:
 receipt = contract.set_number(sender=dev, private=True)
 ```
 
-The `private=True` is available on all contract interaction.
+The `private=True` is available on all contract interactions.
 
 ## Decoding and Encoding Inputs
 
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -101,8 +101,8 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             txn (:class:`~ape.api.transactions.TransactionAPI`): An invoke-transaction.
             send_everything (bool): ``True`` will send the difference from balance and fee.
               Defaults to ``False``.
-            private (bool): ``True`` will use the `eth_sendPrivateTransaction` RPC method.
-              Provider needs to implement `send_private_transaction` method.
+            private (bool): ``True`` with use the
+              :meth:`~ape.api.providers.ProviderAPI.send_private_transaction` method.
             **signer_options: Additional kwargs given to the signer to modify the signing operation.
 
         Returns:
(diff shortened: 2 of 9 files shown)


fubuloubu (dismissed):
Alchemy has support for this method. If you don't mind adding a PR to add this method to `ape-alchemy` we can test this a little bit better in parallel

src/ape/api/accounts.py
```diff
@@ -98,6 +99,8 @@ def call(
             txn (:class:`~ape.api.transactions.TransactionAPI`): An invoke-transaction.
             send_everything (bool): ``True`` will send the difference from balance and fee.
               Defaults to ``False``.
+            private (bool): ``True`` will use the `eth_sendPrivateTransaction` RPC method.
+              Provider needs to implement `send_private_transaction` method.
             **signer_options: Additional kwargs given to the signer to modify the signing operation.
```
antazoey, line 104: - [ ] this method should include a `Raises` section in the docs (see other methods in codebase for example) to say it may raise `APINotImplementedError` from `ape.exceptions`

src/ape/api/providers.py
```diff
@@ -402,6 +402,14 @@ def get_contract_logs(self, log_filter: LogFilter) -> Iterator[ContractLog]:
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
+    @raises_not_implemented
+    def send_private_transaction(  # type: ignore[empty-body]
+        self, txn: TransactionAPI
+    ) -> ReceiptAPI:
+        """
+        Send a transaction through a private mempool (if supported by the Provider)
```
antazoey, line 410: - [ ] args and return doc missing

src/ape/api/accounts.py
```diff
@@ -98,6 +99,8 @@ def call(
             txn (:class:`~ape.api.transactions.TransactionAPI`): An invoke-transaction.
             send_everything (bool): ``True`` will send the difference from balance and fee.
               Defaults to ``False``.
+            private (bool): ``True`` will use the `eth_sendPrivateTransaction` RPC method.
```
antazoey, line 102: it is not necessarily true that it will use `eth_sendPrivateTransaction` RPC.  the provider plugin system allows for other options, such as non-EVM based ecosystems.

- [ ] make doc more generic here

fubuloubu: Probably for local/fork tests can just have it forward to sending a normal transaction, so you can validate a script that has the `private=True` kwarg locally (e.g. it basically just ignores it)

antazoey: that makes sense

antazoey:
@deepsp94 any updates here? still working this?

src/ape/api/accounts.py
antazoey: both done

src/ape/api/providers.py
```diff
@@ -403,6 +404,29 @@ def get_contract_logs(self, log_filter: LogFilter) -> Iterator[ContractLog]:
...
+        Raises:
+            :class:`~ape.exceptions.APINotImplementedError`: If using a non-local
+              network and not implemented by the provider.
+
+        Args:
+            txn (:class:`~ape.api.transactions.TransactionAPI`): The transaction
+              to privately publish.
+
+        Returns:
+            :class:`~ape.api.transactions.ReceiptAPI`
```
antazoey, line 420: Is this actually going to return a Receipt?
From Alchemy's docs:

> eth_sendPrivateTransaction attempts to send your transaction to miners on every block for a maximum of 25 
> blocks. No need to listen for the next block. For more info about this method, check out the [official Flashbots 
> documentation](https://docs.flashbots.net/flashbots-auction/searchers/advanced/rpc-
> endpoint#eth_sendprivatetransaction).

docs/userguides/contracts.md
```diff
@@ -151,6 +151,19 @@ assert contract.get_static_list() == [1, 2, 3]
...
+If you are using a provider that implements `send_private_transaction()`, you are able to use the `private=True` kwarg to privately publish your transaction.
+For example, EVM providers likely will use the `eth_sendPrivateTransaction` RPC to achieve this.
+
+To send a private transaction, do the following:
+
+```python
+receipt = contract.set_number(sender=dev, private=True)
+```
+
+The `private=True` is available on all contract interaction.
```
fubuloubu, line 165:
```suggestion
The `private=True` is available on all contract interactions.
```

src/ape/api/providers.py
```diff
@@ -403,6 +404,29 @@ def get_contract_logs(self, log_filter: LogFilter) -> Iterator[ContractLog]:
...
+        """
+        if self.network.name == LOCAL_NETWORK_NAME or self.network.name.endswith("-fork"):
```
fubuloubu, line 422: Maybe add a warning "`private=True` is set, but bypassing private mempool because we are on a local network"

fubuloubu (approved):
This is looking great!

QA note: test with alchemy plugin update since they support this RPC

fubuloubu:
Another useful thing might be adding a field `private` to `ReceiptAPI` to show if it was sent privately or not. Could come in useful later (when async) if `ReceiptAPI` flow might allow cancellation (e.g. `receipt.cancel()`, which by default submits another self-call transaction at higher gas fees, but here would just call this RPC instead)

src/ape/api/providers.py
```diff
@@ -403,6 +404,29 @@ def get_contract_logs(self, log_filter: LogFilter) -> Iterator[ContractLog]:
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
+    def send_private_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
```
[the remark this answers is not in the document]
antazoey: im good with that

fubuloubu (dismissed):

fubuloubu (approved):

PR #1424, opened 2023-05-02
fix: remove account check for tx serialization

bilbeyt:
### What I did

There was a problem for me that happens when i am using sender as impersonated accounts like in the feature of contract calling another contract. If the gas tracking is on with tests, this check i deleted causing failed gas calculation. I looked up the address it gave me with signature and it is not known, it is not a contract/unlocked account/test account.

fixes: #

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -3,7 +3,6 @@ from enum import Enum, IntEnum
 from typing import IO, Dict, List, Optional, Union
 
 from eth_abi import decode
-from eth_account import Account as EthAccount
 from eth_account._utils.legacy_transactions import (
     encode_transaction,
     serializable_unsigned_transaction_from_dict,
@@ -60,9 +59,6 @@ class BaseTransaction(TransactionAPI):
         unsigned_txn = serializable_unsigned_transaction_from_dict(txn_data)
         signature = (self.signature.v, to_int(self.signature.r), to_int(self.signature.s))
         signed_txn = encode_transaction(unsigned_txn, signature)
-
-        if self.sender and EthAccount.recover_transaction(signed_txn) != self.sender:
-            raise SignatureError("Recovered signer doesn't match sender!")
 
         return signed_txn
 


antazoey:
it is probably better at least for now to handle signature error during the gas tracking stuff

bilbeyt:
> it is probably better at least for now to handle signature error during the gas tracking stuff

it would basically ignore any tx related with those then? currently deleting there on spot actually including them which is better imo

src/ape_ethereum/transactions.py
```diff
@@ -61,9 +60,6 @@ def serialize_transaction(self) -> bytes:
         signature = (self.signature.v, to_int(self.signature.r), to_int(self.signature.s))
         signed_txn = encode_transaction(unsigned_txn, signature)
 
-        if self.sender and EthAccount.recover_transaction(signed_txn) != self.sender:
```
antazoey, line 64: I am not opposed to deleting this, after some careful thought.
I am wondering if there is another way, in case something is actually relying on the signature error (which I kind of doubt, but just being cautious).

**Are we able to detect if the sender is an impersonated account here?** If so, maybe we can skip this check.

fubuloubu: maybe `if isinstance(self.sender, (ContractInstance, ImpersonatedAccount, AddressType))`

antazoey:
would be consider this one instead: https://github.com/ApeWorX/ape/pull/1428 ?
cc @fubuloubu wondering your thoughts on because you have other use-cases for needing this

fubuloubu:
> would be consider this one instead: https://github.com/ApeWorX/ape/pull/1428 ?
> cc @fubuloubu wondering your thoughts on because you have other use-cases for needing this

Yeah, basically the node doesn't enforce signature checks, so neither should we

antazoey:
closing in favor of https://github.com/ApeWorX/ape/pull/1428

fubuloubu:
@bilbeyt let us know if #1428 doesn't solve your problem somehow

bilbeyt:
> @bilbeyt let us know if #1428 doesn't solve your problem somehow

It is working for me thanks, although i wasnt expecting to see some negative gas usage on table, i dont know the reason.

fubuloubu:
> > @bilbeyt let us know if #1428 doesn't solve your problem somehow
> 
> It is working for me thanks, although i wasnt expecting to see some negative gas usage on table, i dont know the reason.

gas usage is a complicated topic, on occasion the refund rules could make certain functions cost a negative amount (although the whole transaction is still net positive in terms of gas use)

Issue #1429, opened 2023-05-05
description of base_fee doesn't match implementation [APE-933]


wakamex:
[base_fee](https://github.com/ApeWorX/ape/blob/f0b5b9d44ac4a50ba6b45ab5c11635b39fafcb01/src/ape/api/providers.py#L292) is defined as:
> The minimum value required to get your transaction included on the **next** block.

Which seems consistent with its usage in [prepare_transaction](https://github.com/ApeWorX/ape/blob/f0b5b9d44ac4a50ba6b45ab5c11635b39fafcb01/src/ape/api/providers.py#L1145) where a transaction that doesn't specify gas has its max gas set to `self.base_fee + txn.max_priority_fee`. If the goal here is to prepare a transaction that will be included on the **next** block, we'd need to use the base fee of the **next **block. So far so good.

The issue is that the current implementation doesn't give us that. Instead it returns the base fee of the **latest** which you can see here: https://github.com/ApeWorX/ape/blob/f0b5b9d44ac4a50ba6b45ab5c11635b39fafcb01/src/ape/api/providers.py#L640

I think this should be fixed to either:
1. Change the docstring of the base_fee property to reflect its implementation, returning the **latest** black fee
2. Change the implementation to match the description as well as the use, by having it return the **next** block fee.

I prefer solution 2 and am submitting a barebones PR right now that will attempt this: #1430

antazoey:
number 2 makes sense.

what we can do:
* today, we can fix it so that prepare_transaction uses the "pending" block
* in 0.7, change `base_fee` (the property) to do this too

wakamex:
>     * today, we can fix it so that prepare_transaction uses the "pending" block
this is the big win for me, because it affects how new transactions are set up, and will hopefully cause fewer failed/hung transactions.

>     * in 0.7, change `base_fee` (the property) to do this too
I'm not sure why you'd want to delay this part, but push the prepare_transaction change earlier.

In my mind, the bigger "feature" change is the change in behavior for creating a new transaction, which the first bullet does. So I'd think, if we were to delay anything, we should delay that change? As you could argue it's a meaningful change to expected behavior, that could throw off some people's workflow (hopefully only for the better, but still 🤞).

[...]

PR #1430, opened 2023-05-05
return next instead of current base fee

wakamex:
### What I did

use a call to web3's fee_history() function to easily get the base fee of the **next block**. shoutout to @BowTiedDevil in discord for pointing to this easy web3-native way of getting the next block base fee! 

fixes: #1429

### How I did it

copy paste BowTiedDevil's code into a repo

### How to verify it

this notebook shows the current behavior: https://github.com/wakamex/ape-stuff/blob/main/test_sanity.ipynb
I'll add another notebook that shows the new behavior.

### Checklist


src/ape/api/providers.py
```diff
@@ -637,17 +637,17 @@ def client_version(self) -> str:
 
     @property
     def base_fee(self) -> int:
-        block = self.get_block("latest")
-        if not hasattr(block, "base_fee"):
-            raise APINotImplementedError("No base fee found in block.")
+        fee_history = self.web3.eth.fee_history(1, self.get_block("latest").number)
+        if len(fee_history["baseFeePerGas"]) < 2:
```
fubuloubu, line 641: Won't this RPC call fail on the previous line if it's unimplemented? Probably needs a test in both scenarios, try it on BSC just to see

src/ape/api/providers.py
```diff
@@ -637,17 +637,26 @@ def client_version(self) -> str:
...
     @property
     def base_fee(self) -> int:
-        block = self.get_block("latest")
-        if not hasattr(block, "base_fee"):
-            raise APINotImplementedError("No base fee found in block.")
-        else:
-            base_fee = getattr(block, "base_fee")
+        latest_block_number = self.get_block("latest").number
+        if latest_block_number is None:
+            raise APINotImplementedError("Could not get block number of latest block.")
```
antazoey, line 643: this might be better as `ProviderError`

antazoey (commented):
thanks, this was a good catch.
would expect some test changes or additions though.

antazoey (commented):
thanks, this was a good catch.
would expect some test changes or additions though.

wakamex:
since [web3's eth_tester doesn't implement FeeHistory](https://github.com/ethereum/web3.py/blob/68cc94769b531149ad33f6c0e5818a1880fc8486/web3/providers/eth_tester/defaults.py#L254), this may throw a kink into... like every test. I got about 50 failing.

src/ape/api/providers.py
antazoey: EthTester does not implement the RPC and ape-geth does.
I tried EthTester using this implementation. With my new changes (see my commits), it now resorts to the old way upon receiving the RPC not implemented error.

(However, since we know EthTester doesnt support it, I just made it short-circuit to the old way right away for perf).

I tested this out on ape-geth as well! It uses the new API which is pretty awesome.
So if you use BSC, it works like my first test where it attempts to use this but fails and falls back to the way that works. Ideally, we do the same for ape-bsc that we do for ape-test by making it short-circuit and not bother.

So we should be good now; bases covered and the thing will resort to the old way when it has to. When possible, plugins should provider their own way when they dont support it.

fubuloubu (approved):

wakamex:
do y'all wanna make 2*base_fee the default for creating a new transaction? would go a long way to making transactions fail less

antazoey:
> do y'all wanna make `2*base_fee` the default for creating a new transaction? would go a long way to making transactions fail less

for all networks?
could we possible have a network config for each, that way can make mainnet less.

i always want testnet deploys to go through, i dont really care the cost.. but mainnet, failure isnt always bad, especially if we can configure it.

wakamex:
I had in mind any network that implements EIP1559. Is there even a way to check this elegantly? For devnet (aka not Goerli) there are other ways of getting your txn through every time (pay lots of gas, set gas cost to 0).

The idea was that setting the "max_fee" higher doesn't actually cost more. It just raises the upper limit. But what you actually end up paying is the same as if you re-submit it every block with `max_fee` of `1*current_base_fee`. That was in the old setup where you knew only the current block's base_fee, and if the actual base_fee came in higher, your txn would fail.

I guess this is an alternative to having accurate base fee for the next block. If we do, then setting max_fee equal to the next block's base_fee is strictly superior. Your transaction will never fail, so no need to over-estimate.

Issue #854, opened 2022-07-03
Add `test` account for running scripts locally


fubuloubu:
### Overview

It is handy to be able to test your scripts to run in a local or mainnet-fork context. However, in this environment you may not have access to an aliased account with any balance to be able to run your script. Thankfully, most local/fork mode providers come with a pre-set list of accounts that are pre-funded for any actions they may take, and additionally they do not require any interaction in order for them to perform signing actions.

However, these test accounts are not directly referable via `ape.cli.account_option`, meaning they cannot be used for testing scripts.

NOTE: This reminds me of an issue that @PatrickAlphaC raised about being difficult to test scripts because his preferred account alias did not have a balance in his local testing environment. Feel free to give us your feedback on this issue!

### Specification

Add an alias named `"TEST"` to `ape.cli.account_option` that is only available in a local or fork mode environment for use in script testing. Additionally, add documentation to the scripts userguide to help people understand that this is the recommended way to test scripts in a local and/or mainnet-fork environment prior to using them live.

```bash
$ ape run deploy --account TEST 0x1C277bD41A276F87D3E92bccD50c7364aa2FFc69
SUCCESS: Contract 'SuperSig' deployed to: 0x274b028b03A250cA03644E6c578D81f019eE1323
```

### Dependencies
n/a

unparalleled-js:
thoughts on doing `--account t0` / `--account t1` etc to refer to a specific test account by index?

this way, if you deploy using `t6` or something, you can invoke another script with the correct account.

unparalleled-js:
or even `--account 0` / `--account 1` etc

fubuloubu:
> or even `--account 0` / `--account 1` etc

I think this works already, but you just get whatever your first account is, depending on the situation. It's not consistent behavior.

fubuloubu:
> thoughts on doing `--account t0` / `--account t1` etc to refer to a specific test account by index?
> 
> this way, if you deploy using `t6` or something, you can invoke another script with the correct account.

I think you can access them from other aliases, but that wasn't the point.

The idea is to have one, idiomatic way to handle testing scripts by executing locally using a specific account

PatrickAlphaC:
Here is what I do for both brownie and ape, I create a file called `helper_functions.py` or something in the scripts folder, and add a function like this:

```python
from ape import networks, accounts
LOCAL_CHAIN_NAME = "local"

def get_account(index=None, id=None):
    if index:
        return accounts[index]
    if id:
[...]

PR #1437, opened 2023-05-16
feat(cli): add support for using test accounts in `account_option()` [APE-945]

fubuloubu:
### What I did

Support using test accounts in `account_option` through special syntax `TEST::{i}`

fixes: #854

By having the `TEST::` prefix, we disambiguate from any "normal" alias (e.g. for built-in loaded accounts), allowing people to make a call to a script for testing purposes using a pre-funded test account that functions on a development or mainnet fork environment. This is safer or more simple than what people are currently doing, which is maintaining their own expansive scripts for doing what is essentially the same thing

### How I did it

Just added support for parsing additional choices to `AccountAliasPromptChoice`

### How to verify it

Try making some different scripts, see if there's any UX gotchas

Example:
```sh
$ ape run scriptname --account TEST::0 --network ::test
```

### Checklist

### DIFF
--- a/docs/userguides/clis.md
+++ b/docs/userguides/clis.md
@@ -18,12 +18,13 @@ This guide is for showcasing utilities that ship with Ape to assist in your CLI 
 
 ## Ape Context Decorator
 
-The `@ape_cli_context` gives you access to all the root Ape objects (`accounts`, `networks` etc.), the ape logger and an `abort` method for stopping execution of your CLI gracefully.
+The `@ape_cli_context` gives you access to all the root Ape objects (`accounts`, `networks` etc.), the ape logger, and an `abort` method for stopping execution of your CLI gracefully.
 Here is an example using all of those features from the `cli_ctx`:
 
 ```python
 import click
 from ape.cli import ape_cli_context
+
 
 @click.command()
 @ape_cli_context()
@@ -37,7 +38,7 @@ def cmd(cli_ctx):
 
 The `@network_option()` allows you to select an ecosystem / network / provider combination.
 When using with the `NetworkBoundCommand` cls, you can cause your CLI to connect before any of your code executes.
-This is useful if your script or command requires a connection to a node provider in order for it to run.
+This is useful if your script or command requires a provider connection in order for it to run.
 
 ```python
 import click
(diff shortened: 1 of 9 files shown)


antazoey:
i will add docs and tests now since this is close

src/ape/cli/choices.py
```diff
@@ -126,13 +126,30 @@ def __init__(
     def convert(
         self, value: Any, param: Optional[Parameter], ctx: Optional[Context]
     ) -> Optional[AccountAPI]:
+        if isinstance(value, str) and value.startswith("TEST::"):
+            try:
+                account_idx = int(value.replace("TEST::", ""))
```
antazoey, line 131: 1 bit of feedback (that i will address):  try to limit what you put in the try / catch to _only_ include the part that will error.

for example, this should be refactored to:

```python
if isinstance(value, str) and value.startswith("TEST::"):
    value_str = value.replace("TEST::", "")
    try:
        account_idx = int(value)
        ....
```

however, even better is avoiding the try catch altogether and just using the `isnumeric()` method.
try / catch is last resort for a situation like this

src/ape/cli/choices.py
```diff
@@ -126,13 +126,30 @@ def __init__(
...
 
         # Prompt the user if they didn't provide a value.
         alias = super().convert(value, param, ctx)
         return accounts.load(alias) if alias else None
 
+    def print_choices(self):
+        super().print_choices()
+        len_test_accounts = len(accounts.test_accounts) - 1
+        click.echo(f"Or 'TEST::account_idx', where `account_idx` is in [0..{len_test_accounts}]")
+        click.echo()
```
antazoey, line 151: rather than an empty call to `echo()`, it's cleaner and simpler to append a newline character to the first stmt

antazoey:
for this to really work right, we also need to update `get_user_selected_account()` to allow for these

docs/userguides/clis.md
```diff
@@ -0,0 +1,132 @@
+# CLIs
```
fubuloubu, line 1: This was so necessary

fubuloubu, line 1: Needs to be added to index tho

antazoey (approved):
someone else also needs to approve since i did a lot on this,
or @fubuloubu  let me know when it looks good to you

antazoey (approved):
someone else also needs to approve since i did a lot on this,
or @fubuloubu  let me know when it looks good to you

fubuloubu (commented):
Approved, let's go!

PR #1438, opened 2023-05-17
refactor: widen PyYAML pin to >=5.0,<7.0 [APE-947]

fubuloubu:
### What I did

`awscli` depends on an older version of pyyaml, and I don't think we rely on any advanced behavior of this dependency, so opening up the pin seems safe

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ setup(
         "PyGithub>=1.54,<2",
         "pytest>=6.0,<8.0",
         "python-dateutil>=2.8.2,<3",
-        "pyyaml>=6.0,<7",
+        "PyYAML>=5.0,<7",
         "requests>=2.28.1,<3",
         "rich>=12.5.1,<13",
         "SQLAlchemy>=1.4.35",


setup.py
```diff
@@ -104,7 +104,7 @@
         "PyGithub>=1.54,<2",
         "pytest>=6.0,<8.0",
         "python-dateutil>=2.8.2,<3",
-        "pyyaml>=6.0,<7",
+        "PyYAML>=5.0,<7",
```
antazoey, line 107: i never realized these were case insensitive
some of the others still using casing tho

should we make all of them lower case or should we stick to the rule of using actual pypi names?
kinda nice to know what the actual pypi name is (with casing)

antazoey (approved):
is good but see comment about casing

antazoey:
@fubuloubu merge this one too?

setup.py
fubuloubu: Would use the pypi name, but can update later no biggie

fubuloubu:
> @fubuloubu merge this one too?

Yes I think this one is good to go, would help me with a pinning issue in chaos

setup.py
antazoey: oh i see, no worries then

Issue #1436, opened 2023-05-15
poll_logs implementation is suboptimal [APE-943]


banteg:
when `start_block` is in the past, the current implementation makes one request for each block before it catches up.
https://github.com/ApeWorX/ape/blob/751e85bba9f29cebc7cea81cb51322acfd6c0fb5/src/ape/contracts/base.py#L635

instead, it could use one `range` query to catch up much faster and only start doing 1-block requests after that.

additionally, i think poll_logs should support `search_topics` and `extra_addresses` to be compatible with other event querying methods.

PR #1439, opened 2023-05-17
perf: minimize unnecessary calls for poll_logs [APE-948]

BobTheBuidler:
### What I did

I added a check to see if start_block is in the past, if so it will yield from self.range called on the full range of blocks between start_block and chain head before proceeding into the new blocks iterator as currently implemented

fixes: #1436 

### How I did it

conceptually separated historical blocks from future blocks and built better handling for historical case

### How to verify it

Just fetch some events

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -618,14 +618,14 @@ class ContractEvent(ManagerAccessMixin):
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
-        height = self.chain_manager.blocks.height
-        if start_block and start_block <= height:
-            yield from self.range(start_block, height)
-            start_block = height + 1
-            
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
+        
+        height = self.chain_manager.blocks.height - required_confirmations
+        if start_block and start_block <= height:
+            yield from self.range(start_block, height)
+            start_block = height + 1
 
         for new_block in self.chain_manager.blocks.poll_blocks(
             start_block=start_block,
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -626,7 +626,7 @@ class ContractEvent(ManagerAccessMixin):
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
-        
+
         # NOTE: We process historical blocks separately here to minimize rpc calls
         height = max(self.chain_manager.blocks.height - required_confirmations, 0)
         if start_block and height > 0 and start_block < height:
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -618,6 +618,11 @@ class ContractEvent(ManagerAccessMixin):
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
+        height = self.chain_manager.blocks.height
+        if height > start_block:
+            yield from self.range(start_block, height)
+            start_block=height + 1
+            
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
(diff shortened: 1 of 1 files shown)


BobTheBuidler:
only thing I'm not satisfied with is the lack of checking for reuired confs for blocks <= chain height, thoughts?

src/ape/contracts/base.py
```diff
@@ -618,6 +618,11 @@ def poll_logs(
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
+        height = self.chain_manager.blocks.height
+        if height > start_block:
+            yield from self.range(start_block, height)
+            start_block=height + 1
```
fubuloubu, line 624:
```suggestion
            start_block = height + 1
```

src/ape/contracts/base.py
```diff
@@ -618,6 +618,11 @@ def poll_logs(
             Iterator[:class:`~ape.types.ContractLog`]
         """
 
+        height = self.chain_manager.blocks.height
+        if height > start_block:
```
fubuloubu, line 622: Just a nit
```suggestion
        if start_block <= height:
```

fubuloubu (approved):

src/ape/contracts/base.py
BobTheBuidler: I am a terrible person

BobTheBuidler:
> only thing I'm not satisfied with is the lack of checking for reuired confs for blocks <= chain height, thoughts?

I do worry about this having conflict with your reuired confs check in the future block handler, curious if you have the same concern

fubuloubu (approved):
On the number of required confs, we don't really have good handling for that today, although I think `range` does respect it I beleive

BobTheBuidler:
> > only thing I'm not satisfied with is the lack of checking for reuired confs for blocks <= chain height, thoughts?
> 
> I do worry about this having conflict with your reuired confs check in the future block handler, curious if you have the same concern

This has been resolved in latest commit

src/ape/contracts/base.py
```diff
@@ -621,6 +621,11 @@ def poll_logs(
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
+        
+        height = self.chain_manager.blocks.height - required_confirmations
+        if start_block is not None and start_block <= height:
```
fubuloubu, line 626:
```suggestion
        if start_block is not None and height > 0 and start_block <= height:
```

src/ape/contracts/base.py
```diff
@@ -621,6 +621,11 @@ def poll_logs(
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
+        
+        height = max(self.chain_manager.blocks.height - required_confirmations, 0)
+        if start_block and height > 0 and start_block < height:
```
antazoey, line 626: would this be better as `start_block is not None` because of `0`?

src/ape/contracts/base.py
```diff
@@ -621,6 +621,11 @@ def poll_logs(
...
+            yield from self.range(start_block, height)
```
antazoey, line 627: this is a nice optimization! but i think this deserves a comment because at first i was like:

`tHeSe bLoCkS aRe aLrEaDy hAnDleD beLoW!!!1`

but after reading the issue, i understand why we pre-handle them here.

src/ape/contracts/base.py
BobTheBuidler: per conversations with @fubuloubu in DM, we did it this way because if start_block == 0 the range call will fail

antazoey:
can someone make `black` happy?

src/ape/contracts/base.py
BobTheBuidler: done

antazoey:
switched to a perf because the functionality is the same, just more efficient now (and i like tracking moments when things become more performant like this)

fubuloubu (approved):

fubuloubu (approved):

banteg (approved):
looks great, thank you

PR #1443, opened 2023-05-23
chore: checking if typing extensions broke core alone [APE-968]

antazoey:
### What I did

latest typing ext broke pydantic and thus ape
checking if broken here are only noticed in plugins
if broken, will have to pin... tho tbh not sure why this dep even shows up since we are post 3.7... someone is forcing us to install this crappy thing and it is terrible and has caused so many problems for us and the Ethereum web3 community.

### Checklist


antazoey:
super confused by all the failures, none of those happen locally

antazoey:
going to try closing this PR and making a new one

PR #1445, opened 2023-05-25
fix: fixes for source tracing [APE-972]

antazoey:
### What I did

Fixes several things that came up from people trying to use the source tracing feature....

1. Add ability to check if compiler supports it (will need this in Coverage feature anyway)
2. Performance improvement by preventing the calculation of the source traceback more than once in certain situations
3. **Critical issue** causing never-ending loop (hanging) when trying to create the source traceback. For this reason, we may want to patch Core ape after this...
4. Better error when txn hash is not found
5. Issue where `contracts_folder` was `None` when it should not have been (only happened when using normal Python REPLs).
6. PC values are now present in the source traceback but also requires that `ape-vyper` PR

### Checklist

### DIFF
--- a/src/ape/api/compiler.py
+++ b/src/ape/api/compiler.py
@@ -1,3 +1,4 @@
+from functools import cached_property
 from pathlib import Path
 from typing import Dict, Iterator, List, Optional, Set, Tuple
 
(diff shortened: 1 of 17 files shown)


antazoey:
- [ ] tests

fubuloubu (approved):

PR #1454, opened 2023-05-30
feat: flattened api [APE-984]

antazoey:
### What I did

fixes: #

### Checklist

### DIFF
--- a/src/ape/api/compiler.py
+++ b/src/ape/api/compiler.py
@@ -169,7 +169,7 @@ class CompilerAPI(BaseInterfaceModel):
     def get_flattened_contract(self, path: Path, **kwargs) -> Content:
         """
         Get the content of a flattened contract via its source path.
-        Plugin implementations handle import resolution, SPX de-duplication,
+        Plugin implementations handle import resolution, SPDX de-duplication,
         and anything else needed. The default implementation of this API
         returns the content of the given contract without doing anything else.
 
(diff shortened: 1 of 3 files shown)


src/ape/api/compiler.py
```diff
@@ -166,6 +166,23 @@ def trace_source(  # type: ignore[empty-body]
             :class:`~ape.types.trace.SourceTraceback`
         """
 
+    def get_flattened_contract(self, path: Path, **kwargs) -> Content:
+        """
+        Get the content of a flattened contract via its source path.
+        Plugin implementations handle import resolution, SPX de-duplication,
```
fubuloubu, line 172: what does "SPX de-duplication" mean?

fubuloubu (approved):

src/ape/api/compiler.py
antazoey: for flattened contracts, you are only supposed to have 1 SPX identifier.
Sometimes, you may get 1 per source file and then you need to remove the duplicate ones.
is really just an example, just kinda explaining why the API generally does since it may seem like all you need to is paste the files on top of each other but actually is often more required steps (for Etherscan verification to be happy anyway)

antazoey: oh shit i meant SPDX

fubuloubu: ahhh

fubuloubu (dismissed):
What would happen if the default implementation raised NotImplementedError instead?

fubuloubu (approved):

PR #1455, opened 2023-05-30
fix: pass through base_err on empty contract logic errors [APE-985]

antazoey:
### What I did

base error was missing on txn failed
looking at this in a debug perspective for some solidity crap

### Checklist


antazoey:
basely much adding since review and i just got done doing some testing to make sure dev message still work, just in case.

PR #1458, opened 2023-05-31
wait and retry when tx not found

z80dev:
### What I did

Added try-catch logic to checking for a transaction receipt instead of failing immediately.
fixes: APE-536

### How I did it

when running transactions back to back, there was often an error for subsequent transactions not being found when querying the rpc endpoint for a transaction receipt. this exception is raised in web3.py, not ape. previously, this caused an immediate failure. I implemented a simple loop to retry instead of failing immediately. In my local testing, 2 or 3 iterations were usually enough.

I verified this by running this with a print statement in the `except` body. any time this message printed was a case where previously we would have insta-failed. in all these cases, the loop ensured execution of the script continued after a few seconds when the rpc node returned the tx receipt.

### How to verify it

Run this script from this repo: [proof of concept](https://github.com/z80dev/optimism-ape-poc/blob/main/scripts/opscript.py)

Not sure how to write tests around this since it depends on the live behavior of remote RPC's. My theory behind why this happens is because RPCs like infura (from bug report) aggregate your requests to a bunch of nodes. while your tx broadcast request hits node A, your get_tx_receipt request hits some other node B, which has never heard of your tx. the speed at which txs propagate may vary network to network, provider to provider, etc. so a simple retry loop is probably the easiest and most general way of handling this

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -23,7 +23,7 @@ from evm_trace import TraceFrame as EvmTraceFrame
 from hexbytes import HexBytes
 from pydantic import Field, root_validator, validator
 from web3 import Web3
-from web3.exceptions import BlockNotFound
+from web3.exceptions import BlockNotFound, TransactionNotFound
 from web3.exceptions import ContractLogicError as Web3ContractLogicError
 from web3.exceptions import MethodUnavailable, TimeExhausted
 from web3.types import RPCEndpoint, TxParams
@@ -1032,7 +1032,18 @@ class Web3Provider(ProviderAPI, ABC):
         except TimeExhausted as err:
             raise TransactionNotFoundError(txn_hash) from err
 
-        txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
+        max_retries = 20
+        for attempt in range(max_retries):
+            try:
+                txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
+                break
+            except TransactionNotFound as e:
+                if attempt < max_retries - 1:  # if this wasn't the last attempt
+                    time.sleep(1)  # Wait for 5 seconds before retrying.
+                    continue  # Continue to the next iteration, effectively retrying the operation.
+                else:  # if it was the last attempt
+                    raise e  # Re-raise the last exception.
+
         receipt = self.network.ecosystem.decode_receipt(
             {
                 "provider": self,


src/ape/api/providers.py
```diff
@@ -1032,7 +1032,18 @@ def get_receipt(
         except TimeExhausted as err:
             raise TransactionNotFoundError(txn_hash) from err
 
-        txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
+        max_retries = 20
```
fubuloubu, line 1035: Can this be a config item?

fubuloubu (commented):
Looking good, okay with this approach

src/ape/api/providers.py
z80dev: We probably want this to be a provider-level config right (as opposed to network)? 

Is 20 alright as the default value?

fubuloubu: provider-level would require updating all providers, I think network-level makes more sense (e.g. update ape-ethereum) since it's most likely to be an issue for a faster chain

src/ape/api/providers.py
```diff
@@ -1032,7 +1032,19 @@ def get_receipt(
...
+        for attempt in range(max_retries):
+            try:
+                txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
+                break
+            except TransactionNotFound as e:
+                if attempt < max_retries - 1:  # if this wasn't the last attempt
+                    time.sleep(1)  # Wait for 1 second before retrying.
+                    continue  # Continue to the next iteration, effectively retrying the operation.
+                else:  # if it was the last attempt
+                    raise e  # Re-raise the last exception.
```
antazoey, line 1046:
```suggestion
                    raise  # Re-raise the last exception.
```

an empty raise is the same thing but preserves the exception stack!

z80dev: woah had no idea! thanks :)

src/ape/api/providers.py
```diff
@@ -1032,7 +1032,19 @@ def get_receipt(
         except TimeExhausted as err:
             raise TransactionNotFoundError(txn_hash) from err
 
-        txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
+        network_config = self.network.config.get(self.network.name)
+        max_retries = network_config.get("max_receipt_retries")
```
antazoey, line 1036: `get()` implies this could return `None`, which it would do if the network config was missing the property.
so we need a default value.

let's set a constant (same place other constants are set like `DEFAULT_LOCAL_TRANSACTION_ACCEPTANCE_TIMEOUT`) and use that value both here as the default:

```
network_config = self.network.config.get(self.network.name, DEFAULT_MAX_RETRIES_TX)
```

as well as in the network config's default:

```python
max_receipt_retries: int = DEFAULT_MAX_RETRIES_TX
```

src/ape_ethereum/ecosystem.py
```diff
@@ -97,6 +97,7 @@ class NetworkConfig(PluginConfig):
     block_time: int = 0
     transaction_acceptance_timeout: int = DEFAULT_TRANSACTION_ACCEPTANCE_TIMEOUT
     default_transaction_type: TransactionType = TransactionType.DYNAMIC
+    max_receipt_retries = 20
```
antazoey, line 100: (see my other comment for more context)

```suggestion
    max_receipt_retries: int = MAX_RETRIES_TX
```

src/ape/api/providers.py
```diff
@@ -1032,7 +1032,19 @@ def get_receipt(
...
+        for attempt in range(max_retries):
+            try:
+                txn = dict(self.web3.eth.get_transaction(HexStr(txn_hash)))
```
antazoey, line 1039: Are you confident this resolves your issue?

It is little odd to me because we already have the receipt at this point but the txn is the one that is missing?
The receipt checking happens above, this part only fetches the txn which is weird that the chain would know about the receipt and not the txn

src/ape_ethereum/ecosystem.py
antazoey, line 100: I am a little confused about the name because it is being used when we are fetching the transaction; not when we are fetching the receipt.

Are we sure we are using this config in the right spot and/or is this the correct name describing when in the tx lifecycle that it gets used?

antazoey (commented):
ive got some questions!

src/ape/api/providers.py
z80dev: Definitely confident, I can reliably reproduce the issue without the retry loop, and it never happens with the loop. In local development and use in a personal project over the weekend I added a print statement in the loop body and saw it pretty much every time I ran a script, and that print statement in the loop body would *only* print in conditions that previously would have been an uncaught exception

I think the confusion is mine in that I shouldn't have used the word `receipt` here when it means something else

it is weird, but not that weird. this probably would never hit if you're talking to a single node. when going through infura/alchemy/other public RPC's, there are probably a number of nodes behind the same RPC endpoint, and its conceivable that one node may know about a tx before other nodes have that data gossiped to them

src/ape_ethereum/ecosystem.py
z80dev: As for the config being in the right spot, that was Doggie's suggestion, if we made this a provider config we would have to update all the provider plugins

As for naming, I think you're right that I picked the wrong name here, I'll rename this to `max_get_transaction_retries`

src/ape/api/providers.py
antazoey: oh that makes tons of sense, now I can see how you would get to this state

antazoey (approved):
Looks 🐬 ing Fantastic, thanks

antazoey:
@z80dev said it was coooo to merge

Issue #1233, opened 2023-01-19
Source the passphrase from an environment variable for unlocking a KeyfileAccount


fubuloubu:
### Overview

For scripting, it would be handy to store the passphrase for an account in relatively safe way that made it easy to retrieve without interaction in an ape session. Probably the best way we can add this feature for the built-in `KeyfileAccount` is by adding a feature to `.unlock()` that would source the passphrase from a custom environment variable for that account if it is available.

### Specification
Add this code to load from an environment variable:

```py
passphrase = passphrase or os.environ.get(f"APE_ACCOUNTS_{self.alias}_PASSPHRASE", None) or ...
```
to here:

https://github.com/ApeWorX/ape/blob/509e3b7491b1651e8c25066fb25574c19f030f85/src/ape_accounts/accounts.py#L85

When you use this feature, it would look like this:

```sh
$ export APE_ACCOUNTS_my-alias_PASSPHRASE="my super secret passphrase"
$ ape run my_script --account my-alias
INFO: Using passphrase for account 'my-alias' from environment variable
...
```

Note: the script would use `account.unlock()` internally somehow

### Dependencies

No dependencies

PR #1460, opened 2023-05-31
feat: load passphrase as environment variable [APE-999]

Aviksaikat:
### What I did

Added the option for setting the password to unlock accounts from env variables

fixes: https://github.com/ApeWorX/ape/issues/1233

### How I did it
Added a check to see if the `password`  is set in the `environment` or not. 

### How to verify it
- Open a console by using `ape console` will be the fastest way to test this. Now follow the steps
```py
# Step 1 -> export the password or save it as env var 
export APE_ACCOUNTS_test_PASSPHRASE="a"
# Step 2 
ape console
# Step 3 -> Run these commands assuming you have an account named test.
account = accounts.load("test")
account.unlock()
```

### Checklist


Aviksaikat:
working on the `type-check` part. My bad I should have updated the `__decrypt_keyfile` method

src/ape_accounts/accounts.py
```diff
@@ -160,7 +175,10 @@ def _prompt_for_passphrase(self, message: Optional[str] = None, **kwargs):
             **kwargs,
         )
 
-    def __decrypt_keyfile(self, passphrase: str) -> HexBytes:
+    def __decrypt_keyfile(self, passphrase: Optional[str]) -> HexBytes:
+        # if no passphrase loaded throw an exception
+        if not passphrase:
+            raise InvalidPasswordError()
```
fubuloubu, line 181: Don't quite understand why this is now Optional :thinking:

Aviksaikat: If you can check the return type of passphrase in method `unlock`, it's optional[str] maybe that's why python is complaining. But before that it successfully ran on my machine without breaking anything.

docs/userguides/accounts.md
```diff
@@ -210,6 +210,28 @@ message = encode_defunct(text="Hello Apes!")
 signature = account.sign_message(message)
 ```
 
+## More Secure Way
+
+By saving your passphrase as `environment` variable. Set your passphrase like this
```
antazoey, line 215:
```suggestion
Another, more secure approach is to use an environment variable.
Set your passphrase in an environment variable by following this template:
```

docs/userguides/accounts.md
```diff
@@ -210,6 +210,28 @@ message = encode_defunct(text="Hello Apes!")
...
+
+```bash
+export APE_ACCOUNTS_test_PASSPHRASE="a"
```
antazoey, line 218:
```suggestion
export APE_ACCOUNTS_<alias>_PASSPHRASE="a"
```

docs/userguides/accounts.md
```diff
@@ -210,6 +210,28 @@ message = encode_defunct(text="Hello Apes!")
...
+```
+
+Where `test` is the name of the account you want to use. Now you can use your account to make any transactions without giving your passphrase again and again
```
antazoey, line 221:
```suggestion
Where `<alias>` is the name of the account you want to use.
Now, you can use your account to make any transactions without subsequently providing your passphrase.
```

src/ape_accounts/accounts.py
antazoey: you may just have to use a different variable name or assert it is not None or something.
let's make it so it is not optional here.

antazoey (commented):
not bad! Thanks

src/ape_accounts/accounts.py
fubuloubu: Okay I see what's happening here. It will never be `None` because of how `_prompt_for_passphrase` works (which should probably have a return type hint)

fubuloubu (dismissed):

docs/userguides/accounts.md
```diff
@@ -210,6 +210,30 @@ message = encode_defunct(text="Hello Apes!")
 signature = account.sign_message(message)
 ```
 
+## More Secure Way
```
antazoey, line 213:
```suggestion
## Keyfile Passphrase Environment Variable (more secure)
```

fubuloubu (dismissed):

fubuloubu (approved):

antazoey (approved):
you da best

PR #1463, opened 2023-06-01
feat: better fallback method support [APE-1008]

antazoey:
### What I did

Allow directly calling an address.

- [x] Requires ethpm-types to get released, doing that now

Fixes: APE-1006

### How I did it

Now, you can call addresses directly (aka invoke fallback methods and stuff like that) via:

```python
contract(sender=me, data=b"yo")
```

### How to verify it

call a fallback method in a contract like i did above

### Checklist


src/ape/api/address.py
```diff
@@ -79,6 +81,30 @@ def __str__(self) -> str:
...
+        Call this address directly.
+
+        Args:
+            **kwargs: Transaction arguments, such as ``sender`` or ``data``.
+
+        Returns:
+            :class:`~ape.api.transactions.ReceiptAPI`
+        """
+
+        converted_kwargs = _convert_kwargs(kwargs, self.conversion_manager.convert)
```
antazoey, line 95: @fubuloubu I feel like I am missing something. This seems to work, but it doesn't seem like there is any real use of the FallbackABI or the ReceiveABI when it comes to this, like that all happens on the EVM side.

Is there anything missing in this PR around the 5 different scenarios of FallbackABI / ReceiveABI combinations or any missing test cases I should add?

I think this PR allows the fallback and receive methods to at least be testable now, which is all I needed here for moving forward with Coverage.

docs/userguides/contracts.md
```diff
@@ -154,6 +154,20 @@ assert contract.get_static_list() == [1, 2, 3]
...
 
+### Default, Fallback, and Direct Calls
+
+To directly call an address, such as invoking a contract's `fallback` or `receive` method, call a contract instance directly:
+
+```python
+from ape import Contract, accounts
+
+sender = accounts.load("dev")
+contract = Contract("0x123...")
```
fubuloubu, line 165: Shouldn't this be
```suggestion
contract = Contract(data="0x123...")
```

src/ape/api/address.py
fubuloubu: After the transaction is created, if the object is a contract instance and has a fallback or receive function we can do some additional validation on the scenarios that are supported.

That might be better off in the ethereum plugin to be honest 🤔

docs/userguides/contracts.md
fubuloubu: oh whoops, duh

docs/userguides/contracts.md
```diff
@@ -154,6 +154,20 @@ assert contract.get_static_list() == [1, 2, 3]
...
+
+# Call the contract's fallback method.
+receipt = contract(sender=sender)
```
fubuloubu, line 168: Might be cool to show other fields here, limit `gas_limit` and `data`

src/ape/api/address.py
antazoey: I think we could put it in a validator from one of the transaction classes defined in `ape_ethereum.transactions`.

fubuloubu: Hmm, maybe could subclass `ContractInstance.__call__` to do this check before using this call instead

antazoey: ya i was thinking that too, but that wouldn't solve the evm-ness entirely, but maybe for now that is all we do and if some zk thing comes up we incorporate an `ape.api`

antazoey:
@fubuloubu GM

so i think if there is no fallback and no receive,  we should still let the user try to call the contact.
mainly, this will help coverage with ensuring the bytecode responsible for reverting does get triggered (which increased  your coverage)... but also, for the same reason we allow __call__ on `BaseAddress`, just helps advanced use-cases.

the only cases im still looking at but so far it seems like they should happen at compile time, maybe in ethpm-types validators... potentially can do something with making sure there is either a payable fallback or a receive if transacting with a `value=` (however, we don't do this for any of the normal methods)

fubuloubu:
> @fubuloubu GM
> 
> so i think if there is no fallback and no receive,  we should still let the user try to call the contact.
> mainly, this will help coverage with ensuring the bytecode responsible for reverting does get triggered (which increased  your coverage)... but also, for the same reason we allow __call__ on `BaseAddress`, just helps advanced use-cases.
> 
> the only cases im still looking at but so far it seems like they should happen at compile time, maybe in ethpm-types validators... potentially can do something with making sure there is either a payable fallback or a receive if transacting with a `value=` (however, we don't do this for any of the normal methods)

I agree with this approach, maybe we can add a warning later just to help clue people in, but for now this is a net improvement

fubuloubu (commented):
One update and then ready to go

src/ape/api/address.py
```diff
@@ -79,6 +81,30 @@ def __str__(self) -> str:
...
+        txn = self.provider.network.ecosystem.create_transaction(
```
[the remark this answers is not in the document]
antazoey: OK done, let me know if that is right.

src/ape/contracts/base.py
```diff
@@ -757,6 +757,27 @@ def __init__(
         self.txn_hash = txn_hash
         self._cached_receipt: Optional[ReceiptAPI] = None
 
+    def __call__(self, *args, **kwargs) -> ReceiptAPI:
```
fubuloubu, line 760: Nice, I like this

And can just bypass this check with `contract.as_transaction()`

fubuloubu (approved):

src/ape/contracts/base.py
antazoey: - [x] Show in tests how to bypass checks

^ Just wanted to make sure that actually worked. It does!

Issue #1465, opened 2023-06-04
Can't encode BlockAPI class to json string [APE-1012]


fubuloubu:
### Environment information

- OS: linux
- Python Version: 3.8.16
- `ape` and plugin versions:

```sh
$ ape --version
0.6.10
```

### What went wrong?

I'm trying to convert Ape `BaseInterfaceModel` classes to a json string, but can't

```py
>>> chain.blocks.head.json()
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[3], line 1
----> 1 chain.blocks.head.json()

File python3.8/site-packages/ape/utils/basemodel.py:132, in BaseInterfaceModel.json(self, *args, **kwargs)
    129 if "exclude_none" not in kwargs:
    130     kwargs["exclude_none"] = True
--> 132 return super().json(*args, **kwargs)

File python3.8/site-packages/pydantic/main.py:504, in pydantic.main.BaseModel.json()

File python3.8/json/__init__.py:234, in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    232 if cls is None:
    233     cls = JSONEncoder
--> 234 return cls(
    235     skipkeys=skipkeys, ensure_ascii=ensure_ascii,
    236     check_circular=check_circular, allow_nan=allow_nan, indent=indent,
    237     separators=separators, default=default, sort_keys=sort_keys,
    238     **kw).encode(obj)

File python3.8/json/encoder.py:199, in JSONEncoder.encode(self, o)
    195         return encode_basestring(o)
    196 # This doesn't pass the iterator directly to ''.join() because the
    197 # exceptions aren't as detailed.  The list call should be roughly
    198 # equivalent to the PySequence_Fast that ''.join() would do.
--> 199 chunks = self.iterencode(o, _one_shot=True)
    200 if not isinstance(chunks, (list, tuple)):
    201     chunks = list(chunks)

File python3.8/json/encoder.py:257, in JSONEncoder.iterencode(self, o, _one_shot)
    252 else:
    253     _iterencode = _make_iterencode(
    254         markers, self.default, _encoder, self.indent, floatstr,
    255         self.key_separator, self.item_separator, self.sort_keys,
    256         self.skipkeys, _one_shot)
--> 257 return _iterencode(o, 0)

File python3.8/site-packages/pydantic/json.py:88, in pydantic.json.pydantic_encoder()

[...]

PR #1466, opened 2023-06-05
fix: bytes serialization issues [APE-1012]

antazoey:
### What I did

Issue where Ape models containing Hexbytes would error.

fixes: #1465 
Fixes: APE-1012
Fixes: APE-1016

### How I did it

The main fix was using ethpm_types basemodel instead of a naked pydantic one.
Additionally, switched remaining usages of `hexbytes.Hexbytes` with `ethpm_types.HexBytes`.
Instead of failing validation, blocks now convert hexbytes.

### How to verify it

```python
chain.blocks.head.json()
```

### Checklist


src/ape_ethereum/ecosystem.py
```diff
@@ -167,6 +168,12 @@ class Block(BlockAPI):
     difficulty: int = 0
     total_difficulty: int = Field(0, alias="totalDifficulty")
 
+    # Type re-declares.
```
fubuloubu, line 171: Is this just to re-arrange the order it serializes as?

fubuloubu (dismissed):

src/ape_ethereum/ecosystem.py
antazoey: no, the `BlockAPI` uses `Any` as the type, but in `ape-ethereum`, we can safely assume it is always going to be HexBytes`.

antazoey: tbh, i dont remember why they are Any in `BlockAPI`; maybe we can be rid of that

fubuloubu: I think bytes/HexBytes is probably an appropiate assumption for those fields abstractly

antazoey: im good with making the change now, shouldnt really break anything except maybe the already broken starknet

fubuloubu (approved):

Issue #1433, opened 2023-05-07
Not calling `.disconnect` within `NetworkManager.fork` creates persistent fork [APE-938]


fubuloubu:
### What went wrong?

A bit ago, we decided that `ProviderContextManager` didn't need to disconnect to save time in setting and reseting subprocess providers:
https://github.com/ApeWorX/ape/blob/main/src/ape/api/networks.py#L572

However, because of this reasoning, when we do something like this:

```py
with networks.fork():
    ...  # Do some transactions in the fork network

...  # Do some stuff on live network

with networks.fork():
    ...  # Inherits context from previous fork
```

We are not reseting the fork between calls to `networks.fork()` (which just uses the `ProviderContextManager` to connect)

### How can it be fixed?

Call disconnect at the end

fubuloubu:
Was thinking more about this today, I think when you pop the latest provider off the stack, it needs to disconnect itself here:
https://github.com/ApeWorX/ape/blob/e6e2aed5cfe44ed4480a496a26f6313e75c0c4b1/src/ape/api/networks.py#L591

PR #1470, opened 2023-06-08
fix(networks): avoid persistent fork [APE-1030]

fubuloubu:
### What I did

When popping a provider off the stack, it should disconnect in order to reset the connection in scenarios where the provider is an ephemeral fork or locally provisioned network

fixes: #1433

### How I did it

Disconnect after popping a provider off the stack

### How to verify it

Noticing now that #933 made a change exactly counter to this, I am thinking that there is a relevant change necessary in `ape_test` that will make these both function properly together. Maybe the behavior should be configurable?

### Checklist

- [ ] Ensure that tests still function appropriately (test providers don't disconnect until the end of a test per #933)
- [ ] Ensure that console usage functions appropriately (repeated calls to `network.fork()` use fresh fork chain)

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -587,8 +587,10 @@ class ProviderContextManager(ManagerAccessMixin):
         if self.empty:
             return
 
-        # Clear last provider
+        # Clear last provider and disconnect
         exiting_provider_id = self.provider_stack.pop()
+        self.connected_providers.pop(exiting_provider_id).disconnect()
+        # If stack is empty, reset to initial state
         if not self.provider_stack:
             self.network_manager.active_provider = None
             return


sabotagebeats (approved):

antazoey (changes requested):
@fubuloubu Disconnecting the provider will cause multi-network testing to not really work.
How else you can change the provider in tests if it disconnects?

Let's say i am running forked nodes across a bunch of live networks in my test suite... if i switch a provider from one to the next, it will disconnect the entire process.

So maybe a config option or something, but the caveat that you have to manage connections yourself or only use 1 provider.

fubuloubu:
> @fubuloubu Disconnecting the provider will cause multi-network testing to not really work. How else you can change the provider in tests if it disconnects?
> 
> Let's say i am running forked nodes across a bunch of live networks in my test suite... if i switch a provider from one to the next, it will disconnect the entire process.
> 
> So maybe a config option or something, but the caveat that you have to manage connections yourself or only use 1 provider.

This is sort of a middle road of what we had before (disconnect and connect so only one connection is maintained at a time) and what we have now (leave everything connected until ape shuts down), the idea being that a connection stays connected as long as it's in the provider stack, and when it's popped off the stack have it disconnect only at that point

I do think some other broader updates are need to support this, but in my mind during testing the way it would work is:
1. connect to default/selected provider at startup
2. run each test (with snapshots)
3. if a provider connection marker is discovered, have it push and then pop when the test goes out of scope (actually kind of similar to how snapshots work)
4. disconnect from selected provider after tests are finished (have to stay connected during report collection and if user wants to debug)

antazoey:
If we disconnect, there is no point to even having a network stack.

```python
def test_multichain(provider, networks):
    # Do initial bridge set up
    <stuff>
    
    # Connect to other network and do stuff
   with networks.other.local.use_provider("test"):
       <stuff>

   # !!!!!! Now, the rest of ALLL tests will fail because there is no more connected provider.

```

antazoey:
How can I deploy a contract in a sessional fixture if popping disconnects?

antazoey:
Show me a project that uses multichain testing in fixtures, individual tests per each network, and tests that switch networks mid test... if all three of the use-cases are fulfilled, we can move forward.

fubuloubu:
> How can I deploy a contract in a sessional fixture if popping disconnects?

The disconnect should happen after the session ends, agree with you there

If you temporarily connect in the middle of a test, it should disconnect at the end of the context manager's scope

I think overall there is some work needed to fix it to work properly under the variety of use cases that we support since the current way test fixtures work relies on this behavior that it never disconnects during the test session, which is hurting it's ability to be used for scripting (e.g. temporary fork to check the outcome of a transaction before commiting to broadcast it to public network)

antazoey:
@fubuloubu Awesome. OK, we can definitely make both use-cases happy, just will need some craftiness

fubuloubu:
> @fubuloubu Awesome. OK, we can definitely make both use-cases happy, just will need some craftiness

Exactly!

antazoey:
so that out multi-chain testing features continue to work, i suggest a solution like this: https://github.com/ApeWorX/ape/pull/1664

which treats the `fork()` method a little differently and actually disconnects in these scenarios, as that seems more fitting for that use-case.

fubuloubu:
Closing in favor of #1664

Issue #1435, opened 2023-05-07
Converter converts potentially integer value strings as hex [APE-940]


fubuloubu:
### What went wrong?

If I put in an integer value as a string, I will get hex as output from the converter

```py
In [19]: chain.conversion_manager.convert(f"100", int)
Out[19]: 256

In [20]: chain.conversion_manager.convert(f"1000", int)
Out[20]: 4096

In [21]: chain.conversion_manager.convert(f"10000", int)
Out[21]: 65536

In [22]: chain.conversion_manager.convert(f"20000", int)
Out[22]: 131072

In [23]: chain.conversion_manager.convert(f"F0000", int)
Out[23]: 983040

In [24]: chain.conversion_manager.convert(f"G0000", int)
ERROR: (ConversionError) No conversion registered to handle 'G0000'.
```

### How can it be fixed?

To convert to a hex value should require `0x` prefix, without a prefix it should just convert it to an integer directly

PR #1481, opened 2023-06-13
fix: hex string converter [APE-1082]

dtdang:
### What I did
Hex to integer conversion now requires a prefix of "0x" for string values otherwise the value is not converted.

fixes: #1435 

### Checklist

### DIFF
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -48,13 +48,18 @@ class HexConverter(ConverterAPI):
 class HexIntConverter(ConverterAPI):
     """
     Convert hex values to integers.
+
+    **NOTE** If value is a ``str``, it must begin with "0x".
     """
 
     def is_convertible(self, value: Any) -> bool:
         return (isinstance(value, str) and is_hex(value)) or isinstance(value, bytes)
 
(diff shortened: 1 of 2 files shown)


src/ape/managers/converters.py
```diff
@@ -48,13 +48,18 @@ def convert(self, value: str) -> bytes:
...
+
+    **NOTE** If value is a ``str``, it must begin with "0x".
     """
 
     def is_convertible(self, value: Any) -> bool:
         return (isinstance(value, str) and is_hex(value)) or isinstance(value, bytes)
 
     def convert(self, value: Any) -> int:
-        return to_int(HexBytes(value))
+        if isinstance(value, bytes) or (value.startswith("0x") and isinstance(value, str)):
```
antazoey, line 59: start with checking if a str before checking if it starts with 0x or else you can get a type error

```suggestion
        if isinstance(value, bytes) or (isinstance(value, str) and value.startswith("0x")):
```

antazoey, line 59: eth_utils also has a `is_0x_prefixed()` utility method you could use

antazoey (approved):

Issue #360, opened 2021-12-21
Testing: Set-Storage RPC-Method


NotPeopling2day:
### Overview

Support development RPC method for `set-storage`

### Specification

https://www.shawntabrizi.com/substrate/querying-substrate-storage-via-rpc/

### Dependencies

#358

PR #1484, opened 2023-06-14
feat: add ability to set raw storage slots [APE-1088]

fubuloubu:
### What I did

Adds a mock API (to be implemented by testing providers) for setting raw storage slots

fixes: #360

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -509,6 +509,19 @@ class ProviderAPI(BaseInterfaceModel):
         """
 
     @raises_not_implemented
+    def set_storage(  # type: ignore[empty-body]
+        self, address: AddressType, slot: int, value: HexBytes
+    ):
+        """
+        Sets the raw value of a storage slot of a contract.
+
+        Args:
+            address (str): The address of the contract.
+            slot (int): Storage slot to write the value to.
+            value: (bytes): The value to overwrite the raw storage slot with.
+        """
+
+    @raises_not_implemented
     def unlock_account(self, address: AddressType) -> bool:  # type: ignore[empty-body]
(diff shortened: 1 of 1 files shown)


src/ape/api/providers.py
```diff
@@ -508,6 +508,19 @@ def set_code(  # type: ignore[empty-body]
             code (:class:`~ape.types.ContractCode`): The new bytecode.
         """
 
+    @raises_not_implemented
+    def set_storage(  # type: ignore[empty-body]
+        self, address: AddressType, slot: int, value: HexBytes
```
antazoey, line 513: I thought I had remembered a virtual machine that didn't use storage slots, maybe it was Cairo?
Just wondering if this method would have a separate meaning then, if we should rename `slot`, stuff like that.

antazoey (approved):
looks good but wondering about systems without using storage slots

src/ape/api/providers.py
fubuloubu: Could just call it `key` I suppose, most VMs come with some sort of persistent key-value store for a stateful blockchain

fubuloubu:
> questions: does eth-tester or geth have implementations of this?

eth-tester doesn't, but you can probably directly set the storage database through the py-evm backend:
https://github.com/ethereum/py-evm/blob/eaa35b2138a65f5822780e514834e5222f7a42cf/eth/vm/logic/storage.py#L66

geth I don't think supports this

antazoey:
> eth-tester doesn't, but you can probably directly set the storage database through the py-evm backend: 

maybe we could convince the Snake Charmers to implement this in `EthTester`

Issue #1441, opened 2023-05-21
value is not a valid integer (type=type_error.integer) [APE-960]


Aviksaikat:
### Environment information

- OS: Linux 
- Python Version: 3.10.9
- `ape` and plugin versions:

```sh
$ ape --version
0.6.9

$ ape plugins list
Installed Plugins:
  solidity    0.6.4
  hardhat     0.6.5
  alchemy     0.6.1
  ganache     0.6.6

```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
name: fallback
plugins:
  - name: solidity
    version: 0.6.3 # version of the plugin
  - name: ganache
  - name: alchemy
ethereum:
  default_network: local
  sepolia_fork:
    default_provider: ganache
    gas_limit: max
    default_transaction_type: 0
ganache:
  server:
    port: auto
  fork:
    ethereum:
      sepolia:
        upstream_provider: alchemy
test:
  mnemonic: test test test test test test test test test test test junk
  number_of_accounts: 10
```

### What went wrong?

- I was trying to invoke a fallback function from this contract
```js
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Delegate {
    
    address public owner;
    
    constructor(address _owner) {
        owner = _owner;
    }
    
    function pwn() public {
        owner = msg.sender;
    }
}

contract Delegation {
    
    address public owner;
    Delegate delegate;
    
    constructor(address _delegateAddress) {
        delegate = Delegate(_delegateAddress);
        owner = msg.sender;
    }
    
    fallback() external {
        (bool result,) = address(delegate).delegatecall(msg.data);
        if (result) {
            this;
        }
    }
}
```

- Tried to create a transaction like this 
```py
data = web3.keccak(text="pwn()")
txn = networks.ecosystem.create_transaction(receiver=delegation, value="1 wei", data=data)
attacker.call(txn)
```

### Error
```py
pydantic.error_wrappers.ValidationError: 2 validation errors for DynamicFeeTransaction
to
  str type expected (type=type_error.str)
value
  value is not a valid integer (type=type_error.integer)
```

- I tried to set type like this but in vain
```py
txn = networks.ecosystem.create_transaction(receiver=delegation, value="1 wei", data=data, type=convert(2, int))
```

### How can it be fixed?

[...]

PR #1487, opened 2023-06-14
fix: convert value for create_transaction [APE-960]

dtdang:
### What I did
Add check for "value" kwarg and convert it if it is not an `int` value in `create_transaction`.

fixes: #1441 

### How to verify it
It can be verified by running this in `ape console`:
`txn = networks.ecosystem.create_transaction(receiver="0x", value="1 wei", data=b"", type=convert(2, int))` 

### Checklist


antazoey (dismissed):
seems good enough! no test needed since is not really an official use-case (but at least now it would work)

real fix was the doc update, thanks.

but lint checks are failing, gotta fix those before we can merge.

antazoey (approved):

Issue #614, opened 2022-03-29
Track if plugins that did not load correctly and don't include them in the `available plugins`


antazoey:
### Environment information

Latest ape
Any OS

### What went wrong?

When a plugin fails to load (e.g. from a breaking change), it shows as an available plugin you can install

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

fubuloubu:
can probably get the set of installed plugins from looking into the site packages

Ninjagod1251:
This is not an issue an anymore

Ninjagod1251:
I didnt mean to close this. How can I break a plugin to fail it? How can we test the plugin to load properly if we do not install it?

fubuloubu:
> I didnt mean to close this. How can I break a plugin to fail it? How can we test the plugin to load properly if we do not install it?

I think one of the easiest ways is to break an `EcosystemAPI` class, since that loads on startup

unparalleled-js:
There are two ways to try:

1. Change the signature of the one of the API methods. This will cause an error saying the plugin is not fully implemented. I am curious if in this case, does it show up in the plugins list output as an available plugin?
2. Similarly, if the plugin has an error during startup. I think to do this, you can add random text to the top of the one source files. This will cause a python syntax error.

antazoey:
the hard part is it would be nice to know that `ape-vyper` is actually a compiler plugin, i wish that data was accessible even when it fails to load correctly

antazoey:
This is no longer happening! Ever since we changed how plugin lookup, using a better source instead of plugin registration, we never show a failed-plugin as an available plugin.

PR #1498, opened 2023-06-22
fix: better warnings for missing compilers

antazoey:
### What I did

The warning for missing compilers needed an iteration... it told me compilers for markdown files were missing, which was weird. Also, I would have liked an additional hint telling me possibly one of the compiler plugins did not load correctly.

### How I did it

* Filter out common extensions like `.md` and `.adoc` from the warning.
* Handle plurality and singularity independently
* Attach suffix for indicating possible failed to load plugins
* Update one of the existing tests to include a check for all of this

### How to verify it

No longer see unnecessary warning about missing compilers for `.md` files and `.adoc` files and stuff
The warning you do you get when something like `ape-vyper` failing to load is a little better. Still if we could get a handle on https://github.com/ApeWorX/ape/issues/614 it could be even better yet.

### Checklist


src/ape_compile/_cli.py
```diff
@@ -42,13 +42,32 @@ def cli(cli_ctx, file_paths: Set[Path], use_cache: bool, display_size: bool):
         return
 
     ext_given = [p.suffix for p in file_paths if p]
-    ext_with_missing_compilers = cli_ctx.project_manager.extensions_with_missing_compilers(
-        ext_given
-    )
+
+    # Filter out common files that we know are not files you can compile anyway,
+    # like documentation files. NOTE: Nothing prevents a ComopilerAPI from using these
```
dtdang, line 47:
```suggestion
    # like documentation files. NOTE: Nothing prevents a CompilerAPI from using these
```

dtdang (approved):

fubuloubu (approved):

Issue #1442, opened 2023-05-22
Permit2 contract causes infinite loop fetching proxy information [APE-962]


fubuloubu:
### Environment information

- OS: linux
- Python Version: 3.10.9
- `ape` and plugin versions:

```sh
$ ape --version
0.6.8

$ ape plugins list
Installed Plugins:
  vyper        0.6.3
  alchemy      0.6.1
  etherscan    0.6.3
  template     0.6.0
  solidity     0.6.3
  foundry      0.6.7
  huff         0.1.2
```

### What went wrong?

The Permit2 implementation contract (`0x000000000022D473030F116dDEE9F6B43aC78BA3` on almost all networks) causes an infinite loop inside of our [proxy info routine](https://github.com/ApeWorX/ape/blob/main/src/ape_ethereum/ecosystem.py#L204)

### How can it be fixed?

I'm guessing that the bytecode has a matching sequence to one of the regex comparisons we are doing, so it should be required to do some additional validation on the length of the bytecode to avoid false positives

banteg:
i've identified where the issue comes from:
1. a call to `masterCopy()` reverts
2. then ape tries to `get_virtual_machine_error ` of a not yet initialized contract.
3. which in turn calls `enrich_error`, which [tries to get the abi](https://github.com/ApeWorX/ape/blob/84b276eff63734f32640f52f87f3b778f6c81eb3/src/ape/managers/compilers.py#L243), so it falls into the inifinite loop.

the solution could be to either propagate `skip_trace=True` up to `enrich_error` or better not try getting partially-initialized abis from the error handler.

PR #1500, opened 2023-06-23
fix: use master copy for gnosis safe proxy [APE-1124]

banteg:
### What I did

reverted #1178 because 1.1.0 was more or less the first public release of gnosis contracts

### How I did it

check for `masterCopy()` and additionally check it matches the address stored in slot 0. keep the check that it's initialized to a non-zero address.

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -267,19 +267,19 @@ class Ethereum(EcosystemAPI):
 
             return ProxyInfo(type=type, target=target)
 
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
Needs QA just to make sure that there isn't any issues with fetching contracts. Also might solve an issue I've seen with getting certain contracts that have specific characteristics to it such as Permit2 (I made an issue somewhere)

banteg:
identified the root cause of #1442 as not related to this code change

banteg:
good to merge since the recursion bug is fixed now

Issue #1221, opened 2023-01-16
Need ability to specify compiler settings to use with dependencies


fubuloubu:
### Overview

https://github.com/safe-global/safe-contracts uses a pretty wide pin for their contracts of anything in solc 0.7.x or 0.8.x series. However, the contracts only compile with 0.7.6 specifically. When specifying a dependency like this, it would be nice to be able to add additional config options to force it to compile with those specific options.

### Specification

```yaml
dependencies:
  - name: safe-contracts
    github: safe-global/safe-contracts
    version: 1.3.0
    settings: # or `config` maybe?
      # Configuring `ape-solidity` so this project compiles
      solidity:
        version: 0.7.6
```

These options would override whatever default config that Ape uses when compiling a dependency.

### Dependencies

n/a

PR #1504, opened 2023-06-27
feat: `ape pm` CLI commands [APE-1133]

antazoey:
### What I did

ugh so this is really needed for me to develop other features with. We are having issues with settings + dependencies and needing to work on that but needing this to develop that. 
i mostly needed the compile cmd but added the others as well.

feat: CLI for listing both project and global packages installed
feat: CLI for installing packages, both from config or project or adhoc given dependencies
feat: CLI for compiling dependencies (and forcing recompile)

### How I did it

create 3 click commands for handling all this stuff

### How to verify it

```
ape pm list
ape pm list --all
ape pm install   # Install all packages listed in local project's config
ape pm install --force   # Force reinstall your project's dependencies
ape pm install gh:OpenZeppelin/openzeppelin-contracts --name OpenZeppelin --version "4.6.0" --force
```

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #1505, opened 2023-06-27
feat: extra settings for dependencies [APE-1134]

antazoey:
### What I did

allows compiler settings config for dependencies

fixes: #1221 

**note**: this goes hand in hand with #1504 so you can easily recompile dependencies with new settings if you need

### How I did it

an optional property on DependencyAPI
use these when configuring compilers during compiling a dependency

### How to verify it

using a PoA testnet? Try this:

```
dependencies:
  - name: OpenZeppelin
    version: 4.6.0
    compiler_settings:
      solidity:
        evm_version: paris
```

### Checklist


src/ape/api/projects.py
```diff
@@ -240,6 +240,11 @@ class DependencyAPI(BaseInterfaceModel):
     A list of glob-patterns for excluding files in dependency projects.
     """
 
+    compiler_settings: Dict = {}
```
fubuloubu, line 243: Was thinking that this could be `config` or `config_override` to show that it could allow config for other plugins, not necessarily just compilers (although they are the most relevant plugins to configure ofc)

fubuloubu (commented):
Looks good, just want to talk this out though

src/ape/api/projects.py
antazoey: ya, it would work with any plugin so probably should use a better name

src/ape/api/projects.py
```diff
@@ -240,6 +240,11 @@ class DependencyAPI(BaseInterfaceModel):
     A list of glob-patterns for excluding files in dependency projects.
     """
 
+    config_override: Dict = {}
+    """
+    A mapping of extension to additional compiler settings to use when compiling.
```
antazoey, line 245:
```suggestion
    Extra settings to include in the dependencies configuration.
```

src/ape/api/projects.py
```diff
@@ -240,6 +240,11 @@ class DependencyAPI(BaseInterfaceModel):
     A list of glob-patterns for excluding files in dependency projects.
     """
 
+    config_override: Dict = {}
+    """
+    Extra settings to include in the dependencies configuration.
```
antazoey, line 245:
```suggestion
    Extra settings to include in the dependency's configuration.
```

fubuloubu (approved):

Issue #1326, opened 2023-03-01
Add `--include-dependencies` to `ape compile` [APE-682]


fubuloubu:
### Overview

`ape compile` will not re-compile dependencies after they're downloaded, so it would be very helpful to have a command that allows you to do this outside of the console

### Specification

```sh
$ ape compile --include-dependencies  # go and see if there are updates to any dependencies

$ ape compile -f --include-dependencies  # force compile all dependencies
```

### Dependencies
n/a

PR #1506, opened 2023-06-27
feat: `--include-dependencies` flag for `ape compile` command and setting [APE-682]

antazoey:
### What I did

fixes: #1326 
Fixes: APE-682

* Adds a new CLI flag for `--include-dependencies` so you compile dependencies during `ape compile` if you choose
* Allow config:

```yaml
compile:
  include_dependencies: true
```

to always do this

* Allow skipping cache for dependency compiling via `dep.compile(use_cache=False)`
* Pass along `use_cache` value so `--force` works with `--include-dependencies` to still compile dependencies even when their contract types are already filled out. This is great for compiling with new settings!

### How I did it

* new config
* new click option
* callback in click option looks at config when it is not set
* during compile command, if the value is true, loop through the project's dependencies and call compile and pass along the use_cache value
* handle use_cache in dep.compile

### How to verify it

you can compile dependencies using `ape compile`

btw this works great with some other PRs i just opened

### Checklist

### DIFF
--- a/src/ape_compile/__init__.py
+++ b/src/ape_compile/__init__.py
@@ -1,24 +1,8 @@
 from ape import plugins
-from ape.api import ConfigEnum, PluginConfig
-
-
-class EvmVersion(ConfigEnum):
-    constantinople = "constantinople"
-    istanbul = "istanbul"
-    paris = "paris"
-    shanghai = "shanghai"
+from ape.api import PluginConfig
 
 
 class Config(PluginConfig):
-    evm_version: EvmVersion = EvmVersion.shanghai
-    """
-    A base EVM version for compiler plugins to look to
-    when their own is missing. **NOTE**: Typically, compiler
-    plugins define their own, and not every compiler is EVM compatible.
-    This is only used as a global setting for the compilers that
-    are EVM compatible.
-    """
-
     include_dependencies: bool = False
     """
     Set to ``True`` to compile dependencies during ``ape compile``.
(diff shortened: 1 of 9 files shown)


src/ape_compile/__init__.py
```diff
@@ -5,10 +5,30 @@
 class EvmVersion(ConfigEnum):
     constantinople = "constantinople"
     istanbul = "istanbul"
+    paris = "paris"
+    shanghai = "shanghai"
 
 
 class Config(PluginConfig):
-    evm_version: EvmVersion = EvmVersion.istanbul
+    evm_version: EvmVersion = EvmVersion.shanghai
```
antazoey, line 13: i never knew this existed and im pretty sure this is not being honored or used anywhere. all compiled that are evm have their own settings here

fubuloubu: Was envisioning that this would work as an more global setting to compiler plugins

fubuloubu: honestly, this is also kind of weird with the plugin system because non-EVM compilers would never use this, so I agree with removing it now as the right outcome

antazoey: k it gone

src/ape_compile/__init__.py
```diff
@@ -1,17 +1,18 @@
...
-
-
-class EvmVersion(ConfigEnum):
-    constantinople = "constantinople"
-    istanbul = "istanbul"
+from ape.api import PluginConfig
 
 
 class Config(PluginConfig):
-    evm_version: EvmVersion = EvmVersion.istanbul
```
fubuloubu, line 11: Would this break someone's config if trhey had it defined for whatever reason?

fubuloubu (commented):
Approved otherwise

src/ape_compile/__init__.py
antazoey: i made it not break and give a warning and we can fully remove in 0.7

fubuloubu (approved):

Issue #760, opened 2022-05-29
Add docs and tests for proxy support feature


fubuloubu:
Enumerating all the proxies in a doc would be nice! There are so many. Even if it happens in the doc str of the method, that is fine. Else, we may need `contracts.md` userguide (in `docs/userguides`). If you go that route, you don't have to thoroughly create it, just initialize it, add the enumeration, and we will fill it in later with non-relevant details to this PR.

_Originally posted by @unparalleled-js in https://github.com/ApeWorX/ape/pull/750#discussion_r881616936_

antazoey:
ok thoughts on not adding tests for every single proxy and just adding them as needed? I tested the first 3 but am exhausted already, even though it helped me find all sorts of other bugs

fubuloubu:
We got a good start more tests = better, but don't have to do them all right away

Let's leave this issue open until it's complete though

antazoey:
I updated the title and description to reflect that it is only to finish testing the proxies now.

itsnevu:
Hey, I'd like to help finish off the proxy test coverage here. I'll pick a few of the remaining untested proxies from the list and add tests for those, then follow up with more if that's welcome.

fubuloubu:
> Hey, I'd like to help finish off the proxy test coverage here. I'll pick a few of the remaining untested proxies from the list and add tests for those, then follow up with more if that's welcome.

Sure, that would be welcome!

itsnevu:
Thanks @fubuloubu. Picking this up with a concrete scope.

Worth flagging for anyone reading this later: #2467 was closed unmerged by the stale bot, so none of those nine proxy tests are in main. @antazoey had already trimmed it to the passing subset and approved it, so that work is salvageable rather than lost.

Current state in main, from tests/functional/geth/test_proxy.py and tests/functional/test_proxy.py:

Tested: Standard, Beacon, UUPS, GnosisSafe, OpenZeppelin, Delegate, Minimal, plus AragonAppUpgradeable which isn't on the list in the description.

Untested: ZeroAge, Clones, Vyper, VyperBeta, CWIA, OldCWIA, SoladyPush0, SoladyCWIA, SudoswapCWIA, SplitsCWIA, Sequence.

[...]

PR #1507, opened 2023-06-28
test: test and document proxy system [APE-197]

antazoey:
### What I did

partially: #760

### Checklist


src/ape_test/provider.py
```diff
@@ -254,3 +254,10 @@ def get_virtual_machine_error(self, exception: Exception, **kwargs) -> VirtualMa
 
         else:
             return VirtualMachineError(base_err=exception, **kwargs)
+
+    @raises_not_implemented
```
fubuloubu, line 258: Won't this still raise? I don't get the comment in the body

antazoey: it is implemented in the base class.
However, we have another handler somewhere that handles the equivalent error from `web3` and raises `APINotImplemented` at that point, as if it had this decorator.
So by adding this decorator, it is small perf by bypassing any call to web3 and just raising immediately, because we know it is not going to work for EthTester, unfortunately.

On the note, I have looked into what it would take a little bit to implement this kind of stuff in EthTester.. I think it needs some extra handling for irregular chain activity like this. At least that is what it seems like Hardhat does.

antazoey: If we were using hardhat or something, I could call `set_code()` instead of deploying a contract for some of these proxies

antazoey: `Web3Provider` implements `get_storage_at()` but it fails in `EthTester`, which maybe is surprising idk

banteg:
when ape supports rpc batching, it would make sense to clamp all code, storage, and call checks into one rpc call

src/ape_test/provider.py
antazoey: good news: https://github.com/ethereum/eth-tester/pull/264
so im going to delete this change so it just works when it comes, and it has no effect now anyways besides short-circuit

fubuloubu (approved):

PR #1510, opened 2023-06-29
fix: issues preventing deploy transaction errors from being very useful [APE-1141]

antazoey:
### What I did

* Fixes an issue where contract logic errors were not enriched during failed deploy txns
* Fixes an issue where contract logic errors did not show traceback during failed deploy txns
* Fixes an issue where could not make assertions in `ape test` for failed deploy txns that were custom
* Fixes an issue where didnt have access to the contract type for failed deploys
* Feature: can now use `ape.reverts()` in scripts.
* Feature: `ape.reverts()` will now return an object that eventually gets its error value set (much like pytest.raises()!!)
* refactor: use walrus operator more often and save some lines of code

NOTE: Requires this PR from `ape-solidity`: https://github.com/ApeWorX/ape-solidity/pull/112
nothing needed in vyper.

### Checklist


src/ape/__init__.py
```diff
@@ -42,6 +43,12 @@
 convert = _ManagerAccessMixin.conversion_manager.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
 
+reverts = RevertsContextManager
```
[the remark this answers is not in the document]
antazoey: ya that is another benefits, i remember someone complaining about that

docs/userguides/testing.md
```diff
@@ -396,6 +396,24 @@ def test_unauthorized_withdraw(contract, hacker):
...
+import ape
+
+def test_error_on_deploy(account):
+    with ape.reverts() as err:
+        ape.project.HasError.deploy(sender=account)
+    
+    # Grab the cached instance using the error's address
+    # and assert the custom error this way.
+    contract = ape.Contract(err.value.address)
+    assert isinstance(err.value, contract.MyError)
```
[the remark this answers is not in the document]
antazoey: Would that make the tests less good ? I remember when we added this feature, we thought it was important to keep `contract_a.MyError` different than `contract_b.MyError` when it comes to asserting. We would lose that if we made it work for ABIs alone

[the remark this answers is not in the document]
antazoey: unless we somehow we made it work only for deploy failures but im not sure that is very easy to do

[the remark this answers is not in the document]
antazoey: like you could do:

```python
import ape


def test_whatever(contract_a, contract_b, account):
    with ape.reverts(contact_a.MyError):
        contract_b.my_method(sender=account)
```

and the test would pass even tho it raises `contract_b.MyError` and not `contract_a.MyError`, so we agreed we need an address here.

Some how the `geth` plugin gives us a receipt before it raises the contract logic error, we are able to cache the type with the address and make this magic happen, but actually not sure that is always the case, especially if it fails when estimating gas.... so perhaps we do need contract container / type / abi support for the handler, but we lose something then... ugh!!!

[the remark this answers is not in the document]
fubuloubu: I was thinking that `contract_a.MyError` would fail if raised from `contract_b`, but `MyContract.MyError` would work if raised from either contract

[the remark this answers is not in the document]
antazoey: OK `MyContract.MyError` works now, it is just an ABI because that is all we need.

docs/userguides/testing.md
```diff
@@ -396,6 +396,47 @@ def test_unauthorized_withdraw(contract, hacker):
         contract.withdraw(sender=hacker)
 ```
 
+You can also use the ABI of the error if you don't have access to an instance:
+
+```python
+import ape
+
+def test_unauthorized(contract, hacker, project):
+    with ape.reverts(project.MyOtherContract.Unauthorized2, addr=hacker.address):
```
fubuloubu, line 405: Just use full `address` for the kwarg

Edit: or wait, is this the event arg?

antazoey: ya that is a custom kwarg, i can use a different example though, and this was kinda copied from the example above so nothing fresh here, can be revised

antazoey: you dont actually assert the address of the contract, that happens via the instance of the contract it is coming off of (see one of the comments i added in that last commit regarding this).

antazoey: changed to `unauth_address`

docs/userguides/testing.md
```diff
@@ -392,10 +392,51 @@ def contract(owner, project):
     return owner.deploy(project.MyContract)
 
 def test_unauthorized_withdraw(contract, hacker):
-    with ape.reverts(contract.Unauthorized, addr=hacker.address):
+    with ape.reverts(contract.Unauthorized, unauth_address=hacker.address):
         contract.withdraw(sender=hacker)
 ```
 
+You can also use the ABI of the error if you don't have access to an instance:
```
fubuloubu, line 399: Is this sentence correct?

fubuloubu (approved):

docs/userguides/testing.md
antazoey: Yes? I think so? is this a trick question?

antazoey: Using the ABI is the `projet.MyContract.MyError` is the way to do if you dont have access to an instance yet, such as during deploys or if you just dont care or if you dont have access to it

antazoey: I have an idea.... standby!

fubuloubu: I meant the way it was phrased seemed weird and didn't make sense to me

antazoey: OK made the statement not true and changed the statement and made it so custom error types are accessible from the container, and not just the ABI alone, this may help with future mocking features we want to consider, similar to what we did for `MockContractLog` perhaps

antazoey: oh well i just did a large-ish refactor because of it haha because i thought you were saying something else. I think maybe was ok though.

antazoey: The sentence is now

```
You can also use custom error types from the contract container (from `ape.project` or the `project` fixture):
```

let me know if it is still unclear!

dtdang (approved):

PR #1512, opened 2023-06-29
fix: manifest would be corrupt when using raw contract types and JSON compiler [APE-1144]

antazoey:
### What I did

I needed to "compile" a build artifact using the `ape-pm` compiler that was previously built with the solidity compiler, like we do in your tests (don't ask why). But it was always corrupting the manifest. I figured out why.

### How I did it

I realized that there is no way the source ID cannot be the name of the file for this to work... else, you will always get a corrupt manifest because the contract types won't match the sources.

### How to verify it

you can compile contract type JSONs

### Checklist


antazoey:
tests dont fail locally for me so im going to wait some others merge and retry this one

fubuloubu (approved):

antazoey:
was previously approved many times and all i did was fix the failing tests

PR #1520, opened 2023-07-04
fix: better default max fee [APE-1160]

banteg:
### What I did

changed the automatic fee to `base_fee * 2 + priority_fee`. this fixes the very tight defaults that have often resulted in txs failing to be submitted when base fee was rising. this also better utilizes eip-1559 since the defaults would be enough even if the base fee is surging for 6 blocks (1.125⁶ ≈ 2)

### Checklist


antazoey (approved):

src/ape/api/providers.py
```diff
@@ -1278,7 +1278,7 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
                 txn.max_priority_fee = self.priority_fee
 
             if txn.max_fee is None:
-                txn.max_fee = self.base_fee + txn.max_priority_fee
+                txn.max_fee = self.base_fee * 2 + txn.max_priority_fee
```
antazoey, line 1281: some providers are using the eth fee history API now for base fee, and some aren't.
Does it matter here? I just want to make sure you know that.

Like sometimes that API doesnt exist, like in eth tester, so it doesnt use that. But some providers do use it.

PR #1522, opened 2023-07-05
docs: fix version selector [APE-1163]

antazoey:
### What I did

noticed the version selector was not really working and it is just broken in general.
this is because it builds the docs before committing to the gh-pages, so it doesnt know all the versions to build, each version had a different list in its list of versions - got weird...

### How I did it

use the web as the source of truth, so you get em all
one issue tho is locally you wont have access to those versions in your _build unless you have built them before

### How to verify it

```
cd docs
```

```python
from conf import get_versions
get_versions()
```

```sh
cd ..
python build_docs.py
python -m http.server --directory "docs/_build/" --bind 127.0.0.1 1337
```

### Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,8 +12,11 @@
 import os
 import re
 import sys
+from functools import lru_cache
 from pathlib import Path
 from typing import List
+
+import requests
 
 sys.path.insert(0, os.path.abspath(".."))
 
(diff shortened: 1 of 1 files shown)


dtdang (dismissed):

dtdang:
When I'm building the docs locally, the previous versions are not being listed in any order descending or ascending, it seems to be in a random order. I think it might also be missing a few versions.

antazoey:
> When I'm building the docs locally, the previous versions are not being listed in any order descending or ascending, it seems to be in a random order. I think it might also be missing a few versions.

ah good catch, it should be sorted

antazoey:
@dtdang ok it seems good now, it sorts descending so you see the later versions first.
we probably could eventually cap these or something or exclude some versions so the list is smaller

dtdang:
> @dtdang ok it seems good now, it sorts descending so you see the later versions first. we probably could eventually cap these or something or exclude some versions so the list is smaller

Eventually to shorten, maybe we should just show the last version of the previous breaking change release (like for 0.5 and 0.4, etc) since the minor releases for previous versions shouldn't be changing too much in between for the docs.

dtdang (approved):

PR #1523, opened 2023-07-05
feat: pyproject.toml approach for mdformat

antazoey:
### What I did

This is kinda fresh and it may be a little hacky, but I'd like to use this approach to cut down the need of a whole config file that does almost nothing.

The reason we have to use a plugin is because `mdformat` is on the fence about supporting this out-of-the-box because of there are non-python projects that use `mdformat` and they prefer delegating to their plugin system (sound familiar?) 

So someone made this plugin and it seems to work and we only have 1 setting. 

### Checklist

### DIFF
--- a/.mdformat.toml
+++ b/.mdformat.toml
@@ -1,1 +1,0 @@
-number = true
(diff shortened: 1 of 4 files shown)


antazoey:
do we want this or not, i guess, given what we know

fubuloubu (approved):
lgtm

PR #1528, opened 2023-07-06
fix: remove italics from last PR [APE-1168]

antazoey:
### What I did

fixes: #

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ setup(
     },
     include_package_data=True,
     install_requires=[
-        "click>=8.1.3,<9",
+        "click>=8.1.3,<8.1.4",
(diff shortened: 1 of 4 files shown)


banteg:
reason: italics would render as inverse text (black text on white background) in the default `tmux` config. it requires `TERM` to be set to `xterm-256color` to render properly. `screen-256color` doesn't support italics.

antazoey:
@banteg also suggested to include ipt names as well; we are doing that now, but leverage ethpm-types

antazoey:
mypy is failing from latest click upgrade

antazoey:
it is easier to just pin click then try to resolve all the "issues" https://github.com/pallets/click/issues/2558

Issue #1525, opened 2023-07-06
Bug: ape account allows 'alias id' to be value of private key


sambacha:
### What went wrong?

ape accepts arbitrary input for alias when setting up ape account. this can lead to users setting their private key to this value. considering the nature of setting up ape account this can be insecure.

### How can it be fixed?

Alias should be limited ideally by less than 32 bytes (64chars). Using some levenshtein comparison would be (overkill/too much effort).

antazoey:
good catch.
let me know if the linked PR suffices for you!

PR #1530, opened 2023-07-07
fix: more alias validation [APE-1171]

antazoey:
### What I did

fixes: #1525 

### Checklist

### DIFF
--- a/src/ape/cli/arguments.py
+++ b/src/ape/cli/arguments.py
@@ -7,14 +7,19 @@ from ape import accounts
 from ape.api import AccountAPI
 from ape.cli.choices import Alias
 from ape.cli.paramtype import AllFilePaths
-from ape.exceptions import AliasAlreadyInUseError
+from ape.exceptions import AccountsError, AliasAlreadyInUseError
 
 _flatten = chain.from_iterable
 
 
-def _require_non_existing_alias(value):
+def _alias_callback(value):
     if value in accounts.aliases:
+        # Alias cannot be used.
         raise AliasAlreadyInUseError(value)
+
+    elif not isinstance(value, str) or len(value) >= 64:
+        raise AccountsError("Alias must be a string less than 64 characters.")
+
     return value
 
 
@@ -35,9 +40,7 @@ def non_existing_alias_argument():
     A ``click.argument`` for an account alias that does not yet exist in ape.
     """
 
-    return click.argument(
-        "alias", callback=lambda ctx, param, value: _require_non_existing_alias(value)
-    )
+    return click.argument("alias", callback=lambda ctx, param, value: _alias_callback(value))
 
 
 def _create_contracts_paths(ctx, param, value):
--- a/src/ape/cli/arguments.py
+++ b/src/ape/cli/arguments.py
@@ -2,12 +2,14 @@ from itertools import chain
 from typing import Optional, Type
 
 import click
+from eth_utils import is_0x_prefixed
 
 from ape import accounts
 from ape.api import AccountAPI
 from ape.cli.choices import Alias
 from ape.cli.paramtype import AllFilePaths
 from ape.exceptions import AccountsError, AliasAlreadyInUseError
+from ape.utils import ZERO_ADDRESS
 
 _flatten = chain.from_iterable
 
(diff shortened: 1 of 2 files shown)


tests/integration/cli/test_accounts.py
```diff
@@ -67,6 +67,20 @@ def test_import_valid_private_key(ape_cli, runner, temp_account, temp_keyfile_pa
     assert temp_keyfile_path.is_file()
 
 
+@run_once
+def test_import_alias_too_long(ape_cli, runner):
+    # Attempt using private key as the alias.
+    key_alias = f"0x{PRIVATE_KEY}"
```
fubuloubu, line 73: technically 66 characters, but I think it works either way

fubuloubu (approved):

fubuloubu:
An alternative solution is just to disallow hex input

antazoey:
> An alternative solution is just to disallow hex input

maybe let's do both? if your alias is that big, you are probably making a mistake anyway

tests/integration/cli/test_accounts.py
antazoey: i changed it to 42 characters and also made it so hex is not allowed... so your alias has to be shorter than an address rather than a private key... idk, seems ok, if you run into these problems you are probably fudging something up anyway i would think.

fubuloubu:
> > An alternative solution is just to disallow hex input
> 
> maybe let's do both? if your alias is that big, you are probably making a mistake anyway

might be limiting for some unknown use case, randomly coming up with a 3 word alias for some sort of automation use case.

I think just not allowing hex is both stronger (catches both private keys and addresses) and also is a clearer limit (you can use anything but hex)

tests/integration/cli/test_accounts.py
fubuloubu: I was picturing using an s3 bucket for storing chaosnet bot keyfiles, and using a 3 word alias to ensure that it's hard for any random person to guess the keyfile's name. If we needed to increase entropy there we'd run into these limits

fubuloubu:
> > I think just not allowing hex is both stronger (catches both private keys and addresses) and also is a clearer limit (you can use anything but hex)
> 
> But what if they paste their pkey without 0x?

Can still detect as hex, if all the chars are in `[0-9a-fA-F]` and there's an even number of them (and I suppose can do a check that it's 20+ chars to prevent false positives)

antazoey:
> Can still detect as hex, if all the chars are in `[0-9a-fA-F]` and there's an even number of them (and I suppose can do a check that it's 20+ chars to prevent false positives)

OK here is what I did - use the `is_hex()` method to check if and only if the alias is equal or longer than an address.
So you can still use short hex-strings for aliases if you want, i think that is fine, right? or no? Like it just seems odd to not allow short hexes, what if my alias is just the first couple bytes of my address or something

src/ape/cli/arguments.py
```diff
@@ -2,19 +2,25 @@
...
 _flatten = chain.from_iterable
 
 
-def _require_non_existing_alias(value):
+def _alias_callback(ctx, param, value):
     if value in accounts.aliases:
+        # Alias cannot be used.
         raise AliasAlreadyInUseError(value)
+
+    elif not isinstance(value, str) or (is_hex(value) and len(value) >= 42):
```
fubuloubu, line 21:
```suggestion
    elif not isinstance(value, str):
        raise AccountsError(f"Alias must be a str, not '{type(value)}'.")
    elif is_hex(value) and len(value) >= 42:
```

tests/integration/cli/test_accounts.py
```diff
@@ -67,6 +67,36 @@ def test_import_valid_private_key(ape_cli, runner, temp_account, temp_keyfile_pa
...
+
+
+@run_once
+def test_import_alias_is_really_long(ape_cli, runner):
+    """
+    For entropy related use-cases regarding alias, we
+    must ensure long aliases are supported.
+    """
+
+    long_alias = "this is a long alias that i am going to use and you cant stop me"
```
antazoey, line 91: @fubuloubu I added a test to make sure that we will always be able to support your sentence-long alias use-case

fubuloubu (approved):

PR #1531, opened 2023-07-07
chore: mypy upgrade [APE-1172]

antazoey:
### What I did

looks like we are finally able to uprade mypy
**this makes having Ape installed in the same environment as other projects like Silverback work together**
now i dont need separate envs necessarily =

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ extras_require = {
     ],
     "lint": [
         "black>=23.3.0,<24",  # Auto-formatter and linter
-        "mypy>=1.4.1,<2",  # Static type analyzer
+        "mypy>=0.991,<2",  # Static type analyzer
         "types-PyYAML",  # Needed due to mypy typeshed
         "types-requests",  # Needed due to mypy typeshed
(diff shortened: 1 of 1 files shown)


fubuloubu (dismissed):

setup.py
```diff
@@ -19,7 +19,7 @@
     ],
     "lint": [
         "black>=23.3.0,<24",  # Auto-formatter and linter
-        "mypy>=0.991,<1",  # Static type analyzer
+        "mypy>=0.991,<2",  # Static type analyzer
```
fubuloubu, line 22: 0.991 lower bound is for another use case?

fubuloubu (approved):

setup.py
antazoey: py-evm still seems to be pinned to a lower version of typing extensions than the rest of the world.
this fails in ci/cd.
but for some reason, locally i can install mypy 1.0 range... so i have no clue...

so i kept it 0.991 so itll work in ci/cd i guess.

PR #1537, opened 2023-07-12
fix: move functions_incremented outside loop [APE-1184]

dtdang:
### What I did
Moved the `functions_incremented` variable outside of the for loop because it wasn't being accessed properly and giving a error: `UnboundLocalError: cannot access local variable 'functions_incremented' where it is not associated with a value`

fixes: #

### Checklist


dtdang:
Wasn't exactly sure how to write a proper test for this change. Tested with other files and contracts to make sure it was still working as intended.

antazoey (approved):

Issue #1540, opened 2023-07-13
Allow specifying `block_id=` for RPC methods that support it [APE-1188]


fubuloubu:
I was trying to fetch the contract code for an address at a specific block, but that feature isn't available:
https://github.com/ApeWorX/ape/blob/adb1e04dedbbfad7c4cc8dfc0f80928b723adbed/src/ape/api/providers.py#L202

Web3py supports it though: https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_code

antazoey:
Note that we are supporting this parameter in many places, but you may have to look at the `Web3Provider`'s kwargs because they were added to avoid making breaking changes for plugins

PR #1547, opened 2023-07-18
feat: support block id for provider `get_code()` [APE-1203]

antazoey:
### What I did

fixes: #1540 

### How I did it

* add missing support
* add block id alias

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -786,8 +786,9 @@ class Web3Provider(ProviderAPI, ABC):
                 The transaction to estimate the gas for.
             kwargs:
                 * ``block_identifier`` (:class:`~ape.types.BlockID`): The block ID
-                  to use when estimating the transaction. Useful for
-                  checking a past estimation cost of a transaction.
+                  to use when estimating the transaction. Useful for checking a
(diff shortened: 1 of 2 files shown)


src/ape/api/providers.py
```diff
@@ -200,12 +200,13 @@ def get_balance(self, address: AddressType) -> int:
         """
 
     @abstractmethod
-    def get_code(self, address: AddressType) -> ContractCode:
+    def get_code(self, address: AddressType, **kwargs) -> ContractCode:
```
antazoey, line 203: so i made sure, we are allowed to add `**kwargs` to abstract methods without it breaking the plugins that override and dont yet have it

fubuloubu (approved):
Works for now. I was actually hoping to make it an explicit kwarg to use in that method via `block_id` but that can come at a breaking change

antazoey:
> Works for now. I was actually hoping to make it an explicit kwarg to use in that method via `block_id` but that can come at a breaking change

If you want this in 0.7, make sure it is in the 0.7 project (or tag, i forget) on linear and itll for sure get in.

Issue #1062, opened 2022-09-21
Fetching all contract events searches through a lot of useless history


fubuloubu:
### What went wrong?

The default value for `start_block` for a query about events starts with block 0, not whatever block was the first block that the contract existed:
https://github.com/ApeWorX/ape/blob/595a63a36ae3b59debba92f8b3e62ade5b9f29dc/src/ape/contracts/base.py#L480

This causes ape to cycle through a lot of history that isn't useful to query, because we know that the contract didn't exist at that point in time:
https://github.com/ApeWorX/ape/blob/595a63a36ae3b59debba92f8b3e62ade5b9f29dc/src/ape/api/providers.py#L872

Typically, this leads my requests to time out if using a remote provider

### How can it be fixed?

If possible, the contract instance object should find a way of determining the block where the contract was (first) deployed, and start it's query from there instead. I say "first" deployed, because a re-initializable proxy can be re-deployed to a given address.

gosuto-inzasheru:
also filed in #1119

PR #1548, opened 2023-07-18
feat: improved contract receipt related logic

antazoey:
### What I did

**Please** review + merge #1547 first! this PR depends on those changes so it has all the same changes but i will rebase after and the diff will look nicer. I am leaving it in draft until that is done

* Adds `chain.get_contract_receipt()` method to help get receipts for contracts.
* Changes `receipt` on `ContractInstance` to no longer be optional. If it does not have it, it will call the new method to get it.
* fixes: #1062  Changes the default `start_block` to use the start block of the contract's receipt when not given for event queries

### How I did it

@fubuloubu wrote the start of the algo, i fixed the bug in it, pasted in it, made mypy happy, added some tests, wired it up in some places that make it work nicely in the bg.

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -591,9 +591,19 @@ class ContractEvent(ManagerAccessMixin):
         stop_block = None
 
         if stop is None:
-            start_block = self.chain_manager.contracts.get_creation_receipt(
-                self.contract.address
-            ).block_number
+            contract = None
+            try:
+                contract = self.chain_manager.contracts.instance_at(self.contract.address)
+            except Exception:
+                pass
+
+            if contract:
+                start_block = contract.receipt.block_number
+            else:
+                start_block = self.chain_manager.contracts.get_creation_receipt(
+                    self.contract.address
+                ).block_number
+
             stop_block = start_or_stop
         elif start_or_stop is not None and stop is not None:
             start_block = start_or_stop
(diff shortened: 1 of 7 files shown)


src/ape/contracts/base.py
```diff
@@ -584,22 +584,25 @@ def range(
             Iterator[:class:`~ape.contracts.base.ContractLog`]
         """
 
+        if not hasattr(self.contract, "address"):
+            yield from []
```
fubuloubu, line 588: Don't think this line is necessary

src/ape/managers/chain.py
```diff
@@ -1645,4 +1645,50 @@ def set_balance(self, account: Union[BaseAddress, AddressType], amount: Union[in
...
+
+        Args:
+            transaction_hash (str): The hash of the transaction.
+
+        Returns:
+            :class:`~ape.apt.transactions.ReceiptAPI`
+        """
         return self.chain_manager.history[transaction_hash]
+
+    def get_contract_receipt(
```
fubuloubu, line 1659: Not a fan of this name or location, could it be `ContractCache.get_creation_receipt(addr)`?

antazoey: Yes, I will make this change!

But I am not 100% convinced `ContractCache` is actually the right place for this because of the frequent use of `BlockContainer` is this logic. That hints to me that this belongs at a higher level than `ContractCache` or something.

fubuloubu: it uses `ProviderAPI.get_code`, not sure where it is more logically situated to better take advantage of the block container (except current chain head), but semantically speaking it's directly related to contracts (we also cache it in the contract)

fubuloubu (approved):

PR #1555, opened 2023-07-21
fix: handle nested structures in global config merge [APE-1218]

antazoey:
### What I did

Before, only top level keys worked, if you overrode one in a project, you had to override everything.
Now, it will merge the objects the best it can, thus global mainnet providers work a lot better as most projects still configure local Ethereum providers and that was causing conflicts / loss of config specs

### How I did it

carefully merge the objects

### How to verify it

can now config Ethereum in both places and have it merge and use both

### Checklist

### DIFF
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -323,7 +323,7 @@ class ConfigManager(BaseInterfaceModel):
 
 
 def merge_configs(base: Dict, secondary: Dict) -> Dict:
-    result = {}
+    result: Dict = {}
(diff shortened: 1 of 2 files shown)


fubuloubu (dismissed):
Probably needs a few more test cases for different scenarios

antazoey:
> Probably needs a few more test cases for different scenarios

All test cases are now added, see comments in test explaining how each part is included.

fubuloubu (approved):

Issue #570, opened 2022-03-14
Control + C doesn't escape `ape test`


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.0

$ ape plugins list
Installed Plugins:
  vyper           0.1.0
  hardhat         0.1.0
  solidity        0.1.0
```

* Python Version: 3.9.9
* OS: linux

### What went wrong?
Typically, I can stop a pytest instance by pressing `<ctrl>+C` while a test is running, but I've noticed that for whatever reason, `ape test` captures this output too hard and I have to keep pressing it until it regesiters and quits with an internal error. This is not great UX

### How can it be fixed?
Debug

unparalleled-js:
Kinda weird, this is my output now:

```
INFO: Starting 'Hardhat node' process.

tests/test_solidity.py ....^CEERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
EERROR: Not connected to a network provider.
```

so it turns off the provider but not pytest

PR #1560, opened 2023-07-24
fix: issue where CTRL-C did not exit tests

antazoey:
### What I did

fixes: #570 
Fixes: APE-

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #1561, opened 2023-07-24
fix: avoid signal [APE-1223]

antazoey:
### What I did

alt and maybe better fix than #1560 

fixes: #570 
Fixes: APE-1223

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -1,52 +1,49 @@
-import signal
 import sys
 
-if "test" not in sys.argv and "pytest" not in sys.argv:
-    signal.signal(signal.SIGINT, lambda s, f: sys.exit(130))
+try:
+    from ape.managers.project import ProjectManager as Project
+    from ape.pytest.contextmanagers import RevertsContextManager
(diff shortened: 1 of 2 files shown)


src/ape/__init__.py
```diff
@@ -1,8 +1,8 @@
 import signal
+import sys
 
-signal.signal(signal.SIGINT, lambda s, f: _sys.exit(130))
-
-import sys as _sys
+if "test" not in sys.argv and "pytest" not in sys.argv:
```
antazoey, line 4: by avoiding signaling when pytest is running, we can rely on the existing handlers

src/ape/api/providers.py
```diff
@@ -1593,14 +1599,6 @@ def connect(self):
         # Register atexit handler to make sure disconnect is called for normal object lifecycle.
         atexit.register(self.disconnect)
 
-        # Register handlers to ensure atexit handlers are called when Python dies.
-        def _signal_handler(signum, frame):
```
antazoey, line 1597: the atexit functions were already running it seems, this wasnt doing anything but causing problems

PR #1566, opened 2023-07-27
fix: remove pin on evm-trace [APE-1230]

istankovic:
### What I did

Remove the pin on evm-trace.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ setup(
         # ** Dependencies maintained by ApeWorX **
         "eip712>=0.2.1,<0.3",
         "ethpm-types>=0.5.3,<0.6",
-        "evm-trace==0.1.0a21",
+        "evm-trace>=0.1.0a22",
     ],
     entry_points={
         "console_scripts": ["ape=ape._cli:cli"],


setup.py
```diff
@@ -123,7 +123,7 @@
         # ** Dependencies maintained by ApeWorX **
         "eip712>=0.2.1,<0.3",
         "ethpm-types>=0.5.3,<0.6",
-        "evm-trace==0.1.0a21",
+        "evm-trace>=0.1.0a22",
```
fubuloubu, line 126: We normally don't leave alpha pins wide, but also this should transition into beta at this point

istankovic: What do you suggest the version constraint be?

fubuloubu: For an alpha it should be `==` but as we discussed in discord, this is fine for now (and we should make a full release of this package so we can do proper pins on minor revisions)

fubuloubu (approved):
TODO: Upgrade `evm-trace` to a fully released package

tracking here: https://github.com/ApeWorX/evm-trace/issues/47

fubuloubu:
@istankovic rebase and it should merge

fubuloubu (approved):

setup.py
antazoey: this was here for a reason!!! lol

antazoey: oh heck ya, web3.py has been updated! awesome, thank you

antazoey: we had this pin because we had released evm-trace a bit too early where it causes install hangs on ape, but i think with the latest ETH based releases are good

PR #1569, opened 2023-07-27
fix: does not add  fixture when it's a doctest [APE-1236]

nazariyv:
### What I did

My previously merged PR wouldn't have fixed the issue because the check for fixturenames was performed before the check if the item is a DoctestItem.

Previous PR: https://github.com/ApeWorX/ape/pull/1567

TLDR; of the issue in previous PR:

image

### How I did it

made the order correct. First check if it's a DoctestItem, if not, then continue with the other check.

### DIFF
--- a/src/ape/pytest/runners.py
+++ b/src/ape/pytest/runners.py
@@ -134,6 +134,7 @@ class PytestApeRunner(ManagerAccessMixin):
         if (
             self.config_wrapper.isolation is False
             or "_function_isolation" in item.fixturenames  # prevent double injection
+            or isinstance(item, pytest.DoctestItem)  # doctests don't have fixturenames
         ):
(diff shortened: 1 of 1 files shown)


src/ape/pytest/runners.py
```diff
@@ -133,6 +133,7 @@ def pytest_runtest_setup(self, item):
         """
         if (
             self.config_wrapper.isolation is False
+            or isinstance(item, pytest.DoctestItem)  # doctests don't have fixturenames
             or "_function_isolation" in item.fixturenames  # prevent double injection
             or isinstance(item, pytest.DoctestItem)  # doctests don't have fixturenames
```
fubuloubu, line 138:
```suggestion
```

fubuloubu (approved):

PR #1588, opened 2023-08-07
feat: network base fee multiplier is now configurable [APE-1273]

antazoey:
### What I did

To prevent issues where the user may not have enough balance to cover the txn, we should have a smaller base fee multiplier.

Downside: less actual safety the tx will go through

The value could be configurable or up to debate though.

### How I did it

noticed a user having an issue where their account was said to not have enough balance to cover a tx.
noticed that other tools use less: https://github.com/wagmi-dev/viem/blob/main/src/utils/transaction/prepareRequest.ts#L81
wonder if we should do the same

### How to verify it

make a tx on a live network

### Checklist


fubuloubu (approved):
Yes would like this configurable at some point

banteg:
i chose 2x as approximately 1.125^6, so the tx would likely go through even the base fee is max growing for six consecutive blocks. since you don't overpay in eip-1559, i don't see how a higher multiplier can be a problem.

i also ran some numbers and the median multiplier people use is around 1.4-1.5. using 2 would land you into 85th percentile.

fubuloubu:
> im against lowering.
> 
> i chose 2x as approximately 1.125^6, so the tx would likely go through even when the base fee is max growing for six consecutive blocks. since you don't overpay in eip-1559, i don't see how a higher multiplier can be a problem.
> 
> i also ran some numbers and the median multiplier people use is around 1.4-1.5. using 2 would land you into 85th percentile.

The flip side of this is that the higher base fee multiplier, the more you have to have to cover in case it goes bigger (even if it doesn't use it), which could prevent people from making transactions sometimes

Perhaps it would be nice to weave this through as a kwarg, so you can on the fly pick a different multiplier e.g. `contract.myMethod(..., base_fee_multiplier=1.4)`

antazoey:
I think it should default to 1.4 and be configurable, but we can wait to change the default until 0.7 and just let it be configurable for now

fubuloubu (approved):

Issue #1396, opened 2023-04-22
Raise warning if `value=` is specified for anything but a `payable` function [APE-874]


fubuloubu:
### Overview

Ape provides no feedback when you attempt to send value to a contract method that may reject it due to having a `nonpayable` check, and relies on compiler-specific errors which may or may not alert a user to scenario

### Specification

Raising a warning and not an error is better, because sometimes you might mean to do this, and it should raise another exception e.g. `ContractLogicError` or something compiler-specific during gas estimation

### Dependencies
n/a

PR #1599, opened 2023-08-10
feat: Raise warning if `value=` is specified for anything but a `payable` function [APE-1283]

Aviksaikat:
### What I did

Raise warning if value= is specified for anything but a payable function

fixes: #1396 (Raise warning if value= is specified for anything but a payable function [APE-874])

### How I did it
Added check to see if `msg.value` is getting passed to a `non-payable` contract

### How to verify it
1. Create a contract with a non-payable constructor like this
```js
pragma solidity ^0.8.19;

contract NonPayableConstructor {

    address public owner;
    constructor() {
        owner = msg.sender;
    }

    function getOwner() external view returns(address){
        return owner;
    }

}
```
3. Create a `test` or `script` with the following contents
```py
#!/usr/bin/python3
from ape import project, accounts

user = accounts[1]
user.set_autosign(True)
user.balance += int(100e18)


def test_NonPayableConstructor():
    npc = project.NonPayableConstructor.deploy(sender=user, value="1 ether")
    
    print(npc.getOwner())

def main():
    test_NonPayableConstructor()
``` 
4. run using `ape run path/to/script` or `ape test -k testName`
5. You will see an expectation like this `ERROR: (ContractError) Sending funds to a non-payable constructor!!`
image

### Checklist

- [x] Documentation has been updated ~~(Only applicable if there is a page for Exceptions)~~ you added a docstring to the excpetion, that will get autogenerated


src/ape/contracts/base.py
```diff
@@ -1300,6 +1300,13 @@ def __call__(self, *args, **kwargs) -> TransactionAPI:
     def deploy(self, *args, publish: bool = False, **kwargs) -> ContractInstance:
         txn = self(*args, **kwargs)
         private = kwargs.get("private", False)
+        has_value = kwargs.get("value")
+
+        if has_value:
+            has_non_payable_constructor = self.contract_type.constructor.is_payable
+            # print(has_non_payable_constructor)
+            if not has_non_payable_constructor:
+                raise ContractError("Sending funds to a non-payable constructor!!")
```
fubuloubu, line 1309:
```suggestion

        if kwargs.get("value") and not self.contract_type.constructor.is_payable
            raise ContractError("Sending funds to a non-payable constructor.")
```

fubuloubu: Also can probably make a new exception in `ape/exceptions.py` for this instead, since it's pre-contract logic

fubuloubu (dismissed):
Needs a test

src/ape/exceptions.py
```diff
@@ -832,3 +832,9 @@ def _get_custom_python_traceback(
         tb_next = tb
 
     return frames[-1]
+
+
+class SendingFundsToNonPayableConstructorError(ApeException):
```
fubuloubu, line 837: I think renaming to this and put it before `TransactionError`
```suggestion
class MethodNonPayableError(ContractError):
```

Aviksaikat:
Hmmm. Why the test on `mac-os, 3.8` failed ? :thinking:

src/ape/api/accounts.py
```diff
@@ -215,6 +221,9 @@ def deploy(
             :class:`~ape.contracts.ContractInstance`: An instance of the deployed contract.
         """
         txn = contract(*args, **kwargs)
```
fubuloubu, line 223: I believe that this line actually front runs the below error with the one in `ape.contracts.base`

Try removing the next line and see if your test still passes

fubuloubu (dismissed):

src/ape/api/accounts.py
Aviksaikat: image
Both `account.deploy` & `project.contract.deploy` test is working. But if I don't put  

```py
if kwargs.get("value") and not contract.contract_type.constructor.is_payable:
            raise MethodNonPayableError("Sending funds to a non-payable constructor.")
```
in line `224` in `src/ape/api/accounts.py` the `test_sending_funds_to_non_payable_constructor_by_accountDeploy` test will fail like you said it's being front run by ` txn = contract(*args, **kwargs)` . I tried to catch the exception here `src/ape/contracts/base.py#L1347` also but for some reason, it was not catching, instead it moved to the next exception i.e. `ContractLogicError`.

src/ape/exceptions.py
```diff
@@ -131,6 +131,12 @@ def __init__(self, message: Optional[str] = None):
         super().__init__(message)
 
 
+class MethodNonPayableError(ContractError):
+    """
+    Raises when sending funds to a non-payable constructor
```
fubuloubu, line 136:
```suggestion
    Raises when sending funds to a non-payable method
```

fubuloubu (dismissed):

fubuloubu (approved):

Issue #1593, opened 2023-08-09
`get_creation_receipt` can't find contracts deployed by contracts


banteg:
### What went wrong?

```python
> chain.contracts.get_creation_receipt('0x35D1b3F3D7966A1DFe207aa4514C12a259A0492B')
ChainError: Failed to find a contract-creation receipt for '0x35D1b3F3D7966A1DFe207aa4514C12a259A0492B'.
```

same with otterscan api
```python
cast rpc ots_getContractCreator 0x35D1b3F3D7966A1DFe207aa4514C12a259A0492B
{"hash":"0xa39edcdeb7310150d7be44b8da94e5043d1cf8600d59f855a65c7a9c035b06a7","creator":"0x403689148fa98a5a6fdcc0b984914ae968d788e5"}
```

### How can it be fixed?

after finding a block when the contract code appeared using binary search, rather than looking through receipts, we need to look at block traces. if we had to emulate the otterscan method, the logic would go something like this:

```python
def get_contract_creator(contract):
    creation_block = get_creation_block(contract)
    block_traces = chain.provider._make_request(
        "trace_replayBlockTransactions", [block, ["trace"]]
    )
    for txn_traces in block_traces:
        for trace in txn_traces:
            if (
                trace["type"] == "create"
                and "error" not in trace
                and trace["result"]["address"] == contract.lower()
            ):
                return {"hash": trace["txnHash"], "creator": trace["action"]["from"]}
```

this would also be much faster than going through all the receipts.

i suggest adding client capabilities detection like discussed earlier and then use the following priortiy:
1. `ots_getContractCreator` when otterscan namespace is available
2. binary search the block + the method i described above when parity trace namespace is available
3. binary search the block + receipts + user warning about this method possibly being inaccurate otherwise
4. fallback to etherscan or some other api when the client doesn't have archive state because we can't binary search.

besides, given we only need to know the creation block in `ContractEvent.range`, it seems wasteful to do so much work identifying the right receipt and then immediately discarding it. 

i suggest we split up `get_creation_receipt` into two functions, one getting a creation block, and another doing something similar to otterscan api. or maybe there could be one func with bool params like `(block, txn_hash, creator)` that would exit early as soon as it collects all the needed info.

antazoey:
i have also proposed that we use the EtherscanAPI when it is available to get the receipt. The issue there is the design, the assumption of how ExplorerAPI will behave always like Etherscan in that regard.

fubuloubu:
[...]

PR #1606, opened 2023-08-17
feat(query): contract creation query support [APE-1296]

fubuloubu:
### What I did

This PR moves the `get_creation_receipt` to be a query that is supported in the query system. The goal is that now that it's in the query system, it can be supported by faster query engines vs. brute force search (such as etherscan) to speed up the capability of the search (and also help resolve the bug noted in #1593)

### How I did it

This is only a partial fix to #1593, there are still more fixes needed to the brute force method in core (so it can resolve the issue mentioned more effectively/efficiently)

### How to verify it

Mostly a refactor, so if tests work :shrug: 

### Checklist


src/ape/managers/chain.py
```diff
@@ -1348,35 +1349,26 @@ def get_creation_receipt(
...
-            )
+            ),
+        )
 
-        else:
+        try:
+            # Get the first contract receipt, which is the first time it appears
+            return next(creation_receipts)
+        except StopIteration:
             raise ChainError(f"Failed to find a contract-creation receipt for '{address}'.")
```
fubuloubu, line 1372: Add to the error that this might be because it's created by another contract, and to install an alternative engine that can handle it, like etherscan

PR #1614, opened 2023-08-22
refactor!: accept any msg type via `AccountAPI.sign_message` [APE-1305]

z80dev:
### What I did

Refactor the definition of sign_message to accept Any. Signers implementing this class can raise when they receive a message type they can't handle.

Account plugins would each individual handle msg objects by type, including how they might display information from them for user feedback in terms of signing. This will improve UX because now we can display the text of a string message to sign (useful for displaying SIWE [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361) authorizations), the complete object body and header for EIP-712 structured messages, and other future types of messages that might become necessary to support within Ape (EIP-4337?)

fixes: APE-1103

### How I did it

Change type signature of sign_message function, and update implementors to match the new definition.

### How to verify it

Run existing test suite

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterator, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Any, Iterator, List, Optional, Type, Union
 
 import click
 from eip712.messages import SignableMessage as EIP712SignableMessage
@@ -54,18 +54,21 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         return None
 
     @abstractmethod
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
         """
         Sign a message.
 
         Args:
-          msg (:class:`~ape.types.signatures.SignableMessage`): The message to sign.
+          msg (Any): The message to sign. Account plugins can handle various types of messages.
+            For example, :class:`~ape.accounts.LocalAccount` can handle
+            :class:`~ape.types.signatures.SignableMessage`
             See these
             `docs <https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.messages.SignableMessage>`__  # noqa: E501
             for more type information on this type.
+          **signer_options: Additional kwargs given to the signer to modify the signing operation.
 
         Returns:
-          :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
+            :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
         """
 
     @abstractmethod
@@ -494,7 +497,7 @@ class ImpersonatedAccount(AccountAPI):
     def address(self) -> AddressType:
         return self.raw_address
 
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
         raise NotImplementedError("This account cannot sign messages")
 
     def sign_transaction(self, txn: TransactionAPI, **kwargs) -> Optional[TransactionAPI]:
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -2,8 +2,9 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterator, List, Optional, Type, Union
 
 import click
-from eip712.messages import SignableMessage as EIP712SignableMessage
+from eip712.messages import EIP712Message, SignableMessage as EIP712SignableMessage
 from eth_account import Account
+from eth_account.messages import encode_defunct
 
 from ape.api.address import BaseAddress
 from ape.api.transactions import ReceiptAPI, TransactionAPI
@@ -261,7 +262,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
 
     def check_signature(
         self,
-        data: Union[SignableMessage, TransactionAPI],
+        data: Union[SignableMessage, TransactionAPI, str, EIP712Message],
         signature: Optional[MessageSignature] = None,  # TransactionAPI doesn't need it
     ) -> bool:
         """
@@ -276,6 +277,10 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
         """
+        if isinstance(data, str):
+            data = encode_defunct(text=data)
+        if isinstance(data, EIP712Message):
+            data = data.signable_message
         if isinstance(data, (SignableMessage, EIP712SignableMessage)):
             if signature:
                 return self.address == Account.recover_message(data, vrs=signature)
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -68,7 +68,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
           **signer_options: Additional kwargs given to the signer to modify the signing operation.
 
         Returns:
-            :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
+            :class:`~ape.types.signatures.MessageSignature` (optional): The signature corresponding to the message.
         """
 
     @abstractmethod
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -6,6 +6,7 @@ from eip712.messages import EIP712Message
 from eip712.messages import SignableMessage as EIP712SignableMessage
 from eth_account import Account
 from eth_account.messages import encode_defunct
+from hexbytes import HexBytes
 
 from ape.api.address import BaseAddress
 from ape.api.transactions import ReceiptAPI, TransactionAPI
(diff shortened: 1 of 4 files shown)


src/ape/api/accounts.py
```diff
@@ -54,18 +54,21 @@ def alias(self) -> Optional[str]:
...
+            For example, :class:`~ape.accounts.LocalAccount` can handle
+            :class:`~ape.types.signatures.SignableMessage`
             See these
             `docs <https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.messages.SignableMessage>`__  # noqa: E501
             for more type information on this type.
+          **signer_options: Additional kwargs given to the signer to modify the signing operation.
 
         Returns:
-          :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
+            :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
```
fubuloubu, line 71:
```suggestion
            :class:`~ape.types.signatures.MessageSignature` (optional): The signature corresponding to the message.
```

src/ape_accounts/accounts.py
```diff
@@ -119,17 +120,20 @@ def delete(self):
...
+            user_approves = self.__autosign or click.confirm(f"{msg}\n\nSign: ")
+            if not user_approves:
+                return None
+            signed_msg = EthAccount.sign_message(msg, self.__key)
+            return MessageSignature(
+                v=signed_msg.v,
+                r=to_bytes(signed_msg.r),
+                s=to_bytes(signed_msg.s),
+            )
+        raise TypeError(f"Can't sign message of type {type(msg)}")
```
fubuloubu, line 136: Can return `None` to signify it didn't sign as well

fubuloubu: Wonder which is better? Remove `Optional` from the signature, and always raise if it can't sign? Or return None

src/ape_accounts/accounts.py
```diff
@@ -119,17 +120,20 @@ def delete(self):
...
-        if not user_approves:
-            return None
-
-        signed_msg = EthAccount.sign_message(msg, self.__key)
-        return MessageSignature(
-            v=signed_msg.v,
-            r=to_bytes(signed_msg.r),
-            s=to_bytes(signed_msg.s),
-        )
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
```
fubuloubu, line 123: oh, would also be nice to accept `eip712.EIP712Message` types here too

src/ape_accounts/accounts.py
z80dev: Is there value in distinguishing between type mismatch and other signature failures (like user rejecting to sign)? We can distinguish between the two by raishing instead of returning None, but maybe we don't need/want to do that

fubuloubu: To help debug the issue, it would be helpful to see that feedback. But None might be more useful for multi-signing flows

Perhaps it can be a warning logg instead?

z80dev: Makes sense, just made that change

fubuloubu (approved):
Changes look good to me, I'll let @antazoey give final approval (and see if she likes the warning log)

src/ape_accounts/accounts.py
```diff
@@ -119,17 +121,25 @@ def delete(self):
...
+        if isinstance(msg, str):
+            # Convert str to SignableMessage for handling below
```
fubuloubu, line 126: Actaully, you got me thinking: maybe this can display the text so when the user goes to confirm, they see the message

src/ape_accounts/accounts.py
```diff
@@ -119,17 +121,25 @@ def delete(self):
...
+            msg = encode_defunct(text=msg)
+        elif isinstance(msg, EIP712Message):
+            # Convert EIP712Message to SignableMessage for handling below
```
fubuloubu, line 129: Same here, you can display the full message instead of just the hashes

src/ape/api/accounts.py
```diff
@@ -258,7 +263,7 @@ def declare(self, contract: "ContractContainer", *args, **kwargs) -> ReceiptAPI:
 
     def check_signature(
         self,
-        data: Union[SignableMessage, TransactionAPI],
+        data: Union[SignableMessage, TransactionAPI, str, EIP712Message],
```
antazoey, line 266: I wonder if we should just make this `Any` as well.
Starknet signs ints. They love ints more than any other data type in that ecosystem. I want ints.

src/ape/api/accounts.py
```diff
@@ -273,6 +278,10 @@ def check_signature(
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
         """
+        if isinstance(data, str):
+            data = encode_defunct(text=data)
+        if isinstance(data, EIP712Message):
```
antazoey, line 283: Should these all be `elif` ?

src/ape_accounts/accounts.py
```diff
@@ -130,11 +132,46 @@ def delete(self):
         self.__decrypt_keyfile(passphrase)
         self.keyfile_path.unlink()
 
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
-        user_approves = self.__autosign or click.confirm(f"{msg}\n\nSign: ")
-        if not user_approves:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
+        if not isinstance(msg, (SignableMessage, EIP712Message, str)):
```
antazoey, line 136: `int` and `bytes` would also be nice.

src/ape_test/accounts.py
```diff
@@ -101,13 +101,17 @@ def alias(self) -> str:
...
-        signed_msg = EthAccount.sign_message(msg, self.private_key)
-        return MessageSignature(
-            v=signed_msg.v,
-            r=to_bytes(signed_msg.r),
-            s=to_bytes(signed_msg.s),
-        )
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
+        if isinstance(msg, str):
+            msg = encode_defunct(text=msg)
+        if isinstance(msg, SignableMessage):
```
antazoey, line 107: `elif` ?

tests/functional/test_accounts.py
```diff
@@ -47,6 +47,18 @@ def test_sign_message(signer):
...
 
 
+def test_sign_string(signer):
+    message = "Hello Apes!"
+    signature = signer.sign_message(message)
+    assert signer.check_signature(message, signature)
+
+
+def test_sign_message_unsupported_type_returns_none(signer):
+    message = 1234
```
antazoey, line 57: Let's use a float or something weirder. I think `int` should be valid personally, just convert to hex first.

```
int_to_str = HexBytes(my_int).hex()
```

src/ape_test/accounts.py
z80dev: not the case here, I should add a comment to clarify

encode_defunct reutns a SignableMessage which we then want to process

the first str check can be thought of as pre-processing a string into a signable message before proceeding

tests/functional/test_accounts.py
z80dev: got it, added support for signing/verifying ints directly, and updated this test case to pass a float

src/ape/api/accounts.py
z80dev: the check for EIP712Message can be an elif (as well as the newly added int check), not the subsequent check for SignableMessage/EIP712SignableMessage

src/ape/api/accounts.py
z80dev: ints have been added, I do agree this list is getting large

src/ape_accounts/accounts.py
```diff
@@ -158,7 +158,7 @@ def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]
             if msg._verifyingContract_:
                 display_msg += f"\tContract: {msg._verifyingContract_}\n"
             if msg._salt_:
-                display_msg += f"\tSalt: {msg._salt_}\n"
+                display_msg += f"\tSalt: {msg._salt_.decode('utf-8')}\n"
```
fubuloubu, line 164: would it be better display it as hex?

src/ape_accounts/accounts.py
```diff
@@ -130,11 +132,51 @@ def delete(self):
...
+        if isinstance(msg, str):
+            # Convert str to SignableMessage for handling below
+            user_approves = self.__autosign or click.confirm(f"Message: {msg}\n\nSign: ")
+            msg = encode_defunct(text=msg)
+        elif isinstance(msg, int):
+            user_approves = self.__autosign or click.confirm(f"Message: {msg}\n\nSign: ")
+            msg = encode_defunct(hexstr=HexBytes(msg).hex())
+        elif isinstance(msg, EIP712Message):
+            # Display message data to user
+            display_msg = "Signing Message\n"
```
fubuloubu, line 151:
```suggestion
            display_msg = "Signing EIP712 Message\n"
```

fubuloubu: can also use the type name of the main message field

fubuloubu: can also use the type name of the main message field

src/ape_accounts/accounts.py
```diff
@@ -130,11 +132,51 @@ def delete(self):
...
+        user_approves = False
+
+        if isinstance(msg, str):
+            # Convert str to SignableMessage for handling below
+            user_approves = self.__autosign or click.confirm(f"Message: {msg}\n\nSign: ")
+            msg = encode_defunct(text=msg)
+        elif isinstance(msg, int):
+            user_approves = self.__autosign or click.confirm(f"Message: {msg}\n\nSign: ")
+            msg = encode_defunct(hexstr=HexBytes(msg).hex())
+        elif isinstance(msg, EIP712Message):
```
fubuloubu, line 149: you want this to fall through to signable message
```suggestion

        if isinstance(msg, EIP712Message):
```

antazoey (dismissed):

fubuloubu (dismissed):

Issue #1571, opened 2023-07-28
pyproject.toml based ape configuration [APE-1238]


antazoey:
### Elevator pitch:

This pitch to document the pros and cons of switching from ape-config.yaml to pyproject.toml, as part of 1.0 release of the framework.

Many python tools are all using the single pyproject.toml file now. If your smart contract project uses `ape test` and `ape run` extensively, `pyproject.toml` is really nice because you can configure all your python tools together.

However, if your project only uses Ape a little bit and it uses Foundry (rust) or Hardhat (javascript) a lot more extensively, forcing the user to use the `pyproject.toml` is a worse experience than having a simple `ape-config.yaml` to configure the tool.

So there are definitely pros and cons to both; you really need to wear the different hats and think about what you would want to do.

Por que no los dos

Well, we have the `ProjectAPI` hook right now which allows a hook for reading configs... But it may be a lot more clearer even after the 1.0 launch where we revamp the plugin system a bit (and the project system honestly).  But in any case, we can create a plugin of some kind that uses `pyproject.toml` and still keep the `ape-config.yaml` as the "normal" way to configure. Many tools do this still!

### Value:

devys

### Dependencies:

knowing how pyproject.toml works i guess

### Design approach:

mirror config in toml
uses known rules for copying them in but still to toml

### Task list:

* Weigh pros and cons
* Decide if need to wait until post plugin revamp
* Support
* Document

### Estimated completion date:

1.0

### Design review:

Do not signoff unless:

- 1. agreed the tasks and design approach will achieve acceptance, and
- 2. the work can be completed by one person within the SLA.
     Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

antazoey:
I kind of wanted to create this issue because I noticed we did not have it, I want community feedback, we are often discussing it, and I found this conversation ins `mdformat` really interesting: https://github.com/executablebooks/mdformat/issues/356

It is not exactly the same issue as ours but ultimately the same decision and the points are interesting to read.

Aviksaikat:
If this is the way we want to proceed then I can take a look at implementing this feature

fubuloubu:
> If this is the way we want to proceed then I can take a look at implementing this feature

I think this would require further research, so if you wanted to research it for us that would really help us in evaluating the solution and any gotchas it might have

Aviksaikat:
[...]

PR #1618, opened 2023-08-23
feat: added option to generate basic `ape-config.yaml` file if `[tool.ape.plugins]` property is specified inside pyproject.toml file [APE-1312]

Aviksaikat:
…

### What I did

Added option to specify plugins in the `ape-config.yaml` file based on the `[tool.ape.plugins]` property of `pyproject.toml` file.

fixes: #1571

### How I did it

Added functions for parsing the `pyproject.toml`  file & look for lines which have the `ape-` initials & then stripping them & put them inside the `ape-config.yaml` file with proper formatting

### How to verify it

1. Create a new folder with this `pyproject.toml` file
```toml
[tool.poetry]
name = "testing ape-config"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"
packages = [{include = "Flash_loan_arbritage"}]

[tool.poetry.dependencies]
python = ">=3.10,<3.11"
eth-ape = "*"
ape-alchemy = "^0.6.2"
ape-solidity= "^0.6.7"
ape-foundry = "^0.6.14"
ape-etherscan = "*"


[tool.ape.plugins]
ape-alchemy = "*"
ape-solidity = "*"
ape-foundry = "*"
ape-etherscan = "*"

# working on adding support for this property
[tool.ape.network]
ecosystem = "ethereum"
default_network = "mainnet-fork"
default_provider = "foundry"

# working on adding support for this property
[tool.ape.tests]
mnemonic = "test test test test test test test test test test test junk"
number_of_test_accounts = 10

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"

```
2. Then create a new virtual environment by doing `poetry install && poetry shell` 
3. Then run `pip install -e <path/to/the/cloned/repo/with/this/revision/`
4. Then run `ape init` 
5. Done

### POC
image
image

### Checklist


src/ape_init/_cli.py
```diff
@@ -7,6 +7,32 @@
 from ape.utils import github_client
 
 
+def read_dependencies_from_toml(file_path):
+    with open(file_path, "r") as file:
```
antazoey, line 11: Use the pathlib for this.

```python
content = file_path.read_text().splitlines()
```

src/ape_init/_cli.py
```diff
@@ -7,6 +7,32 @@
...
+        content = file.readlines()
+
+    in_dependencies = False
+    dependencies = []
+
+    for line in content:
+        stripped_line = line.strip()
```
antazoey, line 18: This should be using the standard tomllib, I would think:
https://docs.python.org/3/library/tomllib.html

src/ape_init/_cli.py
```diff
@@ -59,3 +85,11 @@ def cli(cli_ctx, github):
             project_name = click.prompt("Please enter project name")
             ape_config.write_text(f"name: {project_name}\n")
             cli_ctx.logger.success(f"{project_name} is written in ape-config.yaml")
+
+        pyproject_toml = project_folder / "pyproject.toml"
+        if pyproject_toml.exists():
```
antazoey, line 90: Use `is_file()` instead of `exists()` in this case

antazoey (changes requested):
Some missing things,
and I think there is a helpful standard library to assist with this and simplify the code we have to manage

src/ape_init/_cli.py
fubuloubu: Only in 3.11+, for older versions this is the recommended backport: https://github.com/hukkin/tomli

Aviksaikat:
I messed up this commit.

setup.py
```diff
@@ -112,6 +112,7 @@
         "tqdm>=4.62.3,<5.0",
         "traitlets>=5.3.0",
         "watchdog>=3.0,<4",
+        "tomli>=2.0.1",
```
fubuloubu, line 115: This is how you would specify backports
```suggestion
        "tomli>=2.0.1 ; python_version<3.11",
```

src/ape_init/_cli.py
```diff
@@ -1,11 +1,47 @@
...
+try:
+    import tomllib  # type: ignore[import]
+# backwards compatibility
+except ModuleNotFoundError:
+    import tomli as tomllib
+
+
+def read_dependencies_from_toml(file_path, cli_ctx) -> List[str]:
+
+    with open("./pyproject.toml", "rb") as file:
```
antazoey, line 19: this should all happen in the config manage, not in ape-init plugin.
And then you can use the project_manager.path as the base here instead of cwd

src/ape_init/_cli.py
```diff
@@ -1,11 +1,47 @@
...
+    # Extract the 'tool.poetry.dependencies' section
+    dependencies = data.get("tool", {}).get("poetry", {}).get("dependencies", {})
+
+    # Check each dependency to see if it starts with 'ape-', because we know these are the ape plugins
+    ape_plugins = [dep[4:] for dep in dependencies if dep.startswith("ape-")]
+
+    return ape_plugins
+
+
+def write_ape_config_yml(dependencies: List[str], file_to_write: Path):
```
antazoey, line 39: instead of writing an ape file, the ConfigManager should read the values from the pyproject.toml and treat it the same as if the config file had existed. There is not reason to configure twice, even if one is auto-generated.

antazoey (commented):
nice work but things are in the wrong spot and we should be configuring the tool directly instead of generating a config from a config

Aviksaikat:
any update on this one ?

antazoey:
> any update on this one ?

There is a quite a lot of unaddressed feedback. Basically, we want this to be another source of config rather than creating a config file from it

Aviksaikat:
so let's make it a draft and wait for further feedback.

antazoey:
> so let's make it a draft and wait for further feedback?

There is not much further feedback at this time, since the approach is a bit off from what we want we are not able to provide any more feedback until the approach matches more with the design approach to have the pyproject.toml be soure of config rather than it generating a new config file.

fubuloubu:
> > so let's make it a draft and wait for further feedback?
> 
> We are not able to provide any more feedback until the approach matches more with the design of having the pyproject.toml be source of config rather than it generating a new config file.

to expand on this, make it look something like this I suppose:

```toml
[tools.ape]
plugins = [
    etherscan,
    vyper,
]
```

fubuloubu:
I think this approach should be closed for now. I am in favor of migrating the `ape-config.yaml` file actually into `pyproject.toml` but it needs a lot of further thought

Issue #1622, opened 2023-08-24
Add http_uri and ws_uri properties


fubuloubu:
### Overview

Sometimes it is useful to provide the raw http/https or websocket connections to connect to directly. A `ProviderAPI` implementation should be able to signal if they provide this feature by returning a value for these properties

### Specification

```py
class ProviderAPI(...):
    ...
    def http_uri(self) -> Optional[str]:
        ...  # return http/https URI string if and only if this provider provides a HTTP RPC connection

    def ws_uri(self) -> Optional[str]:
        ...  # return ws/wss URI string if and only if this provider provides a websocket connection
```

### Dependencies

No dependencies

PR #1623, opened 2023-08-24
feat(ProviderAPI): add uri str fields for easier use [APE-1317]

fubuloubu:
### What I did

Adds two small properties that'll make it easier to grab the raw URIs to connect with

fixes: #1622 

### Checklist

- [ ] Need to check all other provider plugins for any issues


antazoey (approved):
I appreciate how we avoided the breaking changes!

PR #1625, opened 2023-08-24
fix: issues causing dev messages from working on vyper 0.3.9 [APE-1320]

antazoey:
### What I did

fixes: #

### Checklist


tests/functional/test_reverts_context_manager.py
```diff
@@ -141,140 +140,3 @@ def test_revert_partial_fails(owner, reverts_contract_instance):
     with pytest.raises(AssertionError):
         with reverts("ze"):
             reverts_contract_instance.revertStrings(0, sender=owner)
-
-
-@geth_process_test
-def test_dev_revert(owner, reverts_contract_instance, geth_provider):
```
antazoey, line 147: i had to move the dev messages check to `ape-vyper` since they now utilize the actual `CompilerAPI` methods for tracing a source; it makes sense they live there anyway, as they will differ from compiler to compiler a lot and this feature doesn't come with Core out of the box. That being said, it'd be nice to have a really simple source tracers for interfaces that treats each call as a single statement (just an idea)

fubuloubu: Hmm. Yeah I like that idea

fubuloubu (approved):

PR #1627, opened 2023-08-25
feat: a config for excluding files from being compile by default [APE-1326]

antazoey:
### What I did

Adds a config so that you can do:

```yaml
compile:
  exclude:
    - testlibs/*  # excludes contracts in `contracts/testlibs/`
    - *Mock.vy  # Exclude vyper files ending in `Mock`.
```

### Checklist


src/ape_compile/__init__.py
```diff
@@ -32,6 +37,10 @@ def warn_deprecate(cls, value):
 
         return None
 
+    @validator("exclude", pre=True)
+    def validate_exclude(cls, value):
+        return value or []
```
fubuloubu, line 42: Is this necessary?

fubuloubu (approved):

src/ape_compile/__init__.py
antazoey: define necessary? it makes an empty list in the config not be `None`.. helped me when commenting stuff out without having to delete the whole config.

antazoey: i was doing this:

```
compile:
  exclude:
    # - *thing*
```

and i had the error. I think it should know i just mean empty list here and not `None`, or at least i dont see the harm in allowing this state.

fubuloubu: ahhhh

you were actively inserting a `None`

fubuloubu (dismissed):

PR #1641, opened 2023-08-30
fix: false-positives bug when making mock-log assertions [APE-1352]

antazoey:
### What I did

Because of a check that ignored all fasley values, log-assertion checks would have false-positives, meaning your tests pass when they shouldn't! dun dun dunnn.

### How I did it

check that `is not None` instead of falsey.

### How to verify it

This passes on main branch believe it or not:

```python
tx = c.setNumber(1, sender=owner)
assert tx.events == [c.NumberChange(newNum=0)]  # actual value was 1 but Ape didn't care
```

however with these changes, it correctly fails.

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -367,7 +367,7 @@ class MockContractLog(BaseContractLog):
         for name, value in self.event_arguments.items():
             # Make sure `value` is not `None` (user explicitly set it `None`)
             # NOTE: `other.event_arguments[name]` will raise `IndexError` only if ABIs don't match
-            if value and value != other.event_arguments[name]:
+            if value is not None and value != other.event_arguments[name]:
                 return False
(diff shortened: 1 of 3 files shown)


src/ape/types/__init__.py
```diff
@@ -367,7 +367,7 @@ def __eq__(self, other: Any) -> bool:
         for name, value in self.event_arguments.items():
             # Make sure `value` is not `None` (user explicitly set it `None`)
             # NOTE: `other.event_arguments[name]` will raise `IndexError` only if ABIs don't match
-            if value and value != other.event_arguments[name]:
+            if value is not None and value != other.event_arguments[name]:
```
antazoey, line 370: bug fix here

fubuloubu (approved):
Good catch!

PR #1644, opened 2023-08-31
fix: shanghai support in local geth [APE-1355]

antazoey:
### What I did

I have finally realized what I was missing to get this to work.
Yay.
This makes the local geth work with shanghai, finally no more PUSH0 VM error when using local geth for testing.

### How I did it

configure `shanghaiTime` in genesis

### How to verify it

passing tests, no more push0 errors

### Checklist


fubuloubu (dismissed):

antazoey:
regarding the large test refactor, this uncovered a bad xdist scoping issue in our test suite. To continue using loadscope, i had to move all geth-related tests under one root folder `geth` so they are part of the same xdist scope

.github/workflows/test.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
             pip install .[test]
 
         - name: Run Tests
-          run: pytest -m "not fuzzing" -s --cov=src -n auto --dist loadscope
```
antazoey, line 98: it only runs on 2 cores anyway and it causes more problems than it is worth

fubuloubu (approved):

PR #1648, opened 2023-09-01
fix: issue where pretty trace tree included two sets of parentheses  [APE-1364]

antazoey:
### What I did

since full name (selector) gets used when methods have the same name, it would cause the pretty tree to look funny.
this PR strips out that nonsense

### Checklist


src/ape/utils/trace.py
```diff
@@ -72,6 +72,10 @@ def parse_as_str(call: "CallTreeNode", stylize: bool = False, verbose: bool = Fa
         and is_0x_prefixed(call.method_id)
         else str(call.method_id or "")
     )
+    if "(" in method:
+        # Only show short name, not ID name
+        # (is full sig when multipl methods with same name).
+        method = method.split("(")[0].strip()
```
antazoey, line 78:
```suggestion
        method = method.split("(")[0].strip() or method
```

src/ape/utils/trace.py
```diff
@@ -72,6 +72,10 @@ def parse_as_str(call: "CallTreeNode", stylize: bool = False, verbose: bool = Fa
         and is_0x_prefixed(call.method_id)
         else str(call.method_id or "")
     )
+    if "(" in method:
+        # Only show short name, not ID name
+        # (is full sig when multipl methods with same name).
```
antazoey, line 77:
```suggestion
        # (it is the full signature when multiple methods have the same name).
```

dtdang (approved):

PR #1654, opened 2023-09-08
fix(deps): update pin of ethpm-types [APE-1383]

fubuloubu:
I was constantly having issues with this, I think the pin should have been updated


antazoey (approved):

antazoey:
i may need to do this again soon

Issue #1559, opened 2023-07-24
Event Logs do not decode structs present in data [APE-1222]


fubuloubu:
### Environment information

- OS: macOS / linux / win
- Python Version: x.x.x
- `ape` and plugin versions:

```sh
$ ape --version
0.6.13

$ ape plugins list
Installed Plugins:
  foundry      0.6.12
  template     0.6.1
  etherscan    0.6.7
  alchemy      0.6.2
  tokens       0.6.1
  solidity     0.6.7
  vyper        0.6.9
  huff         0.1.2
```

### What went wrong?

```py
In [1]: seaport = Contract("0x00000000000000ADc04C56Bf30aC9d3c0aAF14dC")

In [2]: list(seaport.OrderFulfilled.range(chain.blocks.head.number-10, chain.blocks.head.number))
Out[2]: 
[<OrderFulfilled offerer=0xC0bf18A6cFAf6737A097566b15089F354B3365e1 zone=0x004C00500000aD104D7DBd00e3ae0A5C00560C00 orderHash=b'\xc4\xcb\xea\x85@\xab\xbfv\x90\x00\x10\x9f2\xfa\x81\t\xe5\xfcX\x1fo\x87\x84qd\xc12\x94\x1e\xd5\xd4\xeb' recipient=0x71be99c9b5362aD07f7f231bC7a547f4119C6073 offer=((2, '0xe785e82358879f061bc3dcac6f0444462d4b5330', 6127, 1),) consideration=((0, '0x0000000000000000000000000000000000000000', 0, 1452750000000000000, '0xc0bf18a6cfaf6737a097566b15089f354b3365e1'), (0, '0x0000000000000000000000000000000000000000', 0, 37250000000000000, '0x0000a26b00c1f0df003000390027140000faa719'))>,
...
```

### How can it be fixed?

In the ABI, the structs are shown to be "tuples", which is why they decode like this, but it should be possible to figure out what type they "actually" should be potentially

victor-ego:
@fubuloubu @antazoey can you guys have a look at this, please? cheers

PR #1659, opened 2023-09-13
fix: structs in logs [APE-1222]

antazoey:
### What I did

fixes: #1559 
Fixes: APE-1222

### How I did it

ugh don't get me started.
we were trying to decode without access to the ecosystem which causes some issues.
I also had to fix something in MockContractLog for the same thing and also the `__eq__` wasnt working right with structs either

### How to verify it

that one little test i added verifies so much

### Checklist


fubuloubu (approved):
This looks great! Was definitely an issue for me

Issue #1612, opened 2023-08-21
add timestamp column to event query api [APE-1302]


banteg:
### Overview

i would like to be able to do
```python
contract.Event.query('timestamp', 'eventArg')
```

fubuloubu:
Note: `timestamp` is a keyword in both vyper and solidity, so it wouldn't be an event arg (and thus would be safe to add, same as `.block`)

PR #1660, opened 2023-09-13
feat(query): add timestamp to ContractLog [APE-1391]

fubuloubu:
### What I did

Adds the `.timestamp` property to `ContractLog`, allowing users to pull the raw timestamp when querying events

fixes: #1612

### How I did it

This should be safe to add as both Solidity and Vyper use `timestamp` as a reserved keyword, thus it would not appear as an event arg name

### How to verify it

Validate my intution is true re: reserved keyword

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -295,6 +295,10 @@ class ContractLog(BaseContractLog):
         return self.chain_manager.blocks[self.block_number]
 
     @property
+    def timestamp(self) -> int:
+        return self.block.timestamp
+
+    @property
     def _event_args_str(self) -> str:
         return " ".join(f"{key}={val}" for key, val in self.event_arguments.items())
 
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -296,6 +296,11 @@ class ContractLog(BaseContractLog):
 
     @property
     def timestamp(self) -> int:
+        """
+        The UNIX timestamp of when the event was emitted.
+
+        NOTE: This performs a block lookup.
+        """
         return self.block.timestamp
 
     @property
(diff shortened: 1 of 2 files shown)


src/ape/types/__init__.py
```diff
@@ -280,6 +280,10 @@ def validate_address(cls, value):
     def block(self) -> "BlockAPI":
         return self.chain_manager.blocks[self.block_number]
 
+    @property
+    def timestamp(self) -> int:
```
antazoey, line 284: keep in mind, if anyone was relying on the `__getattr__` approach to checking for a custom event property `timestamp`, this will break their stuff. We have work arounds in place for those scenarios of course, but I am always weary about adding properties to our classes that override `__getattr__`.

antazoey (approved):

src/ape/types/__init__.py
fubuloubu: As noted in my PR note, I don't think this is possible because `timestamp` is a reserved keyword

fubuloubu: Originally, I was hoping to allow `<field>.<member>` be a supported column name (which would do `__getattr_(field)` and then `__getattr__(member)` on that)

antazoey: Originally? lol

fubuloubu: More generally I guess haha, I don't think this works (or works well) today

antazoey: I am still good with this PR though.
Though it is a good data science problem: How often is `timestamp` used? What are the chances of breaking existing ape projects? Probably pretty low, and in that case they'd find the PR, hopefully see this comment, and see that you can do this as a work-around

```python
actual_timestamp = log.event_arguments["timestamp"]
assert something with actual_timestamp
```

fubuloubu: codeslaw.app might help us here

fubuloubu: I don't think there are many examples of this

antazoey (approved):

fubuloubu:
still needs tests and a docstring

antazoey (approved):

Issue #1604, opened 2023-08-14
Failed to a decode NodeOperatorAdded log (string field?) [APE-1289]


arwer13:
### Environment information

- OS: macOS
- Python Version: 3.10.12
- `ape` and plugin versions:

```sh
ape --version
0.6.15

ape plugins list
Installed Plugins:
  infura       0.6.3
  etherscan    0.6.8
  solidity     0.6.7
  ganache      0.6.8
  hardhat      0.6.12
  foundry      0.6.14
```

### What went wrong?

Failed to decode a log with known ABI. See event number 136 in https://etherscan.io/tx/0xa4629245311d93a11cedb9143d8b7530057685b4b568a026bac194e162002c13#eventlog

```python
from ape import chain
from ethpm_types.abi import EventABI, EventABIType

operator_added_log = {'address': '0x55032650b14df07b85bF18A3a3eC8E0Af2e028d5',
 'blockHash': HexBytes('0x5f4a2d91be2fbd7c740e279a8061c36d65214e6d3e993e61ebf8f1448c7656b3'),
 'blockNumber': 17885248,
 'data': HexBytes('0x000000000000000000000000000000000000000000000000000000000000001e00000000000000000000000000000000000000000000000000000000000000800000000000000000000000005a8b929edbf3ce44526465dd2087ec7efb59a5610000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b4c61756e63686e6f646573'),
 'logIndex': 10,
 'removed': False,
 'topics': [HexBytes('0xc52ec0ad7872dae440d886040390c13677df7bf3cca136d8d81e5e5e7dd62ff1')],
 'transactionHash': HexBytes('0xa328d01f475108ee2ae271475491a48dadb68ebf6c749277d78336206939f63e'),
 'transactionIndex': 0}

event_abi = EventABI(type='event', name='NodeOperatorAdded', inputs=[EventABIType(name='nodeOperatorId', type='uint256', components=None, internalType=None, indexed=False), EventABIType(name='name', type='string', components=None, internalType=None, indexed=False), EventABIType(name='rewardAddress', type='address', components=None, internalType=None, indexed=False), EventABIType(name='stakingLimit', type='uint64', components=None, internalType=None, indexed=False)], anonymous=False)
 
ContractLogContainer(chain.provider.network.ecosystem.decode_logs([operator_added_log], event_abi))
```

```
[...]

PR #1662, opened 2023-09-14
fix: log decoding issues and missing events [APE-1289]

antazoey:
### What I did

fixes: #1604 
Fixes: APE-1289

### How I did it

* Made it so the events that did not decode would still show up as raw.
So `NodeOperatorAdded` is not still decoding for the tx, going to figure out why exactly before taking this PR out of draft.
So far, all this does and adhere to the advice and still show the events on the tx when they fail to decode.

### How to verify it

You should now see events where the event arguments didnt decode right - they are at least on the receipt now and everything else about them works, including the event name and such...
also the warning makes more sense.

Only thing left:

- [x] figure out why node operator added events isnt decoding

btw, am using some events with strings in it and that didnt cause any bug yet.

### Checklist

### DIFF
--- a/src/ape/utils/abi.py
+++ b/src/ape/utils/abi.py
@@ -3,10 +3,12 @@ from dataclasses import make_dataclass
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from eth_abi import decode, grammar
+from eth_abi.exceptions import DecodingError, InsufficientDataBytes
 from eth_utils import decode_hex
 from ethpm_types import HexBytes
 from ethpm_types.abi import ABIType, ConstructorABI, EventABI, EventABIType, MethodABI
 
+from ape.logging import logger
 from ape.types import AddressType
 
 ARRAY_PATTERN = re.compile(r"[(*\w,? )]*\[\d*]")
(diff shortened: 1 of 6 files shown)


fubuloubu (dismissed):

tests/functional/test_contract_event.py
```diff
@@ -305,10 +305,11 @@ def test_filter_events_with_same_abi(
     """
 
     receipt = contract_with_call_depth.emitLogWithSameInterfaceFromMultipleContracts(sender=owner)
```
fubuloubu, line 307: Is `.events` not cached?

fubuloubu (dismissed):

tests/functional/test_contract_event.py
antazoey: oh it is, youre right

src/ape/utils/abi.py
```diff
@@ -344,21 +346,85 @@ def __init__(self, abi: EventABI):
...
+            hex_value = decode_hex(topic_value)
+
+            try:
+                value = decode([abi_type], hex_value)[0]
+            except InsufficientDataBytes as err:
+                warning_message = f"Failed to decode log topic '{self.event_name}'."
+
+                # Try again with strict=False
+                try:
+                    value = decode([abi_type], hex_value, strict=False)[0]
```
antazoey, line 364: So the log in the linked issue succeeds with `strict=False`

tests/functional/utils/test_abi.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_decoding_with_strict(collection, topics, log_data_missing_trailing_zeroes, caplog):
+    """
+    This test is for a time where Alchemy gave us log data when it was missing trailing zeroes.
+    When using strict=False, it was able to properly decode. In this case, in Ape, we warn
+    the user and still proceed to decode the log.
+    """
+    actual = collection.decode(topics, log_data_missing_trailing_zeroes)
+    expected = {
```
antazoey, line 72: @arwer13 it works now!

fubuloubu (approved):

PR #1665, opened 2023-09-18
test: caplog fixes [APE-1396]

antazoey:
### What I did

we were getting some intermittent failures in tests because of caplog.
i think i found the reason why but dont know 100%

### How I did it

* removed a call `records.pop()` that was potentially causing issues
* removed unneeded calls to `at_level()` which could potentially hide some INFO based logs on other test s

### How to verify it

im hoping just all passing CI both now and in the future can help identify this as a success.

### Checklist

### DIFF
--- a/tests/functional/test_config.py
+++ b/tests/functional/test_config.py
@@ -1,4 +1,3 @@
-import logging
 from typing import Dict, Union
 
 import pytest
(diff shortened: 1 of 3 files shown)


fubuloubu (approved):

antazoey:
ugh - still failed on the merge-to-main run

PR #1666, opened 2023-09-18
feat: handle when local dependency is already a manifest [APE-1397]

antazoey:
### What I did

I was annoyed when I saw ape-safe had to manually copy the manifest into the packages folder, that seemed so unnecessary.
I made it so if you use a local dependency and it is already a manifest, it just copies it in that way.

### How I did it

Check if is a JSON manifest file and if it, copy it on.

### How to verify it

In ape-safe, you can delete that `mkdir` line if you use this PR and this config:

```yaml
dependencies:
  - name: safe-contracts
    local: ./safe-contracts.json
    version: 1.3.0
```

I can open a draft PR after this

### Checklist


fubuloubu (dismissed):
Just a side note about `ape-safe`, we added NPM dependencies to pivot that repo away from using the local uploaded copy eventually (so we can support more than just v1.3.0)

antazoey:
> Just a side note about `ape-safe`, we added NPM dependencies to pivot that repo away from using the local uploaded copy eventually (so we can support more than just v1.3.0)

Yes, I noticed the NPM thing - we can still definitely go that route. But I wish this _could_ have been the temporary fix.

fubuloubu (dismissed):
I'm good with this change if you want to rebase it

PR #1669, opened 2023-09-20
Move polls

z80dev:
### What I did

Moves the logic for polling events and blocks into the Provider API. These new methods no longer also query historical data, they are only used for gathering events/blocks as they come in.

### How I did it

Moved the logic over, and made it so the necessary contract data for polling events comes in as an argument.
Also removed the querying of historical data, calling these new functions now will only pull future events. The `range_*` methods in the provider implementation can still be used to query for historical data, and if useful this could be added to the ProvidersAPI interface.

### How to verify it

Queried for new contract events manually

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterator, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Any, Iterator, List, Optional, Type, Union
 
 import click
 from eip712.messages import SignableMessage as EIP712SignableMessage
@@ -54,18 +54,21 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         return None
 
     @abstractmethod
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
         """
         Sign a message.
 
         Args:
-          msg (:class:`~ape.types.signatures.SignableMessage`): The message to sign.
+          msg (Any): The message to sign. Account plugins can handle various types of messages.
+            For example, :class:`~ape.accounts.LocalAccount` can handle
+            :class:`~ape.types.signatures.SignableMessage`
             See these
             `docs <https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.messages.SignableMessage>`__  # noqa: E501
             for more type information on this type.
+          **signer_options: Additional kwargs given to the signer to modify the signing operation.
 
         Returns:
-          :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
+            :class:`~ape.types.signatures.MessageSignature` (optional): The signed message.
         """
 
     @abstractmethod
@@ -494,7 +497,7 @@ class ImpersonatedAccount(AccountAPI):
     def address(self) -> AddressType:
         return self.raw_address
 
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
         raise NotImplementedError("This account cannot sign messages")
 
     def sign_transaction(self, txn: TransactionAPI, **kwargs) -> Optional[TransactionAPI]:
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -2,8 +2,9 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterator, List, Optional, Type, Union
 
 import click
-from eip712.messages import SignableMessage as EIP712SignableMessage
+from eip712.messages import EIP712Message, SignableMessage as EIP712SignableMessage
 from eth_account import Account
+from eth_account.messages import encode_defunct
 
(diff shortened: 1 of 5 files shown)


src/ape/api/providers.py
```diff
@@ -1591,6 +1625,283 @@ def _handle_execution_reverted(
...
+            columns=list(ContractLog.__fields__.keys()),
+            contract=addresses,
+            event=self.abi,
+            search_topics=search_topics,
+            start_block=start_block,
+            stop_block=stop_block,
+        )
+        yield from self.query_manager.query(contract_event_query)  # type: ignore
+
+    def poll_logs(
```
fubuloubu, line 1877: Technically, for web3py at least, there's a better implementation of this using log filters

src/ape/api/providers.py
```diff
@@ -1591,6 +1625,283 @@ def _handle_execution_reverted(
...
+            columns=list(self.head.__fields__),  # TODO: fetch the block fields from EcosystemAPI
+            start_block=start,
+            stop_block=stop - 1,
+            step=step,
+        )
+
+        blocks = self.query_manager.query(query, engine_to_use=engine_to_use)
+        yield from cast(Iterator[BlockAPI], blocks)
+
+    def poll_blocks(
```
fubuloubu, line 1696: Technically, for web3py at least, there's a better implementation of this using `eth_newBlockFilter`:
https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.filter

fubuloubu (commented):
I think you are basing this off your sign_message work?

src/ape/api/accounts.py
```diff
@@ -54,18 +56,21 @@ def alias(self) -> Optional[str]:
         return None
 
     @abstractmethod
-    def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
```
antazoey, line 59: Are the changed related to signing messages supposed to be on this PR or are those part of https://github.com/ApeWorX/ape/pull/1614?

src/ape/api/accounts.py
```diff
@@ -273,6 +278,10 @@ def check_signature(
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
         """
+        if isinstance(data, str):
+            data = encode_defunct(text=data)
+        if isinstance(data, EIP712Message):
```
antazoey, line 283: these `if` should be `elif`s probably

src/ape/api/accounts.py
```diff
@@ -258,7 +263,7 @@ def declare(self, contract: "ContractContainer", *args, **kwargs) -> ReceiptAPI:
 
     def check_signature(
         self,
-        data: Union[SignableMessage, TransactionAPI],
+        data: Union[SignableMessage, TransactionAPI, str, EIP712Message],
```
antazoey, line 266: should be bytes (hex bytes) also be handled?

src/ape_accounts/accounts.py
```diff
@@ -119,11 +121,21 @@ def delete(self):
...
+    def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
+        if not isinstance(msg, (SignableMessage, EIP712Message, str)):
+            logger.warning("Unsupported message type, (type=%r, msg=%r)", type(msg), msg)
             return None
 
+        user_approves = self.__autosign or click.confirm(f"Message: {msg}\n\nSign: ")
+        if isinstance(msg, str):
+            # Convert str to SignableMessage for handling below
+            msg = encode_defunct(text=msg)
+        elif isinstance(msg, EIP712Message):
```
antazoey, line 133: future design idea: I wonder if we should have our conversion system also supply hooks for signing messages, like we do for tx but for messages.

then we could literally build an eip712 plugin and wouldnt need to add these types every in core.

for example, what if starknet or something offers a whole new spec for message signing, we could also implement conversion hooks there...

tests/functional/test_accounts.py
```diff
@@ -50,6 +50,12 @@ def test_sign_message(signer):
     assert signer.check_signature(message, signature)
 
 
+def test_sign_message_unsupported_type_returns_none(signer):
+    message = 1234
```
antazoey, line 54: why not allow signing integers?

antazoey (commented):
lookin good, but please address the linting and commit messages problems.
also i am not sure why the tests are failing so you may need to investigate that as well.

antazoey:
`mypy` is failing because `setuptools-scm` updated and it is failing to generate the version file.
I tried fixing it all that I could but I am left pinning it <8

fubuloubu:
> `mypy` is failing because `setuptools-scm` updated and it is failing to generate the version file. I tried fixing it all that I could but I am left pinning it <8

just merged a fix for this, @z80dev please rebase

src/ape/api/accounts.py
fubuloubu: I just don't think this should be a part of the diff for this refactor

src/ape/api/accounts.py
```diff
@@ -258,7 +260,7 @@ def declare(self, contract: "ContractContainer", *args, **kwargs) -> ReceiptAPI:
...
```
antazoey, line 263: we can rebase such that this is not in the diff?

PR #1670, opened 2023-09-20
refactor: move web3 loggers and fix test failures [APE-1404]

antazoey:
### What I did

* Moves web3 loggers to web3 provider
* Fixes a type on a query validation method (was receiving tuples here even tho it says list, but really any sequence should work)
* Soft-rename (non-breaking) of the logger class
* Issue with plugin installs when given dir, only noticed in a xfail test
* create a testing assert log fixtures for our tests here so operations can be retried, this should mitigate the xdist related issues with caplog hopefully
* 
### How to verify it

gimme a passing build on merge to main

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0"]
+requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0,<8"]
 
 [tool.mypy]
 exclude = ["build/", "dist/", "docs/", "tests/integration/cli/projects/"]
--- a/src/ape/logging.py
+++ b/src/ape/logging.py
@@ -104,13 +104,12 @@ class ClickHandler(logging.Handler):
 
 class CliLogger:
     _mentioned_verbosity_option = False
+    _extra_loggers: Dict[str, logging.Logger] = {}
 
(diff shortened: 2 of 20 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0"]
+requires = ["setuptools>=51.1.1", "wheel", "setuptools_scm[toml]>=5.0,<8"]
```
antazoey, line 2: It seems like setuptools-scm 8.0.1 no longer creates the version file. It is possibly a bug though.
I am suspecting it was introduced by this PR: https://github.com/pypa/setuptools_scm/pull/870

Note: I have tried renaming to the new config key `version_file` but that also did not work.

For now, we have to pin to `<8` until we can figure out why or how to fix this.

fubuloubu: Yeah, vyper ran into this too

src/ape/logging.py
```diff
@@ -262,7 +258,10 @@ def _get_level(level: Optional[Union[str, int]] = None) -> str:
     return level
 
 
-logger = CliLogger.create()
+logger = ApeLogger.create()
+
+# TODO: Can remove this type alias after 0.7
```
fubuloubu, line 263: Okay this works

fubuloubu (approved):

PR #1673, opened 2023-09-25
feat!: rewrite poll methods with new API [APE-1411]

z80dev:
### What I did

Implemented new poll_blocks and poll_events methods in the ProviderAPI

fixes: #1669 

### How I did it

Used the new Web3.py filter objects for polling of new blocks and events. The event polling uses the block polling for confirmations so as to not repeat that logic in both functions.

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -635,14 +635,17 @@ class ProviderAPI(BaseInterfaceModel):
         stop_block: Optional[int] = None,
         required_confirmations: Optional[int] = None,
         new_block_timeout: Optional[int] = None,
-    ) -> Iterator[BlockAPI]: # type: ignore[empty-body]
+    ) -> Iterator[BlockAPI]:  # type: ignore[empty-body]
         """
         Poll new blocks.
+
         **NOTE**: When a chain reorganization occurs, this method logs an error and
         yields the missed blocks, even if they were previously yielded with different
         block numbers.
+
         **NOTE**: This is a daemon method; it does not terminate unless an exception occurs
         or a ``stop`` is given.
+
         Args:
             start_block (Optional[int]): The block number to start with. Defaults to the pending
               block number.
@@ -654,8 +657,42 @@ class ProviderAPI(BaseInterfaceModel):
             new_block_timeout (Optional[float]): The amount of time to wait for a new block before
               timing out. Defaults to 10 seconds for local networks or ``50 * block_time`` for live
               networks.
+
         Returns:
             Iterator[:class:`~ape.api.providers.BlockAPI`]
+        """
+
+    @raises_not_implemented
+    def poll_logs(  # type: ignore[empty-body]
+        self,
+        stop_block: Optional[int] = None,
+        address: Optional[AddressType] = None,
+        topics: Optional[List[Union[str, List[str]]]] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[ContractLog]:  # type: ignore[empty-body]
+        """
+        Poll new blocks. Optionally set a start block to include historical blocks.
+
+        **NOTE**: This is a daemon method; it does not terminate unless an exception occurrs.
+
+        Usage example::
+
+            for new_log in contract.MyEvent.poll_logs():
+                print(f"New event log found: block_number={new_log.block_number}")
+
+        Args:
+            stop_block (Optional[int]): Optionally set a future block number to stop at.
+              Defaults to never-ending.
+            required_confirmations (Optional[int]): The amount of confirmations to wait
+              before yielding the block. The more confirmations, the less likely a reorg will occur.
+              Defaults to the network's configured required confirmations.
+            new_block_timeout (Optional[int]): The amount of time to wait for a new block before
+              quitting. Defaults to 10 seconds for local networks or ``50 * block_time`` for live
+              networks.
+
+        Returns:
+            Iterator[:class:`~ape.types.ContractLog`]
         """
 
     @raises_not_implemented
@@ -1387,7 +1424,12 @@ class Web3Provider(ProviderAPI, ABC):
                     stop_block,
                 )
 
-    def poll_blocks(self, stop_block: Optional[int] = None, required_confirmations: Optional[int] = None, new_block_timeout: Optional[int] = None) -> Iterator[BlockAPI]:
+    def poll_blocks(
+        self,
+        stop_block: Optional[int] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]:
         filter = self.web3.eth.filter("latest")
         network_name = self.network.name
         block_time = self.network.block_time
@@ -1407,7 +1449,11 @@ class Web3Provider(ProviderAPI, ABC):
         last_yielded_height = None
 
         while True:
-            if stop_block is not None and last_yielded_height is not None and last_yielded_height >= stop_block:
+            if (
+                stop_block is not None
+                and last_yielded_height is not None
+                and last_yielded_height >= stop_block
+            ):
                 break
             changes = self.web3.eth.get_filter_changes(filter.filter_id)
             for new_block_hash in changes:
@@ -1416,9 +1462,15 @@ class Web3Provider(ProviderAPI, ABC):
                 if last_yielded_height and confirmed_block_number < last_yielded_height:
                     num_blocks_behind = last_yielded_height - confirmed_block_number
                     if num_blocks_behind > required_confirmations:
-                        logger.error(f"{num_blocks_behind} Block reorganization detected. Try adjusting the required network confirmations")
+                        logger.error(
+                            f"{num_blocks_behind} Block reorganization detected. "
+                            + "Try adjusting the required network confirmations"
+                        )
                     else:
-                        logger.warning(f"{num_blocks_behind} Block reorganization detected. Reorg is within the required network confirmations")
+                        logger.warning(
+                            f"{num_blocks_behind} Block reorganization detected. "
+                            + "Reorg is within the required network confirmations"
+                        )
                         last_yielded_height = confirmed_block_number
                         continue
                 confirmed_block = self.web3.eth.get_block(confirmed_block_number)
@@ -1428,34 +1480,38 @@ class Web3Provider(ProviderAPI, ABC):
                         raise RuntimeError(
                             f"Timed out waiting for block {confirmed_block_number} to be available."
                         )
-                    logger.warning(
-                        f"Block {confirmed_block_number} not found. Waiting 1 seconds."
-                    )
+                    logger.warning(f"Block {confirmed_block_number} not found. Waiting 1 seconds.")
                     time.sleep(1)
                     confirmed_block = self.web3.eth.get_block(confirmed_block_number)
                 if last_yielded_height and confirmed_block.number < last_yielded_height:
                     num_blocks_behind = last_yielded_height - confirmed_block.number
-                    logger.error(f"{num_blocks_behind} Block reorganization detected. Try adjusting the required network confirmations")
+                    logger.error(
+                        f"{num_blocks_behind} Block reorganization detected. "
+                        + "Try adjusting the required network confirmations"
+                    )
                 last_yielded_height = confirmed_block.number
                 yield self.network.ecosystem.decode_block(confirmed_block)
 
     def poll_logs(
-            self,
-            stop_block: Optional[int] = None,
-            address: Optional[AddressType] = None,
-            topics: Optional[List[Union[str, List[str]]]] = None,
-            required_confirmations: Optional[int] = None,
-            new_block_timeout: Optional[int] = None) -> Iterator[ContractLog]:
+        self,
+        stop_block: Optional[int] = None,
+        address: Optional[AddressType] = None,
+        topics: Optional[List[Union[str, List[str]]]] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[ContractLog]:
         required_confirmations = (
             required_confirmations or self.provider.network.required_confirmations
         )
         for block in self.poll_blocks(stop_block, required_confirmations, new_block_timeout):
-            for log in self.web3.eth.get_logs({
-                "fromBlock": block.number,
-                "toBlock": block.number,
-                "address": address,
-                "topics": topics
-            }):
+            for log in self.web3.eth.get_logs(
+                {
+                    "fromBlock": block.number,
+                    "toBlock": block.number,
+                    "address": address,
+                    "topics": topics,
+                }
+            ):
                 yield ContractLog.parse_obj(log)
 
     def block_ranges(self, start=0, stop=None, page=None):
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -1477,7 +1477,7 @@ class Web3Provider(ProviderAPI, ABC):
                 start_time = time.time()
                 while confirmed_block is None:
                     if time.time() - start_time > timeout:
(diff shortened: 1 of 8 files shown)


src/ape/api/providers.py
```diff
@@ -598,6 +598,35 @@ def get_transaction_trace(  # type: ignore[empty-body]
...
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]: # type: ignore[empty-body]
+        """
+        Poll new blocks.
+        **NOTE**: When a chain reorganization occurs, this method logs an error and
+        yields the missed blocks, even if they were previously yielded with different
+        block numbers.
+        **NOTE**: This is a daemon method; it does not terminate unless an exception occurs
+        or a ``stop`` is given.
+        Args:
```
antazoey, line 615: you need newlines in this doc for all this to render properly.
new lines before all the little sections.

(i know when you copy+paste docs sometimes, it will get rid of these on accident, that may be what happened here)

src/ape/api/providers.py
```diff
@@ -1316,6 +1345,76 @@ def _find_txn_by_account_and_nonce(
                     stop_block,
                 )
 
+    def poll_blocks(self, stop_block: Optional[int] = None, required_confirmations: Optional[int] = None, new_block_timeout: Optional[int] = None) -> Iterator[BlockAPI]:
```
antazoey, line 1348: lint

src/ape/api/providers.py
```diff
@@ -1316,6 +1345,76 @@ def _find_txn_by_account_and_nonce(
...
+                    logger.warning(
+                        f"Block {confirmed_block_number} not found. Waiting 1 seconds."
+                    )
+                    time.sleep(1)
+                    confirmed_block = self.web3.eth.get_block(confirmed_block_number)
+                if last_yielded_height and confirmed_block.number < last_yielded_height:
+                    num_blocks_behind = last_yielded_height - confirmed_block.number
+                    logger.error(f"{num_blocks_behind} Block reorganization detected. Try adjusting the required network confirmations")
+                last_yielded_height = confirmed_block.number
+                yield confirmed_block
```
antazoey, line 1398: you might need to something like `self.network.ecosystem.decode_block(confirmed_block)`

src/ape/api/providers.py
antazoey, line 1348: i believe we are going to need to update the `ChainManager` methods (or the `QueryManager` methods, or both) to delegate to these new methods

src/ape/api/providers.py
```diff
@@ -1316,6 +1345,76 @@ def _find_txn_by_account_and_nonce(
...
+            stop_block: Optional[int] = None,
+            address: Optional[AddressType] = None,
+            topics: Optional[List[Union[str, List[str]]]] = None,
+            required_confirmations: Optional[int] = None,
+            new_block_timeout: Optional[int] = None) -> Iterator[ContractLog]:
+        required_confirmations = (
+            required_confirmations or self.provider.network.required_confirmations
+        )
+        for block in self.poll_blocks(stop_block, required_confirmations, new_block_timeout):
+            yield from self.web3.eth.get_logs({
```
antazoey, line 1411:
```suggestion
            for log in self.web3.eth.get_logs({
            ...
            yield ContractLog.parse_obj(log)
```

something like that?

src/ape/contracts/base.py
```diff
@@ -691,6 +691,14 @@ def poll_logs(
...
+        # NOTE: Now we process the rest
+        yield from self.provider.poll_logs(
+            stop_block=stop_block,
+            address=self.contract.address,
+            required_confirmations=required_confirmations,
+            new_block_timeout=new_block_timeout,
+        )
+
         # NOTE: Now we process the rest
         for new_block in self.chain_manager.blocks.poll_blocks(
```
fubuloubu, line 703: can remove this and below I think

src/ape/api/providers.py
```diff
@@ -598,6 +598,72 @@ def get_transaction_trace(  # type: ignore[empty-body]
             Iterator(:class:`~ape.type.trace.TraceFrame`): Transaction execution trace.
         """
 
+    @raises_not_implemented
+    def poll_blocks(  # type: ignore[empty-body]
+        self,
+        stop_block: Optional[int] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]:  # type: ignore[empty-body]
```
antazoey, line 607: only the top type ignore (line 602) is needed i believe

```suggestion
    ) -> Iterator[BlockAPI]: 
```

src/ape/api/providers.py
```diff
@@ -1316,6 +1382,96 @@ def _find_txn_by_account_and_nonce(
...
+                            f"{num_blocks_behind} Block reorganization detected. "
+                            + "Reorg is within the required network confirmations"
+                        )
+                        last_yielded_height = confirmed_block_number
+                        continue
+                confirmed_block = self.web3.eth.get_block(confirmed_block_number)
+                start_time = time.time()
+                while confirmed_block is None:
+                    if time.time() - start_time > timeout:
+                        raise RuntimeError(
```
antazoey, line 1438:
```suggestion
                        raise ProviderError(
```

src/ape/api/providers.py
```diff
@@ -1316,6 +1382,96 @@ def _find_txn_by_account_and_nonce(
...
+            if (
+                stop_block is not None
+                and last_yielded_height is not None
+                and last_yielded_height >= stop_block
+            ):
+                break
+            changes = self.web3.eth.get_filter_changes(filter.filter_id)
+            for new_block_hash in changes:
+                block = self.web3.eth.get_block(new_block_hash)
+                confirmed_block_number = block.number - required_confirmations
```
antazoey, line 1419: Are we able to adjust the filter to account for the confs? Right now, it is set to `"latest"`, but could we refactor such that we account for the confs at the filter level and then just yield from there.

I think that will prevent some unnecessary `get_block()` requests.

src/ape/api/providers.py
```diff
@@ -1316,6 +1382,96 @@ def _find_txn_by_account_and_nonce(
                     stop_block,
                 )
 
+    def poll_blocks(
+        self,
+        stop_block: Optional[int] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]:
+        filter = self.web3.eth.filter("latest")
```
antazoey, line 1391: Are we able to accounts for the confs here, at the filter level?

antazoey (commented):
I've a question re: confirmations at the filter level!

src/ape/api/providers.py
z80dev: Updated ChainManager's poll methods to use the provider poll methods for non-historical data

src/ape/api/providers.py
z80dev: I also explored ways of doing this. The options web3.py gives us are:

- 'latest'
- 'pending'
- 'earliest'
- 'safe'
- 'finalized'

it sounds like 'safe' or 'finalized' would be what we want. Unfortunately, web3.py doesn't expose any way of customizing what 'safe' or 'finalized' means. The docs also say:

> Though `"latest"` and `"safe"` block identifiers are not yet part of the specifications for `eth_newFilter`, they are supported by web3.py and may or may not yield expected results depending on the node being accessed.

we can maybe specify a block number and re-create a new filter each time we're going to fetch a new set of blocks. This would require us only pinging the node for the latest block number, subtracting our number of required confs, and fetching that old block. 

Currently, we poll each block twice: once when its the newest block, and once to get its number. This is because when we poll for blocks, we only get a blockhash back, not a block number. In order to get the block number, we fetch this block by its hash.

We should be able to use `eth_getBlockNumber` and reduce this double-fetching of blocks, but it is not compatible with our current block polling. Instead of being notified of a new block, we would loop and call `eth_getBlockNumber` until the number increases.

antazoey: > we can maybe specify a block number and re-create a new filter each time we're going to fetch a new set of blocks. This would require us only pinging the node for the latest block number, subtracting our number of required confs, and fetching that old block.

This is kinda what i was imagining

> We should be able to use eth_getBlockNumber and reduce this double-fetching of blocks, but it is not compatible with our current block polling. 

Keep in mind if you in Web3Provider, you are allowed to assume EVM like behavior and RPCs (provided they are standard enough)

z80dev: Implemented polling logic the way you suggested here

src/ape/api/providers.py
```diff
@@ -598,6 +598,76 @@ def get_transaction_trace(  # type: ignore[empty-body]
...
+    ) -> Iterator[BlockAPI]:
+        """
+        Poll new blocks.
+
+        **NOTE**: When a chain reorganization occurs, this method logs an error and
+        yields the missed blocks, even if they were previously yielded with different
+        block numbers.
+
+        **NOTE**: This is a daemon method; it does not terminate unless an exception occurs
+        or a ``stop`` is given.
```
antazoey, line 616:
```suggestion
        or a ``stop_block`` is given.
```

src/ape/api/providers.py
```diff
@@ -598,6 +598,76 @@ def get_transaction_trace(  # type: ignore[empty-body]
...
+
+        Args:
+            start_block (Optional[int]): The block number to start with. Defaults to the pending
+              block number.
+            stop_block (Optional[int]): Optionally set a future block number to stop at.
```
antazoey, line 621: **note for myself as the reviewer**: so the doc here makes it sound like it's inclusive, meaning the you enter the last block you want here and not the end_range value. This is how `poll_blocks` worked before because only the method named `range()` behave like the python builtin `range()` method

src/ape/api/providers.py
```diff
@@ -1316,6 +1386,109 @@ def _find_txn_by_account_and_nonce(
...
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]:
+        network_name = self.network.name
+        block_time = self.network.block_time
+        wait_time = block_time / 2
+        timeout = (
+            (
+                10.0
+                if network_name == LOCAL_NETWORK_NAME or network_name.endswith("-fork")
```
antazoey, line 1401: **note for future work or myself**: At this point, we should have an `is_local` property on `NetworkAPI`.

src/ape/api/providers.py
```diff
@@ -1316,6 +1386,109 @@ def _find_txn_by_account_and_nonce(
...
+                stop_block is not None
+                and last_yielded_height is not None
+                and last_yielded_height >= stop_block
+            ):
+                break
+            latest_block = self.web3.eth.block_number
+            if last_yielded_height is None:
+                last_yielded_height = latest_block - required_confirmations
+            elif latest_block - required_confirmations <= last_yielded_height and reorg_height is None:
+                print("continuing")
```
antazoey, line 1426: **TODO**: Replace with a more proper `logger.debug()` line or remove, or something else

src/ape/api/providers.py
```diff
@@ -1316,6 +1386,109 @@ def _find_txn_by_account_and_nonce(
...
+        while True:
+            if (
+                stop_block is not None
+                and last_yielded_height is not None
+                and last_yielded_height >= stop_block
+            ):
+                break
+            latest_block = self.web3.eth.block_number
+            if last_yielded_height is None:
+                last_yielded_height = latest_block - required_confirmations
```
antazoey, line 1424: **warning**: potentially dangerous to set a variable implying something was yield _before_ anything got yielded.. the actually yielding takes places much later. I can forsee an issue with this but I don't know if that is the problem or not

src/ape/api/providers.py
```diff
@@ -1316,6 +1386,109 @@ def _find_txn_by_account_and_nonce(
...
+            elif latest_block - required_confirmations <= last_yielded_height and reorg_height is None:
+                print("continuing")
+                time.sleep(wait_time)
+                continue
+            else:
+                for block_num in range(
+                    last_yielded_height + 1, latest_block - required_confirmations + 1
```
antazoey, line 1431: **confusion**: We are ranging past last yielded height but we never actually yield anything yet.

src/ape/api/providers.py
```diff
@@ -1316,6 +1386,109 @@ def _find_txn_by_account_and_nonce(
...
+                ):
+                    confirmed_block = self.web3.eth.get_block(block_num)
+                    if reorg_height and confirmed_block and confirmed_block.number == reorg_height:
```
antazoey, line 1434: **question or nit**: maybe just better to explicitly check for `not None` here for `reorg_height` and `confirmed_block` rather than a truthy test?

There are other similar instances below

antazoey (commented):
Adding these comments in case they help but I am going to debug now.

src/ape/managers/chain.py
```diff
@@ -271,18 +269,6 @@ def poll_blocks(
...
-                10.0
-                if network_name == LOCAL_NETWORK_NAME or network_name.endswith("-fork")
-                else 50 * block_time
-            )
-            if new_block_timeout is None
-            else new_block_timeout
-        )
-
         if required_confirmations is None:
             required_confirmations = self.network_confirmations
```
antazoey, line 274: **todo**: On line 279, `block = None`. You can delete it - it is no longer necessary to declare ahead of time like that.

src/ape/contracts/base.py
```diff
@@ -699,17 +699,12 @@ def poll_logs(
             start_block = height + 1
```
antazoey, line 699: Oh, is line 691 a problem? It seems like would prevent `0` confirmations from working.

src/ape/api/providers.py
antazoey: https://github.com/ApeWorX/ape/pull/1730

src/ape/api/providers.py
antazoey: * note, added comments and renames and things are little clearer now

src/ape/api/providers.py
antazoey: did `not None` everywhere now

src/ape/api/providers.py
```diff
@@ -1362,6 +1432,127 @@ def _find_txn_by_account_and_nonce(
...
+        stop_block: Optional[int] = None,
+        required_confirmations: Optional[int] = None,
+        new_block_timeout: Optional[int] = None,
+    ) -> Iterator[BlockAPI]:
+        # Wait half the time as the block time
+        # to get data faster.
+        block_time = self.network.block_time
+        wait_time = block_time / 2
+
+        # The timeout occurs when there is chain activity
```
antazoey, line 1446:
```suggestion
        # The timeout occurs when there is no chain activity
```

src/ape/api/providers.py
```diff
@@ -1362,6 +1432,127 @@ def _find_txn_by_account_and_nonce(
...
+            time_waiting = time.time() - last.time
+            if time_waiting > timeout:
+                raise ProviderError("Timed out waiting for next block.")
+
+        # Begin the daemon.
+        while True:
+            # The next block we want is simply 1 after the last.
+            next_block = last.number + 1
+
+            # Use an "adjused" head, based on the required confirmations.
```
antazoey, line 1481: this comment i meant to put above `adjusted_head = `

src/ape/api/providers.py
```diff
@@ -1362,6 +1432,127 @@ def _find_txn_by_account_and_nonce(
...
+            head = self.get_block("latest")
+
+            try:
+                adjusted_head = self.get_block(head.number - required_confirmations)
+            except Exception:
+                # TODO: I did encounter this sometimes in a re-org, needs better handling
```
antazoey, line 1487: @z80dev I hit this sometimes in reorg in some race condition, like a block-not-found error of some kind, where it reorgs in between getting the head and adjusted head

antazoey (approved):

PR #1675, opened 2023-09-26
refactor: BaseModel getattr logic sharing [APE-1415]

antazoey:
### What I did

So as I am working on some pydantic v2 refactor stuff for !0.7, I am noticing some other smaller refactors that would really make it a lot easier, stuff we can do now... stuff we are likely going to have to do anyway...

So this is one of those things.
Right now, we have some weird logic for __getattr__ and everyone is kinda wildwildwesting it but for the most part have to keep in mind the same gotchas. I am finding that v2 of pydantic has even more gotchas in this regard. So, we can isolated those to one place with this new system of allowing getattr / getitem overloads.

### How I did it

A new method in BaseModel `__ape_extra_attributes__` that returns an iterator of this new class containing the attributes, some metadata, and what parts to use it in. It also has a handy method for checking such attributes.
So the models that need custom __getattr__, e.g. events and projects and stuff, they yield some of these bad girls and off we go, no more needing to call `__getattribute__ and being all funky there.

Examples:

```python
    def __ape_extra_attributes__(self) -> Iterator[ExtraModelAttributes]:
        yield ExtraModelAttributes(
            name=self.name,
            attributes=self.contracts,
            include_getattr=True,
            include_getitem=True,
        )


    def __ape_extra_attributes__(self) -> Iterator[ExtraModelAttributes]:
        yield ExtraModelAttributes(
            name=self.event_name,
            attributes=self.event_arguments,
            include_getattr=True,
            include_getitem=True,
        )

```

### Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,6 +5,7 @@ exclude =
 	docs
 	build
 	.eggs
+	tests/integration/cli/projects
 per-file-ignores =
     # Need signal handler before imports
     src/ape/__init__.py: E402
--- a/src/ape/api/config.py
+++ b/src/ape/api/config.py
@@ -40,7 +40,7 @@ class PluginConfig(BaseSettings):
         return cls(**update(default_values, overrides))
 
     def __getattr__(self, attr_name: str) -> Any:
-        # allow hyphens in plugin config files
+        # Allow hyphens in plugin config files.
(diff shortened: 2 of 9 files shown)


fubuloubu (dismissed):
Very interesting solution, I like the code de-dup

fubuloubu (approved):

PR #1676, opened 2023-09-26
refactor: move transaction convert methods to conversion manager [APE-1416]

antazoey:
### What I did

Moves `_convert_args` and `_convert_kwargs` to the `ConversionManager`.

Reasoning: The `ConversionManager` is just dying for my tasks and these methods are awkwardly placed and they require passing the conversion manager around anyway, so it jus makes sense to have them be in the ConversionManager

Also, does other small typing related changes to help make the diff for Pydantic v2 much smaller.

### How I did it

moved the methods

### How to verify it

things still work

### Checklist

### DIFF
--- a/src/ape/api/query.py
+++ b/src/ape/api/query.py
@@ -94,7 +94,7 @@ def extract_fields(item, columns):
 
 
 class _BaseQuery(BaseModel):
-    columns: List[str]
+    columns: Sequence[str]
 
     # TODO: Support "*" from getting the EcosystemAPI fields
 
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -1,5 +1,6 @@
 import shutil
 from pathlib import Path
+from urllib.parse import urlparse
 
 import pytest
 import yaml
@@ -183,7 +184,10 @@ def test_meta(temp_config, project):
         assert project.meta.license == "MIT"
         assert project.meta.description == "test"
         assert project.meta.keywords == ["testing"]
-        assert str(project.meta.links["apeworx.io"]).startswith("https://apeworx.io")
+
+        actual_url = urlparse(project.meta.links["apeworx.io"])
+        assert actual_url.hostname == "apeworx.io"
+        assert actual_url.scheme == "https"
 
 
 def test_brownie_project_configure(config, base_projects_directory):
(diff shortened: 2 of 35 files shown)


tests/functional/test_project.py
```diff
@@ -183,7 +183,7 @@
         assert project.meta.license == "MIT"
         assert project.meta.description == "test"
         assert project.meta.keywords == ["testing"]
-        assert "https://apeworx.io" in project.meta.links["apeworx.io"]
+        assert str(project.meta.links["apeworx.io"]).startswith("https://apeworx.io")
```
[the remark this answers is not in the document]
antazoey: Should be addressed here [54ae3ae](https://github.com/ApeWorX/ape/pull/1676/commits/54ae3aec28eeef4f8c478d38a4ea22e596e037e9)

src/ape/api/query.py
```diff
@@ -94,7 +94,7 @@ def extract_fields(item, columns):
 
 
 class _BaseQuery(BaseModel):
-    columns: List[str]
+    columns: Sequence[str]
```
fubuloubu, line 97: Would this work appropriately as a base model?

fubuloubu (approved):

dtdang (approved):

src/ape/api/query.py
antazoey: Yes you can! it is kinda neat. Pydantic will validate it is a sequence but if you give it a list, it will do its normal list validation or if you give it a tuple, it does the normal tuple validation. `mypy` handles the rest in how it gets used and stuff.

Note: we were already passing tuples in here, but they were getting unnecessary converted to lists. Pydantic v2 + mypy told me all about it, even though it is just as much happening now as it is during the refactor (one of those things).

antazoey: https://docs.pydantic.dev/latest/concepts/performance/#sequence-vs-list-or-tuple-mapping-vs-dict

antazoey: You can find examples of Sequence in the v1 docs too like here https://docs.pydantic.dev/1.10/usage/types/#typing-iterables

fubuloubu: Okay great! Just making sure the research was done here 🫡

Issue #1668, opened 2023-09-20
Solidity Custom Error is not showing properly when tx reverts [APE-1402]


n1punp:
### Environment information

- OS: macOS
- Python Version: 3.11.3
- `ape` and plugin versions:

```sh
$ ape --version
0.6.19

$ ape plugins list
Installed Plugins:
  alchemy      0.6.4
  ens          0.6.1
  etherscan    0.6.10
  infura       0.6.3
  template     0.6.1
  foundry      0.6.16
  vyper        0.6.11
  arbitrum     0.6.2
  tokens       0.6.2
  hardhat      0.6.13
  solidity     0.6.9
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
name: Test

foundry:
  fork:
    arbitrum:
      mainnet:
        upstream_provider: <URL>
```

### What went wrong?

Test.sol source:

```
pragma solidity 0.8.17;

contract Test {
    error Revert(uint x);
    event Transfer(address indexed spender, address indexed receiver, uint256 amount);

    function test() external {
        revert Revert(1);
    }
}
```

To reproduce:
1. Spin up ape console: `ape console --network arbitrum:mainnet-fork:foundry`
2. Run
```
$ alice = accounts.test_accounts[0]
>
$ test = project.Test.deploy(sender=alice)
>
$ test.test(sender=alice)
> ERROR: (TransactionError) Gas estimation failed: 'intrinsic gas too high'. This transaction will likely revert. If you wish to broadcast, you must set the gas limit manually.
```

Also tried running with
```  
$ test.test(sender=alice, gas_limit=2000000)
> ...long lines of errors...
```

Also tried:
```
$ test.test.call()
```

This produces error message for revert strings (e.g. for `require` revert pattern, but not for custom error). It'll just show `ERROR: (ContractLogicError) Transaction failed.` 

### How can it be fixed?

Decode the RPC response return value (which is already available in the DEBUG verbosity -- just need to find a way to decode + display it). This is what I got from running `test.test.call()` with DEBUG verbosity:

```
...
[...]

PR #1680, opened 2023-09-27
fix: issue with custom errors on estimate gas and static fee txns [APE-1421]

antazoey:
### What I did

fixes: #1668
Fixes: APE-1402

### How I did it

Honor networks gas settings during tx prepartion (only was issue on the Static fee txns)

### How to verify it

Here is a screenshot showing it working, i am using the same example as from the linked issue:

Screenshot 2023-09-27 at 15 31 54

### Checklist


src/ape/api/providers.py
```diff
@@ -1429,6 +1430,12 @@ def prepare_transaction(self, txn: TransactionAPI) -> TransactionAPI:
 
             txn.gas_limit = gas
 
+        elif gas_limit == "max":
+            txn.gas_limit = self.max_gas
```
fubuloubu, line 1434: this is a mild change in behavior?

fubuloubu (approved):

src/ape/api/providers.py
antazoey: without it, i end up in a situation where a txns have none gas, like the validator never runs?
i cant quite figure out why this is happening.
I have added some tests for some ecosystem methods to try and investigate.

As soon as I have a better understanding, I will report back.

I dont think this is really a change in behavior though - more like a bug fix... the problem is that im not sure we should have hit this bug because the gas limit should have gotten set before this point for the local networks. Yet, for some reason in non EIP-1559 networks, it was getting into this state where it skipped paying attention to the network setting....

Ill close this comment with a reasoning as soon as its discovered

tests/functional/test_ecosystem.py
```diff
@@ -307,3 +307,19 @@ def test_decode_return_data_non_empty_padding_bytes(ethereum):
     )
     with pytest.raises(DecodingError):
         ethereum.decode_returndata(abi, raw_data)
+
+
+@pytest.mark.parametrize("type_", TransactionType)
```
fubuloubu, line 312: maybe clearer
```suggestion
@pytest.mark.parametrize("tx_type", TransactionType)
```

fubuloubu (approved):

src/ape/api/providers.py
fubuloubu: okay just wanted to double check that it was a bug fix

antazoey: OK - update: The top-level bug in all the L2 plugins.
You can see the fix here: https://github.com/ApeWorX/ape-arbitrum/pull/20

This fix is still nice because it handles the case where you are calling methods in a weird one (kinda anti-normal flow).
See my new test in `test_provider` to see what i mean, the setup of the test.

However, the biggie here is that I need to update all the L2s that dont use EIP-1559 to handle gas limits appropriately in their `create_transaction()` override.

fubuloubu (approved):

fubuloubu (dismissed):

fubuloubu (approved):

antazoey:
tests pass locally of course

fubuloubu (approved):

PR #1681, opened 2023-09-27
perf: lazily load plugins [APE-1422]

antazoey:
### What I did

This shaves almost a second off of `time ape`. None of the plugins are loaded if they are not needed (but they are loaded if they are needed).

### How I did it

Use iterators mostly.

### Checklist

### DIFF
--- a/src/ape/cli/choices.py
+++ b/src/ape/cli/choices.py
@@ -241,7 +241,7 @@ def get_networks(
     network: _NETWORK_FILTER = None,
     provider: _NETWORK_FILTER = None,
 ) -> LazySequence:
-    # NOTE: Use str-keys and a lru_cache.
+    # NOTE: Use str-keys an lru_cache.
     return _get_networks_sequence_from_cache(
         _network_filter_to_key(ecosystem),
         _network_filter_to_key(network),
--- a/src/ape/cli/choices.py
+++ b/src/ape/cli/choices.py
@@ -9,7 +9,7 @@ from click import Choice, Context, Parameter
 from ape import accounts, networks
 from ape.api.accounts import AccountAPI
 from ape.exceptions import AccountsError
-from ape.types import LazySequence
+from ape.types import _LazySequence
 
 ADHOC_NETWORK_PATTERN = re.compile(r"\w*:\w*:https?://\w*.*")
 
@@ -36,7 +36,7 @@ class Alias(click.Choice):
         # NOTE: we purposely skip the constructor of `Choice`
         self.case_sensitive = False
         self._account_type = account_type
-        self.choices = LazySequence(self._choices_iterator)
+        self.choices = _LazySequence(self._choices_iterator)
 
     @property
     def _choices_iterator(self) -> Iterator[str]:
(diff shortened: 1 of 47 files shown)


src/ape/plugins/__init__.py
```diff
@@ -134,23 +138,16 @@ def __init__(self) -> None:
...
 
-                logger.warn_from_exception(err, f"Error loading plugin package '{module_name}'.")
+        # NOTE: Returns tuple because this shouldn't change.
+        return tuple(installed_plugin_module_names.union(core_plugin_module_names))
 
     def __repr__(self):
         return f"<{self.__class__.__name__}>"
 
     def __getattr__(self, attr_name: str) -> Iterator[Tuple[str, Tuple]]:
+        self._register_plugins()
```
fubuloubu, line 149: add a note here

fubuloubu (approved):

fubuloubu:
seeing about a 12% performance improvement with this on my machine :rocket:

antazoey:
> seeing about a 12% performance improvement with this on my machine 🚀

my goal rn is to not load plugins for any of the helps.
i think `ape ledger` still does because of some of account tools, so im going to fix that now

- [ ] fix account tool lazy stuff

but disclaimer here: if you run any cmd that loads plugins, there shouldnt be any performance gains by this PR alone. in the future, we want to load only plugins needed but thatll likely require a bigger refactor

src/ape/plugins/__init__.py
antazoey: [4361ccc](https://github.com/ApeWorX/ape/pull/1681/commits/4361cccbfa812544baceb1bf657fb21257023805)

antazoey:
> seeing about a 12% performance improvement with this on my machine 🚀

You will see a bigger one on commands like `ape run --help` (sub command helps).

Main: `ape run  4.26s user 0.75s system 121% cpu 4.102 total`
Branch: `ape run  2.45s user 0.54s system 144% cpu 2.071 total`

src/ape/cli/choices.py
```diff
@@ -36,16 +38,19 @@ def __init__(self, account_type: Optional[Type[AccountAPI]] = None):
         self._account_type = account_type
 
     @property
-    def choices(self) -> List[str]:  # type: ignore
+    def choices(self) -> Iterator[str]:  # type: ignore
```
antazoey, line 41: I should maybe use the LazySequence for all these.
We are using iterators here but technically click says they have to be sequences. Things have been working but technically click can assume sequences and our stuff may suddenly be failing. 

Click should consider supporting iterators explicitly, that would make our lives easier. But for now, we have a custom type that matches the Sequence protocol while still being as effective as an iterator in its laziness.

src/ape/types/__init__.py
```diff
@@ -415,6 +418,64 @@ def __contains__(self, val: Any) -> bool:
         return any(log == val for log in self)
 
 
+_T = TypeVar("_T")  # _LazySequence generic.
+
+
+class _LazySequence(Sequence[_T]):
```
antazoey, line 424: at this time, this is only really used in internal stuff in the framework for performance reasons; i tried using `collections.deque()` but it wasnt lazily enough and so i went with this but if there is way to make deque not iterate at all, that would be nice, or an alternative data structure

fubuloubu (approved):

Issue #1688, opened 2023-10-03
multiple abi issues [APE-1432]


banteg:
### Environment information

- OS: macOS
- Python Version: 3.11.4
- `ape` and plugin versions:

```sh
$ ape --version
0.6.21

$ ape plugins list
Installed Plugins
  alchemy      0.6.4
  ens          0.6.1
  etherscan    0.6.10
  foundry      0.6.16
  hardhat      0.6.13
  infura       0.6.3
  solidity     0.6.9
  template     0.6.1
  tokens       0.6.2
  vyper        0.6.11
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
# ...copy and paste result of above command here...
```

### What went wrong?

i found several bugs when trying to use this contract on mainnet

```python
composable_cow = Contract("0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74")
start_block = 17883049
stop_block = 18272118
```

1. incorrect decoding of tuples in logs
```python
print(composable_cow.ConditionalOrderCreated)
# ConditionalOrderCreated(address indexed owner, (address,bytes32,bytes) params)
# notice how params has a tuple type

logs = list(composable_cow.ConditionalOrderCreated.range(start_block, stop_block))
log = logs[0]
print(log)
# ConditionalOrderCreated(owner=0xE7602Ca44f83a5E9Ba8BD14125dDcb295f3D63BD params=[['0x6cf1e9ca41f7611def408122793c358a3d11e5a5', b'snip', b"snip"]])
# bug 1: params got wrapped into double list for no reason
```

2. cannot pass a list to call hander where a tuple is expected
```python
# now let's try to encode the params from the event for a call
# note: changing to log['params'][0] mitigating bug 1 results in the same error
composable_cow.getTradeableOrderWithSignature.encode_input(log['owner'], log['params'], b'', [])
# bug 2: cannot pass a list where a tuple is expected
# AttributeError: 'list' object has no attribute 'handler'
```

3. incorrect implementation of aggegate3 in multicall
```python
call = multicall.Call()
call.add(composable_cow.getTradeableOrderWithSignature, log['owner'], tuple(log['params'][0]), b'', [], allowFailure=True)
next(call())
# bug 3: incorrect unwrapping of result from aggregate3
# AttributeError: 'list' object has no attribute 'returnData'
```

### How can it be fixed?

[...]

PR #1694, opened 2023-10-05
fix: multiple ABI encoding and decoding bug fixes

antazoey:
### What I did

fixes: #1688
Fixes: APE-1432

### Checklist


fubuloubu (dismissed):

src/ape_ethereum/multicall/handlers.py
```diff
@@ -42,23 +50,24 @@ def inject(cls):
             @pytest.fixture(scope="session")
             def use_multicall():
                 # NOTE: use this fixture any test where you want to use a multicall
-                multicall.BaseMulticall.deploy()
+                multicall.BaseMulticall.inject()
```
fubuloubu, line 53:
```suggestion
                return multicall.BaseMulticall.inject()
```

src/ape_geth/provider.py
```diff
@@ -376,6 +383,10 @@ def get_call_tree(self, txn_hash: str) -> CallTreeNode:
         self._can_use_parity_traces = True
         return tree
 
+    def set_code(self, address: AddressType, code: ContractCode) -> bool:
+        # NOTE: Does no work on regular Geth nodes, will still raise APINotImpemented.
+        return self._make_request("debug_setCode", [address, code])
```
fubuloubu, line 388: Does this only work on geth dev?

antazoey: no it doesnt work anywhere, i just deleted

src/ape_ethereum/multicall/handlers.py
```diff
@@ -152,20 +168,43 @@ def add(self, call: ContractMethodHandler, *args, **kwargs):
 
     @property
     def returnData(self) -> List[HexBytes]:
-        if not self._result:
+        result = self._result  # Decl for typing reasons.
+        if not result:
             raise NotExecutedError()
 
-        elif isinstance(self._result, tuple):
```
antazoey, line 158: im not 100% but it seems like it was checking for one type but handling the other type, so i broke it into 2 checks to handle each `Tuple[int, List[HexBytes]]` and `List[Tuple[bool, HexBytes]]` respectively and now each should work, but this is the part i need the most testing help with.

The last case though still is not really annotated in the `_result` type - the one that is a Struct object

src/ape_ethereum/multicall/handlers.py
```diff
@@ -23,14 +25,20 @@
 
 
 class BaseMulticall(ManagerAccessMixin):
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        address: AddressType = MULTICALL3_ADDRESS,
+        supported_chains: List[int] = SUPPORTED_CHAINS,
```
fubuloubu, line 31: Never pass a list as a default arg: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

This could be okay here, but definitely requries comments to ensure nobody updates this constant value

src/ape_ethereum/multicall/handlers.py
antazoey: OK I actually figured out what was going now. the bug only happens when using a single call in a multi-call, the results were already uplifted in a sense, so with some extra handling there, we are good to go and i fixed by mistake comments. 👍🏻

fubuloubu (approved):

PR #1700, opened 2023-10-13
fix: issue when switching providers when disconnecting [APE-1460]

antazoey:
### What I did

fixed bug with Disconnect => reconnect with new settings was bugged when using the same Python session.
also adds `settings` object which is useful for getting all provider settings mashed together.
Also, fixes an issue where settings were not updated when connecting to an already-connected provider with new settings.

**note**: as I was fixing this, I had some random failing tests pop up; bugs highlighted with the networks yaml not actually filtering. I had to fix this in order to get a passing build on this PR.

That caused this PR to be a little out of scope but I don't think it is worth opening more PRs at this point; it is all network related anyway.

### How I did it

Delete provider from connected providers when disconnecting

### How to verify it

in combo with 

https://github.com/ApeWorX/ape-foundry/pull/74

you can do:

```python
In [4]: with networks.parse_network_choice("ethereum:local:foundry", provider_settings={"port": 8999}) as prv:
   ...:     print(prv.uri)
http://127.0.0.1:8999
```

You can even switch multiple times to have many ongoing foundrys:

```python
with networks.parse_network_choice("ethereum:local:foundry", provider_settings={"port": 8999}) as prv:
   ...:     print(prv.uri)
   ...:     with networks.parse_network_choice("ethereum:local:foundry", provider_settings={"port": 8993}) as prv2:
   ...:         print(prv.uri)
   ...:         print(prv2.uri)
```

Outputs:

```
INFO: Connecting to existing 'anvil' process.
http://127.0.0.1:8999
WARNING: `port` setting is deprecated. Please use `host` key that includes the port.
INFO: Starting 'anvil' process.
http://127.0.0.1:8999
http://127.0.0.1:8993

```

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -196,8 +196,8 @@ class ProviderAPI(BaseInterfaceModel):
     @property
     def connection_id(self) -> Optional[str]:
         """
-        A connection ID to uniquely identify and manage multiple 
-        connections to providers, especially when working with multiple 
+        A connection ID to uniquely identify and manage multiple
+        connections to providers, especially when working with multiple
         providers of the same type, like multiple Geth --dev nodes.
         """
 
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -214,7 +214,7 @@ class ProviderAPI(BaseInterfaceModel):
 
         # NOTE: If other provider settings are different, ``.update_settings()``
         #    should be called.
-        return f"{self.network_choice}:-{chain_id}"
+        return f"{self.network_choice}:{chain_id}"
 
     @abstractmethod
     def update_settings(self, new_settings: dict):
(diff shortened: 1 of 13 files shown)


antazoey:
im still working something out in foundry..

src/ape/api/providers.py
```diff
@@ -184,6 +184,38 @@ def ws_uri(self) -> Optional[str]:
...
+            if chain_id := self.settings.get("chain_id"):
+                pass
+
+            else:
+                # A connection is required to obtain a chain ID for this provider.
+                return None
+
+        # NOTE: If other provider settings are different, ``.update_settings()``
+        #    should be called.
+        return f"{self.network_choice}:-{chain_id}"
```
fubuloubu, line 217: I think chain ID won't change if you have two different fork nodes of the same network, only the port changes

antazoey: you can make it change tho

antazoey: but ill make the SubprocessProvider handle this , that is a good call out.

antazoey:
update: making progress on this but haven't completed or am ready to push yet but hoping to have something cleaned up soon. Apologies for the delay!

src/ape/api/providers.py
antazoey: Now, the command affects the ID, so if you run multiple nodes with different commands, it recognized it is different.

If a provider doesn't include the port in the command, itll have to handle independently in it's plugin implementation code.

antazoey:
Somehow it works with hardhat but I think some more updates are needed there anyway.

fubuloubu:
checked this out in working with https://github.com/ApeWorX/silverback/tree/feat/add-backtesting

seems to work!

Issue #1701, opened 2023-10-16
`KeyfileAccount` should have a publicKey method


Aviksaikat:
### Overview

- It will be nice to have a `account.public_key` method to the `KeyfileAccount` or `AccountAPI` object.

### Specification

- I have something in my mind. If I can get a green signal then I will dive right in.

### Approach
- Using the `keyfile` property of the `KeyfileAccount` object I would like to implement this feature.
- Something like this 
```py
self.wallet = self.keyfile
passphrase = self._prompt_for_passphrase(default="")
# Decrypt the private key
self.private_key = Account.decrypt(self.wallet, passphrase)

# Derive the public key from the private key
self.pk = keys.PrivateKey(private_key)
self.public_key = self.pk.public_key
``` 

### Dependencies

- eth_account
- eth_keys
I believe both of them are already in use.

fubuloubu:
Yeah, this seems fine. Implementation suggestion would be to add this property to `KeyfileAccount`:

```py
class KeyfileAccount(AccountAPI):
    ...

    @property
    def public_key(self) -> HexBytes:
        ... # logic you outlined above
```

antazoey:
I am just perplexed that we have to decrypt an account's private key in order to obtain the public key... that just seems backwards?

fubuloubu:
> I am just perplexed that we have to decrypt an account's private key in order to obtain the public key... that just seems backwards?

Not really, and address is a keccak hash so it is not able to be obtained like this, so you need to to run the key derivation to get the public key again

But yes, you can obtain the public key during the address derivation process

Aviksaikat:
I have made some changes to cache the public key in the `keyfile` please review this pr https://github.com/ApeWorX/ape/pull/1702

PR #1702, opened 2023-10-18
feat: `public_key` method added to `KeyfileAccount` class [APE-1466]

Aviksaikat:
### What I did

fixes: #1701

### How to verify it

```py
from ape import accounts
a = accounts[0]
a.public_key
# Enter your password
>> HexBytes("0x33abababbababababbabababababababba..")
```

### Checklist

### DIFF
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -5,7 +5,7 @@ from typing import Iterator, Optional
 
 import click
 from eth_account import Account as EthAccount
-from eth_keys import keys # type: ignore
+from eth_keys import keys  # type: ignore
 from eth_utils import to_bytes
 from ethpm_types import HexBytes
 
@@ -95,12 +95,19 @@ class KeyfileAccount(AccountAPI):
 
     @property
     def public_key(self) -> HexBytes:
+        if "public_key" in self.keyfile:
+            return HexBytes(bytes.fromhex(self.keyfile["public_key"]))
         key = self.__key
 
         # Derive the public key from the private key
         pk = keys.PrivateKey(key)
         # convert from eth_keys.datatypes.PublicKey to str to make it HexBytes
         publicKey = str(pk.public_key)
+
+        key_file_data = self.keyfile
+        key_file_data["public_key"] = publicKey[2:]
+
+        self.keyfile_path.write_text(json.dumps(key_file_data))
 
         return HexBytes(bytes.fromhex(publicKey[2:]))
 
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -5,6 +5,7 @@ from typing import Iterator, Optional
 
 import click
 from eth_account import Account as EthAccount
+from eth_keys import keys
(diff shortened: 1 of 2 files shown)


src/ape_accounts/accounts.py
```diff
@@ -92,6 +93,17 @@ def __key(self) -> HexBytes:
 
         return key
 
+    @property
+    def public_key(self) -> HexBytes:
+        key = self.__key
+
+        # Derive the public key from the private key
+        pk = keys.PrivateKey(key)
```
antazoey, line 101: could we save this to the account's JSON so we don't have to decrypt it like this? it is odd to require a private key to get a public key. for accounts that dont yet have it in the JSON, we could do this approach here and the save it in the JSON so next time decrypting is no required.

antazoey (commented):
looks good, but the public key should maybe just be in the JSON unencrypted so your passphrase isnt required to get the public key.

thoughts anyone?

src/ape_accounts/accounts.py
```diff
@@ -92,6 +93,17 @@ def __key(self) -> HexBytes:
 
         return key
 
+    @property
+    def public_key(self) -> HexBytes:
```
antazoey, line 97:
```
    @property
    def public_key(self) -> HexBytes:
        if "public_key" in self.keyfile:
            return self.keyfile["public_key"]

        # Derive public key
        passphrase = self._prompt_for_passphrase(default="")
        key = self.__get_key(passphrase=passphrase)
        self.keyfile_path.unlink()

        pk = keys.PrivateKey(key)
        public_key = str(pk.public_key)
        data = json.dumps(EthAccount.encrypt(key, passphrase))
        data["public_key"] = pk

        self.keyfile_path.write_text(json.dumps(data))

        return pk
```

src/ape_accounts/accounts.py
```diff
@@ -92,6 +93,17 @@ def __key(self) -> HexBytes:
...
+        key = self.__key
+
+        # Derive the public key from the private key
+        pk = keys.PrivateKey(key)
+        # convert from eth_keys.datatypes.PublicKey to str to make it HexBytes
+        publicKey = str(pk.public_key)
```
antazoey, line 103: keep_it_snake_case

src/ape_accounts/accounts.py
antazoey: but this requires changing `__key` to `__get_key(passphrase=None)`

fubuloubu:
> looks good, but the public key should maybe just be in the JSON unencrypted so your passphrase isnt required to get the public key.
> 
> thoughts anyone?

can do that, the keyfile format just needs a minimum number of members to work, you can add more

fubuloubu (approved):

Aviksaikat:
I have opened an issue in the `eth_keys` repo as well so that we can it can be a general property of the `keyfile`. But the repo seems to be very inactive.

antazoey (approved):
LGTM! This is the best we can do until eth-keyfile updates

Issue #688, opened 2022-05-03
Allow CompilerAPI to compile things ad-hoc


fubuloubu:
### Overview

Sometimes it's handy to be able to compile a quick smart contract in a given language. `CompilerAPI` should allow that

### Specification

Add a method `compile_code` to `CompilerAPI`

```py
class CompilerAPI:
    ...
    def compile_code(code: str) -> ContractContainer:
    ...
```

### Dependencies
No deps

fubuloubu:
Not sure if method should return exactly one container, or a list

antazoey:
what about something like:

```python
def compile_code(*code: str) -> Iterator[ContractContainer]:
```

so you can give it multiplie codes and it gives you multiple contract  containers

fubuloubu:
> what about something like:
> 
> ```python
> def compile_code(*code: str) -> Iterator[ContractContainer]:
> ```
> 
> so you can give it multiplie codes and it gives you multiple contract  containers

I think too complicated, and you can just replicate with this:

```py
containers = map(compilers["vyper"].compile_code, codes)
```

PR #1705, opened 2023-10-27
feat: adhoc compiler settings [APE-1474]

antazoey:
### What I did

fixes: #1704

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #1707, opened 2023-10-27
feat: adhoc code compile [APE-1476]

antazoey:
### What I did

fixes: #688 
fixes: #789 

Fixes: APE-203
Fixes: APE-192

### Checklist

### DIFF
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -126,7 +126,7 @@ container = compilers.compile_source(
    "vyper",
    CODE,
    settings={"vyper": {"version": "0.3.7"}}, 
-   contract_type_overrides={"contractName": "MyContract"}
+   contractName="MyContract",
 )
 
 owner = accounts.test_accounts[0]
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -125,7 +125,7 @@ CODE = """
 container = compilers.compile_source(
    "vyper",
    CODE,
-   settings={"version": "0.3.7"}, 
+   settings={"vyper": {"version": "0.3.7"}}, 
    contract_type_overrides={"contractName": "MyContract"}
 )
 
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -110,3 +110,26 @@ vyper.compile([Path("path/to/contract.vy")])
 solidity = compilers.get_compiler("solidity", settings=settings["solidity"])
 vyper.compile([Path("path/to/contract.sol")])
 ```
+
+## Compile Source Code
+
+Instead of compiling project source files, you can compile code (str) directly:
+
+```python
+from ape import accounts, compilers
+
+CODE = """
+   ... source code here
+"""
+
+container = compilers.compile_source(
+   "vyper",
+   CODE,
+   settings={"version": "0.3.7"}, 
+   contract_type_overrides={"contractName": "MyContract"}
+)
+
+owner = accounts.test_accounts[0]
+
+instance = container.deploy(sender=owner)
+```
--- a/src/ape/api/compiler.py
+++ b/src/ape/api/compiler.py
@@ -111,7 +111,8 @@ class CompilerAPI(BaseInterfaceModel):
             code (str): The code to compile.
             base_path (Optional[pathlib.Path]): Optionally provide the base path, such as the
               project ``contracts/`` directory. Defaults to ``None``. When using in a project
-              via ``ape compile``, gets set to the project's ``contracts/`` directory.
+              via ``compilers.compile_source()``, gets set to the project's ``contracts/``
+              directory.
             **contract_type_overrides: ContractType overrides
               (needed if unable to gather required properties from source code alone, such as
               ``contractName``).
--- a/src/ape/api/compiler.py
+++ b/src/ape/api/compiler.py
@@ -102,7 +102,7 @@ class CompilerAPI(BaseInterfaceModel):
         self,
         code: str,
         base_path: Optional[Path] = None,
-        contract_type_overrides: Optional[Dict] = None,
+        **contract_type_overrides,
     ) -> ContractType:
         """
         Compile a program.
@@ -112,7 +112,7 @@ class CompilerAPI(BaseInterfaceModel):
             base_path (Optional[pathlib.Path]): Optionally provide the base path, such as the
               project ``contracts/`` directory. Defaults to ``None``. When using in a project
               via ``ape compile``, gets set to the project's ``contracts/`` directory.
-            contract_type_overrides (Optional[Dict]): ContractType overrides
+            **contract_type_overrides: ContractType overrides
               (needed if unable to gather required properties from source code alone, such as
               ``contractName``).
 
--- a/src/ape/api/compiler.py
+++ b/src/ape/api/compiler.py
@@ -102,7 +102,7 @@ class CompilerAPI(BaseInterfaceModel):
         self,
         code: str,
         base_path: Optional[Path] = None,
-        contract_type_overrides: Optional[Dict] = None,
+        **kwargs,
     ) -> ContractType:
         """
         Compile a program.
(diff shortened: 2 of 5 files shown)


fubuloubu (approved):

antazoey:
- [ ] before merging, let me get a working `ape-vyper` example / PR up

src/ape/managers/compilers.py
```diff
@@ -179,6 +181,37 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
 
         return contract_types_dict
 
+    def compile_source(self, compiler_name: str, code: str, **kwargs) -> ContractContainer:
```
antazoey, line 184: Now I am thinking about this in combo with https://github.com/ApeWorX/ape/pull/1705 and wondering if we should only have a compiler settings optional argument here

docs/userguides/compile.md
```diff
@@ -81,3 +81,26 @@ Alternatively, configure it to always happen:
...
+from ape import accounts, compilers
+
+CODE = """
+   ... source code here
+"""
+
+container = compilers.compile_source(
+   "vyper",
+   CODE,
+   settings={"version": "0.3.7"},
```
fubuloubu, line 99: shouldn't this be: 
```suggestion
   settings={"vyper": {"version": "0.3.7"}}, 
```

src/ape/api/compiler.py
```diff
@@ -75,12 +75,36 @@ def compile(
             List[:class:`~ape.type.contract.ContractType`]
         """
 
+    @raises_not_implemented
+    def compile_code(  # type: ignore[empty-body]
+        self,
+        code: str,
+        base_path: Optional[Path] = None,
+        contract_type_overrides: Optional[Dict] = None,
```
fubuloubu, line 83: what do you think about this?
```suggestion
        **contract_type_overrides: dict,
```

src/ape/managers/compilers.py
```diff
@@ -179,6 +180,37 @@ def compile(self, contract_filepaths: List[Path]) -> Dict[str, ContractType]:
...
+        settings: Optional[Dict] = None,
+        contract_type_overrides: Optional[Dict] = None,
+    ) -> ContractContainer:
+        """
+        Compile the given program.
+
+        Usage example::
+
+            code = '[{"name":"foo","type":"fallback", "stateMutability":"nonpayable"}]'
+            contract_type = compilers.compile_source("ethpm", code, contractName="MyContract")
```
fubuloubu, line 196: like this actually, or this needs to change

src/ape/api/compiler.py
antazoey: it works for now

antazoey: the type is wrong though

antazoey: actually - the reason i didnt do this (or changed it from this) is because this would require the user to always pass `base_path` when only doing some overrides

antazoey: oh wait nvm, im mistaken

src/ape/managers/compilers.py
antazoey: this is the manager, not compiler API, but ya this doc is wrong now

src/ape/api/compiler.py
antazoey: I made this change, but it is kinda be funny looking in practice, because it looks like we have a mix of snake case and camel case kwargs:

```python
contract_type = self.compile_code(
    path.read_text(),
    base_path=base_path,
    sourceId=source_id,
)
```

src/ape/managers/compilers.py
antazoey: so manager i didnt want this because it is not clear which dict gets the **, the settings or the overrides, so it seems nicer to just have 2 dicts are optional args.

in general, i tend to prefer this anyway to avoid issue mentioned here: https://github.com/ApeWorX/ape/pull/1707#discussion_r1376633835

docs/userguides/compile.md
```diff
@@ -110,3 +110,26 @@ vyper.compile([Path("path/to/contract.vy")])
...
+
+CODE = """
+   ... source code here
+"""
+
+container = compilers.compile_source(
+   "vyper",
+   CODE,
+   settings={"vyper": {"version": "0.3.7"}}, 
+   contract_type_overrides={"contractName": "MyContract"}
```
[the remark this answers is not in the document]
antazoey: yes it does, this is the manager.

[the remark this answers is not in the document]
antazoey: i didnt do kwargs on the manager, just the api, because there are 2 dicts here and it seems confusing which one is **kwargs and which one isnt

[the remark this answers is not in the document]
antazoey: plus the manager doesnt even return a contract type - it returns a container, so the context is even more removed when using the **kwargs method at the manager level, so it only at api level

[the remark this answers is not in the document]
antazoey: here is the manager signature:

```
 def compile_source(
        self,
        compiler_name: str,
        code: str,
        settings: Optional[Dict] = None,
        contract_type_overrides: Optional[Dict] = None,
    ) -> ContractContainer:
```

here is api signature:

```python
    def compile_code(
        self,
        code: str,
        base_path: Optional[Path] = None,
        **contract_type_overrides,
    ) -> ContractType:
```

I think i'd like avoiding ** altogether but idc.
as long as i can pass both settings and overrides separately, the use-case is covered.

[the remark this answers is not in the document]
fubuloubu: yeah, that does seem kind of confusing altogether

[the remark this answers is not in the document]
antazoey: how so?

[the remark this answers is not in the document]
antazoey: tell me the signatures you prefer.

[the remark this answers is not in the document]
antazoey: the `**` based part is what it makes confusing imo, it is much clearer when using dictionaries.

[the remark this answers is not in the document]
fubuloubu: the kwargs in the API interface will allow it to change a bit more easily over time.

I think doing the same in the manager interface to match would be helpful. the settings override being a dict I don't think is that confusing if the contract type overrides are kwargs. in fact, might be less confusing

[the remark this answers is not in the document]
antazoey: I am going to change it back to that, let me know if you have any other ideas though 👍🏻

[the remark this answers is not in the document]
antazoey: kwargs causes name clashing between contract type overrides and actual method kwargs. if anything it makes change harder because now if we attributes to ContractType, we have to make sure they are not clashing with kwarguments not part of the overrides.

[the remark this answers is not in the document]
antazoey: I anticipate problems with this... but ok...

src/ape/api/compiler.py
```diff
@@ -97,12 +97,35 @@ def compile(
...
+        """
+        Compile a program.
+
+        Args:
+            code (str): The code to compile.
+            base_path (Optional[pathlib.Path]): Optionally provide the base path, such as the
+              project ``contracts/`` directory. Defaults to ``None``. When using in a project
+              via ``compilers.compile_source()``, gets set to the project's ``contracts/``
+              directory.
+            **kwargs: Additional overrides.
```
fubuloubu, line 116: Would be helpful to add where these overrides are applied

src/ape/managers/compilers.py
```diff
@@ -183,19 +184,61 @@ def compile(
...
+                "ethpm",
+                code,
+                contractName="MyContract",
+            )
+
+        Args:
+            compiler_name (str): The name of the compiler to use.
+            code (str): The source code to compile.
+            settings (Optional[Dict]): Compiler settings.
+            **kwargs: Additional overrides.
```
fubuloubu, line 226: Same

fubuloubu (approved):

Issue #915, opened 2022-07-27
Better error messaging w/ Ape and Brownie in the same environment


parseb:
### Environment information

* OS: linux 
* Python Version: 3.8.10 
* `ape` and plugin versions:

```sh
$ ape --version
# 0.3.5

$ ape plugins list
  solidity    0.3.2
  foundry     0.2.0
  vyper       0.3.0
  infura      0.3.0
  alchemy     0.3.0

```

* Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
# name: gasJar
```

### What went wrong?
Installed and trying to explore.
I get this error on ape test.
The docs mention an ape-test module, not sure if that's separate... tried to install but that failed. 
Since this is a framework I assume that's already in though. Beside the point tho.

Anyway. Installed. ape test fails with below trace.
I don't have the time to look into in atm. Feel free to close without response. Just wanted to dump this for potential future reference.

* ape test
```
`parseb@:~/Desktop/github/gasJar$ ape test
Traceback (most recent call last):
  File "/home/parseb/.local/bin/ape", line 8, in <module>
    sys.exit(cli())
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/parseb/.local/lib/python3.8/site-packages/ape/_cli.py", line 40, in invoke
    return super().invoke(ctx)
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/parseb/.local/lib/python3.8/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/parseb/.local/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
[...]

PR #1708, opened 2023-10-30
feat: better error when pytest option collision [APE-1479]

antazoey:
### What I did

fixes: #915 

### Checklist

### DIFF
--- a/src/ape/pytest/plugin.py
+++ b/src/ape/pytest/plugin.py
@@ -14,14 +14,14 @@ from ape.utils import ManagerAccessMixin
 
 
 def pytest_addoption(parser):
-    def add_option(name, **kwargs):
+    def add_option(*names, **kwargs):
         try:
-            parser.addoption(name, **kwargs)
+            parser.addoption(*names, **kwargs)
         except ValueError as err:
             if "already added" in str(err):
                 # A different plugin added an option
                 # with this name. It should be ok to share it.
-                logger.warning(f"Multiple pytest options named '{name}'.")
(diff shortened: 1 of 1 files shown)


src/ape/pytest/plugin.py
```diff
@@ -14,33 +14,42 @@
...
-    )
-    parser.addoption(
+    def add_option(*names, **kwargs):
+        try:
+            parser.addoption(*names, **kwargs)
+        except ValueError as err:
+            if "already added" in str(err):
+                # A different plugin added an option
+                # with this name. It should be ok to share it.
+                logger.warning(f"Multiple pytest options named '{names[0]}'.")
```
antazoey, line 24: ugh, no longer liking this because now we are potentially telling other plugins wrong things...
perhaps it is better to raise a nicer error instead.

fubuloubu (dismissed):
Like this change!

src/ape/pytest/plugin.py
```diff
@@ -14,33 +15,44 @@
...
+        except ValueError as err:
+            name_str = ", ".join(names)
+            if "already added" in str(err):
+                raise ConfigError(
+                    f"Another pytest plugin besides `ape_test` uses an option with "
+                    f"one of '{name_str}'. Note that Ape does not support being "
+                    f"installed alongside Brownie; please use separate environments!"
+                )
+
+            raise ConfigError(f"Failing add option {name_str}: {err}") from err
```
antazoey, line 30:
```suggestion
            raise ConfigError(f"Failed adding option {name_str}: {err}") from err
```

Issue #1297, opened 2023-02-11
Documentation is not clear about provider context manager [APE-618]


PoCk3T-SPAI:
### Environment information

- OS: macOS / linux / win
- Python Version: x.x.x
- `ape` and plugin versions:

### Environment information

- apeworx/ape:latest Docker image

```sh
$ ape --version
0.6.2

$ ape plugins list
Installed Plugins:
  tokens       0.6.0
  alchemy      0.6.0
  trezor       0.6.0
  ledger       0.6.0
  ens          0.6.0
  infura       0.6.0
  hardhat      0.6.0
  etherscan    0.6.0
  foundry      0.6.0
  vyper        0.6.1
  avalanche    0.1.0a3
  solidity     0.6.0
  bsc          0.6.0
  polygon      0.6.0
  template     0.6.0
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
dependencies:
  - name: OpenZeppelin
    github: OpenZeppelin/openzeppelin-contracts
    version: 4.8.1    

  - name: Chainlink
    github: smartcontractkit/chainlink
    branch: master
    contracts_folder: contracts/src/v0.8
    
solidity:
  import_remapping:
    - "@openzeppelin=OpenZeppelin/4.8.1"
    - "@chainlink=Chainlink/master"

geth:
  bsc:
    testnet:
      uri: https://bsc-testnet.public.blastapi.io 
    mainnet:
      uri: https://bsc-mainnet.public.blastapi.io 
  avalanche:
    fuji:
      uri: https://api.avax-test.network/ext/bc/C/rpc # https://ava-testnet.public.blastapi.io/ext/bc/C/rpc
    mainnet:
      uri: https://api.avax.network/ext/bc/C/rpc # https://ava-mainnet.public.blastapi.io/ext/bc/C/rpc
  fantom:
    testnet:
      uri: https://fantom-testnet.public.blastapi.io
    opera:
      uri: https://fantom-mainnet.public.blastapi.io
  polygon:
    mumbai:
      uri: https://polygon-mumbai.g.alchemy.com # https://polygon-testnet.public.blastapi.io
    mainnet:
      uri: https://polygon-mainnet.public.blastapi.io
```

### What went wrong?

In ape console:
```python
In [16]: with networks.polygon.mumbai.use_provider("geth") as provider:
    ...:         provider.connect()
    ...:
[...]

Issue #657, opened 2022-04-12
Add Doc Tests


fubuloubu:
### Elevator pitch:
Test the docs

### Value:
Ensures docs are up to date with the latest ape

### Dependencies:
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

### Design approach:
Add tests to the docs where an ape feature is being tested

### Task list:

* [ ] Tasks go here

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

fubuloubu:
We do doctests via `sphinx-ape`

PR #1710, opened 2023-10-30
docs: document ProviderContextManager [APE-1481]

antazoey:
### What I did

[APE-618] #1297

fixes: #1297 
Fixes: APE-618

### Checklist

### DIFF
--- a/docs/userguides/networks.md
+++ b/docs/userguides/networks.md
@@ -138,12 +138,11 @@ block = chain.provider.get_block("latest")
 
 ## Provider Context Manager
 
-Use the `ProviderContextManager` to change network-context in Python.
-For example, if you are using a script connected to one network, you can change connection in the middle of the script by using the context manager:
+Use the `ProviderContextManager` to change the network-context in Python.
+For example, if you are using a script with a default network connection, you can change connection in the middle of the script by using the provider context manager:
 
 ```python
 from ape import chain, networks
-
 
 def main():
     start_provider = chain.provider.name
--- a/docs/userguides/networks.md
+++ b/docs/userguides/networks.md
@@ -138,7 +138,7 @@ block = chain.provider.get_block("latest")
 
 ## Provider Context Manager
 
-Use the `ProviderContextManager` to change the network-context in Python.
+Use the [ProviderContextManager](../methoddocs/api.html#ape.api.networks.ProviderContextManager) to change the network-context in Python.
 When entering a network for the first time, it will connect to that network.
 **You do not need to call `.connect()` or `.disconnect()` manually**.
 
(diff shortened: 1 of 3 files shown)


docs/userguides/networks.md
```diff
@@ -135,3 +135,73 @@ from ape import chain
...
+
+```python
+from ape import networks
+
+# Same as doing `networks.ethereum.local.use_provider("test")`.
+with networks.parse_network_choice("ethereum:local:test") as provider:
+    print(provider)
+```
+
+**A note about disconnect**: Providers do not disconnect until the very end of your Python session.
```
antazoey, line 181: note: am not against re-designing this as a breaking change but this is how it works today.

We could probably flip it though, like `keep_alive=True` to keep certain connections from disconnecting and then we can have users do that in their pytest fixtures.

But this is a breaking change; the network disconnect functionality has been around since we were really  focused on `ape-starknet` so I apologize the conflicting use-cases, but I at least want this documented.

docs/userguides/networks.md
```diff
@@ -135,3 +135,73 @@ from ape import chain
...
+**You do not need to call `.connect()` or `.disconnect()` manually**.
+
+For example, if you are using a script with a default network connection, you can change connection in the middle of the script by using the provider context manager:
+
+```python
+from ape import chain, networks
+
+def main():
+    start_provider = chain.provider.name
+    with networks.ethereum.mainnet.use_provider("geth") as provider:
```
fubuloubu, line 152: Might be helpful yo add an example using default provider

src/ape/pytest/runners.py
```diff
@@ -133,7 +133,8 @@ def pytest_runtest_setup(self, item):
         """
         if (
             self.config_wrapper.isolation is False
-            or isinstance(item, pytest.DoctestItem)  # doctests don't have fixturenames
+            # doctests don't have fixturenames
```
fubuloubu, line 136: Wait. Are doctests actually running?

fubuloubu (dismissed):

src/ape/pytest/runners.py
antazoey: Not us, but in Ape projects that have both

fubuloubu: Ah, would be sweet to have doc tests set up

antazoey: So would a ton of ice cream right now

antazoey: The issue is tracked here: https://github.com/ApeWorX/ape/issues/657
It would be a good amount of work I think getting those running how we want.

docs/userguides/networks.md
antazoey: done

fubuloubu (approved):

PR #1713, opened 2023-10-31
feat: account list filter

antazoey:
### What I did

fixes: #

### Checklist


docs/userguides/clis.md
```diff
@@ -158,3 +158,23 @@ def create_account(alias):
...
+Use the `account_list` kwarg to filter the choices by a given sequence of accounts.
+
+```python
+import click
+from ape import accounts
+from ape.cli import existing_alias_argument
+from ape_accounts.accounts import KeyfileAccount
+
+def _valid_accounts():
+   return [a for a in accounts if a.alias.startswith("root_signer_")]
```
fubuloubu, line 173: what's `root_signer_`?

docs/userguides/clis.md
```diff
@@ -158,3 +158,23 @@ def create_account(alias):
...
+
+@click.command()
+@existing_alias_argument(account_type=KeyfileAccount, account_list=_valid_accounts())
```
fubuloubu, line 176: also curious if using a key filter function (e.g. `Callable[[AccountAPI], bool]` signature function) would be more flexible

src/ape_test/accounts.py
```diff
@@ -51,7 +51,7 @@ def _dev_accounts(self) -> List[GeneratedDevAccount]:
     @property
     def aliases(self) -> Iterator[str]:
         for index in range(self._num_of_accounts):
-            yield f"dev_{index}"
+            yield f"TEST::{index}"
```
fubuloubu, line 54: might make this breaking, but I am fine with that

docs/userguides/clis.md
antazoey: I was just trying to come up with made up example, I can make that more obvious though

src/ape_test/accounts.py
antazoey: it never worked before anyway, so it is no really breaking. You couldn't do anything with this alias. The only place anything remote related to the alias happens is `get_user_selected_account` but that required TEST:: prefix anyway, so this just makes those the same, which somehow became apparent that they weren't in this PR

docs/userguides/clis.md
antazoey: That would work! However, it may be a little less efficient. Maybe we can have it accept either a callable or a list?

fubuloubu: it might be more efficient because it doesn't have to loop `accounts` first

src/ape_test/accounts.py
fubuloubu: sounds good, as long as it's thought through

fubuloubu (approved):

docs/userguides/clis.md
antazoey: Updated! `account_type` can now be a callable, a sequence, or the original Type.

If you already have a list, you avoid looking at accounts entirely:

```python
selected_signer = get_user_selected_account(account_type=safe.local_signers)
```

fubuloubu (approved):

PR #1714, opened 2023-10-31
feat: upstream provider and forked network tools [APE-1485]

antazoey:
### What I did

realizing it was absolute pain to need to do anything with the upstream provider, even at the plugin level.
This aids that by adding properties and such to a new class for forked networks and a method for getting that context as well.

- [ ] show case in ape-safe and ape-foundry
- [ ] update in all ecosystem plugins

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -48,7 +48,7 @@ from .config import PluginConfig
 
 if TYPE_CHECKING:
     from .explorers import ExplorerAPI
-    from .providers import BlockAPI, ProviderAPI
+    from .providers import BlockAPI, ProviderAPI, UpstreamProvider
     from .transactions import ReceiptAPI, TransactionAPI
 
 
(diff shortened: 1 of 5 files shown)


src/ape/api/networks.py
```diff
@@ -1046,6 +1046,54 @@ def verify_chain_id(self, chain_id: int):
...
+    def upstream_provider(self) -> Optional["UpstreamProvider"]:
+        """
+        The provider used when requesting data before the local fork.
+        Set this in your config under the network settings.
+        When not set, will attempt to use the default provider, if one
+        exists.
+        """
+
+        config_choice = self._network_config.get("upstream_provider")
+        if provider_name := config_choice or self.upstream_network.default_provider:
```
fubuloubu, line 1068: shouldn't it always have a default?

fubuloubu (approved):
love this

src/ape/api/networks.py
antazoey: `.default_provider` is optional!

Let's say for example you are forking some network that does not have any providers associated with it. It would fail in that case.

I am not sure that is very easy in today's ecosystem... but with the plugin system, anything is possible.

antazoey: We could raise an error instead and change this to not be Optional? But technically, None fits the bill

fubuloubu: Yeah, I think the corner case that you have a network with no providers then it should just raise

antazoey: Ape does ship with networks without providers, so it is quite normal! But none of the upstream providers..
All the forked networks themselves are provider-less.

antazoey: done

antazoey:
I addressed feedback and need to merge because my I need for other work and is hard to manage without!
If more feedback arises, please let me know :)

Issue #1590, opened 2023-08-08
manifest produced by `ape compile` does not contain compiler info


0xalpharush:
### Environment information

- OS: macOS 
- Python Version: 3.10
- `ape` and plugin versions:

```sh
$ ape --version
0.6.15 # I also tried 0.6.16.dev9+gabe4e96b

$ ape plugins list
Installed Plugins:
  solidity    0.6.7
  hardhat     0.6.12
  vyper       0.6.9
```
- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
name: yearn-v3

plugins:
  - name: solidity
    version: 0.6.7
  - name: vyper
    version: 0.6.9
  - name: hardhat
    version: 0.6.12

default_ecosystem: ethereum
dependencies:
  - name: openzeppelin
    github: OpenZeppelin/openzeppelin-contracts
    version: 4.7.3

solidity:
  import_remapping:
    - "@openzeppelin/contracts=openzeppelin/v4.7.3"

ethereum:
  local:
    default_provider: hardhat
```

### What went wrong?

I am trying to parse the build manifest for ape support in https://github.com/crytic/crytic-compile, but the compiler info is missing
Reproduce with:
```
$ git clone https://github.com/0xalpharush/yearn-vaults-v3 && cd yearn-vaults-v3
$ python3 -m venv venv
$ source venv/bin/activate
$ python3 -m pip install -r requirements.txt
$ yarn
$ ape plugins install .
$ ape compile
$ cat .build/__local__.json | jq 'keys'
[
  "contractTypes",
  "manifest",
  "sources"
]
```
With python:
```
$ python
>>> from ethpm_types import PackageManifest
>>> import json
>>> print(PackageManifest.parse_obj(json.load(open(".build/__local__.json"))).compilers)
None

```
This field is missing https://github.com/ApeWorX/ethpm-types/blob/dffcb6b2fe17f858c95d07499aafe1d66e939dfa/ethpm_types/manifest.py#L124

### How can it be fixed?
I think it may need to be added here https://github.com/ApeWorX/ape/blob/abe4e96bc3ce7132e486913a7b0a4a19394f3015/src/ape/managers/project/types.py#L177

Aviksaikat:
hmm, it seems like there are many bugs in `ape compile`

fubuloubu:
> hmm, it seems like there are many bugs in `ape compile`

[...]

PR #1719, opened 2023-11-01
refactor: use build artifacts from compilers in `project.compiler_data` [APE-1495]

antazoey:
### What I did

partially fixes https://github.com/ApeWorX/ape/issues/1590

### How I did it

* Creates root properties for the location and decoding of such data (this is the part that `ape-solidity` and `ape-vyper` depend on to _actually_ accomplish the goal.
* Refactor manifest extract (used in publishing) to rely on the build artifacts instead of making shit up
* Test: We can actually test compiler data in core now because of this because we just create the raw JSON and test that it decoded properly. In the soon-coming PRs for `ape-solidity` and `ape-vyper`, there will be more end-to-end testing because they are responsible for creating such data and we can see that both it'll be created and show up in the extracted manifest correctly

### Checklist

### DIFF
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -111,6 +111,25 @@ solidity = compilers.get_compiler("solidity", settings=settings["solidity"])
 vyper.compile([Path("path/to/contract.sol")])
 ```
 
+### Settings Artifacts
+
+Compiler build artifacts can be found in the the `<project>/.build/compilers.json` file.
+Each compiler in your project has an associated [ethpm_types.source.Compiler](https://github.com/ApeWorX/ethpm-types/blob/main/ethpm_types/source.py) build artifact.
+This data contains the versions and settings used for each contract in your project.
+For example, assume you have`ape-solidity` creating a contract named `Test` from file `Test.sol`.
+The structure of `.builds/compilers.json` file would be:
+
+```json
+[
+   {
+      "contractTypes": ["Test"],
+      "name": "solidity",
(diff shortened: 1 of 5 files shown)


src/ape/managers/compilers.py
```diff
@@ -42,6 +43,21 @@ def __getattr__(self, name: str) -> Any:
 
         raise ApeAttributeError(f"No attribute or compiler named '{name}'.")
 
+    @property
+    def compilers_json_file(self) -> Path:
+        return self.project_manager.local_project._cache_folder / "compilers.json"
+
+    @property
+    def compilers_json(self) -> Dict[str, Compiler]:
```
antazoey, line 51: - [ ] needed version layer

fubuloubu: so the keys here are arbitrary, can be something like `{compiler plugin}-{compiler version}` so that if more than one version is required then the key is supported

docs/userguides/compile.md
```diff
@@ -111,6 +111,23 @@ solidity = compilers.get_compiler("solidity", settings=settings["solidity"])
 vyper.compile([Path("path/to/contract.sol")])
 ```
 
+### Settings Artifacts
+
+Compiler build artifacts can be found in the the `<project>/.build/compilers.json` file.
```
fubuloubu, line 116: actually, not sure if this location makes sense, because the `Compiler` object in the manifest is not expected to be a link to another file.

can probably just be stored directly in the `<project>/.build/__local__.json` file

antazoey: I though we were going to keep everything separate so people to be choosier about the files they need to host and stuff? Didn't we chat about that recently on GitHub somewhere?

antazoey: I suppose the contract types being separate is probably ok enough for now.

antazoey: I'm realizing this is a much bigger refactor ... Dependencies heavily rely on the compiler data already and messing with __local__.json is giving me a chicken and egg problem right now. Closing for now.

PR #1720, opened 2023-11-02
fix: issue where ape wouldn't recompile changed contracts [APE-1498]

antazoey:
### What I did

Thank you @z80dev for pointing this out.
I don't see any GH issues or anything for it? This was a serious issue that got unnoticed and wasn't captured in tests.

### How I did it

Use `load_contracts()` instead of the contracts dict in a method used for contract container access in the ProjectManager.
But - the real good part is actually the integration test that prevents this issue from returning, at least in scripts.

### How to verify it

* Have deploy script
* Have a contract
* Make change to contract
* Run deploy script
* It better have re-compiled!

### Checklist


tests/integration/cli/test_run.py
```diff
@@ -161,3 +161,24 @@ def test_scripts_module_already_installed(ape_cli, runner, project, mocker):
...
+    # Make a change to the contract
+    contract = project.contracts_folder / "contract.json"
+    method_name = project.TestContractVy.contract_type.view_methods[0].name
+    new_method_name = f"f__{method_name}__"
+    new_contract_text = contract.read_text().replace(method_name, new_method_name)
+    contract.write_text(new_contract_text)
+
+    # Run the script. It better recompile first!
+    result = runner.invoke(ape_cli, ["run", "output_contract_view_methods"])
+    assert new_method_name in result.output
```
antazoey, line 184: *note*: This FAILS on main or if you comment out the fix! so yay.

fubuloubu (dismissed):

fubuloubu (approved):

PR #1731, opened 2023-11-07
docs: fix plugin install command [APE-1519]

alex-damjanovic:
### What I did

Added `.` in the plugins install command in the docs as it requires `.` for plugins to be installed from ape-config.yaml. 

### How I did it

Added a simple dot in the docs.

### How to verify it

The plugins won't be installed if there is no `.`. 

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -132,7 +132,7 @@ plugins:
 Install these plugins by running command:
 
 ```bash
-ape plugins install
+ape plugins install .
 ```
 
 ## Testing


antazoey:
I thought both worked?

antazoey:
more info: the CLI can take a path arg to a config file, so the `.` means uses the cwd.
However, the default arg is also to just look at the cwd, so that is why I thought they both worked

alex-damjanovic:
> I thought both worked?

ERROR: You must give at least one requirement to install. This is the case with me.

PR #1733, opened 2023-11-09
fix: run console under providercontext [APE-1529]

z80dev:
### What I did

Run the interactive console under a Provider context

fixes: #581 

### How I did it

Before launching the console, run under a new provider context manager because the old one exits

### How to verify it

run a script with `ape run -I ...` and evaluate `ape.chain.provider`

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -577,7 +577,12 @@ class ProviderContextManager(ManagerAccessMixin):
     connected_providers: Dict[str, "ProviderAPI"] = {}
     provider_stack: List[str] = []
     disconnect_map: Dict[str, bool] = {}
-    recycled_provider: ClassVar[Optional["ProviderAPI"]] = None
+
+    # We store a provider object at the class level for use when disconnecting
+    # due to an exception, when interactive mode is set. If we don't hold on
+    # to a reference to this object, the provider is dropped and reconnecting results
+    # in losing state when using a spawned local provider
+    _recycled_provider: ClassVar[Optional["ProviderAPI"]] = None
 
     def __init__(
         self,
@@ -595,16 +600,21 @@ class ProviderContextManager(ManagerAccessMixin):
         return not self.connected_providers or not self.provider_stack
 
     def __enter__(self, *args, **kwargs):
-        if self.recycled_provider is not None:
-            self._provider = self.recycled_provider
-            ProviderContextManager.recycled_provider = None
+        # If we have a recycled provider available, this means our last exit
+        # was due to an exception during interactive mode. We should resume that
+        # same connection, but also clear the object so we don't do this again
+        # in later provider contexts, which we would want to behave normally
+        if self._recycled_provider is not None:
+            # set inner var to the recycled provider for use in push_provider()
+            self._provider = self._recycled_provider
+            ProviderContextManager._recycled_provider = None
         return self.push_provider()
 
-    def __exit__(self, *args, **kwargs):
-        if not self._disconnect_on_exit and args[0] is not None and not self._skipped_disconnect:
-            self._skipped_disconnect = True
+    def __exit__(self, exception, *args, **kwargs):
+        if not self._disconnect_on_exit and exception is not None:
+            # We want to skip disconnection when exiting due to an exception in interactive mode
             if provider := self.network_manager.active_provider:
-                ProviderContextManager.recycled_provider = provider
+                ProviderContextManager._recycled_provider = provider
         else:
             self.pop_provider()
 
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -576,20 +576,36 @@ class ProviderContextManager(ManagerAccessMixin):
     connected_providers: Dict[str, "ProviderAPI"] = {}
     provider_stack: List[str] = []
     disconnect_map: Dict[str, bool] = {}
+    recycled_provider = None
 
-    def __init__(self, provider: "ProviderAPI", disconnect_after: bool = False):
+    def __init__(
+        self,
+        provider: "ProviderAPI",
+        disconnect_after: bool = False,
+        disconnect_on_exit: bool = True,
+    ):
         self._provider = provider
         self._disconnect_after = disconnect_after
+        self._disconnect_on_exit = disconnect_on_exit
+        self._skipped_disconnect = False
 
     @property
     def empty(self) -> bool:
         return not self.connected_providers or not self.provider_stack
 
     def __enter__(self, *args, **kwargs):
+        if self.recycled_provider is not None:
+            self._provider = self.recycled_provider
+            ProviderContextManager.recycled_provider = None
         return self.push_provider()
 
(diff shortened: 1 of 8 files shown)


src/ape_run/_cli.py
```diff
@@ -76,13 +76,15 @@ def invoke(self, ctx: Context) -> Any:
...
                 # Attempt to use source-traceback style printing.
-                if not isinstance(err, ApeException) or not handle_ape_exception(
-                    err, [ctx.obj.project_manager.path]
-                ):
-                    err_info = traceback.format_exc()
-                    click.echo(err_info)
-
-                self._launch_console()
+                network_value = ctx.params.get("network") or networks.default_ecosystem.name
+                with networks.parse_network_choice(network_value):
```
fubuloubu, line 80: so the problem here with disconnecting and reconnecting is if the network context gets dumped

I think there has to be a way of extending the context such that the interactive mode stays under the main test context

z80dev: yep, ended up putting the provider in a classvar to 'recycle' it in the right conditions

src/ape/api/networks.py
```diff
@@ -576,20 +577,36 @@ class ProviderContextManager(ManagerAccessMixin):
     connected_providers: Dict[str, "ProviderAPI"] = {}
     provider_stack: List[str] = []
     disconnect_map: Dict[str, bool] = {}
+    recycled_provider: ClassVar[Optional["ProviderAPI"]] = None
```
antazoey, line 580: Can we make this "protected" (underscore prefix)?

antazoey, line 580: This needs comments everywhere

src/ape/api/networks.py
```diff
@@ -576,20 +577,36 @@ class ProviderContextManager(ManagerAccessMixin):
...
 
     def __enter__(self, *args, **kwargs):
+        if self.recycled_provider is not None:
+            self._provider = self.recycled_provider
+            ProviderContextManager.recycled_provider = None
         return self.push_provider()
 
     def __exit__(self, *args, **kwargs):
-        self.pop_provider()
+        if not self._disconnect_on_exit and args[0] is not None and not self._skipped_disconnect:
```
antazoey, line 604: maybe assign `args[0]` to a variable name just for the sake of knowing what is going on and why we care about this not being `None`.. and a comment too

src/ape/cli/commands.py
```diff
@@ -14,5 +14,9 @@ class NetworkBoundCommand(click.Command):
 
     def invoke(self, ctx: Context) -> Any:
         value = ctx.params.get("network") or networks.default_ecosystem.name
-        with networks.parse_network_choice(value):
+        interactive = ctx.params.get("interactive")
+        # If not found, check the parent context.
+        if interactive is None and ctx.parent:
+            interactive = ctx.parent.params.get("interactive")
```
antazoey, line 20: what about `parent.parent` ? Should this be recursive?

antazoey (commented):
Yay, I am thrilled this will finally function correctly.
I added some thoughts, I think since this is so weird we just need to be super clear about what's going on where.

The most important piece left is a new or updated test . See the existing test `tests/integration/cli/test_run.py::test_run_interactive`!

src/ape/cli/commands.py
z80dev: yeah the real goal is to see if any parent context has this flag set, I made this a recursive function now

src/ape/api/networks.py
z80dev: args[0] is an exception if one exists, named an arg for it

src/ape/api/networks.py
z80dev: added comments explaining the goal and use

z80dev: yep good catch 👍🏼

z80dev:
added a test verifying state changes at the node level persist across exception boundaries

fubuloubu (approved):

tests/integration/cli/projects/script/scripts/error_cli.py
```diff
@@ -1,7 +1,12 @@
...
 
+import ape
+
 
 @click.command(short_help="Use a subcommand")
 def cli():
     local_variable = "test foo bar"  # noqa[F841]
+    acct = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"
+    provider = ape.chain.provider
+    provider.set_balance(acct, 100000000)
```
antazoey: **issue** this may require foundry, so perhaps we can just mine a bunch of blocks or something and assert we are above a certain point

antazoey: ** nit ** can use more readable values i think, like "10 ETH", might require the chain method or wrapping in conversion convert

tests/integration/cli/test_run.py
```diff
@@ -100,13 +100,14 @@ def test_run_interactive(ape_cli, runner, project):
...
     # Show that the variable namespace from the script is available in the console.
-    user_input = "local_variable\nexit\n"
+    user_input = "local_variable\nprovider = ape.chain.provider\nprovider.get_balance(acct)\nexit\n"
 
     result = runner.invoke(ape_cli, ["run", "--interactive", scripts[0].stem], input=user_input)
     assert result.exit_code == 0, result.output
 
     # From script: local_variable = "test foo bar"
     assert "test foo bar" in result.output
+    assert "100000000" in result.output
```
antazoey: ** problem ** a lo of our tests run on 2 cores in CI and so if you arent using isolated accounts, you can have the same account spending, thus this number would be < 100,000,000.

src/ape/cli/commands.py
```diff
@@ -6,6 +6,16 @@
 from ape import networks
 
 
+def check_parents_for_interactive(ctx: Context) -> bool:
```
antazoey, line 9: **praise**: The name of the method cracks me up!

src/ape_run/_cli.py
```diff
@@ -76,13 +76,15 @@ def invoke(self, ctx: Context) -> Any:
...
                 # Attempt to use source-traceback style printing.
-                if not isinstance(err, ApeException) or not handle_ape_exception(
-                    err, [ctx.obj.project_manager.path]
-                ):
-                    err_info = traceback.format_exc()
-                    click.echo(err_info)
-
-                self._launch_console()
+                network_value = ctx.params.get("network") or networks.default_ecosystem.name
+                with networks.parse_network_choice(network_value, disconnect_on_exit=False):
```
antazoey, line 80: **idea**: We could add a comment here like "NOTE: We must be connected to the provider before launching the console so that the active provider is set from the last context.

antazoey (approved):
LGTM!

I added some comments and I probably should have waited for this to release Ape 0.6.25, but we could definitely just release 0.6.26, doesn't hurt.

src/ape/cli/commands.py
z80dev: 🤣

Issue #1727, opened 2023-11-04
nice to have accounts.sign_message return the hashed message as well.


Aviksaikat:
### Overview

- As ape is delegating most of the account calls to `eth_accounts`; the `accounts.sign_message` should return the hashed message as well. So that we can get the message signature.

![image](https://github.com/ApeWorX/ape/assets/31238298/c10b575a-4ecb-43e5-9875-f45013e774cf)
Something like this.

### Dependencies

N/A

Aviksaikat:
Want me to take a look ?

fubuloubu:
> Want me to take a look ?

could test adding it here: https://github.com/ApeWorX/ape/blob/main/src/ape/types/signatures.py#L21

PR #1734, opened 2023-11-11
feat: messageHash, signature added to the `_Signature` class [APE-1536]

Aviksaikat:
### What I did

Added the `messageHash` & `signature` to the `_Signature` dataclass

fixes: #1727

### How I did it

```py
[...]
def sign_message(self, msg: SignableMessage) -> Optional[MessageSignature]:
        signed_msg = EthAccount.sign_message(msg, self.private_key)
        return MessageSignature(
            _messageHash=signed_msg.messageHash,
            v=signed_msg.v,
            r=to_bytes(signed_msg.r),
            s=to_bytes(signed_msg.s),
            _signature=signed_msg.signature,
        )
```

### How to verify it
```py
>>> ape console
from eth_account.messages import encode_defunct
message = encode_defunct(text="Hello Apes!")
signer = accounts.test_accounts[0]

signature = signer.sign_message(message)
signature.messageHash
>>> HexBytes('0x46b937e48a6aaec0a17aac9dc326950632c949b1f1f9300073b1ef55b905c278')
In [6]: signer.sign_message(message).signature
Out[6]: HexBytes('0x8a1826611000533319d940bd2b5cd90d6a5ea40015e68fb65252cb7eb2b656450cb90a84b287f25eaff5a2b4259245486f985a30ac88b712af2afb661d0f5c161c')


```

### Checklist

### DIFF
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -239,20 +239,18 @@ signature = account.sign_message(message)
 Get the signed message signature as well as message hash signed by the account
 
 ```py
-from ape import accounts
-from eth_account.messages import encode_defunct
+from ape.types.signatures import recover_signer
 
-account = accounts.load("<ALIAS>")
-account.set_autosign(True)
-
-# Now, you will not be prompted to sign messages or transactions
-message = encode_defunct(text="Hello Apes!")
+[...]
 signature = account.sign_message(message)
 
 # signature
 print(signature.encode_vrs())
 # message hash
-print(signature.messageHash)
+print(signature.message_hash)
+
+# recover the signer from
+signer = recover_signer(message, signature.encode_vrs())
 ```
 
 ## Hardware Wallets
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -241,7 +241,6 @@ Get the signed message signature as well as message hash signed by the account
 ```py
 from ape.types.signatures import recover_signer
 
-[...]
 signature = account.sign_message(message)
 
 # signature
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -4,7 +4,7 @@ Accounts in Ape come from [AccountAPI](../methoddocs/api.html#ape.api.accounts.A
 There are typically two types of accounts:
 
 1. Test accounts
-1. Live network accounts
+2. Live network accounts
 
 Test accounts are useful for local network testing and debugging contracts.
 Live network accounts are for interacting with live blockchains and should be secured.
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -249,8 +249,6 @@ print(signature.encode_vrs())
 # message hash
 print(signature.message_hash)
 
-# recover the signer from
-signer = recover_signer(message, signature.encode_vrs())
 ```
 
 ## Hardware Wallets
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -246,8 +246,8 @@ signature = account.sign_message(message)
 
 # signature
 print(signature.encode_vrs())
-# message hash
-print(signature.message_hash)
+# get original signer
+print(recover_signer(...))
 ```
 
 ## Hardware Wallets
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -4,7 +4,7 @@ Accounts in Ape come from [AccountAPI](../methoddocs/api.html#ape.api.accounts.A
 There are typically two types of accounts:
 
 1. Test accounts
-2. Live network accounts
+1. Live network accounts
 
 Test accounts are useful for local network testing and debugging contracts.
 Live network accounts are for interacting with live blockchains and should be secured.
@@ -250,7 +250,7 @@ message = encode_defunct(text="Hello Apes!")
 signature = account.sign_message(message)
 
 # signature
-print(signature.signature)
+print(signature.encode_vrs())
 # message hash
 print(signature.messageHash)
 ```
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -248,7 +248,6 @@ signature = account.sign_message(message)
 print(signature.encode_vrs())
 # message hash
 print(signature.message_hash)
-
 ```
 
 ## Hardware Wallets
--- a/docs/userguides/accounts.md
+++ b/docs/userguides/accounts.md
@@ -234,6 +234,27 @@ message = encode_defunct(text="Hello Apes!")
 signature = account.sign_message(message)
 ```
 
+## Signed Message
+
+Get the signed message signature as well as message hash signed by the account
+
+```py
+from ape import accounts
+from eth_account.messages import encode_defunct
+
+account = accounts.load("<ALIAS>")
+account.set_autosign(True)
+
+# Now, you will not be prompted to sign messages or transactions
+message = encode_defunct(text="Hello Apes!")
+signature = account.sign_message(message)
+
+# signature
+print(signature.signature)
+# message hash
+print(signature.messageHash)
+```
+
 ## Hardware Wallets
 
 Because of the plugin system in Ape, we are able to support other types of accounts including hardware wallet accounts.
--- a/src/ape/types/signatures.py
+++ b/src/ape/types/signatures.py
@@ -22,7 +22,7 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
-    messageHash: Optional[bytes] = None
+    message_hash: Optional[bytes] = None
 
     def __iter__(self) -> Iterator[Union[int, bytes]]:
         # NOTE: Allows tuple destructuring
--- a/src/ape/types/signatures.py
+++ b/src/ape/types/signatures.py
@@ -22,7 +22,7 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
-    _messageHash: Optional[SignableMessage] = None
+    messageHash: Optional[SignableMessage] = None
 
     def __iter__(self) -> Iterator[Union[int, bytes]]:
         # NOTE: Allows tuple destructuring
@@ -38,7 +38,6 @@ class _Signature:
 
     def encode_rsv(self) -> bytes:
         return _left_pad_bytes(self.r, 32) + _left_pad_bytes(self.s, 32) + to_bytes(self.v)
-
 
 
 class MessageSignature(_Signature):
--- a/src/ape/types/signatures.py
+++ b/src/ape/types/signatures.py
@@ -3,7 +3,6 @@ from typing import Iterator, Optional, Union
 from eth_account import Account
 from eth_account.messages import SignableMessage
 from eth_utils import to_bytes, to_hex
-from hexbytes import HexBytes
 from pydantic.dataclasses import dataclass
 
 from ape.types import AddressType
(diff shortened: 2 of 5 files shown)


Aviksaikat:
I have made the initial draft but there is an issue with the `sign_transaction` method as it's using the `TransactionSignature` which is inheriting from the `_Signautre` dataclass. As a signed transaction contains different values  there is no `messageHash` or `signature` for them

src/ape/types/signatures.py
```diff
@@ -22,6 +23,8 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
+    _messageHash: HexBytes
+    _signature: HexBytes
```
fubuloubu, line 27: this field should not be added

src/ape/types/signatures.py
```diff
@@ -38,6 +41,12 @@ def encode_vrs(self) -> bytes:
     def encode_rsv(self) -> bytes:
         return _left_pad_bytes(self.r, 32) + _left_pad_bytes(self.s, 32) + to_bytes(self.v)
 
+    def messageHash(self) -> HexBytes:
+        return self._messageHash
+
+    def signature(self) -> HexBytes:
+        return self._signature
```
fubuloubu, line 48: we already have `encode_*` methods for properly encoding the signature for different applications

src/ape/types/signatures.py
```diff
@@ -22,6 +23,8 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
+    _messageHash: HexBytes
```
fubuloubu, line 26: I think this might be better like this:
```suggestion
    _msg: Optional[SignableMessage] = None
```

fubuloubu: And add a note that this value can be added by downstream applications to cache this property for recovery purposes

src/ape/types/signatures.py
fubuloubu: These may need to be documented better to show how the encode methods can be used in smart contract applications

Aviksaikat: how about we keep the both ? if people want to get the signature directly without calling any further functions?

fubuloubu: I don't think this particular function adds a lot of value, and also maybe confuse people who are seeking the encoded signatures (where it is useful to call out which way you are encoding)

src/ape/types/signatures.py
```diff
@@ -38,6 +39,9 @@ def encode_vrs(self) -> bytes:
     def encode_rsv(self) -> bytes:
         return _left_pad_bytes(self.r, 32) + _left_pad_bytes(self.s, 32) + to_bytes(self.v)
 
+    def messageHash(self) -> SignableMessage:
+        return self._messageHash
```
fubuloubu, line 43:
```suggestion
```

src/ape/types/signatures.py
```diff
@@ -22,6 +22,7 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
+    _messageHash: Optional[SignableMessage] = None
```
fubuloubu, line 25:
```suggestion
    messageHash: Optional[SignableMessage] = None
```

fubuloubu: also I think this should be `Optional[HexBytes]`

fubuloubu (dismissed):
Should a test showing the recover_message works on it

fubuloubu (dismissed):

fubuloubu:
@Aviksaikat please fix linting and testing issues

Aviksaikat:
locally the tests are working. idk what is failing

src/ape/types/signatures.py
```diff
@@ -22,6 +23,7 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
+    messageHash: Optional[HexBytes] = None
```
mikeshultz, line 26: suggestion: I suspect setting the type as `bytes` here may fix your tests.  You can still set the value as `HexBytes` since it's a derivative of `bytes` and pydantic will serialize it as such.

Aviksaikat: done

docs/userguides/accounts.md
```diff
@@ -234,6 +234,27 @@ message = encode_defunct(text="Hello Apes!")
...
+from eth_account.messages import encode_defunct
+
+account = accounts.load("<ALIAS>")
+account.set_autosign(True)
+
+# Now, you will not be prompted to sign messages or transactions
+message = encode_defunct(text="Hello Apes!")
+signature = account.sign_message(message)
+
+# signature
```
fubuloubu, line 252: Probably don't need this set up given the rest of the docs 
```suggestion
```

docs/userguides/accounts.md
```diff
@@ -234,6 +234,27 @@ message = encode_defunct(text="Hello Apes!")
 signature = account.sign_message(message)
 ```
 
+## Signed Message
+
+Get the signed message signature as well as message hash signed by the account
```
fubuloubu, line 239: Can we reformat this section to show how to use the new property to recover a message signer?

fubuloubu (approved):
Update the docs and we are good to go

docs/userguides/accounts.md
Aviksaikat: I don't think we need `messageHash` to verify a contract. Ex. usecase would be to verify contracts. As mentioned here https://eips.ethereum.org/EIPS/eip-191

fubuloubu: > I don't think we need `messageHash` to recover a signer. Ex. use case would be to verify contracts. As mentioned here https://eips.ethereum.org/EIPS/eip-191

did you mean to link something else?

Aviksaikat: Nope. I wanted to show this EIP to make sure I'm understanding it correctly. here is a use case I found using web3.py https://github.com/0xMaka/w3py/blob/d4c6d945002fa3a6990dba19cd8c6ee970156dbe/permit/pysign.py

fubuloubu: I don't think that's an EIP

In general, most signature-based EIPs require passing individual data elements through the contract call so they can be individually verified before they are hashed together

Aviksaikat: Do you have any examples in mind ?

src/ape/types/signatures.py
```diff
@@ -22,6 +22,7 @@ class _Signature:
     v: int
     r: bytes
     s: bytes
+    messageHash: Optional[bytes] = None
```
antazoey, line 25:
```suggestion
    message_hash: Optional[bytes] = None
```

Can we do this?

docs/userguides/accounts.md
Aviksaikat: I'm still confused about the use case

fubuloubu: Well that's sorta my point, I don't think I know of a single production example where this happens in practice

I thought this field was for help in doing recovery off-chain

Aviksaikat: Maybe something like this? Say we have an anonymous messaging system or something like that.
1. Alice sends a message & the system stores the message_hash. 
2. Bob sends the same message & pretends to be Alice. 
3. By verifying the message_hash we can say that Bob is doing something suspicious 

All without passing any signatures ? If it makes sense.

docs/userguides/accounts.md
```diff
@@ -234,6 +234,23 @@ message = encode_defunct(text="Hello Apes!")
...
+
+```py
+from ape.types.signatures import recover_signer
```
antazoey, line 242: Is this import needed? `recover_signer` is imported but not even used in his example.

docs/userguides/accounts.md
```diff
@@ -234,6 +234,23 @@ message = encode_defunct(text="Hello Apes!")
...
+
+[...]
+signature = account.sign_message(message)
+
+# signature
+print(signature.encode_vrs())
+# message hash
+print(signature.message_hash)
+
```
antazoey, line 251:
```suggestion
```

docs/userguides/accounts.md
Aviksaikat: The final docs are not updated. It's a draft & waiting for dog

antazoey: Waiting for the busiest person at ApeWorX instead of working with me who is directly trying to help you now is not a good strategy.

Aviksaikat: Don't get me wrong I could use all the help I get 😅

Aviksaikat: I may be the first person in the world to get a thumbs down in a pr conversation 😶

docs/userguides/accounts.md
Aviksaikat: Can you suggest what examples I should put here ?

antazoey (dismissed):

docs/userguides/accounts.md
```diff
@@ -234,6 +234,22 @@ message = encode_defunct(text="Hello Apes!")
...
 ```
 
+## Signed Message
+
+Get the signed message signature as well as message hash signed by the account
+
+```py
+from ape.types.signatures import recover_signer
+
+[...]
```
fubuloubu, line 244:
```suggestion
```

docs/userguides/accounts.md
```diff
@@ -234,6 +234,22 @@ message = encode_defunct(text="Hello Apes!")
...
+signature = account.sign_message(message)
+
+# signature
+print(signature.encode_vrs())
+# message hash
+print(signature.message_hash)
```
fubuloubu, line 250: Was picturing something where you could recover the signer using only the signature object as part of the example:
```suggestion
# get original signer
print(recover_signer(...))
```

Aviksaikat: I don't think we can use `message_hash` to do that off chain. We have to use `ecrecover`

antazoey:
After looking at this closer, I agree field 0 is missing from our class and this PR adds it.
This is why I was struggling to calculate the hash in eip712 messages - field 0 is missing.

https://eth-account.readthedocs.io/en/stable/eth_account.html#eth_account.datastructures.SignedMessage.messageHash

so you can't properly formulate the hash, you have to rely on eth-account internal framework.

That is why in my eip712 PR https://github.com/ApeWorX/eip712/pull/40 I had to pre-pend field 0 to the hash calculation.

Unless I am misunderstanding but looking at the eth-accounts, docs it seems like that is what we need.

So for the docs example feedback, you could do what I did in the eip712 PR as an example:


```
full_hash = HexBytes(keccak(b"".join([msg.message_hash, *msg]))
```python

and as breaking change, we could include it in the `__iter__`

Aviksaikat:
As my discussion with @fubuloubu on discord, we have decided to keep `message_hash` as property as well as the `message` that is being passed when signing & we can use them to recover the `public key` with a method inside the account class.

fubuloubu:
> As per my discussion with @fubuloubu on Discord, we have decided to keep `message_hash` as property as well as the `message` that is being passed when signing & we can use them to recover the `public key` with a method inside the account class.

I didn't necessarily want to keep both fields, just enough of the data such that we could add a `.signer` computed property if the data was available

antazoey:
> I didn't necessarily want to keep both fields, just enough of the data such that we could add a `.signer` computed property if the data was available

Honestly, more confused by the eth-account saying the messageHash is field 0. what is field 0? it doesn't seem like that's a thing or it seems like is talking about something other than what it is

Aviksaikat:
What will be the purpose of the `.signer` property ?

fubuloubu:
> What will be the purpose of the `.signer` property @fubuloubu ?

Have the signer address for a signature without needing to have the original message

Aviksaikat:
isn't `acccount.address` will do the work ?

fubuloubu:
> isn't `acccount.address` will do the work ?

If you know what account signed it

Issue #1736, opened 2023-11-14
bug: BaseModel error: AttributeError: 'super' object has no attribute '__getattr__'


mikeshultz:
### Environment information

- OS: linux
- Python Version: 3.10.13
- `ape` and plugin versions:

```sh
$ ape --version
0.6.25

$ ape plugins list
# ...copy and paste result of above command here...
```

**NOTE**: unable to provide the last one due to another (probably unrelated?) error in plugin iteration.

### What went wrong?

Accessing non-ape or model props on `BaseModel` results in an `AttributeError`.  Here's an repro example:

```python
from ape.utils.basemodel import BaseModel


class MyModel(BaseModel):
    foo: str
    bar: int


m = MyModel(foo="abc", bar=123)
print("dump:", m.model_dump())
```

And raises the following:

```
Traceback (most recent call last):
  File "/home/mike/.venvs/ape-test/lib/python3.10/site-packages/ape/utils/basemodel.py", line 206, in __getattr__
    return super().__getattr__(name)
AttributeError: 'super' object has no attribute '__getattr__'. Did you mean: '__setattr__'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mike/dev/silverback/model_test.py", line 10, in <module>
    print("dump:", m.model_dump())
  File "/home/mike/.venvs/ape-test/lib/python3.10/site-packages/ape/utils/basemodel.py", line 227, in __getattr__
    raise ApeAttributeError(message)
ape.exceptions.ApeAttributeError: 'MyModel(foo='abc', bar=123)' has no attribute 'model_dump'
```

This occurs on both Pydantic v1 and v2.

Interestingly, `m.dict()` does work.  Presumably because `__getattr__()` is never called?  I'm not entirely sure.

### How can it be fixed?

I'm not entirely sure, maybe `__getattr__()` shouldn't exist on Ape's `BaseModel`.  It seems that `__getattr__()` [was removed from `BaseModel` back in v0.32](https://github.com/pydantic/pydantic/pull/712).

@antazoey thoughts?

antazoey:
Just a note: `model_dump()` is NOT available on v1 so an AttributeError seems expected in that case.
I do notice that there is no super().__getattr__() though so that part can be reworked a little but I don't think it is causing any issues?

I guess I am a little confused on the main problem you are facing. `dict()` should work on v1 and for v2 proper you need to call `model_dump()`. If you are using v1 back imports, you must call dict().

[...]

PR #1739, opened 2023-11-16
fix: use getattribute [APE-1539]

antazoey:
### What I did

fixes: #1736 

### Checklist

### DIFF
--- a/src/ape/utils/basemodel.py
+++ b/src/ape/utils/basemodel.py
@@ -203,7 +203,7 @@ class BaseModel(_BaseModel):
         """
 
         try:
-            return super().__getattr__(name)
+            return super().__getattribute__(name)
         except AttributeError:
             extras_checked = set()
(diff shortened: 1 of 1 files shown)


mikeshultz (approved):
That sure is some confusing method naming there.  LGTM

fubuloubu:
> That sure is some confusing method naming there. LGTM

https://stackoverflow.com/questions/3278077/difference-between-getattr-and-getattribute

antazoey:
> That sure is some confusing method naming there. LGTM

You mean the dunders??

mikeshultz:
> > That sure is some confusing method naming there. LGTM
> 
> You mean the dunders??

Yeah.  Having both `__getattr__` and `__getattribute__` with fairly different behaviors seems like a bad call.  Maybe it makes sense under the hood but from an API user perspective, it's confusing.

PR #1742, opened 2023-11-17
feat: add ecosystem to build file structure in ProjectAPI [APE-1540]

NotPeopling2day:
### What I did

Allow build structure to support multiple ecosystems.

### How I did it

- Added ecosystem to build file structure 
-  ".build" / "ethereum" / "__local__.json"

### How to verify it

- Updated and passes tests

### Checklist

- [X] New test cases have been added
- [X] Documentation has been updated

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -158,7 +158,8 @@ class ProjectAPI(BaseInterfaceModel):
 
     @property
     def _cache_folder(self) -> Path:
-        folder = self.contracts_folder.parent / ".build"
+        current_ecosystem = self.network_manager.network.ecosystem.name
+        folder = self.contracts_folder.parent / ".build" / current_ecosystem
         # NOTE: If we use the cache folder, we expect it to exist
         folder.mkdir(exist_ok=True, parents=True)
         return folder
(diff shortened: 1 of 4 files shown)


fubuloubu (commented):
In general, many networks copy ethereum and can use the same compiled code across many networks. Some networks are stuck on previous EVM versions, where it would make sense to compile a source with a different compiler configuration (actually the same compiler version but using a different setting in config for evm version). Some networks (like zksync) use entirely different compilers for the same source. And some networks define their own custom compilers for their own programming languages (in this case, there is no potential for conflict because the sources will be different)

I think there should be a different approach, possible an update required in ethpm-types, where contract types could be compiled using multiple compiler configurations (compiler version + settings, as well as different compilers). The data from these multiple compilations should be stored under that contract type keyed by compiler configuration in a structure like:

```
contractTypes:
  [name]:
    [compiler-key]:
      ... # artifact data
```

This way, multiple compilations for a given contract type can be switched depending on the compiler configuration required for deployment on a given network

To assist in that discovery, it could make sense to also store a relational mapping of networks to compiler configs, or update the `Deployment` item in ethpm-type to contain the `compiler-key` in addition to the `contractName` (which keys back to `contractTypes`)

PR #1744, opened 2023-11-17
fix: do not fetch history if there is no history in poll_blocks()

### LINKED ISSUE #1743: bug: BlockQuery validation error when calling poll_blocks close to head
### Environment information

- OS: linux
- Python Version: 3.11.5
- `ape` and plugin versions:

```sh
(ape) [mike@vemik ape]$ ape --version
0.6.26.dev2+gf2511ae6
(ape) [mike@vemik ape]$ ape plugins list
Installed Plugins
  optimism    0.6.6
```

### What went wrong?

When calling `chain.blocks.poll_blocks(start_block=head+1)` when tracking close to the head, it will call `chain.block.range()` with an invalid non-inclusive range.

https://github.com/ApeWorX/ape/blob/f2511ae6064e9d369a2e16c9271f2cab6820aabd/src/ape/managers/chain.py#L290-L293

For instance, if the chain is at block 420, and we call `chain.blocks.poll_blocks(start_block=420)`.  Here's an example test case that will fail:

```python
def test_poll_blocks_future(chain_that_mined_5, eth_tester_provider, owner, PollDaemon):
    blocks = list(
        chain_that_mined_5.blocks.poll_blocks(start_block=chain_that_mined_5.blocks.head.number + 1)
    )
    assert len(blocks) == 5
    assert blocks[0].number == 5
```

With this error:

```
chain_that_mined_5 = <ChainManager (id=1337)>, eth_tester_provider = <test chain_id=1337>
owner = <TestAccount 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C>, PollDaemon = <class 'tests.functional.conftest.PollDaemonThread'>

    def test_poll_blocks_head(chain_that_mined_5, eth_tester_provider, owner, PollDaemon):
>       blocks = list(
            chain_that_mined_5.blocks.poll_blocks(start_block=chain_that_mined_5.blocks.head.number + 1)
        )

tests/functional/test_block_container.py:160: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/ape/managers/chain.py:296: in poll_blocks
    for block in self.range(start_block, self.height - required_confirmations + 1):
src/ape/managers/chain.py:225: in range
    query = BlockQuery(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

__pydantic_self__ = BlockQuery()
data = {'columns': ['num_transactions', 'hash', 'number', 'parent_hash', 'size', 'timestamp', ...], 'start_block': 6, 'step': 1, 'stop_block': 5}
values = {}, fields_set = set()
validation_error = ValidationError(model='BlockQuery', errors=[{'loc': ('__root__',), 'msg': "stop_block: '5' cannot be less than start_block: '6'.", 'type': 'value_error'}])

    def __init__(__pydantic_self__, **data: Any) -> None:
        """
        Create a new model by parsing and validating input data from keyword arguments.
    
        Raises ValidationError if the input data cannot be parsed to form a valid model.
        """
        # Uses something other than `self` the first arg to allow "self" as a settable attribute
        values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
        if validation_error:
>           raise validation_error
E           pydantic.v1.error_wrappers.ValidationError: 1 validation error for BlockQuery
E           __root__
E             stop_block: '5' cannot be less than start_block: '6'. (type=value_error)

../../.venvs/ape/lib/python3.11/site-packages/pydantic/v1/main.py:341: ValidationError
```

IMO, this isn't really an invalid query.  It's just saying "we want to start tracking the chain at this block" that just happens to be in the future a bit.

### How can it be fixed?

 I'm a little torn.  Here's a few options:

- Skip historical fetching in `poll_logs()` when `start_block` <= `self.height`
- Allow `stop_block` to be less than or equal `start_block` and just do `max(start_block, stop_block)` in the `BlockQuery` instantiation

I'm partial to the latter, but that alters the behavior of `range()` which might not be ideal.  The former is probably lowest impact with the least funky logic.  I'll see about putting together a PR for the latter but let me know if there are other opinions.

mikeshultz:
### What I did

This alters the behavior of `chain.blocks.poll_blocks()` to prevent an out of bounds `range()` call.

Fixes #1743

### How to verify it

Could use a sanity check to make sure I'm not logically skipping a block with the change.

### Checklist

Do docs need to be updated?  I believe this is just a bugfix not a behavior change.

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -287,9 +287,10 @@ class BlockContainer(BaseManager):
         # Get number of last block with the necessary amount of confirmations.
         block = None
 
-        if start_block is not None:
+        head_minus_confirms = self.height - required_confirmations
+        if start_block is not None and start_block <= head_minus_confirms:
             # Front-load historical blocks.
-            for block in self.range(start_block, self.height - required_confirmations + 1):
+            for block in self.range(start_block, head_minus_confirms + 1):
                 yield block
 
         if block:
--- a/tests/functional/test_block_container.py
+++ b/tests/functional/test_block_container.py
@@ -147,3 +147,22 @@ def test_poll_blocks_timeout(
     with pytest.raises(ChainError, match=r"Timed out waiting for new block \(time_waited=1.\d+\)."):
         with PollDaemon("blocks", poller, lambda x: None, lambda: False):
             time.sleep(1.5)
+
+
+def test_poll_blocks_future(chain_that_mined_5, eth_tester_provider, owner, PollDaemon):
+    blocks: Queue = Queue(maxsize=3)
+    poller = chain_that_mined_5.blocks.poll_blocks(
+        start_block=chain_that_mined_5.blocks.head.number + 1
+    )
+
+    with PollDaemon("blocks", poller, blocks.put, blocks.full):
+        # Sleep first to ensure listening before mining.
+        time.sleep(1)
+        eth_tester_provider.mine(3)
+
+    assert blocks.full()
+    first = blocks.get().number
+    second = blocks.get().number
+    third = blocks.get().number
+    assert first == second - 1
+    assert second == third - 1


fubuloubu (approved):

fubuloubu:
@antazoey just double check and we can merge

antazoey (approved):
LGTM

antazoey:
im still not sure why the issue suddenly appeared... but here is the fix, you can rebase after this: https://github.com/ApeWorX/ape/pull/1745

mikeshultz:
> im still not sure why the issue suddenly appeared... but here is the fix, you can rebase after this: #1745

I think it just showed up now because the exception fixed in #1739 was masking this issue.

PR #1745, opened 2023-11-21
fix: enum values were raw [APE-1551]

antazoey:
### What I did

* Issue where all enums were values instead of enum types
* Prevents exceptions in `__repr__`
* Fix issues where plugin type was a str instead of enum
* add some tests
* fix issue where geth connection ID was not very identifiable

### Checklist

### DIFF
--- a/src/ape_plugins/utils.py
+++ b/src/ape_plugins/utils.py
@@ -5,7 +5,7 @@ from functools import cached_property
 from typing import Iterator, List, Optional, Sequence, Set, Tuple
 
 from ape.__modules__ import __modules__
-from ape._pydantic_compat import root_validator
+from ape._pydantic_compat import root_validator, validator
 from ape.logging import logger
 from ape.plugins import clean_plugin_name
 from ape.utils import BaseInterfaceModel, get_package_version, github_client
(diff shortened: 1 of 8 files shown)


src/ape_plugins/utils.py
```diff
@@ -373,14 +373,27 @@ def __bool__(self) -> bool:
...
+            # Prevent exceptions happening in repr()
+            logger.log_debug_stack_trace()
+            return "<PluginGroup>"
 
     def __str__(self) -> str:
         return self.to_str()
 
+    @validator("plugin_type")
+    def validate_plugin_type(cls, value):
+        return PluginType(value) if isinstance(value, str) else value
```
mikeshultz, line 388: question: So was something not initializing with a valid enum?

mikeshultz (approved):
LGTM

Thanks for tracking that down.  Sure don't see Python segfaults very often!

mikeshultz:
I jumped the gun on approval, I guess.  Tests sure don't seem happy about that change.

src/ape_plugins/utils.py
antazoey: somehow yes but I don't see where! I can't find he instance where we do that anywhere in our code, that's why im quite confused... it seems like sometimes the `Enum` part doesnt work in pydantic.

antazoey:
> I jumped the gun on approval, I guess. Tests sure don't seem happy about that change.

it should be good now

fubuloubu (approved):

antazoey:
please merge ASAP

PR #1755, opened 2023-12-05
fix: issue where `KeyfileAccount` would not stay unlocked [APE-1586]

antazoey:
### What I did

In ape-safe, I was prompted `"Leave unlocked?"` over and over again and it was because signers were re-checked and reloaded and thus lost the "yes" to my prompt... So it is would be nice if Ape remembered I said "yes" to leave the account unlocked.

### How I did it

Only create the `KeyfileAccount` class once.

### How to verify it

see second test. The first one was me trying to figure out but is still a nice test

### Checklist

### DIFF
--- a/src/ape/managers/project/manager.py
+++ b/src/ape/managers/project/manager.py
@@ -752,7 +752,8 @@ class ProjectManager(BaseManager):
 
         if destination.is_file():
             logger.debug("Deployment already tracked. Re-tracking.")
-            destination.unlink()
+            # NOTE: missing_ok=True to handle race condition.
+            destination.unlink(missing_ok=True)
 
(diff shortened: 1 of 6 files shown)


antazoey:
I guess one may want it to be relocked on when calling `load()` again? in which case, the caching can happen at the ape-safe level.
But it is just prone to be a problem again in future applications...

There is `.lock()` method you can and should call anyway.. to relock the account... It was just frustrating to have always bash "yes, leave unlocked" and it is just never stayed unlocked-

fubuloubu (dismissed):

antazoey:
was approved before, all i did was make tests happy since

PR #1758, opened 2023-12-07
fix: allow `None` values in `Ethereum.create_transaction()`

antazoey:
### What I did

Handle some explicit-None issues... like int ops on None and weird pydantic ValidationError

### How I did it

filter them out so Pydantic uses defaults for None

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -639,7 +639,9 @@ class Ethereum(EcosystemAPI):
         if isinstance(kwargs.get("chainId"), str):
             kwargs["chainId"] = int(kwargs["chainId"], 16)
 
-        elif "chainId" not in kwargs and self.network_manager.active_provider is not None:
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -660,7 +662,16 @@ def create_transaction(self, **kwargs) -> TransactionAPI:
...
+
+        # This causes problems in pydantic for some reason.
+        if "gas_price" in kwargs and kwargs["gas_price"] is None:
+            del kwargs["gas_price"]
+
+        # None is not allowed, the user likely means `b""`.
+        if "data" in kwargs and kwargs["data"] is None:
+            kwargs["data"] = b""
 
         return txn_class(**kwargs)
```
antazoey, line 676: note: can't remove all None values because for some reason, the validators get all fudged up with gas_limit when i do... so every field is handled case by case

Issue #1572, opened 2023-07-28
Have `NetworkBoundCommand` automatically add `network_option` if it is not there. [APE-1239]


antazoey:
### Overview

This would be a breaking chance for using `NetworkBoundCommand` without `network_option()`, but that is a rare case.
I think it would save so much code if we can find a way to have `NetworkBoundCommand` automatically use the network option.
Currently, using `NetworkBoundCommand` without `network_option` means `**run this command on the default network**`.

### Specification

```python
import click
from ape.cli import NetworkBoundCommand

@click.command(cls=NetworkBoundCommand)
def cli(network):
    """Notice I did not have to use @network_option"""
    print(f"Connected to {network}")
```

**NOTE** : could do this at the same time as https://github.com/ApeWorX/ape/issues/1401

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

Issue #1401, opened 2023-04-23
Expand `network_option` argument to be more useful [APE-881]


fubuloubu:
### Overview

Currently, `network_option` will provide a variable `network` that just reflects the flag as passed, whether it's the default e.g. `"ethereum"` or a specific triple e.g. `"ethereum:local:test"`

However, this is not too helpful, because often I want to know the name of the ecosystem, network (and maybe the provider)

### Specification

At a basic level of consistency, this flag should expand to what the current triple is that it's using, even if a default

It would additional be nicer if I could specific `ecosystem`, `network`, and `provider` *optionally* so that I can leverage whatever I need (introspect the context) e.g.:
```py
@click.command(cls=NetworkBoundCommand)
@network_option()
def stuff(ecosystem, network, provider, ...):
# or def stuff(ecosystem, network, ...):
# or def stuff(network, ...):
# or def stuff(...):
# etc....
    ...
```

Further, it would be *most* useful if these were references to the specific objects, instead of just their names, so that I could use them. I often use this feature expecting it to be a `NetworkAPI` subclass, not realizing it's just a string and not even the string I want

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

antazoey:
What if it just returned the `provider`? Some users may not want so many params in their tests.
And you could do this:

```
provider.network
provider.network.ecosystem
```

to access the same objects.

antazoey:
Oh I realize you mean to fill in these optionally. I see now. That is awesome actually.

fubuloubu:
> realizing you mean optionally fill in those params. That may be do-able, that would be a nice feature

Yeah, I mean half the time I don't want _any_ of those arguments in my function, I just want it to connect for me and that's it

PR #1764, opened 2023-12-10
feat!: network customization and tooling improvements

antazoey:
### What I did

* `NetworkBoundCommand` renamed to `ConnectedProviderCommand` and no longer requires explicitly setting the network option (base wasn't removed and warns a DeprecatedWarning with instructions how to migrate and it has the legacy behavior).
* Both `@network_option` and `ConnectedProviderCommand` allow key-args in your callback for `ecosystem`, `network,` and `provider` for each instance respectively. You don't need to use any of them either (yay, can delete all the `_ = network  # needed for NetworkBoundCommand` lines).
* Move another class to `ape_ethereum` called `EthereumNodeProvider`.. It is exactly what was in ape-geth, however ape-geth still sorta re-exports because `EthereumNodeProvider` is the default class for custom networks and such and just sounds better and in some cases may be a better base class than `Web3Provider` alone. So on the node, `adhoc` neworks are rebranded as `custom` networks so itll be smoother to add the config-based ones with names and chaind IDs and such.
* Adds some other Deprecated methods back to less the damager of 0.7 on scripts and emit Deprecated warnings and TODOs to delete on 0.8 which **is going to happen**.

fixes: #1572 
Fixes: APE-1239
fixes: #1401 
Fixes: APE-881
Fixes: APE-486

### How I did it

when you use `@network_option` now, it analyzes the fn  you give it via `inspect` and then it checks out `sys.argv` to parse the value, since at option time it does not have access to it. It then turns the callback fn to a partial with the fields filled out.

`ConnectedProviderCommand` (f.k.a NetworkBoundCommand) changes were much easier to implement because it has access to code responsible for pass kwargs into your command callback.. so at that time, it is able to put the right ones in, also via using `inspect` to see what the callback is even requesting.

But because `NetworkBoundCommand` no longer needs `network_option` and most likely never will use it together because it is pointless unless you are being super custom with it, it will never have the partially-filled callback.. Thus it utiliizes is own way of passing in the instances..

it was a little more complex than i anticipated, mostly because Options are parsed so early on. If this only happened in `ConnectedProviderCommand`, it'd be easier, but I got it working nonetheless.

### Checklist

### DIFF
--- a/docs/userguides/clis.md
+++ b/docs/userguides/clis.md
@@ -63,14 +63,18 @@ def cli(cli_ctx):
 
 ## Network Tools
 
-The [@network_option()](../methoddocs/cli.html#ape.cli.options.network_option) allows you to select an ecosystem, network, and provider combination.
+The [@network_option()](../methoddocs/cli.html#ape.cli.options.network_option) allows you to select an ecosystem, network, and provider.
 To specify the network option, use values like:
 
 ```shell
---network ethereu
+--network ethereum
 --network ethereum:sepolia
 --network ethereum:mainnet:alchemy
+--network ::alchemy
 ```
+
+When omitting values, leave the semi-colons in the network choice for parsing.
+And omitted values means to use the defaul value, so if you omit the ecosystem out-of-the-box, it will default to Ethereum (unless you have configured a different default ecosystem).
 
 Use `ecosystem`, `network`, and  `provider` argument names in your command implementation to have access to the parsed network option:
 
@@ -95,7 +99,7 @@ def cmd_2(ecosystem, network, provider):
 
 The [ConnectedProviderCommand](../methoddocs/cli.html#ape.cli.commands.ConnectedProviderCommand) automatically uses the `--network` option and connects to the network before any of your code executes and disconnected after.
 This is useful if your script or command requires a provider connection in order for it to run.
-Additionally, specify `ecosystem`, `network`, or `provider` in your command function if you need any of those instances in your ``ConnectedProviderCommand`.
+Additionally, specify `ecosystem`, `network`, or `provider` in your command function if you need any of those instances in your `ConnectedProviderCommand`.
 
 ```python
 import click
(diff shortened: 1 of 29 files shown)


antazoey:
Now, we won't need unused `network` values in our CLIs like:

```
_ = network  # Used for NeworkBoundCommand
```

And also if we do need it, it is actually useful.

antazoey:
after this PR, I want to all config-based adhoc network customization so it doesnt have to say "adhoc" and you can tell it the name and chain Id and whatever else

docs/userguides/clis.md
```diff
@@ -63,28 +63,60 @@ def cli(cli_ctx):
...
+--network ethereum
+--network ethereum:sepolia
+--network ethereum:mainnet:alchemy
+--network ::foundry
+```
+
+To use default values automatically, omit sections of the choice, but leave the semi-colons for parsing.
+For example, `::test` means use the default ecosystem and network and the `test` provider.
+
+Use `ecosystem`, `network`, and `provider` argument names in your command implementation to access their corresponding class instances:
```
fubuloubu, line 79: Nice

fubuloubu (approved):
General thoughts:

1. When this goes out, we need to educate folks how to upgrade their CLIs or it may cause problems
2. Conversely, we can have `NetworkBoundCommand` just raise an error demonstrating how to upgrade across the 0.7.x series

And a question:
1. I can use `@network_option` without `ConnectedProviderCommand` and it won't connect?

antazoey:
> 1. I can use `@network_option` without `ConnectedProviderCommand` and it won't connect?

Correct, using the option does not connect, but that isn't a changing behavior, it already works that way.

And yes, good point, I can some deprecated methods back in and do the deprecated warning thing.

antazoey:
@fubuloubu A good example of using `@network_option` without being connected is the `ape cache init` methods, which don't need to connect but sill needs the network value given for creating the DB. I made sure that still works.

Also :

- [x] Put back `NetworkBoundCommand` with deprecation warning and legacy behavior.

antazoey:
- [ ] fix a wrong-instance issue

fubuloubu (approved):

Issue #1741, opened 2023-11-17
Update the `urllib3` module version or remove the upper bound requirement


Aviksaikat:
### Overview

Provide a simple overview of what you wish to see added. Please include:

Update the `urllib3` module version or remove the upper-bound.
Here 
https://github.com/ApeWorX/ape/blob/f2511ae6064e9d369a2e16c9271f2cab6820aabd/setup.py#L115

### Dependencies

```sh
urllib3
```

PR #1771, opened 2023-12-12
feat: update setup.py, removal of urllib3 module version upper boundary requirement [APE-1610]

popraf:
Update setup.py - removal of urllib3 module version requirement

### What I did

fixes: #1741

### How I did it

Updated `setup.py` by removal of `urllib3` module version upper boundary requirement
### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ setup(
         "SQLAlchemy>=1.4.35",
         "tqdm>=4.62.3,<5.0",
         "traitlets>=5.3.0",
-        "urllib3>=1.26.16,<2",
+        "urllib3>=1.26.16",
         "watchdog>=3.0,<4",
         # ** Dependencies maintained by Ethereum Foundation **
         "eth-abi>=4.1.0,<5",


fubuloubu (approved):

fubuloubu:
Can you give a quick explanation why the library pin is troublesome?

Aviksaikat:
> Can you give a quick explanation why the library pin is troublesome?

Hey, thanks for looking into this. As ape is not using urllib heavily I think it's safe to remove the upper bound as it creates conflicts with other libraries that depend on specific versions of urllib.

antazoey:
We could also bump it completely to the 2.0 range, as we are about to do big breaking change release

setup.py
```diff
@@ -112,7 +112,7 @@
         "SQLAlchemy>=1.4.35",
         "tqdm>=4.62.3,<5.0",
         "traitlets>=5.3.0",
-        "urllib3>=1.26.16,<2",
+        "urllib3>=1.26.16",
```
antazoey, line 115:
```suggestion
        "urllib3>=2.1.0,<3",
```

antazoey (commented):
bump up instead?

setup.py
antazoey: i bumped here instead: https://github.com/ApeWorX/ape/pull/1773

PR #1781, opened 2023-12-18
fix: issue with interactive scripts would not be connected to right provider

antazoey:
### What I did

Two bugs prevented this feature from working for me, I tried both on main and on the current released Ape.

1. interactive mode was honored in some cases, for the released mode I think that was only the default network.
2. The logic for checking for the interactive flag in the parents context had an issue because it used `False` as a default and then checked for None before looking in parents, and that never happened, so some cases interactive mode was not known during recycling provider.

### Checklist

### DIFF
--- a/src/ape/cli/commands.py
+++ b/src/ape/cli/commands.py
@@ -1,26 +1,43 @@
 import inspect
 import warnings
-from typing import Any, List
+from typing import Any, List, Optional
 
 import click
 from click import Context
 
 from ape import networks
-from ape.api import ProviderAPI
+from ape.api import ProviderAPI, ProviderContextManager
 from ape.cli.choices import NetworkChoice
 from ape.exceptions import NetworkError
 
 
-def check_parents_for_interactive(ctx: Context) -> bool:
-    interactive: bool = ctx.params.get("interactive", False)
-    if interactive:
-        return True
+def get_param_from_ctx(ctx: Context, param: str) -> Optional[Any]:
+    if value := ctx.params.get(param):
+        return value
 
     # If not found, check the parent context.
-    if interactive is None and ctx.parent:
-        return check_parents_for_interactive(ctx.parent)
+    elif parent := ctx.parent:
+        return get_param_from_ctx(parent, param)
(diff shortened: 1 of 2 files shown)


src/ape/cli/commands.py
```diff
@@ -1,26 +1,43 @@
...
 
 from ape import networks
-from ape.api import ProviderAPI
+from ape.api import ProviderAPI, ProviderContextManager
 from ape.cli.choices import NetworkChoice
 from ape.exceptions import NetworkError
 
 
-def check_parents_for_interactive(ctx: Context) -> bool:
-    interactive: bool = ctx.params.get("interactive", False)
```
antazoey, line 15: bug 1: was defaulting to False here when it shouldnt be because it hasnt checked the parents yet.. you can see below we check for parents only if None here (unspecified)

src/ape/cli/commands.py
```diff
@@ -1,26 +1,43 @@
...
+        return get_param_from_ctx(parent, param)
+
+    return None
+
+
+def parse_network(ctx: Context) -> ProviderContextManager:
+    interactive = get_param_from_ctx(ctx, "interactive")
+    provider = get_param_from_ctx(ctx, "network")
+    if provider is not None and isinstance(provider, ProviderAPI):
+        return provider.network.use_provider(provider, disconnect_on_exit=not interactive)
```
antazoey, line 29: bug2: ` disconnect_on_exit=not ` was only on one of the cases here, it didnt occur on the `elif` statements below... so that setting didn't work

src/ape/cli/commands.py
```diff
@@ -1,26 +1,43 @@
...
-        return check_parents_for_interactive(ctx.parent)
-
-    return False
+    elif parent := ctx.parent:
+        return get_param_from_ctx(parent, param)
+
+    return None
+
+
+def parse_network(ctx: Context) -> ProviderContextManager:
```
antazoey, line 25: refacorted these into their own methods so i can just easily without having to actually connect

PR #1788, opened 2023-12-19
fix: issue where would not get a transaction error when expected

antazoey:
### What I did

after replaying a tx to try and get a proper web3.py extracted error, if that don't work, we were not erroring.

### How I did it

if get to a certain point, call `receipt.raise_for_status()`

### How to verify it

trust me lol...
i dont know how the chain gets in this condition but it seems like it happens on foundry in a forked network when using an impersonated account, or at least that is when i have seen it.

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -362,7 +362,7 @@ class Ethereum(EcosystemAPI):
         elif "input" in data and isinstance(data["input"], str):
             data["input"] = HexBytes(data["input"])
 
-        block_number = data.get("block_number") or data.get("blockNumber")
+        block_number = data.get("block_number", data.get("blockNumber"))
         if block_number is None:
             raise ValueError("Missing block number.")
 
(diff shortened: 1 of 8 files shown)


src/ape_ethereum/transactions.py
```diff
@@ -262,7 +262,11 @@ def decode_logs(
             Union[List[Union[EventABI, "ContractEvent"]], Union[EventABI, "ContractEvent"]]
         ] = None,
     ) -> ContractLogContainer:
-        if abi is not None:
+        if not self.logs:
```
antazoey, line 265: author's note: the changes in the file are nice, but are a product of investigating missing events on a receipt.

so the receipt was actually failing super early, and thus didnt have the logs, so no actual change here but 1 small perf and some refactoring.

src/ape_ethereum/ecosystem.py
```diff
@@ -362,7 +362,7 @@ def decode_receipt(self, data: Dict) -> ReceiptAPI:
         elif "input" in data and isinstance(data["input"], str):
             data["input"] = HexBytes(data["input"])
 
-        block_number = data.get("block_number") or data.get("blockNumber")
+        block_number = data.get("block_number", data.get("blockNumber"))
```
fubuloubu, line 365: non-blocking: what's the issue here?

fubuloubu (approved):

src/ape_ethereum/ecosystem.py
antazoey: just slightly more correct in the odd world if block 0 had logs.
if we were in that case (which tbh not sure is possible in any Eths), it would use the value of `None` here instead.

so more correctness

Issue #1790, opened 2023-12-19
`AddressType` Validation Error


Aviksaikat:
### Environment information

- OS: linux
- Python Version:3.9.0
- `ape` and plugin versions:

```sh
$ ape --version
0.7.0

$ ape plugins list
# ...copy and paste result of above command here...
N/A
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
# ...copy and paste result of above command here...
N/A
```

### What went wrong?

Please include information like:

I'm using ape's utilities rather than the module so there's not plugins and `ape-config`. 
So I made a class like this 

```py
class SOCReader(BaseModel): # pydantic BaseModel
    """
    SOCReader model.

    Attributes:
      owner: The owner of the SOCReader.
      download: A function to download a single owner chunk.
    """

    owner: AddressType # here owner is of type AddressType
```

- Input value `307838643337363634343066306437623934396135653332393935643039363139613766383665363332` or `b'0x8d3766440f0d7b949a5e32995d09619a7f86e632'` when wrapped using `HexBytes` of `eth_pydantic_types` . Now when I'm passing the value to `owner` variable like this I'm getting the following error

```py
address = b'0x8d3766440f0d7b949a5e32995d09619a7f86e632' # type: HexBytes 
SOCReader(owner=address)
```

```py
   return SOCReader(owner=make_hex_eth_address(canonical_owner), download=download)
~/site-packages/ape/types/address.py:36: in __eth_pydantic_validate__
    cls.conversion_manager.convert(value, AddressType)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ConversionManager>
value = HexBytes('0x307838643337363634343066306437623934396135653332393935643039363139613766383665363332')
type = typing.Annotated[ChecksumAddress, <class 'ape.types.address._AddressValidator'>]

    def convert(self, value: Any, type: Union[Type, Tuple, List]) -> Any:
        """
        Convert the given value to the given type. This method accesses
        all :class:`~ape.api.convert.ConverterAPI` instances known to
        `ape`` and selects the appropriate one, so long that it exists.
    
        Raises:
[...]

PR #1791, opened 2023-12-19
test: prove that `AddressType` works

antazoey:
### What I did

add tests to prove AddressType works for https://github.com/ApeWorX/ape/issues/1790

### Checklist

### DIFF
--- a/tests/functional/test_types.py
+++ b/tests/functional/test_types.py
@@ -3,8 +3,10 @@ from typing import Dict
 import pytest
 from eth_utils import to_hex
 from ethpm_types.abi import EventABI
+from hexbytes import HexBytes
+from pydantic import BaseModel
 
-from ape.types import ContractLog, LogFilter, SignableMessage, TransactionSignature
+from ape.types import AddressType, ContractLog, LogFilter, SignableMessage, TransactionSignature
 from ape.utils import ZERO_ADDRESS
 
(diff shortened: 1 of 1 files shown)


tests/functional/test_types.py
```diff
@@ -147,3 +149,24 @@ def test_signature_from_rsv_and_vrs(signature):
...
+    # Show str works.
+    instance_str = MyModel(addr=owner.address)
+    assert instance_str.addr == owner.address
+
+    # Show hex bytes work.
+    instance_hex_bytes = MyModel(addr=HexBytes(owner.address))
+    assert instance_hex_bytes.addr == owner.address
+
+    # Show raw bytes work.
+    instance_bytes = MyModel(addr=bytes.fromhex(owner.address[2:]))
```
antazoey, line 167: @Aviksaikat This is the case you are in.  If you are using raw bytes, make sure it looks the same as doing `bytes.fromhex()` on the hex str _without_ the 0x prefix... or even simpler, you can just use the hex str directly!

thank you

fubuloubu: In general, we are using Pydantic v2 now, which is less forgiving on various "implicit conversions" that may have been happening before

Aviksaikat: Cool. :+1:

antazoey: Bad things can also happen if we try to handle bytes encoded hex strings like `b"0x123"`, the mixing of the formats leads to ambiguity.

PR #1798, opened 2023-12-21
fix: use helpful `IndexError` in `AccountManager.__getitem__()`  when conversion error occurs

antazoey:
### What I did

issue 1: __getitem__ is supposed to raise IndexError when it fails but i was getting conversion errors.
issue 2: I accidentally did `ape.accounts["safe"]` instead of `ape.accounts.containers["safe"]` once and the error was confusing for a second and also was not an index error.
issue 3: when ape-ens or conversion plugin not installed, the index error should hint at that.

### How I did it

catch and handle `ConversionError` in account manager `__getitem__` implementation.

### How to verify it

```python
ape.accounts["BOBO"]
# then without ens installed
ape.accounts["antazoey.eth"]
```

### Checklist

### DIFF
--- a/src/ape/managers/accounts.py
+++ b/src/ape/managers/accounts.py
@@ -338,7 +338,7 @@ class AccountManager(BaseManager):
             account_id = self.conversion_manager.convert(account_str, AddressType)
         except ConversionError as err:
             prefix = f"No account with ID '{account_str}'"
-            if account_str.endswith(".ens"):
+            if account_str.endswith(".eth"):
                 suffix = "Do you have `ape-ens` installed?"
             else:
                 suffix = "Do you have the necessary conversion plugins installed?"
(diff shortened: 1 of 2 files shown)


src/ape/managers/accounts.py
```diff
@@ -329,7 +334,16 @@ def __getitem_str(self, account_str: str) -> AccountAPI:
             :class:`~ape.api.accounts.AccountAPI`
         """
 
-        account_id = self.conversion_manager.convert(account_str, AddressType)
+        try:
+            account_id = self.conversion_manager.convert(account_str, AddressType)
+        except ConversionError as err:
+            prefix = f"No account with ID '{account_str}'"
+            if account_str.endswith(".ens"):
```
fubuloubu, line 341:
```suggestion
            if account_str.endswith(".eth"):
```

fubuloubu (dismissed):

PR #1803, opened 2024-01-02
fix: decoding issues that affect tracing features such as coverage

### LINKED ISSUE #1777: Tests fail with --coverage while they are 100% pass without --coverage
### Environment information

- OS: ubuntu 22.*
- Python Version: 3.8.10
- `ape` and plugin versions: 0.6.20

```
Installed Plugins:
  infura       0.6.0
  alchemy      0.6.0
  template     0.6.1
  foundry      0.6.11
  solidity     0.6.6
  tokens       0.6.1
  vyper        0.6.11
  hardhat      0.6.13
  polygon      0.6.5
  etherscan    0.6.7
  ens          0.6.1
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
name: myContract

default_ecosystem: ethereum

# number_of_accounts: 5

plugins:
  - name: infura
  - name: alchemy
  - name: vyper
  - name: ens
  - name: etherscan 
  - name: hardhat

vyper:
  evm_version: istanbul  # bor doesn't support latest forks


geth:
  ethereum:
    mainnet:
      uri: http://myServer:8545
      chain_id: 109

hardhat:
  request_timeout: 30  # Defaults to 30
  fork_request_timeout: 600  # Defaults to 300
  fork:
    ethereum:
      mainnet:
        upstream_provider: geth

test:
  coverage:
    reports:
      terminal:
        verbose: true  # Show verbose coverage information in the terminal.

```

### What went wrong?

When I run my tests without `--coverage`, I get 100% PASS as a result. When I add --coverage, I get errors. 
It happens when I call my function that accepts an address as the only input parameter.

So in short:

`def myFunction(_address: address):`

And I test it with `myContract.myFunction(address_to_check, sender=address_in_adminlist)`

Ape tells me:

`from field must match key's 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C, but it was 0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7`

*Long version below*

```
def test_Profile(my_contract, owner, not_owner):
    my_contract.storeProfile('mango', owner , 1, 'hi Im nobody bla bla bla', sender=owner)
    assert my_contract.getProfile(owner, sender=owner) == ('mango', owner, 1,'hi Im nobody bla bla bla')
    assert my_contract.getProfile(owner, sender=not_owner) == ('mango', owner, 1,'hi Im nobody bla bla bla')
```
The contract function is literally this: Return a stored struct for the account that got passed.
There is no assertion that only the owner address can request their profile.

```
@view
@external
def getProfile(account: address) -> Profile:
    return self.profileStore[account]
```

And the error message is 

```
/home/ethbox/Vyper-Main/myContract/tests/test_myContract.py:70: in test_Profile
    assert my_contract.getProfile(owner, sender=not_owner) == ('mango', owner, 1,'hi Im nobody bla bla bla')
E   TypeError: from field must match key's 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C, but it was 0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7
        myContract = <myContract 0x274b028b03A250cA03644E6c578D81f019eE1323>
        not_owner  = <TestAccount 0xc89D42189f0450C2b2c3c61f58Ec5d628176A1E7>
        owner      = <TestAccount 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C>

```

And the way profiles are stored:

```py
struct Profile:
    # userID: uint256 
    userName: String[64]
    pfpContract: address
    pfpID: uint256
    bio: String[256]

profileStore: HashMap[address, Profile]
```

So for some reason it doesn't allow not_owner to call the getProfile function and pass not_owner as the sender and requesting the profile created by sender. Which doesn't make sense with the vyper function.
Also it should either succeed or fail in both scenarios, wether I test with or without --coverage.
So not consistent. 

To make things more exciting I fork shibarium with ape-hardhat (even though it's called ethereum in the ape-config. I just flipped the IP)
`ape test --network ethereum:mainnet-fork:hardhat --coverage`

### How can it be fixed?

Not entirely sure, seems like the address passed as a parameter (owner) is overwritten with the sender= address (not_owner).
This only gets triggered when I use --coverage.

antazoey:
### What I did

maybe resolves #1777

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -448,13 +448,13 @@ class Ethereum(EcosystemAPI):
         input_types = [parse_type(i.model_dump(mode="json")) for i in abi.inputs]
 
         try:
-            raw_input_values = decode(raw_input_types, calldata)
-            input_values = [
-                self.decode_primitive_value(v, t) for v, t in zip(raw_input_values, input_types)
-            ]
+            raw_input_values = decode(raw_input_types, calldata, strict=False)
         except InsufficientDataBytes as err:
             raise DecodingError(str(err)) from err
 
+        input_values = [
+            self.decode_primitive_value(v, t) for v, t in zip(raw_input_values, input_types)
+        ]
         arguments = {}
         index = 0
         for i, v in zip(abi.inputs, input_values):
@@ -467,10 +467,14 @@ class Ethereum(EcosystemAPI):
     def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
         output_types_str_ls = [o.canonical_type for o in abi.outputs]
 
-        try:
-            vm_return_values = decode(output_types_str_ls, raw_data)
-        except (InsufficientDataBytes, NonEmptyPaddingBytes) as err:
-            raise DecodingError(str(err)) from err
+        if raw_data:
+            try:
+                vm_return_values = decode(output_types_str_ls, raw_data, strict=False)
+            except (InsufficientDataBytes, NonEmptyPaddingBytes) as err:
+                raise DecodingError(str(err)) from err
+        else:
+            # Use all zeroes.
+            vm_return_values = tuple([0 for _ in output_types_str_ls])
 
         if not vm_return_values:
             return vm_return_values
--- a/tests/functional/test_ecosystem.py
+++ b/tests/functional/test_ecosystem.py
@@ -1,4 +1,3 @@
-import sys
 from typing import Any, Dict
 
 import pytest
--- a/tests/functional/test_ecosystem.py
+++ b/tests/functional/test_ecosystem.py
@@ -1,7 +1,8 @@
+import sys
 from typing import Any, Dict
 
 import pytest
-from eth_pydantic_types import HexBytes
+from eth_pydantic_types import HashBytes32, HexBytes
 from eth_typing import HexAddress, HexStr
 from ethpm_types.abi import ABIType, EventABI, MethodABI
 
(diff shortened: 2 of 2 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -467,10 +467,14 @@ def decode_calldata(self, abi: Union[ConstructorABI, MethodABI], calldata: bytes
...
-        except (InsufficientDataBytes, NonEmptyPaddingBytes) as err:
-            raise DecodingError(str(err)) from err
+        if raw_data:
+            try:
+                vm_return_values = decode(output_types_str_ls, raw_data, strict=False)
+            except (InsufficientDataBytes, NonEmptyPaddingBytes) as err:
+                raise DecodingError(str(err)) from err
+        else:
+            # Use all zeroes.
+            vm_return_values = tuple([0 for _ in output_types_str_ls])
```
antazoey, line 477: it seems like the user is getting hardhat to give them empty bytes for a 0 return in some weird case that i havent reproduced

src/ape_ethereum/ecosystem.py
```diff
@@ -467,10 +467,14 @@ def decode_calldata(self, abi: Union[ConstructorABI, MethodABI], calldata: bytes
     def decode_returndata(self, abi: MethodABI, raw_data: bytes) -> Tuple[Any, ...]:
         output_types_str_ls = [o.canonical_type for o in abi.outputs]
 
-        try:
-            vm_return_values = decode(output_types_str_ls, raw_data)
-        except (InsufficientDataBytes, NonEmptyPaddingBytes) as err:
-            raise DecodingError(str(err)) from err
+        if raw_data:
+            try:
+                vm_return_values = decode(output_types_str_ls, raw_data, strict=False)
```
antazoey, line 472: second change: use `strict=False` here to see if that resolves the problem. I remember having to do this somewhere else because of node oddities (probably hardhat!)

fubuloubu (approved):

Issue #1800, opened 2023-12-29
Dependency of a dependency using `branch: master` can't download


Schlagonia:
### Environment information

- OS:  linux 
- Python Version: 3.9.5
- `ape` and plugin versions:

```sh
$ ape --version
0.7.1

```

- Contents of your `ape-config.yaml` 

```sh

dependencies:
  - name: yearn-vaults
    github: yearn/yearn-vaults-v3
    ref: v3.0.1
    exclude: 
     - test/

```

### What went wrong?

- The yearn-vaults-v3 is an Ape repo that is being imported as a dependency which has its own dependency that is a foundry repo.
- The tagged version of yearn-vaults-v3 ape-config is using ape version== 0.6.3 and still using "branch" in the config and trying to use the master branch.

```sh

dependencies:
  - name: tokenized-strategy
    github: yearn/tokenized-strategy
    branch: master
    contracts_folder: src

```

compilation fails when using Ape >= 0.7.0 with folowing error:
![image](https://github.com/ApeWorX/ape/assets/89420541/f2598c38-3a51-4943-bb92-d41ec9e3e28d)
![image](https://github.com/ApeWorX/ape/assets/89420541/e58506de-aebc-4364-83ef-3d1880795032)

fubuloubu:
May be time to accelerate the "package manifest" approach to solving problems like this

antazoey:
I am guessing `branch:` is supposed to be `ref:` ?

antazoey:
Yes, changing it to `ref:` works. `ref:` was supposed to be made required and branch was deleted.  Well, branch got deleted but ref never got required, and that caused the code to try and download weird non-released files and thus break this logic here but if it didnt break here, itd break some where else.

I fixed the error handling of all that by making ref required on the attached PR.

to fix your issues specifically, replace `branch:` with `ref:` in your `ape-config.yaml`.

fubuloubu:
I think part of the problem here is that the dependency is an older version of ape, but it is code frozen to <0.7

antazoey:
> I think part of the problem here is that the dependency is an older version of ape, but it is code frozen to <0.7

An old Ape project's code should not matter. That only thing that would affect it is the ape-config.yaml file in the project, but you can use `config_overrides` in the top-level project to fix any issues there.

The attached PR will at least give you a better exception for the real error.

fubuloubu:
> > I think part of the problem here is that the dependency is an older version of ape, but it is code frozen to <0.7
[...]

PR #1805, opened 2024-01-02
fix: issues with GitHub dependencies when not specifying `ref:` or `version:`

antazoey:
### What I did

fixes: #1800 

Make ref or version required, at least one or other

### How I did it

simply make `ref:` required so the error is correct.

### Checklist


antazoey:
ok i was a bit wrong, ref should still be optional i see... hmm...

fubuloubu (approved):

PR #1810, opened 2024-01-03
refactor: move `__getattr__` overload to new mixin and only use where needed

antazoey:
### What I did

This should also be a performance gain.
But more importantly, it lessens the chance of problems occurring from this because now we are only using `__getattr__` overloading where extras are.

### How I did it

* Make new mixin
* Use new mixin in all spots that implement `__ape_attributes_extras__`

### Checklist

### DIFF
--- a/src/ape_cache/query.py
+++ b/src/ape_cache/query.py
@@ -1,12 +1,12 @@
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator, List, Optional, cast
 
 from sqlalchemy import create_engine, func
 from sqlalchemy.engine import CursorResult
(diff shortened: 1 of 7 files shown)


src/ape_cache/query.py
```diff
@@ -432,7 +433,7 @@ def _get_block_txns_data(
                     new_dict["receiver"] = b""
                 elif col == "block_hash":
                     new_dict["block_hash"] = query.block_id
-                elif col == "signature":
+                elif col == "signature" and val.signature is not None:
```
antazoey, line 436: good news: by doing this, typing issues revealed themselves through `mypy`.

fubuloubu (approved):

Issue #1824, opened 2024-01-06
Contract interaction: Difference between transaction and call not documented


miohtama:
### What went wrong?

Currently, [Contract interaction](https://docs.apeworx.io/ape/stable/userguides/contracts.html#contract-interaction) documentation does not explain the difference between a transaction and a call, and how the framework user is supposed to differentiate between these two.

<img width="785" alt="image" src="https://github.com/ApeWorX/ape/assets/49922/18e53dc6-9cef-4469-9a5d-8389a61904c1">

PR #1828, opened 2024-01-08
docs: document difference between CALL versus TRANSACT in contract-interaction guide

antazoey:
### What I did

fixes: #1824 

### Checklist

### DIFF
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -181,7 +181,8 @@ At the RPC level, Ethereum calls are performed using the `eth_call` RPC and tran
 
 ### Transactions
 
-The following example demonstrates invoking a contract's method in Ape as a transaction:
+The following example demonstrates invoking a contract's method in Ape as a transaction.
+However, take note that there is a [separate guide](./transactions.html) which fully covers transactions in Ape.
 
 ```python
 from ape import accounts, Contract
@@ -204,6 +205,8 @@ Notice that transacting returns a [ReceiptAPI](../methoddocs/api.html#ape.api.tr
 ```python
 assert receipt.return_value == 123
 ```
+
+For more general information on transactions in the Ape framework, see [this guide](./transactions.html).
 
 ### Calls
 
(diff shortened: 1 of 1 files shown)


docs/userguides/contracts.md
```diff
@@ -171,13 +171,33 @@ def set_number(num: uint256):
...
+The following examples demonstrate both making a call and invoking a transaction to a deployed version of the contract above.
+First, here is an example of invoking a function (creating a transaction):
 
 ```python
-assert contract.get_static_list() == [1, 2, 3]
+from ape import accounts, Contract
 
-# Mutable calls are transactions and require a sender
-receipt = contract.set_number(sender=dev)
+sender = accounts.load("<ALIAS>")
```
fubuloubu, line 188: I usually like to call this something like `account` so `sender=account` appears more demostrative

docs/userguides/contracts.md
fubuloubu, line 1: Would add another section about `.call` on mutable transactions, which convert them to regular calls

And also something about the return types of calls vs. transactions (e.g. Receipts)

docs/userguides/contracts.md
```diff
@@ -171,13 +171,105 @@ def set_number(num: uint256):
...
+account = accounts.load("<ALIAS>")
+contract = Contract("0x...")
+
+# CALL: A sender is not required for calls!
 assert contract.get_static_list() == [1, 2, 3]
+```
+
+### Calling Transactions and Transacting Calls
+
+You can treat transactions as calls as visa-versa.
```
fubuloubu, line 230:
```suggestion
You can treat transactions as calls and vice-versa.
```

docs/userguides/contracts.md
```diff
@@ -171,13 +171,105 @@ def set_number(num: uint256):
...
+```python
+from ape import Contract
+
+contract = Contract("0x...")
+
+result = contract.addBalance.call(123)
+assert result == "123"  # The return value gets forwarded from the contract.
+```
+
+Oppositely, you may want to measure a call as if it were a transaction, in which case you can use the `.transact` attribute on the contract call handler:
```
fubuloubu, line 252:
```suggestion
Similarly, you may want to measure a call as if it were a transaction, in which case you can use the `.transact` attribute on the contract call handler:
```

fubuloubu (approved):
Looking good

fubuloubu (approved):

Issue #1855, opened 2024-01-10
"Transfer value meets or exceeds account balance" error message needs more context


miohtama:
### Overview

- Currently, if there is an out-of-balance error message, it is not helpful enough to resolve the issue.

```
 (AccountsError) Transfer value meets or exceeds account balance.
Are you using the correct provider/account combination?
(transfer_value=1000000000, balance=0).
```

### Specification

While the error message is technically correct, for the user, it is not possible to tell if the provider/account combination is correct because this combination is missing from the error message, not giving enough context to fix the error and understand what Ape framework is trying to do.

The error message should tell:

- What account address is it trying to check
- What chain it is trying to check
- (And if possible): What provider it is using to do this check

PR #1860, opened 2024-01-10
feat: add more details to account transfer error

antazoey:
### What I did

fixes: #1855 

### Checklist

### DIFF
--- a/src/ape_run/_cli.py
+++ b/src/ape_run/_cli.py
@@ -12,7 +12,7 @@ from click import Command, Context, Option
 
 from ape import networks, project
 from ape.cli import ConnectedProviderCommand, verbosity_option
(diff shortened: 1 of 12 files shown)


src/ape_run/_cli.py
```diff
@@ -252,6 +252,7 @@ def _launch_console(self):
     default=False,
     help="Drop into interactive console session after running",
 )
+@network_option()
```
fubuloubu, line 255: this isn't breaking right?

fubuloubu (approved):

src/ape_run/_cli.py
antazoey: oops, this was research on something else, didnt mean to commit

fubuloubu (approved):

PR #1862, opened 2024-01-11
chore: allow rich v13 as dependency

mikeshultz:
### What I did

Expanded the allowed dependency version of the rich package to include v13.

### How to verify it

Output looks the same on tree list commands and tests pass.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(
         "python-dateutil>=2.8.2,<3",
         "PyYAML>=5.0,<7",
         "requests>=2.28.1,<3",
-        "rich>=12.5.1,<13",
+        "rich>=12.5.1,<14",
         "SQLAlchemy>=1.4.35",
         "tqdm>=4.62.3,<5.0",
         "traitlets>=5.3.0",


antazoey (approved):

fubuloubu (approved):
Do we want to lower pin against v13?

mikeshultz:
> Do we want to lower pin against v13?

I'm not sure what you mean by this.

fubuloubu:
> > Do we want to lower pin against v13?
> 
> I'm not sure what you mean by this.

Upgrade the pin from 12.5.1 to whatever latest 13 is

mikeshultz:
> > > Do we want to lower pin against v13?
> > 
> > 
> > I'm not sure what you mean by this.
> 
> Upgrade the pin from 12.5.1 to whatever latest 13 is

After wrestling with dependencies with overly restrictive dependency specs over the last week I'm not especially wanting to do the same to others just because I'm toying around with something.  Also, our API exposure to rich is super limited on this one so I think we're pretty safe here for the time being.

Issue #1729, opened 2023-11-05
Add public RPCs as default geth provider links [APE-1514]


fubuloubu:
### Overview

Currently when using the geth plugin there are no configured networks that come out of the box. It owupd be nice if you can connect to a public RPC out of the box

### Specification

A large collection of networks are specified here:
https://github.com/pcaversaccio/hardhat-project-template-ts/blob/main/hardhat.config.ts#L141

### Dependencies

No dependencies. Would be nice to have multiple networks, which would require a change

antazoey:
Can you elaborate? Currently, you can connect to any public RPC out of the box in ape-geth.

fubuloubu:
> Can you elaborate? Currently, you can connect to any public RPC out of the box in ape-geth.

pre-configure the ape-geth plugin to connect to these public RPCs if it does not auto-detect one on the typical port

antazoey:
ohhh now I get it.
Yes, that totally makes sense actually. I think there some other related issues in some of the L2 plugins for this.

mikeshultz:
Good source of information: https://github.com/ethereum-lists/chains/tree/master/_data/chains

Thinking maybe we could either build our set from these to be included in our release packages or download on-the-fly when needed.  Mild preference for the former for performance and security reasons.

PR #1866, opened 2024-01-16
feat: add default public RPC endpoints for default configured networks

mikeshultz:
### What I did

Adds RPC endpoints for default configured networks sourced from official Ethereum Foundation repository.

fixes: #1729

### How I did it

Added a script to generate a source file that has constants for the RPC endpoints.

### How to verify it

```bash
ape console --network ethereum:mainnet
```

### Checklist


mikeshultz:
https://github.com/mikeshultz/ape/blob/027305b71bcb99b2cebb8b049e9e66e2ae0f3a8d/src/ape_geth/provider.py#L209-L210

@antazoey thoughts on adding more default ecosystem configurations here?  Script currently also includes base, polygon, polygon-zkevm, and gnosis.  Could expand/contract that if we want, too.

src/ape_geth/provider.py
```diff
@@ -197,9 +199,9 @@ def wait(self, *args, **kwargs):
...
 class GethNetworkConfig(PluginConfig):
     # Make sure you are running the right networks when you try for these
-    mainnet: Dict = DEFAULT_SETTINGS.copy()
-    goerli: Dict = DEFAULT_SETTINGS.copy()
-    sepolia: Dict = DEFAULT_SETTINGS.copy()
+    mainnet: Dict = {"uri": random.choice(PUBLIC_CHAIN_RPCS["ethereum"]["mainnet"])}
+    goerli: Dict = {"uri": random.choice(PUBLIC_CHAIN_RPCS["ethereum"]["goerli"])}
+    sepolia: Dict = {"uri": random.choice(PUBLIC_CHAIN_RPCS["ethereum"]["sepolia"])}
     # Make sure to run via `geth --dev` (or similar)
     local: Dict = {**DEFAULT_SETTINGS.copy(), "chain_id": DEFAULT_TEST_CHAIN_ID}
```
antazoey, line 206: i think here was have to add a custom validator again or something to check for any network that isn't part of Ethereum

src/ape_geth/provider.py
```diff
@@ -197,9 +199,9 @@ def wait(self, *args, **kwargs):
...
```
antazoey, line 207: we have to handle the non-eth networks via a custom validator or something

antazoey (commented):
cool! I have some ideas but can be discussed or changed later perhaps

scripts/update_rpc.py
```diff
@@ -0,0 +1,132 @@
...
+        "mainnet": 1101,
+        "testnet": 1442,
+    },
+}
+INCLUDE_PROTOCOLS = ["http", "https"]
+SOURCE_URL = "https://raw.githubusercontent.com/ethereum-lists/chains/master/_data/chains/"
+CHAIN_CONST_FILE = Path(__file__).parent.parent / "src" / "ape_geth" / "chains.py"
+
+
+class Chain(BaseModel):
```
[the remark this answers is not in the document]
mikeshultz: That is like a packaging step away from that comment.  Though not sure it's worth the time now.

src/ape_geth/provider.py
mikeshultz: Couple of questions.  Are we responsible for non-eth networks here?  Or would the ecosystem plugin handle that?  I don't think your change allows for custom ecosystems.  Though maybe I should test against one?

I wrote a little validator for the `GethNetworkConfig` but since it's only used for Ethereum, it's basically pointless so I'm not planning on committing it.

```python
    @model_validator(mode="after")
    def verify_networks(self) -> "GethEcosystemConfig":
        if self.__pydantic_extra__:
            for aname in self.__pydantic_extra__.keys():
                rpcs = PUBLIC_CHAIN_RPCS["ethereum"].get(aname)
                if rpcs:
                    self.__pydantic_extra__[aname] = {"uri": random.choice(rpcs)}
                    continue
                # Fallback to default settings.
                self.__pydantic_extra__[aname] = {**DEFAULT_SETTINGS.copy()}
        return self
```

So, is any change necessary here?

mikeshultz: Or maybe you're not talking about custom networks.  I'll test against ape_optimism or something.

antazoey: i think both custom networks and the plugins use `GethNetworkConfig`

mikeshultz: Yeah, I don't think custom networks are relevant though since they won't ever match a public RPC.  Working on optimism now though.  Should have something shortly.

mikeshultz: I updated the `Geth` provider to handle dynamic networks.  Tiny bit of code duplication with `EthereumNodeProvider` but I didn't want to fully override it either.

src/ape_geth/provider.py
```diff
@@ -471,4 +471,23 @@ def build_command(self) -> List[str]:
...
+    @property
+    def uri(self) -> str:
+        uri = super().uri
+        ecosystem = self.network.ecosystem.name
+        network = self.network.name
+
+        # If we didn't find one in config, look for a public RPC.
+        if not uri or uri == DEFAULT_SETTINGS["uri"]:
+            # Do not override explicit configuration
+            if hasattr(self.config, ecosystem):
```
antazoey, line 483: you can also use `.get` on configs

mikeshultz: Would that improve this somehow?  Same amount of code but it would be a truthy value check instead of a strict bool check, I think.  Unless I'm missing something.

antazoey: I would do this:

```python
if ecosystem in self.config:
```

if it lets you, it reads better.

antazoey: but no, just a nit i guess.
`hasattr` and `getattr` are cool but always feel hacky to me

antazoey: or this:

```python
if cfg := self.config.get(ecosystem)
```

mikeshultz: [7d3cc1f](https://github.com/ApeWorX/ape/pull/1866/commits/7d3cc1f66005738fbe7fdb3a2ff7ea5c4215a983)

src/ape_geth/provider.py
antazoey: I think that works

src/ape_geth/provider.py
```diff
@@ -469,4 +471,23 @@ def build_command(self) -> List[str]:
...
 class Geth(EthereumNodeProvider):
-    pass
+    @property
+    def uri(self) -> str:
+        uri = super().uri
+        ecosystem = self.network.ecosystem.name
+        network = self.network.name
+
+        # If we didn't find one in config, look for a public RPC.
+        if not uri or uri == DEFAULT_SETTINGS["uri"]:
```
antazoey, line 481: A simple test here for this new logic would be complete the story

src/ape_geth/provider.py
```diff
@@ -469,4 +471,23 @@ def build_command(self) -> List[str]:
...
+            # Do not override explicit configuration
+            if ecosystem in self.config:
+                # Shape of this is odd.  Pydantic model containing dicts
```
antazoey, line 484: it is not too odd, it isnt much different than using `additonalProperties` in OpenAPI swagger docs

antazoey (approved):
looks good - a test would be cool though.

antazoey:
this makes forking in test work out-of-the-gate without having to configure a default provider or set a URI.
<3

src/ape_geth/provider.py
mikeshultz: Feel free to add a comment under mine disputing my assertion.  :smirk:

src/ape_geth/provider.py
mikeshultz: How's this?  8c2cf999

mikeshultz: Disliking that these [make actual RPC calls though](https://github.com/ApeWorX/ape/actions/runs/7588778920/job/20672029316?pr=1866).  Thoughts?

scripts/update_rpc.py
```diff
@@ -0,0 +1,150 @@
...
+from typing import Any, Dict, List, Optional
+from urllib.parse import urljoin
+
+import requests
+from pydantic import BaseModel
+
+from ape.logging import logger
+
+CHAIN_IDS = {
+    "base": {
```
antazoey, line 19: do they have any of these:

1. avalanche
2. arbitrum
3. bsc
4. fantom

im just looking through some of the other plugins we have

src/ape_geth/provider.py
antazoey: We could do this:

```python
def test_default_public_uri(config):
    cfg = config.get_config("ethereum").sepolia
    assert cfg.uri in PUBLIC_CHAIN_RPCS["ethereum"]["sepolia"]
```

It is simpler, won't be flakey, and still tests the changes, I think, since the URI is calculated at the config level.

If you prefer the integration test though, we could catch the exception and handle it like task-failed-successfully.

antazoey (commented):
test is going to be too flakey as-is, but it is good.
i added 2 options i could think of.

src/ape_geth/provider.py
mikeshultz: Good idea.  Done in [f8dcccd](https://github.com/ApeWorX/ape/pull/1866/commits/f8dcccd3e7bdf95adeb24ff9d6c6805b4570e5b9)

scripts/update_rpc.py
mikeshultz: [29f449b](https://github.com/ApeWorX/ape/pull/1866/commits/29f449be15c76966192b177f21b2014b02319b6d)

antazoey (approved):
this is great but I dont want people to open PRs here adding more URIs so before that happens, let's make the chain enum package thing.

Issue #1278, opened 2023-02-02
Make `ManagerAccessMixin.provider` accessible also as a classmethod


fubuloubu:
### What went wrong?

As of #1224, we started using the active provider in some of our pydantic model validators, for example:

https://github.com/ApeWorX/ape/blob/97264edf16f51e9909ccc064aebfb30ab9362ee6/src/ape/api/transactions.py#L52-L55

However, these are classmethods and not a method on a normal instance, so we can't take advantage of `ManagerAccessMixin.provider`, which resolves this issue in the same way:

https://github.com/ApeWorX/ape/blob/97264edf16f51e9909ccc064aebfb30ab9362ee6/src/ape/utils/basemodel.py#L59-L74

We need a way to access this property from a class as well as an instance, so that we can reduce the tech debt of working around this.

### How can it be fixed?

A `combomethod` is a common way of resolving this:
https://stackoverflow.com/questions/2589690/creating-a-method-that-is-simultaneously-an-instance-and-class-method

Fortunately, the fine folks at EF maintain a version of this code in `eth-utils`: 
https://github.com/ethereum/eth-utils/blob/master/eth_utils/decorators.py#L10-L22

fubuloubu:
Turns out, not that simple. But a similar approach can work w/ `.provider`

PR #1873, opened 2024-01-19
feat: make `.provider` on `@classproperty` on all manager-access models

antazoey:
### What I did

fixes: #1278 

### Checklist

### DIFF
--- a/src/ape/utils/basemodel.py
+++ b/src/ape/utils/basemodel.py
@@ -2,6 +2,7 @@ from abc import ABC
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     ClassVar,
     Dict,
     Iterator,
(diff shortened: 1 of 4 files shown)


src/ape/utils/basemodel.py
```diff
@@ -35,6 +36,14 @@
     from ape.pytest.runners import PytestApeRunner
 
 
+class classproperty(object):
```
fubuloubu, line 39: There's nothing in Python stdlib that can acheive this?

antazoey: no, and also this wont work with setters. setters is where things seem to get complicated, which is why im guessing they dont have this

in our case , we only need a getter though

fubuloubu (approved):

Issue #1740, opened 2023-11-17
Fork at block height


fubuloubu:
### Overview

Add an optional `block_number` kwarg (that accepts negative index height) which allows `network.fork` calls to connect to a fork at a specific block height

### Specification

```py
with networks.fork("provider", block_number=-100):
    ...  # create a fork 100 blocks ago
```

### Dependencies

N/a

salparadi:
This would be awesome... is a nice feature of brownie to be able to set up a fork-atblock network and then easily pass a block # in during a script's execution. Hope this gets implemented!

fubuloubu:
> This would be awesome... is a nice feature of brownie to be able to set up a fork-atblock network and then easily pass a block # in during a script's execution. Hope this gets implemented!

I believe you can do this with config, same as with Brownie. Is there a Brownie feature that I don't know about that let's you ad-hoc fork at a block height like that?

salparadi:
Maybe I misunderstood the feature. My thinking was that in brownie I can create two "fork" networks. One that was just from current block (mainnet-fork) and then one that was at a specific block height (mainnet-fork-atblock), Passing in the block number to ganache>infura. On command line I'd just export BLOCK="18720551" before loading the console, and in a script I'd set os.environ["BLOCK"] = "18720551" then load the network.

```
  - name: Ganache-CLI (Mainnet Fork at Block)
    id: mainnet-fork-atblock
    cmd: ganache-cli
    host: http://127.0.0.1
    explorer: https://api.etherscan.io/api
    timeout: 120
    cmd_settings:
      chain_id: 1
      port: 8545
      gas_limit: 20000000
      accounts: 10
      evm_version: grayGlacier
      mnemonic: brownie
      fork: https://mainnet.infura.io/v3/$WEB3_INFURA_PROJECT_ID@$BLOCK
```

My understanding in Ape is there can't be multiple forks for a provider like foundry. Maybe I can accomplish what I'm after though with this. If I just want to fork at current height, just set BLOCK=. Setting it to nothing?

```
foundry:
  host: auto
  fork:
    ethereum:
      mainnet:
        upstream_provider: https://rpc.ankr.com/eth
        block_number: ${BLOCK}
```

fubuloubu:
[...]

PR #1883, opened 2024-01-24
feat: adds block_number arg to network.fork()

mikeshultz:
### What I did

Adds `block_number` argument to `network.fork()`, including negative behind numbers.

fixes: #1740

### How I did it

When the kwarg is provided, it will update the expected config structure for the fork provider with the given block number

### How to verify it

```python
from ape import chain, networks
from ape.cli import ConnectedProviderCommand
import click


@click.command(cls=ConnectedProviderCommand)
def cli():
    live_block = chain.provider.get_block("latest")

    # with networks.fork("provider", block_number=-100):
    with networks.fork("foundry", block_number=18569420):
        block = chain.provider.get_block("latest")
        assert block.number == 18569420

    with networks.fork("foundry", block_number=-100):
        block = chain.provider.get_block("latest")
        assert block.number == live_block.number - 100
```

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -8,7 +8,7 @@ from ape.api import EcosystemAPI, ProviderAPI, ProviderContextManager
 from ape.api.networks import NetworkAPI
 from ape.exceptions import ApeAttributeError, EcosystemNotFoundError, NetworkError
 from ape.managers.base import BaseManager
-from ape.utils.misc import dict_overlay
+from ape.utils.misc import _dict_overlay
 from ape_ethereum.provider import EthereumNodeProvider
 
 
@@ -103,7 +103,7 @@ class NetworkManager(BaseManager):
                 block_number = self.provider.get_block("latest").number + block_number
 
             # Ensure block_number is set in config for this network
-            dict_overlay(
+            _dict_overlay(
                 provider_settings,
                 {
                     "fork": {
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -8,6 +8,7 @@ from ape.api import EcosystemAPI, ProviderAPI, ProviderContextManager
 from ape.api.networks import NetworkAPI
 from ape.exceptions import ApeAttributeError, EcosystemNotFoundError, NetworkError
 from ape.managers.base import BaseManager
+from ape.utils.misc import dict_overlay
 from ape_ethereum.provider import EthereumNodeProvider
 
 
@@ -79,6 +80,7 @@ class NetworkManager(BaseManager):
         self,
         provider_name: Optional[str] = None,
         provider_settings: Optional[Dict] = None,
+        block_number: Optional[int] = None,
     ) -> ProviderContextManager:
         """
         Fork the currently connected network.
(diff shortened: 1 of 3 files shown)


src/ape/managers/networks.py
```diff
@@ -95,6 +97,21 @@ def fork(
         forked_network = self.ecosystem.get_network(f"{self.network.name}-fork")
         provider_settings = provider_settings or {}
 
+        if block_number is not None:
+            # Negateive block_number means relative to HEAD
```
antazoey, line 101:
```suggestion
            # Negative block_number means relative to HEAD
```

src/ape/utils/misc.py
```diff
@@ -490,9 +490,33 @@ def is_zero_hex(address: str) -> bool:
         return False
 
 
+def dict_defaults(mapping: Dict[str, Any], defaults: Dict[str, Any], depth=0):
```
antazoey, line 493: I don't see this method being used, is it still needed?

src/ape/utils/misc.py
```diff
@@ -490,9 +490,33 @@ def is_zero_hex(address: str) -> bool:
...
+    """Conditionally overlay given default on a dict if no value is set"""
+    for key, value in defaults.items():
+        if key not in mapping:
+            mapping[key] = value
+        elif isinstance(value, dict):
+            dict_defaults(mapping[key], value, depth + 1)
+    return mapping
+
+
+def dict_overlay(mapping: Dict[str, Any], overlay: Dict[str, Any], depth=0):
```
antazoey, line 503: **thought**: Could make this an internal util method like `_dict_overlay` so we don't have to officially support it from a library perspective and can use it internally as needed. Otherwise, we need a full doc-str here and stuff.

antazoey, line 503:
```suggestion
def dict_overlay(mapping: Dict[str, Any], overlay: Dict[str, Any], depth: int=0):
```

src/ape/utils/misc.py
mikeshultz: Speculative future use.  I'll axe it.  Just didn't want to write it again.

src/ape/utils/misc.py
mikeshultz: I assumed this module was internal use.  Will rename.  Is it fine in this module or is there another spot for 'em?

mikeshultz: I thought it was customary to leave off the type hint when you have an explicit non-None default.  Though I'm not sure where I got that from.  Might be mixing that up with Rust or something.  :shrug:

mikeshultz:
I don't really know how to write a test for this.  I don't see any other tests for the `fork()` context manager and don't really want to create a test that makes real RPC calls.  Might take the suggestion and just test in foundry plugin.

src/ape/utils/misc.py
antazoey: this module is good

antazoey (approved):
LGTM

For testing, we can add some after https://github.com/ApeWorX/ape/pull/1883 merges

PR #1884, opened 2024-01-25
feat: Use `eth_createAccessList` RPC in type=1 prepared transaction

antazoey:
### What I did

now when using type=1, access lists are automatically prepared on the tx

**question**: should we automatically set access list on type=2? I am too afraid to make that change I think but there isnt really a good way to do in Ape otherwise

### Checklist


fubuloubu (approved):

fubuloubu:
> **question**: should we automatically set access list on type=2? I am too afraid to make that change I think but there isnt really a good way to do in Ape otherwise

Maybe something for v0.8?

PR #1885, opened 2024-01-25
fix: issue when forking past genesis

antazoey:
### What I did

The error when trying to fork a non-forkable network was not quite good enough for me, so I fixed it.

Also, adds missing tests for `netowork.fork` context stuff.

### How I did it

Catch the network error and raise a different one with a more custom message for the `.fork()` method use-case.

### How to verify it

connect to `local` and then do:

```python
with networks.fork():
   ...
```

it should fail something like 

```
NetworkError: Unable to fork network local
```

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -6,7 +6,12 @@ import yaml
 
 from ape.api import EcosystemAPI, ProviderAPI, ProviderContextManager
 from ape.api.networks import NetworkAPI
-from ape.exceptions import ApeAttributeError, EcosystemNotFoundError, NetworkError
+from ape.exceptions import (
+    ApeAttributeError,
+    EcosystemNotFoundError,
+    NetworkError,
+    NetworkNotFoundError,
+)
 from ape.managers.base import BaseManager
 from ape.utils.misc import _dict_overlay
 from ape_ethereum.provider import EthereumNodeProvider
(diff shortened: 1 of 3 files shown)


tests/functional/test_network_manager.py
```diff
@@ -289,3 +289,66 @@ def test_ecosystems_include_custom(networks, custom_networks_config_dict, temp_c
...
+    Happy-path fork test when specifying the provider.
+    """
+    ctx = networks.fork(provider_name="mock")
+    assert ctx._disconnect_after is True
+    with ctx as provider:
+        assert provider.name == "mock"
+        assert provider.network.name == "sepolia-fork"
+
+
+def test_fork_with_provider_settings(networks, mock_sepolia, mock_fork_provider):
```
antazoey, line 347: @mikeshultz we can test your [new feature similarly](https://github.com/ApeWorX/ape/pull/1883) similarly

src/ape/managers/networks.py
```diff
@@ -92,15 +97,19 @@ def fork(
         Returns:
             :class:`~ape.api.networks.ProviderContextManager`
         """
-        forked_network = self.ecosystem.get_network(f"{self.network.name}-fork")
-        provider_settings = provider_settings or {}
+        try:
+            forked_network = self.ecosystem.get_network(f"{self.network.name}-fork")
+        except NetworkNotFoundError as err:
+            raise NetworkError(f"Unable to fork network '{self.network.name}'.") from err
```
mikeshultz, line 103: though: Cool, I wasn't aware of the `raise ... from` syntax.

tests/functional/conftest.py
```diff
@@ -695,3 +696,53 @@ def mock_compile(paths, base_path=None):
...
+    """
+    actual = ethereum.sepolia_fork.__dict__.pop("providers", {})
+    mock_provider = mocker.MagicMock()
+    mock_provider.name = "mock"
+    mock_provider.network = ethereum.sepolia_fork
+
+    # Have to this because providers are partials.
+    def fake_partial(*args, **kwargs):
+        mock_provider.partial_call = (args, kwargs)
+        return mock_provider
```
mikeshultz, line 741: question: What are partials?

mikeshultz (approved):

tests/functional/conftest.py
antazoey: They are classes that are partially initialized, it is a functools thing iicr.. we use them for providers because we can initialize some of the provider at plugin-load time and the rest gets filled in upon connection.

src/ape/managers/networks.py
antazoey: Ya it preserves the exception stack when raising a new error in an except block, it is nifty

src/ape/managers/networks.py
```diff
@@ -90,17 +95,24 @@ def fork(
...
+
         provider_settings = provider_settings or {}
 
         if block_number is not None:
             # Negative block_number means relative to HEAD
             if block_number < 0:
-                block_number = self.provider.get_block("latest").number + block_number
+                block_number = max(0, self.provider.get_block("latest").number)
```
antazoey, line 115: bugfix here: when accidentally forking past genesis, it would still be negative.. 
now you are just bound to genesis

hmm, wondering if we should do a warning here?

antazoey: ok exception better, did that instead

fubuloubu (approved):

PR #1888, opened 2024-01-25
fix: circular import issues from accounts plugins

antazoey:
### What I did

Replace top-level imports with ManagerAccessMixin to prevent circular imports in account plugins.

### Checklist

### DIFF
--- a/src/ape/cli/arguments.py
+++ b/src/ape/cli/arguments.py
@@ -3,16 +3,16 @@ from itertools import chain
 import click
 from eth_utils import is_hex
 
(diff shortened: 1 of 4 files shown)


src/ape/cli/utils.py
fubuloubu, line 1: added an empty file?

fubuloubu (dismissed):

src/ape/cli/utils.py
antazoey: deleted an empty file

fubuloubu (approved):

PR #1890, opened 2024-01-28
test: adds timeout and skips some breaking integration tests

mikeshultz:
### What I did

- Adds pytest-timeout to prevent some failing tests from running forever.
- Skips plugin installation integration tests because they will often re-install ape (and dependencies), which means the following tests run against an unwanted versions of packages.

@antazoey we spoke about the latter a bit the other day.  I think I narrowed the behavior down to these tests.  I'm not entirely sure if it's only local test runs or if it may affect CI as well.  Slight suspicion it might be related to my local `src/ape/version.py` file but not sure.  I'd be happy skipping on local only if we'd prefer.

### Checklist

### DIFF
--- a/tests/integration/cli/test_plugins.py
+++ b/tests/integration/cli/test_plugins.py
@@ -129,29 +129,25 @@ def test_list_does_not_repeat(ape_plugins_runner, installed_plugin):
     assert "ethereum" not in result.available_plugins
 
 
-@pytest.mark.skip(reason="Installation may replace local packages with unwanted versions")
-@github_xfail()
+@pytest.mark.pip
 def test_upgrade(ape_plugins_runner, installed_plugin):
     result = ape_plugins_runner.invoke(["install", TEST_PLUGIN_NAME, "--upgrade"])
     assert result.exit_code == 0
 
 
-@pytest.mark.skip(reason="Installation may replace local packages with unwanted versions")
-@github_xfail()
+@pytest.mark.pip
 def test_upgrade_failure(ape_plugins_runner):
     result = ape_plugins_runner.invoke(["install", "NOT_EXISTS", "--upgrade"])
     assert result.exit_code == 1
 
 
-@pytest.mark.skip(reason="Installation may replace local packages with unwanted versions")
-@github_xfail()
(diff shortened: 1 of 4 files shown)


antazoey (approved):

tests/integration/cli/test_plugins.py
```diff
@@ -129,24 +129,28 @@ def test_list_does_not_repeat(ape_plugins_runner, installed_plugin):
...
 
 
+@pytest.mark.skip(reason="Installation may replace local packages with unwanted versions")
 @github_xfail()
 def test_upgrade(ape_plugins_runner, installed_plugin):
     result = ape_plugins_runner.invoke(["install", TEST_PLUGIN_NAME, "--upgrade"])
     assert result.exit_code == 0
 
 
+@pytest.mark.skip(reason="Installation may replace local packages with unwanted versions")
```
antazoey, line 139: I think we should either

1. replace the `@github_xfail()` with this skip, could also store it as variable `github_skip = `
2. edit the `@github_xfail()` to include a manual marker (also registering it and stuff), this way allows us to run the tests manually if we want, which you may if you are editing the plugin commands a bunch

3. option 3 is delete the tests and only rely on the functional tests for plugin stufff.... but I kinda like having integration tests here, even if they don't run in CI

mikeshultz: What do you think about e7783af?

antazoey: It seems to skip them, but you want to add `@run_once` back in for those.
The integration tests are (oddly, imo) parametrized across projects.
To force an integ test to only run once (1 project) instead of n times you have to use `@run_once`.

Right now, when I run the module tests, it says this:

```
55 skipped, 3 xpassed, 1 warning
```

But if i import `run_once` and decorate them with that (note: github_xfail has this decor builtin to it), then it says:

```
 5 skipped, 3 xpassed, 1 warning
```

mikeshultz: Ah, didn't realize that.  A feature of the `@github_xfail` decorator lost.  Added.

antazoey (approved):

Issue #1046, opened 2022-09-11
Ape core reinstalls itself when there's a plugin installed that requires older versions of core


fubuloubu:
### What went wrong?

When you install a plugin that is pinned to an older version of ape via `ape plugins install ...`, the installation will silent install the older version of ape specified in the pin

### How can it be fixed?

1. When installing a plugin without specifying a version, a minimum pin should be specified using ape core's version e.g. `f"ape-plugin>={ape.version}"`
2. When installing a plugin and specifying a version, that version should be >= the current version of ape, and tell the user to downgrade ape core if it isn't
3. We should probably have better methods to upgrade/downgrade all of ape core and plugins at once

sabotagebeats:
Is this seeming to be the same issue? 

```
$ ape plugins install solidity vyper infura alchemy ens 
INFO: Installing solidity...
SUCCESS: Plugin 'solidity==0.5.0' has been installed.
INFO: Installing vyper...
SUCCESS: Plugin 'vyper==0.5.0' has been installed.
INFO: Installing infura...
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ape-vyper 0.5.0 requires eth-ape<0.6.0,>=0.5.0, but you have eth-ape 0.4.5 which is incompatible.
ape-solidity 0.5.0 requires eth-ape<0.6,>=0.5.0, but you have eth-ape 0.4.5 which is incompatible.
SUCCESS: Plugin 'infura==0.4.0' has been installed.
INFO: Installing alchemy...
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ape-infura 0.4.0 requires eth-ape<0.5.0,>=0.4.0, but you have eth-ape 0.5.1 which is incompatible.
SUCCESS: Plugin 'alchemy==0.5.0' has been installed.
INFO: Installing ens...
SUCCESS: Plugin 'ens==0.5.0' has been installed.
$ ape --version
0.5.1
```

fubuloubu:
Yes, that's exactly the issue. It actually is downgrading eth-ape to 0.4.6 and then upgrading again to 0.5.1,depending on the order of the plugins. Trying installing infura last

ef3n9r98:
Hi @fubuloubu, I know some devs who'd be interested in working on this. Is there a budget for fixing this issue, and if so, how much? 

[...]

PR #1891, opened 2024-01-29
fix: issue where installing a plugin would change Ape version

antazoey:
### What I did

fixes: #1046 

### Checklist

### DIFF
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -1,11 +1,13 @@
 import subprocess
 import sys
 from pathlib import Path
-from typing import List, Tuple
+from typing import Any, Dict, List, Tuple
 
 import click
+from packaging.version import Version
 
 from ape.cli import ape_cli_context, skip_confirmation_option
+from ape.logging import logger
 from ape.managers.config import CONFIG_FILE_NAME
 from ape.utils import github_client, load_config
 from ape_plugins.utils import (
@@ -13,6 +15,8 @@ from ape_plugins.utils import (
     PluginMetadata,
     PluginMetadataList,
     PluginType,
+    _pip_freeze,
+    ape_version,
 )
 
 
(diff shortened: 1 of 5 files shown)


tests/integration/cli/utils.py
```diff
@@ -142,6 +142,7 @@ def wrapper(f):
         f = pytest.mark.xfail(
             strict=False, reason="GitHub rate limiting issues or plugin install issues"
         )(f)
+        f = pytest.mark.manual()(f)
```
antazoey, line 145: @mikeshultz An alternative to the skip idea here https://github.com/ApeWorX/ape/pull/1890

src/ape_plugins/_cli.py
```diff
@@ -223,3 +199,55 @@ def uninstall(cli_ctx, plugins, skip_confirmation):
...
+    for plugin in _pip_freeze.get_plugins():
+        logger.info(f"Updating {plugin} ...")
+        name = plugin.split("=")[0].strip()
+        subprocess.call([sys.executable, "-m", "pip", "install", f"{name}{spec}", "--quiet"])
+
+    # This check is for verifying the update and shouldn't actually do anything.
+    logger.info("Updating Ape core ...")
+    completed_process = subprocess.run(
+        [sys.executable, "-m", "pip", "install", f"eth-ape{spec}", "--quiet"]
+    )
```
mikeshultz, line 243: question: Would this still not produce the same issue?  If you're running this it will install packaged ape rather than local editable, wouldn't it?

src/ape_plugins/utils.py
```diff
@@ -298,6 +383,62 @@ def check_installed(self, use_cache: bool = True):
...
+        result_handler = ModifyPluginResultHandler(self)
+        pip_arguments = [sys.executable, "-m", "pip", "install"]
+
+        if upgrade:
+            logger.info(f"Upgrading '{self.name}' plugin ...")
+
+            # NOTE: A simple --upgrade flag may upgrade the plugin
+            # to a version outside Core Ape's. Thus, we handle it
+            # with a version-specifier instead.
+            # NOTE: There can issues when --quiet is not at the end.
```
mikeshultz, line 414: suggestion: Might be useful to note what issues could occur.

tests/integration/cli/utils.py
mikeshultz: I can't seem to find docs on this, can you share?  Seems more useful than a strict skip.

mikeshultz (commented):
Took a quick pass because I was curious.  lmk if you want a tester.

src/ape_plugins/_cli.py
antazoey: I am attempting to address the 3rd bullet here in the issue link with the new command, it is not related to the primary issue but supposed to be a way to upgrade the full suite easily... What issue do you mean regarding the editable install?

.github/workflows/test.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
             pip install .[test]
 
         - name: Run Tests
-          run: pytest -m "not fuzzing" -s --cov=src -n auto --dist loadgroup
+          run: pytest -m "not (fuzzing or manual)" -s --cov=src -n auto --dist loadgroup
```
antazoey, line 98: @mikeshultz the new custom marker is used here and it is also registered in pyproject.toml. it is not a direct pytest feature but just a pattern ive used in the past for similar scenarios.

mikeshultz: I've been wondering what the `-m` option was.  Interesting, thanks.

tests/integration/cli/utils.py
mikeshultz: Is there an alternative to this that would require explicit action to run rather than explicit action to skip?

src/ape_plugins/_cli.py
mikeshultz: The issue that made me create #1890 that skips the plugin installation integration tests.  If that's unrelated, then ignore.

src/ape_plugins/utils.py
antazoey: It just can't be first, so comment was kinda confusing so i just removed it.

tests/integration/cli/utils.py
antazoey: rebase after your work!

fubuloubu (approved):

mikeshultz (approved):
LGTM

Issue #1335, opened 2023-03-09
Option to specify location of '.cache' folder [APE-703]


0rtis:
### Overview

Add an option to specify th location of the `.cache` folder. At the moment, it is located at the root of the `contracts` folder. This can cause issue with other framework (ex: Hardhat will parse all contracts in the `contracts` folder, including `.cache` resulting in very long processing time)

### Specification

Ideally, it would be an option on the ape config file like `cache_folder: mycache` (similar to `contracts_folder`).

unparalleled-js:
This may also affect compiler plugins like `ape-solidity`. We would need to make sure they work when the `.cache` folder is outside of the `contracts/` folder may require providing multiple base-paths).

mikeshultz:
So what all ends up in the `.cache` folder?  Is it just ethpm/dependency stuff or is there other uses?

antazoey:
> So what all ends up in the `.cache` folder? Is it just ethpm/dependency stuff or is there other uses?

Primarily it gets used by `ape-solidity`. Your dependency source files get plopped there from the cache and then ape-soldity uses remappings to refer to the files there from your import statements. That is why it defaults to the `contracts/` folder - it works best for importing

mikeshultz:
> > So what all ends up in the `.cache` folder? Is it just ethpm/dependency stuff or is there other uses?
> 
> Primarily it gets used by `ape-solidity`. Your dependency source files get plopped there from the cache and then ape-soldity uses remappings to refer to the files there from your import statements. That is why it defaults to the `contracts/` folder - it works best for importing

Wondering if solc going to have a problem with this.  I vaguely recall some sticking points that might make this challenging.  For instance, directory paths affecting some kind of source hashing (libraries maybe?) or contract validation (memory is foggy).  Also, I didn't think solc would compile source files outside of the contracts path.

Maybe there's some workarounds.  Just need to wire it up and test it, I guess.

antazoey:
> > > So what all ends up in the `.cache` folder? Is it just ethpm/dependency stuff or is there other uses?
> > 
> > 
> > Primarily it gets used by `ape-solidity`. Your dependency source files get plopped there from the cache and then ape-soldity uses remappings to refer to the files there from your import statements. That is why it defaults to the `contracts/` folder - it works best for importing
> 
[...]

PR #1897, opened 2024-01-31
feat: adds compiler_cache_folder config var

mikeshultz:
### What I did

Adds the `cache_folder` config var for ape_compile. Compiler plugins will leverage this for various temporary build-storage (mostly dependency sources).

Ref: #1335

### Checklist

### DIFF
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -85,7 +85,14 @@ compile:
 ## Settings
 
 Generally, configure compiler plugins using your `ape-config.yaml` file.
-For example, when using the `vyper` plugin, you can configure settings under the `vyper` key:
+One setting that applies to many compiler plugins is `compiler_cache_folder`, which holds dependency source files the compiler uses when compiling your contracts.
+By default, the folder is in your `contracts/.cache` folder but there are times you may want to move this to another location.
+
+```yaml
+compiler_cache_folder: .cache
+```
+
+As another example, when using the `vyper` plugin, you can configure settings under the `vyper` key:
 
 ```yaml
 vyper:
--- a/docs/userguides/compile.md
+++ b/docs/userguides/compile.md
@@ -85,11 +85,20 @@ compile:
 ## Settings
 
 Generally, configure compiler plugins using your `ape-config.yaml` file.
-One setting that applies to many compiler plugins is `compiler_cache_folder`, which holds dependency source files the compiler uses when compiling your contracts.
+One setting that applies to many compiler plugins is `cache_folder`, which holds dependency source files the compiler uses when compiling your contracts.
 By default, the folder is in your `contracts/.cache` folder but there are times you may want to move this to another location.
+Paths are relative to the project directory.
+For instance, to move the dependency cahce to the root project directory:
 
 ```yaml
-compiler_cache_folder: .cache
+compile:
+   cache_folder: .cache
+```
+
+```{caution}
+Changing the location of the dependency cache folder may alter the the output bytecode of your contracts from some compilers.
+Specifically, the [solc compiler will apend a hash of the input metadata to the contract bytecode](https://docs.soliditylang.org/en/latest/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode) which will change with contract path or compiler settings changes.
+This may impact things like contract verification on existing projects.
 ```
 
 As another example, when using the `vyper` plugin, you can configure settings under the `vyper` key:
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
 
 
 CONFIG_FILE_NAME = "ape-config.yaml"
+DEFAULT_COMPILER_CACHE_FOLDER = ".cache"
 
 
 class DeploymentConfig(PluginConfig):
@@ -108,6 +109,11 @@ class ConfigManager(BaseInterfaceModel):
     (differs by project structure).
     """
 
+    compiler_cache_folder: Optional[Path] = None
+    """
+    Path to contract dependency cache directory (e.g. `contracts/.cache`)
+    """
+
     dependencies: List[DependencyAPI] = []
     """A list of project dependencies."""
 
@@ -156,6 +162,9 @@ class ConfigManager(BaseInterfaceModel):
             self.dependencies = cache.get("dependencies", [])
             self.deployments = cache.get("deployments", {})
             self.contracts_folder = cache.get("contracts_folder", self.PROJECT_FOLDER / "contracts")
+            self.compiler_cache_folder = cache.get(
+                "compiler_cache_folder", self.contracts_folder / DEFAULT_COMPILER_CACHE_FOLDER
+            )
             return cache
 
         # First, load top-level configs. Then, load all the plugin configs.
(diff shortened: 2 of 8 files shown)


docs/userguides/compile.md
```diff
@@ -85,7 +85,14 @@ compile:
 ## Settings
 
 Generally, configure compiler plugins using your `ape-config.yaml` file.
-For example, when using the `vyper` plugin, you can configure settings under the `vyper` key:
+One setting that applies to many compiler plugins is `compiler_cache_folder`, which holds dependency source files the compiler uses when compiling your contracts.
+By default, the folder is in your `contracts/.cache` folder but there are times you may want to move this to another location.
+
+```yaml
+compiler_cache_folder: .cache
```
antazoey, line 92: idea: we could move this to the `ape_compile` plugin, so then youd do this:

```yaml
compile:
    cache_folder: .cache
```

src/ape/managers/project/manager.py
```diff
@@ -95,6 +95,16 @@ def contracts_folder(self) -> Path:
 
         return folder
 
+    @property
+    def compiler_cache_folder(self) -> Path:
+        """
+        The path to the project's compiler source cache folder.
+        """
+        # NOTE: as long as config has come up properly, this should not be None
```
antazoey, line 103: idea: could return the default if it is None

src/ape/managers/project/manager.py
```diff
@@ -95,6 +95,16 @@ def contracts_folder(self) -> Path:
...
+        if self.config_manager.compiler_cache_folder is None:
+            raise ProjectError("Compiler cache folder not set in config.")
```
antazoey, line 105: concern: this error message implies you must configure the cache folder even though that is not the case.

docs/userguides/compile.md
```diff
@@ -85,7 +85,14 @@ compile:
 ## Settings
 
 Generally, configure compiler plugins using your `ape-config.yaml` file.
-For example, when using the `vyper` plugin, you can configure settings under the `vyper` key:
+One setting that applies to many compiler plugins is `compiler_cache_folder`, which holds dependency source files the compiler uses when compiling your contracts.
+By default, the folder is in your `contracts/.cache` folder but there are times you may want to move this to another location.
```
antazoey, line 89: If any caveats, gotchas, or  findings from ape-solidity work goes, we should maybe include here.. like 

"NOTE: If using ape-solidity, your hashes are different now or wha have yous

antazoey:
Related: https://github.com/ApeWorX/ape-solidity/issues/62
When we utilize these changes in `ape-solidity`, can close that issue

docs/userguides/compile.md
mikeshultz: I like the way this looks.  Just curious, is it generally kosher for plugins to reference other plugins' config?

src/ape/managers/project/manager.py
mikeshultz: This error shouldn't happen under normal operation.  And I really don't want to do a second default that may conflict with the source of truth and also should never happen.

I'm actually a little more inclined to use an assert here rather an error with in a dead code path.  It was mostly to shut up the type checker.

docs/userguides/compile.md
mikeshultz: Will keep that in mind.  I don't have anything definitive to add at the moment, though.

docs/userguides/compile.md
antazoey: sometimes it makes sense, like in this case.

antazoey: other examples: many test providers, like ape-hardhat and ape-foundry, read stuff from the ape-test config

src/ape/managers/project/manager.py
antazoey: Oh, I see, ya because even if the user didn't configure (left as None), it would still be set after the first loading of the config.

An typing assert works, or maybe a builtin error like `raise TypeError("Expected non-None cache folder")`

all just a nit at this point

docs/userguides/compile.md
mikeshultz: Got a little weird, but got something working.  Check it out: [ff6a456](https://github.com/ApeWorX/ape/pull/1897/commits/ff6a4562432d6b17545f583342ea68665c4468f1)

src/ape/managers/project/manager.py
mikeshultz: Moved to an assert

src/ape/managers/config.py
```diff
@@ -216,7 +217,9 @@ def _plugin_configs(self) -> Dict[str, PluginConfig]:
             user_override = user_config.pop(plugin_name, {}) or {}
             if config_class != ConfigDict:
                 # NOTE: Will raise if improperly provided keys
-                config = config_class.from_overrides(user_override)  # type: ignore
+                config = config_class.from_overrides(  # type: ignore
```
antazoey, line 220: maybe a comment here like "Passing self in case loading plugin config depends on root config items" or something

antazoey (dismissed):
I tried using ManagerAccessMixin myself just to make sure and yeah you can't do it because it's loading still.

I can't really think of another way to do it either and this way if this problem again, we at least have a solution for it (rather than  1-off fix each time).

LGTM

antazoey (approved):

mikeshultz:
*DO NOT MERGE* this PR.  Still working through some stuff and this may not be final.

antazoey:
> **DO NOT MERGE** this PR. Still working through some stuff and this may not be final.

I converted to a Draft just to make it less likely to merge until it's ready

src/ape_compile/__init__.py
```diff
@@ -23,6 +26,56 @@ class Config(PluginConfig):
...
+        # time this prop is accessed.
+        assert self._config_manager is not None
+        assert self.cache_folder is not None
+
+        # If the dependency cache folder is configured, to be outside of the contracts dir, we want
+        # to use the projects folder to be the base dir for copmilation.
+        if self._config_manager.contracts_folder not in self.cache_folder.parents:
+            return self._config_manager.PROJECT_FOLDER
+
+        # Otherwise, we're defaulting to contracts folder for backwards compatibility. Chaning this
```
antazoey, line 48: question: Thoughts on for 0.8 using the project path as the base regardless, as a breaking change?

src/ape_compile/__init__.py
```diff
@@ -23,6 +26,56 @@ class Config(PluginConfig):
     Source exclusion globs across all file types.
     """
 
+    cache_folder: Optional[Path] = None
+    """
+    Path to contract dependency cache directory (e.g. `contracts/.cache`)
+    """
+
+    @property
+    def base_dir(self) -> Path:
```
antazoey, line 35: suggestion: rename to `base_path` to be consistent with the compiler API parameter names

src/ape_compile/__init__.py
antazoey, line 48:
```suggestion
        # Otherwise, we're defaulting to contracts folder for backwards compatibility. Changing this
```

src/ape_compile/__init__.py
```diff
@@ -23,6 +26,56 @@ class Config(PluginConfig):
...
+        elif not self.cache_folder.is_absolute():
+            self.cache_folder = project_folder / self.cache_folder
+
+        # Do not allow escape of the project folder for security and functionality reasons. Paths
+        # outside the relative compilation root are not portable and will cause bytecode changes.
+        project_resolved = project_folder.resolve()
+        cache_resolved = self.cache_folder.resolve()
+        if project_resolved not in cache_resolved.parents:
+            raise ValueError(
+                f"cache_folder must be a child of the project directory. "
```
antazoey, line 75:
```suggestion
                "cache_folder must be a child of the project directory. "
```

antazoey (approved):
LGTM - some nits, 1 question

My biggest takeaway is that we should probably change in 0.8 to using the base path as the project path. I know it is a big change; maybe we can offer another config so users can opt in the past approach when necessary.

src/ape_compile/__init__.py
mikeshultz: I have no issues with making this change.  We may want to document the old way somehow so users can reconstruct their project to match the old ways (or just instruct them to use an old version of ape and its deps, if possible). It's hard to get a sense how important repeatable builds are.

I also think it's worth discussing the idea in #1912. The functionality in it could be used for some kinds of build stages to assemble the compilation directory structure. Then it wouldn't matter where users put things. And we wouldn't need to really put a dependency cache anywhere semi-permanently. It's possible we could introduce this in a way to not be breaking.

antazoey (dismissed):

antazoey (approved):

Issue #22, opened 2021-04-13
General: Replace PyGithub w/ GHapi


fubuloubu:
PyGithub is LGPL licensed, switch to Apache license here:

https://github.com/fastai/ghapi

unparalleled-js:
Is `GPLv2` license okay?

unparalleled-js:
`pygit2` is `GPLv2` and is a more generic `git` client

fubuloubu:
> `pygit2` is `GPLv2` and is a more generic `git` client

I believe GPLv2 is copy left, meaning you have to have a license with equal or greater rights or your in violation. Hmm... Afraid to audit this lol

PR #1915, opened 2024-02-06
refactor!: swap out PyGithub library for requests

antazoey:
### What I did

Replace PyGithub lib with raw requests

fixes: #22 

### Checklist


fubuloubu:
Not sure how actively developed this is, but this might work too (since ghapi is dead):
https://github.com/sigmavirus24/github3.py

antazoey:
> Not sure how actively developed this is, but this might work too (since ghapi is dead): https://github.com/sigmavirus24/github3.py

Pros:
*  Good license
* More trusted code

Nuetral:
* Probably won't make code less complex (removing PyGithub2 didn't change much either)

Bad:

* github3.py has a large source dist size: 36 MB. That is higher than most libraries by a lot. For example, click is 336.1 kB and Ape is 1.3 MB.
* Dependency Hell

I know some of our users are trying to keep Ape's Docker image small, so avoiding a dependency may be good.
But the lib seems pretty good for the most par, I wonder about the size though.

fubuloubu:
> github3.py has a large source dist size: 36 MB

No way we are using this lol

fubuloubu (approved):

Issue #1917, opened 2024-02-07
crash on long event query after refusing to connect to localhost


wakamex:
### Environment information

- OS: linux (Ubuntu 23.10)
- Python Version: 3.10.13
- `ape` and plugin versions:

```sh
$ ape --version
0.7.8.dev6+g52f431bf

$ ape plugins list
  alchemy      0.7.1.dev10+gf332896.d20240201
  arbitrum     0.7.2
  base         0.7.1
  etherscan    0.7.1.dev1+g89b7d26
  optimism     0.7.2
  polygon      0.7.1
  solidity     0.7.1
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
cat: ape-config.yaml: No such file or directory
```

### What went wrong?

Please include information like:

Here are different commands with their associated outputs
```py
# %%
from ape import Contract, networks
from dotenv import load_dotenv

load_dotenv()

# %%
# this works on my local node
with networks.parse_network_choice('ethereum:local:http://localhost:8545') as provider:
    v2_abi = '[{"anonymous": false,"inputs": [{ "indexed": true, "name": "reportTimestamp", "type": "uint256" },{ "indexed": false, "name": "timeElapsed", "type": "uint256" },{ "indexed": false, "name": "preTotalShares", "type": "uint256" },{ "indexed": false, "name": "preTotalEther", "type": "uint256" },{ "indexed": false, "name": "postTotalShares", "type": "uint256" },{ "indexed": false, "name": "postTotalEther", "type": "uint256" },{ "indexed": false, "name": "sharesMintedAsFees", "type": "uint256" }],"name": "TokenRebased","type": "event"}]'
    steth = Contract("0xae7ab96520de3a18e5e111b5eaab095312d7fe84",abi=v2_abi)  # Lido v2 main contract
    events2 = steth.TokenRebased.query("*")
print(events2.head())
output="""
1  0xe9ba59b2b3e83d93df52528d5e17c802973f5de34602...                  3  
2  0xbf52777e4dd583d52104be96f7da420be977faeee97c...                 10  
3  0x5489a6364dc63a9b2614c7665d15827fa3b100c80dcb...                212  
4  0x54b86d0ca8df20f49ad14c3fa2899b9b21cbe144b1df...                  6  
"""

# %%
# ignores localhost, connects to public node, sometimes fails right away
[...]

PR #1919, opened 2024-02-07
fix: provider settings RPC was not being used when it was the default

antazoey:
### What I did

In ape-geth, if was given a provdider-settings URI, use that no matter what
Also, when generating config defaults, if something is running on the local, still use that as the default just in case.
Then, finally, if the default is available, use that instead of again checking for random public

fixes: #1917 

### Checklist

### DIFF
--- a/src/ape_geth/provider.py
+++ b/src/ape_geth/provider.py
@@ -473,7 +473,7 @@ class Geth(EthereumNodeProvider):
     @property
     def uri(self) -> str:
         if "uri" in self.provider_settings:
-            # If specifying in Python, user no matter what.
+            # If specifying in Python, use no matter what.
             return self.provider_settings["uri"]
 
         uri = super().uri
@@ -484,7 +484,7 @@ class Geth(EthereumNodeProvider):
         if not uri or uri == DEFAULT_SETTINGS["uri"]:
             # Do not override explicit configuration
             if ecosystem in self.config:
-                # Shape of this is odd.  Pydantic model containing dicts
+                # Shape of this is odd. Pydantic model containing dicts
                 if network_config := self.config[ecosystem].get(network):
                     if "uri" in network_config:
                         return network_config["uri"]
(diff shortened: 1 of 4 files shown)


fubuloubu (dismissed):

src/ape_geth/provider.py
```diff
@@ -196,11 +204,27 @@ def wait(self, *args, **kwargs):
         self.proc.wait(*args, **kwargs)
 
 
+class _RPCFactory:
+    @cached_property
+    def _default_available(self) -> bool:
+        return _check_url_available(DEFAULT_SETTINGS["uri"])
```
mikeshultz, line 210: suggestion: If we're going to go the HTTP request route, maybe we should at least do a chain_id check to make sure it's the right network?

mikeshultz (commented):
I get that the public RPC change was "breaking" in that it changed behavior so it may be a bug.  But I'm wondering if this is any more correct.

Say that I happen to be running a local anvil/hardhat node that I'm using for unrelated dev purposes.  Then I run `ape console --network ethereum:mainnet` to do *otherwork*. I wouldn't really want it to connect to my local node just because it was responds to a HEAD request.

src/ape_geth/provider.py
antazoey: Good point, I have an idea that is much simpler that avoids all this. Brb.

tests/functional/geth/test_provider.py
```diff
@@ -40,22 +41,31 @@ def test_uri(geth_provider):
...
-    config = {"geth": {"ethereum": {"local": {"uri": value}}}}
+    config = {"geth": {"ethereum": {"local": {"uri": value}, "mainnet": {"uri": value}}}}
     try:
         with temp_config(config):
+            # Assert we use the config value.
             assert geth_provider.uri == value
+
+            # Assert provider settings takes precedence.
+            expected = DEFAULT_SETTINGS["uri"]
+            for network in ("local", "mainnet"):
```
antazoey, line 64: `mainnet` example here fails without the change
now, no longer weird - all that is changing here is that if the user specifies the value in `--network ethereum:mainnet:http://localhost:8545` it will use that value instead of the random public one.

But if the user does not do that, it will still use the public one for mainnet and the local one for local by default

mikeshultz: That works for me if it works for you.

mikeshultz (approved):
LGTM

PR #1925, opened 2024-02-09
perf: lazy model extras and IPython canary short-circuitry

antazoey:
### What I did

perf: evaluates extras lazily
perf: short-circuit during IPython canary check

the combo of these makes it not compile at weird times! people kept saying ape always compiles so much, I think this was the cause.

watch this:

go to ape main branch in an uncompiled projet and launch a console:

```sh
ape console

# THEN TYPE `project`
[1] project
```

it immediately starts compiling the project :[ but why doh

### How I did it

* first i tried __repr__ in project manager, didnt matter though :/ but it seemed like it helped kinda
* then i added the lazy eval for the extras. this is great. now, itll calculate the extras if you ask it to making ape seem a lot faster
* finally, what the heck ipython was triggering stuff in Ape it shouldnt have, like compiling, via this canary attribute: https://stackoverflow.com/questions/56068348/what-is-ipython-canary-method-should-not-exist
so to handle this, I short-circuit with what it actually wants. It just wants to make sure we are not saying we have all the full attribute name space because that would crash IPython, so it just wants an attribute error. by raising early, we avoid all the stuff we thought we were to begin with!

for example, when you do `project.MyContract`, if the contract is not found, Ape will compile the missing sources and then check again. So `ape console` then typing `project`, IPython sends this canary message and ape is like "whattttt this aint something i know, but i realize there are uncompiled contracts here so lemme just go away and compile dose and then check again.... ok, i still didnd find `_ipython_canary_bullshit_` so ima just raise AttributeError. and then IPython is like "wow thanks, i was just making sure you ain't bat-shit nuts and claiming you have all these random attributes because that would be really bad for me" and then Ape is like "we gooooood"... and so me as a dev is like, why not just formalize our little chat, skip the check for uncompiled sources because we know it aint in there, we have done this every single time we launch `ape console` for the past 3 years or whatever, so heyyy, performance me upppp. happy friday.

### DIFF
--- a/src/ape/api/projects.py
+++ b/src/ape/api/projects.py
@@ -366,7 +366,7 @@ class DependencyAPI(ExtraAttributesMixin, BaseInterfaceModel):
     def __ape_extra_attributes__(self) -> Iterator[ExtraModelAttributes]:
         yield ExtraModelAttributes(
             name=self.name,
-            attributes=self.contracts,
+            attributes=lambda: self.contracts,
             include_getattr=True,
             include_getitem=True,
             additional_error_message="Do you have the necessary compiler plugins installed?",
--- a/src/ape/managers/project/manager.py
+++ b/src/ape/managers/project/manager.py
@@ -536,7 +536,12 @@ class ProjectManager(BaseManager):
         try:
             return self.__getattribute__(attr_name)
         except AttributeError:
-            if not self._getattr_contracts:
+            # NOTE: Also handles IPython attributes such as _ipython_display_
+            if (
+                not self._getattr_contracts
+                or attr_name.startswith("_ipython_")
+                or attr_name.startswith("_repr_")
+            ):
                 # Raise the attribute error as if this method didn't exist.
                 raise
 
--- a/src/ape/managers/project/manager.py
+++ b/src/ape/managers/project/manager.py
@@ -537,14 +537,10 @@ class ProjectManager(BaseManager):
             return self.__getattribute__(attr_name)
         except AttributeError:
             # NOTE: Also handles IPython attributes such as _ipython_display_
-            if (
-                not self._getattr_contracts
-                or attr_name.startswith("_ipython_")
-                or attr_name.startswith("_repr_")
-            ):
+            if not self._getattr_contracts:
                 # Raise the attribute error as if this method didn't exist.
                 raise
-
+        breakpoint()
         try:
             # NOTE: Will compile project (if needed)
             if contract := self._get_contract(attr_name):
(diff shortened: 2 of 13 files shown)


src/ape/api/projects.py
```diff
@@ -366,7 +366,7 @@ def __repr__(self):
     def __ape_extra_attributes__(self) -> Iterator[ExtraModelAttributes]:
         yield ExtraModelAttributes(
             name=self.name,
-            attributes=self.contracts,
+            attributes=lambda: self.contracts,
```
fubuloubu, line 369: It knows to eval a function to obtain attribute names every time? Or does it once and caches? (if the later, it should be represented in a comment)

src/ape/managers/project/manager.py
```diff
@@ -56,6 +57,9 @@ def __init__(
     def __str__(self) -> str:
         return f'Project("{self.path}")'
 
+    def __repr__(self):
+        return "<ProjectManager>"
```
fubuloubu, line 61: Would love to see the path here too

src/ape/api/projects.py
antazoey: it does once and caches

antazoey: but ya, realizing it shouldnt do it that way since it may be cached already, so ill fix that

src/ape/managers/project/manager.py
```diff
@@ -526,7 +536,12 @@ def _get_attr(self, attr_name: str) -> Any:
         try:
             return self.__getattribute__(attr_name)
         except AttributeError:
-            if not self._getattr_contracts:
+            # NOTE: Also handles IPython attributes such as _ipython_display_
+            if (
+                not self._getattr_contracts
+                or attr_name.startswith("_ipython_")
```
antazoey, line 542: i really wish i can figure out a better way...

antazoey: yeah, ugh, need to make this all smarter

antazoey:
Check this out.
Running from main branch:

```
In [1]: %time project
CPU times: user 3 µs, sys: 1 µs, total: 4 µs
Wall time: 5.96 µs
Out[1]: INFO: Compiling 'manifest_dependency.json'.
WARNING: Unable to parse ContractType from 'manifest_dependency.json'.
INFO: Compiling 'manifest_dependency.json'.
WARNING: Unable to parse ContractType from 'manifest_dependency.json'.
INFO: Compiling 'manifest_dependency.json'.
WARNING: Unable to parse ContractType from 'manifest_dependency.json'.
INFO: Compiling 'manifest_dependency.json'.
WARNING: Unable to parse ContractType from 'manifest_dependency.json'.
<ape.managers.project.manager.ProjectManager at 0x121f03be0>
```

WTF

Now, running from this branch:

```
In [2]: %time project
CPU times: user 3 µs, sys: 0 ns, total: 3 µs
Wall time: 11 µs
Out[2]: <ProjectManager $HOME/PycharmProjects/more_apes/1/ape>
```

fubuloubu (approved):

PR #1928, opened 2024-02-12
feat: EIP-4844 (type 3) transactions initial support

antazoey:
### What I did

This allows us to decode transactions and receipts from EIP-4844

There are still some issues however that will require updating eth-account and other eth-prefixed python packages.
For example, the hash calculation fails because of an unknown type (3).

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -47,6 +47,7 @@ class TransactionType(Enum):
     STATIC = 0
     ACCESS_LIST = 1  # EIP-2930
     DYNAMIC = 2  # EIP-1559
+    SHARED_BLOB = 3  # EIP-4844
 
 
 class AccessList(BaseModel):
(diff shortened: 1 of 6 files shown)


antazoey:
Before we can sign: https://github.com/ethereum/eth-account/issues/249

src/ape_ethereum/transactions.py
```diff
@@ -404,3 +424,13 @@ def _decode_ds_note(self, log: Dict) -> Optional[ContractLog]:
             transaction_hash=log["transactionHash"],
             transaction_index=log["transactionIndex"],
         )
+
+
+class SharedBlobReceipt(Receipt):
+    blob_gas_used: int
+    blob_gas_price: int
+
```
fubuloubu, line 432: can't actually fetch the blob without access to a consensus node, right?

fubuloubu (approved):

src/ape_ethereum/transactions.py
antazoey: I used Alchemy to get the receipt.

fubuloubu: Oh I mean is the actual blob data fetchable via RPC? AFAIK this type of transaction is just a commitment to a particular blob, which must be fetched from a consensus node

antazoey: i have no idea how to get the blob but the tx.data is the like the proof or something (summarizing discussion from elsewhere)

PR #1930, opened 2024-02-13
feat: log print and console.log calls made during transactions

mikeshultz:
### What I did

Logs output from calls to Vyper's `print()` and `hardhat/console.sol`'s `console.log()` calls.

fixes: #685
fixes: https://github.com/ApeWorX/ape-hardhat/issues/125

### How I did it

After a transaction, Ape will go through the call trace looking for relevant staticcalls.  If it finds them, it'll decode and log the data to the console.

### How to verify it

TBD

### Checklist


src/ape_ethereum/provider.py
```diff
@@ -917,6 +917,11 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
             # a VM error.
             receipt.raise_for_status()
 
+        # TODO: Optional configuration?
+        if txn.data:
```
mikeshultz, line 921: Need something better since not all contract executions will have data (fallback, for instance).  I guess this will require an RPC call for `eth_getCode()` to verify that it's a contract.  I bet there's some useful utils somewhere for this, maybe as part of the `Address` type?

src/ape_ethereum/provider.py
```diff
@@ -917,6 +917,11 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
+            # Look for and print any contract logging
+            log_print(receipt)
```
mikeshultz, line 923: I'm not really big on the placement of this because it requires implementation in all relevant providers.  In fact, there's a bit of redundant code between all the providers I checked.  For instance, the history addition.

https://github.com/ApeWorX/ape-foundry/blob/0b6b55f370c6bb1d460c7100a07b5b718b1496d7/ape_foundry/provider.py#L576

I'm wondering of there should be a `post_transaction(tx, receipt)` method on `Web3Provider` that gets calls on all implementations.  Or maybe there's a better idea?  Just seems annoyingly redundant to have to call this same function from every provider that implements `send_transaction()`.

src/ape_ethereum/provider.py
antazoey: yep, there is a `is_contract` property available on `ContractInstance` via the `Address` base.

mikeshultz: Abstraction spelunking... 

- [`ape.types.address.AddressType`](https://github.com/ApeWorX/ape/blob/main/src/ape/types/address.py)
- [`eth_pydantic_types.address.Address`](https://github.com/ApeWorX/eth-pydantic-types/blob/main/eth_pydantic_types/address.py)
- [`eth_typing.ChecksumAddress`](https://github.com/ethereum/eth-typing/blob/main/eth_typing/evm.py)
- [`ape.api.address.Address`](https://github.com/ApeWorX/ape/blob/2ef1d7901126120d171b5681f5ea57f3c0350b67/src/ape/api/address.py#L183) (and `BaseAddress`)

:sweat_smile:

src/ape_ethereum/provider.py
mikeshultz: I added the hook.  That way subclassed providers will also have this functionality.

src/ape_ethereum/_print.py
fubuloubu, line 1: Curious if we could refactor this into more of an "enrichment" type thing on receipts by adding an API method to `ReceiptAPI` (e.g. `ReceiptAPI.show_debug_logs()`)

@antazoey what do you think about that?

mikeshultz: is `ReceiptAPI` properly ecosystem agnostic?  I wouldn't really expect this use to be outside of the Ethereum/EVM ecosystem any time soon.

Also, if we do add a public API, maybe we should have a method/prop that returns the strings that users can manipulate rather than just logging them (e.g. `debug_logs(self) -> List[str]`).  Then we could just use it to log them in the provider.

fubuloubu: > is `ReceiptAPI` properly ecosystem agnostic? I wouldn't really expect this use to be outside of the Ethereum/EVM ecosystem any time soon.

Yes, broadly it represents the state of a transaction post-network broadcast

> Also, if we do add a public API, maybe we should have a method/prop that returns the strings that users can manipulate rather than just logging them (e.g. debug_logs(self) -> List[str]). Then we could just use it to log them in the provider.

Not sure what this means

mikeshultz: > Yes, broadly it represents the state of a transaction post-network broadcast

And this console/print stuff is Ethereum specific, parsing the call tree with EVM addresses and method signatures.  Is `ReceiptAPI` an appropriate place for it?

> Not sure what this means

What part?  I was proposing a method signature and how to use it if we went this route.

fubuloubu: > > Yes, broadly it represents the state of a transaction post-network broadcast
> 
> And this console/print stuff is Ethereum specific, parsing the call tree with EVM addresses and method signatures. Is `ReceiptAPI` an appropriate place for it?

It would be an appropriate place for it if properly designed. A lot of these methods in `ReceiptAPI` are (or should be) agnostic to how the computations are performed, and describe the resulting data structures we need to perform the features in Ape core, such as displaying call trees (this is a base assumption of Ape's environment system, that addresses can call other addresses and there is data in this tree structure that can be enriched for display), collecting gas reports (another assumption is that there is resource utilization with these calls), and coverage data for contracts (the assumption being that accounts that are targets of calls which contain handling logic can have source code associated with it which we want to see various details of how well covered that source is). At least, these are guiding principles for the `ReceiptAPI` object, you may find instances where it's not perfect, but our goal is to make it independent of the specifics of how any one ecosystem formats code (let's say it's a zkVM that formats code differently) or performs transaction-level tracing.

> 
> > Not sure what this means
> 
> What part? I was proposing a method signature and how to use it if we went this route.

What types of strings would users manipulate? Can you give some examples?

mikeshultz: > > And this console/print stuff is Ethereum specific, parsing the call tree with EVM addresses and method signatures. Is `ReceiptAPI` an appropriate place for it?
> 
> It would be an appropriate place for it if properly designed. [...]

I guess this is what I'm getting at.  What would be the appropriate design?  Add the method to `ReceiptAPI` returning None or raising NotImplemented and then implement this version in [`ape_ethereum.transactions.Receipt`](https://github.com/ApeWorX/ape/blob/2ef1d7901126120d171b5681f5ea57f3c0350b67/src/ape_ethereum/transactions.py#L184)?  Looks like that's more or less how `call_tree` is done.

> > What part? I was proposing a method signature and how to use it if we went this route.
> 
> What types of strings would users manipulate? Can you give some examples?

The logged data.  `console.log(data)`.  I can just put together some code and you can provide feedback.

antazoey: btw, EcosystemAPIs can and do define their own ReceiptAPIs... It is based on what they return in `decode_receipt()`. An example is in Arbitrum, they have a receipt for their weird transaction type because it has different fields. Starknet had its own, etc.

And yeah I agree, getting logs is just as good as showing them because in `ape console` (or any python shell), you get something, it shows that thing's repr.

mikeshultz: Updated `ReceiptAPI` to add a new prop and method in b5659baf.  lmk what you both think.

src/ape/api/transactions.py
```diff
@@ -294,6 +305,18 @@ def validate_txn_hash(cls, value):
     def call_tree(self) -> Optional[Any]:
         return None
 
+    @cached_property
+    def debug_logs(self) -> Iterable[Tuple[Any]]:
```
fubuloubu, line 309: I think we can make this more streamlined
```suggestion
    def debug_logs(self) -> Iterable[str]:
```

src/ape/api/transactions.py
```diff
@@ -294,6 +305,18 @@ def validate_txn_hash(cls, value):
...
+        """Return any debug log data outputted by the transaction"""
+        return []
+
+    def print_debug_logs(self):
+        """
+        Log debug logs to the console
+        """
+        for ln in self.debug_logs:
+            logger.info("[CONTRACT-LOG] " + " ".join(map(str, ln)))
```
fubuloubu, line 318:
```suggestion
        logger.info(f"[CONTRACT-LOG]\n  {'\n  '.join(self.debug_logs)}")
```

src/ape_ethereum/_print.py
```diff
@@ -0,0 +1,104 @@
...
+            # Empty string as first arg?
+            pass
+    return False
+
+
+def console_log(method_abi: MethodABI, calldata: str) -> Tuple[Any]:
+    """Return logged data for console.log() calls"""
+    bcalldata = decode_hex(calldata)
+    data = ape.networks.ethereum.decode_calldata(method_abi, bcalldata)
+    return tuple(data.values())
```
fubuloubu, line 82:
```suggestion
    return f"console.log({', '.join(data.values())})"
```

src/ape_ethereum/_print.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def vyper_print(calldata: str) -> Tuple[Any]:
```
fubuloubu, line 85:
```suggestion
def vyper_print(calldata: str) -> str:
```

src/ape_ethereum/_print.py
```diff
@@ -0,0 +1,104 @@
...
+    """Return logged data for print() calls"""
+    bcalldata = decode_hex(calldata)
+    schema, payload = decode(["string", "bytes"], bcalldata)
+    data = decode(schema.strip("()").split(","), payload)
+    return tuple(data)
```
fubuloubu, line 90:
```suggestion
    return f"print({', '.join(data)})"
```

src/ape_ethereum/_print.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def extract_debug_logs(call_tree: CallTreeNode) -> Iterable[Tuple[Any]]:
```
fubuloubu, line 93:
```suggestion
def extract_debug_logs(call_tree: CallTreeNode) -> Iterable[str]:
```

src/ape_ethereum/transactions.py
```diff
@@ -207,6 +208,21 @@ def failed(self) -> bool:
     def call_tree(self) -> Optional[CallTreeNode]:
         return self.provider.get_call_tree(self.txn_hash)
 
+    @cached_property
+    def debug_logs(self) -> Iterable[Tuple[Any]]:
```
fubuloubu, line 212:
```suggestion
    def debug_logs(self) -> Iterable[str]:
```

src/ape_ethereum/transactions.py
```diff
@@ -207,6 +208,21 @@ def failed(self) -> bool:
...
+        """
+
+        # Some providers do not implement this, so skip
+        try:
+            self.call_tree
+        except APINotImplementedError:
+            logger.debug("Call tree not available, skipping print log extraction")
+            return []
+
+        return extract_debug_logs(self.call_tree) if self.call_tree is not None else []
```
fubuloubu, line 224:
```suggestion
        return list(extract_debug_logs(self.call_tree)) if self.call_tree is not None else []
```

fubuloubu (commented):
trying to flatten the structure a bit

src/ape/api/transactions.py
```diff
@@ -294,6 +294,18 @@ def validate_txn_hash(cls, value):
     def call_tree(self) -> Optional[Any]:
         return None
 
+    @cached_property
+    def debug_logs(self) -> List[List[Any]]:
+        """Return any debug log data outputted by the transaction"""
+        return []
+
+    def print_debug_logs(self):
```
antazoey, line 313: To align with other similar methods on this class:

```suggestion
    def show_debug_logs(self):
```

src/ape/api/transactions.py
```diff
@@ -294,6 +294,18 @@ def validate_txn_hash(cls, value):
...
+        """
+        Log debug logs to the console
+        """
+        for ln in self.debug_logs:
+            logger.info("[CONTRACT-LOG] ".join(map(str, ln)))
```
antazoey: worry: `CONTRACT-LOG` seems like an overloaded term which may be confusing. It makes me think it is a "contract log" like an event.

src/ape_ethereum/transactions.py
```diff
@@ -207,6 +214,34 @@ def failed(self) -> bool:
...
+        except APINotImplementedError:
+            logger.debug("Call tree not available, skipping print log extraction")
+            return lines
+
+        # Extract any Vyper print() calls to output
+        for call in extract_prints(self):
+            if call.inputs is not None:
+                lines.append(vyper_print(call.inputs))
+
+        # Extract any Hardhat console.log() calls to output
```
antazoey: thought: The whole reason we made the plugin system is to handle all these one-offs in their respective plugins.
I understand the architecture might not be accommodating for this right now, but it's something to think about. 

Because of suddenly Foundry invents its own log format, now we have to adjust here.

And users who are not using Hardhat, this is still checked for.

fubuloubu: some complicating factors here: foundry uses hardhat style console.log, and vyper has a compat mode that also exposes it. so, may not be as simple or relegated to the hardhat plugin only, the behavior is more widely adopted across ethereum ecosystem tooling

src/ape/api/transactions.py
mikeshultz: They're not all strings

mikeshultz: (just noticed the below)

Wouldn't the native types be more useful rather than making them all flat strings?

src/ape/api/transactions.py
mikeshultz: I just took the recommendation.  What do you think would be better?

src/ape_ethereum/transactions.py
mikeshultz: Yeah, it's not really a hardhat thing.  It's just a contract they wrote and others sort of standardized to.  It's very possible to never use hardhat but use their console.sol contract for this stuff.

mikeshultz:
Trying to bring the discussion into the PR from commit-comments: The reason it's not just a flat list of strings is because I think the user might want to access these vars and not just use them as debug strings.  So say a solididity func does something like this:

```solidity
console.log("line one", 1, address(1));
console.log("line two", 2, address(2));
```

Then `receipt.debug_logs()` should output something like this:

```python
[
    ("line one", 1, ChecksumAddress(0x0000000000000000000000000000000000000001)),
    ("line two", 2, ChecksumAddress(0x0000000000000000000000000000000000000002)),
]
```

That seems more useful to me than just giving them `"list one 1 0x0000000000000000000000000000000000000001\nline two 2 0x0000000000000000000000000000000000000002"` that they may then need to parse if they're looking for programatically-useful types.

src/ape/api/transactions.py
fubuloubu: Have to convert them somewhere, rather do it before caching it than afterwards

mikeshultz: `show_debug_logs()` is pretty unlikely to be accessed multiple times.  The `debug_logs` prop might be if leveraged in a user script or console.  Could cache both if you think it would be beneficial.

As a programmer, getting access to typed data is significantly more useful to me than a list of strings.  Converting `debug_logs` would essentially prevent that access or force them to re-implement what we already have.  Maybe we could compromise and provide both?

```python
    @cached_property
    def debug_logs_typed(self) -> Iterable[Tuple[Any]]:
        ...

    @cached_property
    def debug_log_lines(self) -> Iterable[str]:
        ...
```

src/ape_ethereum/transactions.py
antazoey: Ah got it, I understand now

src/ape/api/transactions.py
antazoey: `DEBUG-LOG` I think is better, I saw that somewhere else and I knew it was different than `.logs`.

src/ape/api/transactions.py
```diff
@@ -294,6 +294,25 @@ def validate_txn_hash(cls, value):
...
+        Return any debug log data outputted by the transaction as strings suitable for printing
+        """
+        return [" ".join(map(str, ln)) for ln in self.debug_logs_typed]
+
+    def show_debug_logs(self):
+        """
+        Output debug logs to logging system
+        """
+        for ln in self.debug_logs_lines:
+            logger.info("[DEBUG-LOG] " + ln)
```
antazoey, line 314: nit: i love f-strings, they are so readable

```suggestion
            logger.info(f"[DEBUG-LOG] {ln}")
```

src/ape/api/transactions.py
```diff
@@ -294,6 +294,25 @@ def validate_txn_hash(cls, value):
     def call_tree(self) -> Optional[Any]:
         return None
 
+    @cached_property
+    def debug_logs_typed(self) -> List[Tuple[Any]]:
+        """Return any debug log data outputted by the transaction"""
```
antazoey, line 299: nit: most of our docs end in periods i think

```suggestion
        """Return any debug log data outputted by the transaction."""
```

src/ape/api/transactions.py
```diff
@@ -294,6 +294,25 @@ def validate_txn_hash(cls, value):
...
+        return []
+
+    @cached_property
+    def debug_logs_lines(self) -> List[str]:
+        """
+        Return any debug log data outputted by the transaction as strings suitable for printing
```
antazoey, line 305: nit: for properties, word 'return' is kinda unnecessary.

```suggestion
        Any debug log data outputted by the transaction as strings suitable for printing.
```

src/ape_ethereum/transactions.py
```diff
@@ -207,6 +208,26 @@ def failed(self) -> bool:
...
         return self.provider.get_call_tree(self.txn_hash)
 
+    @cached_property
+    def debug_logs_typed(self) -> List[Tuple[Any]]:
+        """
+        Extract messages to console outputted by contracts via print() or console.log() statements
+        """
+
+        try:
+            assert self.call_tree is not None
```
antazoey, line 218: instead of assert + catching assertion error, check if it None later on.

When compiling python to an optimized mode, asserts are removed, they should not be used this way.

tests/integration/cli/test_test.py
```diff
@@ -140,6 +140,8 @@ def run_gas_test(
         remainder = "\n".join(expected[actual_len:])
         pytest.fail(f"Expected contains more than actual:\n{remainder}")
 
+    print("actual:", "\n".join(actual))
```
antazoey, line 143: - [ ] delete print statements

antazoey (commented):
lookin good

src/ape_ethereum/transactions.py
mikeshultz: Spaced on this.  Don't know why I like this pattern so much.  Will change.

tests/integration/cli/test_test.py
mikeshultz: This was intentional instrumentation because I wasn't immediately able to repro a CI test issue.

mikeshultz:
This test error has been painful to get to the bottom of.  It only shows up when running the full suite of tests the same way the CI runs them (which takes forever so each repro is like 15 minutes).

Apparently the gas report is picking up the `printing` and `console` contracts in some fashion, for only one test (and not always?).  Even though as far as I can tell, none of the test projects used in the integration tests make any Vyper `print()` calls.  And the abstraction of pytest running pytest in the failing test sure isn't making sussing it out easy either.

The contracts I've added, are part of the geth functional tests, which I wouldn't expect to impact the integration tests and their (isolated?) projects.

@antazoey any tips on this failure?  Look familiar by chance?  Any background/overview that might help me understand these tests?

mikeshultz:
I am flummoxed and could use a second brain on this.  This test error is gnarly.  This is basically what I know so far:

- Happens when either or both of my new tests are enabled
- Only occurs when running the full test suite (or a large combination of tests I wasn't able to figure out).
- Only the last of the transactions from one of the tests is added to the gas report (so `print_uint()` or `print_uint_compat()` not both)
- Only occurs if pytest is run multi-process
- Running the tests via `ape test --gas --network ethereum:local:geth` in the same generated project on the same running geth instance passes and produces the expected gas report without mention of the printing contract

Vague ideas I might still want to chase up: 

- I wonder if the gas report uses on-disk files that might be conflicting with multiple pytest subprocesses.  Though I'd expect to see more race behavior instead of it being the same test failing every time.
- Some cached prop in ape somewhere?  Though with subprocesses, I wouldn't expect shared memory, unless these tests are consistently grouped together in one of the pytest workers.
- Some strange behavior of having pytest in pytest in pytest causing vague hard to define errors that are impossible to research
- Gremlins

I'm running out of ideas on this one though and might move on to something else until I get some kind of inspiration on how to proceed.

antazoey:
I seem to repro the problem by running:

```python
pytest tests/functional/geth/test_trace.py tests/integration/cli/test_test.py::test_gas_flag_in_tests
```

No x-dist needed and don't need to run the full suite.

mikeshultz:
> I seem to repro a problem by running:
> 
> ```python
> pytest tests/functional/geth/test_trace.py::test_printing_debug_logs_compat tests/integration/cli/test_test.py::test_gas_flag_in_tests
> ```

Swear I tried this but can't find it in my history.  Can confirm that it repros the issue locally though.

antazoey (approved):

Issue #1935, opened 2024-02-16
"size" field missing for Blocks from `newHeads` subscription method


fubuloubu:
### Environment information

- OS: linux
- Python Version: 3.11.7
- `ape` and plugin versions:

```sh
$ ape --version
0.7.9

$ ape plugins list
Installed Plugins
  alchemy      0.7.1
  etherscan    0.7.1
  infura       0.7.1
  optimism     0.7.2
```

### What went wrong?

The field `"size"` is not available in the serialized data structure, at least for Optimism using an Infura provider (both up to date):

```py
  File "bin/silverback", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/ape/cli/commands.py", line 96, in invoke
    return self._invoke(ctx, provider=provider)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/ape/cli/commands.py", line 133, in _invoke
    return ctx.invoke(self.callback or (lambda: None), **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/silverback/_cli.py", line 100, in run
    asyncio.run(runner.run())
  File "python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
[...]

PR #1936, opened 2024-02-16
fix(ape-ethereum): `.size` missing from `Block` for `newHeads` subs

fubuloubu:
### What I did

TEMP HACK: https://github.com/ApeWorX/ape/issues/1935

### How I did it

Realized that `eth_subscribe:newHeads` does not provide this info, so put a sentinel value in it's place

### How to verify it

Silverback apps work with Infura

### Checklist


antazoey (approved):

src/ape_ethereum/ecosystem.py
```diff
@@ -551,6 +551,11 @@ def decode_block(self, data: Dict) -> BlockAPI:
         if "transactions" in data:
             data["num_transactions"] = len(data["transactions"])
 
+        if "size" not in data:
+            # NOTE: Due to an issue with `eth_subscribe:newHeads` on Infura
```
antazoey, line 555: Question: How can we re-design block API to handle a transaction-less state better.

fubuloubu: maybe combo of `_size: Optional[int] = Field(None, alias="size")` and a property `size(self) -> int` that either returns the cached value or triggers another call to fetch `.transactions` in order to compute the size from them

antazoey:
@fubuloubu Any reason we've been holding off on merging this one?

fubuloubu:
> @fubuloubu Any reason we've been holding off on merging this one?

Oh right! Lack of testing

Actually learning that IPC supports subscriptions is pretty cool, probably makes it easier to test

But for now, testing is probably not a deal breaker

antazoey (approved):

PR #1946, opened 2024-03-05
feat: add conversion for fixed ABI types

BowTiedDevil:
### What I did

Ape has no converter to handle Vyper's `decimal` type, which is translated to the `fixed168x10` ABI type. Testing a contract function with a `decimal` type argument will fail with a `ConversionError`.

#### `ape_ethereum` plugin:
- Add `Decimal` as the native Python type for ABI types with "fixed" in the name.

#### `ape` base:
- Add a converter to support using string-formatted float values instead of requiring a strictly `Decimal` value. This converter will translate a "1.0" string in-place. Did not add a `float` converter, because floating-point accuracy will cause subtle errors. 

#### Tests:
- Added simple string -> Decimal converter test.

### Checklist:

### DIFF
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -192,9 +192,9 @@ class StringDecimalConverter(ConverterAPI):
     """
 
     def is_convertible(self, value: Any) -> bool:
-        # Matches only string-formatted floats with an optional sign character (+/-),
-        # leading and trailing zeros are optional.
-        return isinstance(value, str) and re.fullmatch("[+-]?[0-9]*[.][0-9]*", value)
+        # Matches only string-formatted floats with an optional sign character (+/-).
+        # Leading and trailing zeros are required.
+        return isinstance(value, str) and re.fullmatch("[+-]?[0-9]+[.][0-9]+", value) is not None
 
     def convert(self, value: str) -> Decimal:
         return Decimal(value)
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -194,7 +194,7 @@ class StringDecimalConverter(ConverterAPI):
     def is_convertible(self, value: Any) -> bool:
         # Matches only string-formatted floats with an optional sign character (+/-).
         # Leading and trailing zeros are required.
-        return isinstance(value, str) and re.fullmatch("[+-]?[0-9]+[.][0-9]+", value) is not None
+        return isinstance(value, str) and re.fullmatch(r"[+-]?\d+\.\d+", value) is not None
 
     def convert(self, value: str) -> Decimal:
         return Decimal(value)
--- a/src/ape/managers/converters.py
+++ b/src/ape/managers/converters.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime, timedelta, timezone
 from decimal import Decimal
 from typing import Any, Dict, List, Sequence, Tuple, Type, Union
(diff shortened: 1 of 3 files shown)


src/ape/managers/converters.py
```diff
@@ -185,6 +186,20 @@ def convert(self, value: Union[str, datetime, timedelta]) -> int:
...
 
+class StringDecimalConverter(ConverterAPI):
+    """
+    Convert string-formatted floating point values to `Decimal` type.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        # Matches only string-formatted floats with an optional sign character (+/-),
+        # leading and trailing zeros are optional.
+        return isinstance(value, str) and re.fullmatch("[+-]?[0-9]*[.][0-9]*", value)
```
fubuloubu, line 197: Only nit is if theres a simpler regex to run here that picks up decimal values

Or even just trying to cast to decimal and catching exceptions

fubuloubu (approved):

src/ape/managers/converters.py
BowTiedDevil: I've reworked the regex to require digits before and after the decimal, because a "." input was slipping through.

fubuloubu: using `\d` might even be clearer (lots of alpha [here](https://docs.python.org/3/howto/regex.html))

fubuloubu: also make sure to use `\.` and not just `.` (which is "any character")

fubuloubu:
Also, general testing advice, using a regex is usually great paired with fuzzing to find corner cases for your expression (given positive and negative examples of what you want it to handle)

src/ape/managers/converters.py
BowTiedDevil: That's interesting, I used \d and got a depreciation warning from the re module. I didn't look further into it, but will take another look now.

BowTiedDevil: Resolved with ba341bb

fubuloubu (dismissed):

tests/functional/conversion/test_decimal.py
```diff
@@ -0,0 +1,22 @@
+from decimal import Decimal
+
+import pytest
+
+
+def test_converting_formatted_float_strings_to_decimal(convert):
```
antazoey, line 6: nit: I tend to use `-k` to run multiple tests and it works great if following this pattern:

```
test_<method-name>_<rest>`
```

```suggestion
def test_convert_formatted_float_strings_to_decimal(convert):
```

antazoey (approved):

src/ape/managers/converters.py
```diff
@@ -185,6 +186,20 @@ def convert(self, value: Union[str, datetime, timedelta]) -> int:
...
 
+class StringDecimalConverter(ConverterAPI):
+    """
+    Convert string-formatted floating point values to `Decimal` type.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        # Matches only string-formatted floats with an optional sign character (+/-).
+        # Leading and trailing zeros are required.
+        return isinstance(value, str) and re.fullmatch(r"[+-]?\d+\.\d+", value) is not None
```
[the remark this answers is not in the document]
BowTiedDevil: Confirmed, `re.fullmatch` will check the full string - it _should_ be equivalent to sandwiching the above expression with `^...$` but I'm not experienced enough with regex to know for sure if that expression has edge cases.

Added a note to the comment with b31222a

tests/functional/conversion/test_decimal.py
BowTiedDevil: I reformatted the test names per your suggestion, and added some more to check other bad strings per @fubuloubu's comment above.

tests/functional/conversion/test_decimal.py
```diff
@@ -0,0 +1,56 @@
...
+            convert(test_string, Decimal)
+
+
+def test_convert_strings_with_token_names(convert):
+    test_strings = [
+        "0.999 DAI",
+        "10.0 USDC",
+    ]
+    for test_string in test_strings:
+        with pytest.raises(Exception):
```
antazoey, line 44: The actual exception should go in these spots... ConversionError?

BowTiedDevil: Good call, I've swapped `Exception` to `ConversionError` and included a pattern match in 99f61b8

antazoey (approved):

fubuloubu (approved):

PR #1948, opened 2024-03-05
feat(ethereum): add builder pattern support to eth multicall

fubuloubu:
### What I did

fixes: #

### Checklist


src/ape_ethereum/multicall/handlers.py
```diff
@@ -165,6 +177,7 @@ def add(self, call: ContractMethodHandler, *args, **kwargs):
 
         super().add(call, *args, **kwargs)
         self.abis.append(_select_method_abi(call.abis, args))
+        return self
 
     @property
     def returnData(self) -> List[HexBytes]:
```
antazoey, line 183: question: how come this is camelCase?

antazoey (approved):

src/ape_ethereum/multicall/handlers.py
fubuloubu: To match the name that it has in the contract

I felt like it would be expected when working with that property that it matched, would feel awkward to have to remember it changes slightly

fubuloubu (commented):
More technically accurate

src/ape_ethereum/multicall/handlers.py
antazoey: Both might be nice, but ya it makes sense..
But also - a comment explaining the veer from the norm would be appreciated

fubuloubu: For the purposes of the expected usage, it's kinda of "internal", so I don't think we need both

But yes definitely explaining in a comment is a good suggestion

fubuloubu:
```suggestion
    def returnData(self) -> List[HexBytes]:
        # NOTE: this property is kept camelCase to align with the raw EVM struct
```

antazoey (approved):

Issue #1945, opened 2024-03-04
Support call traces on reverted txns


fp-crypto:
### Overview

At present, reverted txns do not provide a call trace. This functionality would be invaluable.

### Dependencies

mikeshultz:
Couple clarifying questions here:

- Which network and provider are you using?  
- What behavior were you expecting from Ape?

Digging into this issue today, I found something that might be what you were looking for. Would this suit your needs?

```python
from ape.exceptions import ContractLogicError
try:
    # My test contract call that reverts
    contract.revertExtern(True, sender=deployer)
except ContractLogicError as err:
    err.txn.show_trace()
    raise
```

This should give you an output something like this:

```
Call trace for '0xd0cb99a4fd37ae95c67d789cc3aa6f55b2ceb217e013dee3651614ba2f34455c'
reverted with message: "revertCall"
tx.origin=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
Reverter.revertExtern(go=True) -> <?> [27675 gas]
└── 0xa16E02E87b7454126E5E10d957A927A7F5B5d2be.0x3381f788(0x0000000000000000000000000000000000000000
    000000000000000000000001) -> 
    0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000
    00000000000000000000000000000000000000000a72657665727443616c6c0000000000000000000000000000000000
    0000000000 [354 gas]
```

I don't think this functionality is obvious.  It's something I just found when digging into this report.  Definitely some room for improvement here either in the API or in documentation at least.

fubuloubu:
was thinking a flag `ignore_revert=True` would be handy here for several use cases:

```py
from ape_ethereum import TransactionStatusEnum

tx = contract.revertExtern(True, sender=deployer, ignore_revert=True)  # skips `tx.raise_for_status()`
assert tx.status is TransactionStatusEnum.NO_ERROR, tx.show_trace()
# OR
breakpoint()  # check stuff off of `tx` manually before resuming
# OR just always proceeded despite a failure
```

fubuloubu:
It should probably show traces without `ignore_revert` set as the default behavior, but one nice thing about the override flag ignoring that behavior is we have a workaround in case it is broken

mikeshultz:
Alright, so traces should print to console on every revert regardless.  Will put something together.

[...]

PR #1864, opened 2024-01-15
refactor!: TraceAPI

antazoey:
### What I did

fixes: #1587 
Fixes: APE-1389

Wanna try other providers? These ones are ready:

https://github.com/ApeWorX/ape-foundry/pull/85
https://github.com/ApeWorX/ape-alchemy/pull/63

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #1949, opened 2024-03-06
feat: show trace on failed transactions in Web3Provider

mikeshultz:
### What I did

Shows the call trace on a failed transaction.

Ref: #1945

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -1116,7 +1116,17 @@ class Web3Provider(ProviderAPI, ABC):
                 **params,
             )
         )
-        return self.compiler_manager.enrich_error(result)
+        enriched = self.compiler_manager.enrich_error(result)
+
+        # Show call trace if availble
+        if enriched.txn:
+            # Apparently this is not always a receipt on a failed tx
+            if isinstance(enriched.txn, TransactionAPI) and enriched.txn.receipt:
+                enriched.txn.receipt.show_trace()
+            elif isinstance(enriched.txn, ReceiptAPI):
+                enriched.txn.show_trace()
+
+        return enriched
 
 
 class EthereumNodeProvider(Web3Provider, ABC):
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -1120,7 +1120,7 @@ class Web3Provider(ProviderAPI, ABC):
 
         # Show call trace if availble
         if enriched.txn:
-            # Apparently this is not always a receipt on a failed tx
+            # Unlikely scenario where a transaction is on the error even though a receipt exists.
             if isinstance(enriched.txn, TransactionAPI) and enriched.txn.receipt:
                 enriched.txn.receipt.show_trace()
             elif isinstance(enriched.txn, ReceiptAPI):


src/ape_ethereum/provider.py
```diff
@@ -1116,7 +1116,17 @@ def _handle_execution_reverted(
                 **params,
             )
         )
-        return self.compiler_manager.enrich_error(result)
+        enriched = self.compiler_manager.enrich_error(result)
+
+        # Show call trace if availble
+        if enriched.txn:
+            # Apparently this is not always a receipt on a failed tx
```
antazoey, line 1123: correct, failed calls for example do not have an associated txn.

src/ape_ethereum/provider.py
```diff
@@ -1116,7 +1116,17 @@ def _handle_execution_reverted(
...
+            if isinstance(enriched.txn, TransactionAPI) and enriched.txn.receipt:
+                enriched.txn.receipt.show_trace()
```
antazoey, line 1125: when using ape-vyper, it already shows the source-traceback (the vyper lines last ran before the revert). How does this look when used with that feature? (im about to test and see for myself)

antazoey: answered my own question, I think it looks pretty good when together:

```
SUCCESS: Contract 'VyperContract' deployed to: 0x5FbDB2315678afecb367f032d93F642f64180aa3
Call trace for '0xa438858260fca5d95d2e2c0a6e37f9dea771c9738b102d8ebad696fa0c997490'
tx.origin=0x70997970C51812dc3A010C7d01b50e0d17dc79C8
VyperContract.setNumber(num=123) [23728 gas]

  File "/Users/jules/ApeProjects/ape-playground/scripts/rev.py", line 8, in main
    contract.setNumber(123, sender=notme)
  File "/Users/jules/ApeProjects/ape-playground/contracts/VyperContract.vy", line 84, in 
setNumber
    assert msg.sender == self.owner, "!authorized"
    ^^^^^^^^^^^^^^^^^^^^^^^

ERROR: (ContractLogicError) !authorized
```

You can see first your call trace gets printed, than the normal source traceback, and then the error (containing the revert message)

antazoey (approved):

antazoey (approved):

src/ape_ethereum/provider.py
```diff
@@ -1150,7 +1150,17 @@ def _handle_execution_reverted(
                 **params,
             )
         )
-        return self.compiler_manager.enrich_error(result)
+        enriched = self.compiler_manager.enrich_error(result)
+
+        # Show call trace if availble
+        if enriched.txn:
+            # Unlikely scenario where a transaction is on the error even though a receipt exists.
+            if isinstance(enriched.txn, TransactionAPI) and enriched.txn.receipt:
```
antazoey, line 1158: we may also want to handle the scenario where a txn exists but a receipt does not, such as during gas estimation

mikeshultz: You said this in the ape-hardhat PR as well and I guess I'm just not picking up what you're putting down.  What are you suggesting I do in that case?

antazoey: Well, I was going to write down how to do it, but as I began to write down, I realize it is just easier to wait for 0.8 and [the refactor that is happening there](https://github.com/ApeWorX/ape/pull/1864).

So it is not a blocker - but this may not alleviate some of users problems if they are estimating gas and it seems like they are.

But here is some info in case it helps come up with something:

1. The `.trace` may be available already on the txn. In fact, this should be the case if the provider supports `debug_traceCall` as of [this PR](https://github.com/ApeWorX/ape/pull/1951).
2. If the trace is not available on the error for some annoying reason, the same strategy can be applied in [the PR I just mentioned](https://github.com/ApeWorX/ape/pull/1951) to create it (using `debug_traceCall`).
3. Once the trace is procured, a call tree can be created from it using evm-trace lib and stuff, a think a provider does this somewhere as a last resort. But - this is the part that is especially going to be easier with the refactor and which is why I realized it may be ok holding off.
4. Once you have a call tree, you can do the same thing the Receipt is doing to `show` it, if you will.

And again, I realize this process of a bit whonky and the refactor is going to help; I get also that excuse is probably getting hold which is why I have been working on 0.8 a bunch (getting there!).

mikeshultz: I'm going to just grab the low hanging fruit on this one since you're changing all this stuff anyway.  Let me know if you want me to follow up after/for 0.8.

antazoey (approved):

Issue #1113, opened 2022-10-29
Load a Project from a Manifest


fubuloubu:
### Overview

I have a package manifest, I would love to be able to load it via `Project.from_manifest(manifest: ethpm.PackageManifest) -> Project`

### Specification

In the `ape` namespace, there is a class re-import of `ProjectManager` called `Project`, so we would add a method `from_manifest` that would be able to load from a manifest type.

---

Upon further investigation, it turns out that the internal structure of the project manager isn't set up well to replicate this type of use case. Here is one suggestion for a set of changes to make in order to better support these kinds of use cases:

1. Use `ProjectType` sub-types (`ApeProject`, `BrownieProject`, etc.) to parse out a manifest from dependencies (assuming they are not going to change)
2. Change `ProjectManager` to be a superclass of `ApeProject`, and add methods for managing an active local project's cached manifest
3. Replicate structure of a project with dependencies by having each dependency be a `ProjectType` with a finalized manifest that the local `ProjectManager` can work with

NOTE: Only `ProjectManager` superclass can mutate it's encapsulated project. Dependencies would recursively build a final manifest for themselves (e.g. dependencies of dependencies would build a finalized manifest including linking extra manifests) so that the local project was working against a final set of dependency files

NOTE: A `ProjectManager` may not actually have any local `contracts/` files to work with, which is okay. ethPM supports this use case of dependency-only manifests (which we can store in local cache, but never deploy)

### Dependencies
n/a

unparalleled-js:
We can actually achieve pretty straightforwardly by adding a method to `ProjectManager`. I am not sure about the proposed changes.

* The part about `ProjectManager` being a super-type of `ProjectAPI` may not be needed.
* What are you calling `ProjectType` is currently called `ProjectAPI` - can name it whatever though...
* Dependencies already just-about this way, only with a lot more steps.

unparalleled-js:
Plugins, like compiler implementations, are driven by config values. For example, you setup remappings there and even compiler versions.

Somehow, those needs to be captured and used without the need of `config_manager` in order to have several instances of `ProjectAPI` classes without switching the active project context.

fubuloubu:
> What are the types of thing you would do with the `project` you get from `Project.from_manifest()`?

- Load a project from a file on disk (as mentioned, using a data file from package resources)
- Fetch a contract type or dependency from that package

[...]

Issue #660, opened 2022-04-15
Convert `CompilerAPI.compile` method to consume and produce iterators 


fubuloubu:
### Elevator pitch:
Change the methods to consume and produce iterators, so that files can be compiled in batches

### Value:
It would then be possible to compile a set of files and issue warnings if they fail to compile, so that the rest of the project can be compiled without the failure stopping execution. It will also lend itself well later if an async compile process is chosen.

### Task list:

* [ ] Tasks go here

### Design review:

Do not signoff unless:
- 1) agreed the tasks and design approach will achieve acceptance, and
- 2) the work can be completed by one person within the SLA.
Design reviewers should consider simpler approaches to achieve goals.

(Please leave a comment to sign off)

PR #1960, opened 2024-03-11
refactor!: project system refactors

antazoey:
### What I did

fixes: #1113 
fixes: #660 

required this PR: https://github.com/ApeWorX/ape-solidity/pull/134
ape-vyper needs the same but havent done yet.

btw, just about everything is probably broken still, but i got all project system kinda working, so opening so no one think i just ghosted, because this has been a lot of refactoring

```python
from ape import project, Project

# Still refer to the root project.
project.MyNFT

# Refer to projects from other locations.
other_project = Project("../path/to/other/project")
other_project.MyContractThere

# Use manifests as project.
manifest_project = Project.from_manifest("../path/to/manifest.json")
manifest_project.SomeContractThere

# Dependencies are better now.
# Install directly from Python if you care to (CLI + a config is still much better).
project.dependencies.install({"name": "apetrace", "github": "antazoey/ApeTrace", "ref": "main"})

# Refer to dependencies anywhere, even if they are not listed from this project.
# Note: listing is required for ape-solidity remappings to work though.
dependency = project.dependencies.get_dependency("apetrace", "main")

# Access dependency projects the exact same as regular projects:
dependency.project.ContractFromDependency
```

### Checklist


docs/userguides/dependencies.md
```diff
@@ -1,6 +1,6 @@
 # Dependencies
 
-Ape downloads and caches dependencies in the `.ape/packages/<name>/<version-id>` directory where `<name>` refers to the name of the dependency and `<version-id>` refers to the version or branch of the package.
+Ape downloads and caches dependencies in the `.ape/packages/projects/<name>/<version-id>` directory where `<name>` refers to the name of the dependency and `<version-id>` refers to the version or branch of the package.
```
fubuloubu, line 3: What other folders appear under there now?

antazoey: `apis`: for caching dependency APIs, which is needed for using dependencies outside of a project, being able to fetch a dependency outside a project, makes the CLI commands just work better in all regard... 

`manifests`: which are the individual manifest for each dependency project, works the same `__local__.json` for the local one.

fubuloubu: I was thinking that we can use "global names" (the names that each dependency type uses e.g. `github: OpenZeppelin/openzeppelin-contracts` instead of the "local names" (the name we give to the dependency within the project) inside this directory. Those "local names" are really only important when referencing them inside of source remappings for the project using the compiler plugin config

antazoey: I think that makes sense but it might be more clear then to have "name: " in the config be "import_key: " or "remapping_key: " or something, rather than have 2 names

antazoey:
- [ ] figure out how to put back ContractType separation

So right now on this PR, the notion of the separated `ContractType` artifact .jsons from the .build folder has been removed.
Mostly because it very annoying to work with.. the contract types need to be managed in 2 spots.

My idea is to output ABIs to a subfolder in .build, maybe if and only if you configure Ape to do so, and the source of truth remains in the manifest, and the ABIs just get places (and replaced) upon ever compilation via looking at all types in the manifest _after_ compilation finishes. This way, users can still host these files and they are even smaller and they are unrelated to the rest of Ape... but am open to other ideas or even just putting back as-is. And by making it config driven, it becomes opt-in and most centered to the idea of minimization.

src/ape/managers/project.py
```diff
@@ -0,0 +1,1599 @@
...
+            config_override (Optional[Dict]): Optionally provide a config override.
+
+        Returns:
+            :class:`~ape.managers.project.ProjectManifest`
+        """
+        return Project.from_manifest(manifest, config_override=config_override)
+
+    @classmethod
+    @contextmanager
+    def create_sandbox(cls, config_override: Optional[Dict] = None) -> Iterator["LocalProject"]:
```
fubuloubu, line 868: Suggestion: Not sure if `sandbox` is accurately descriptive, something containing `temporary` may be better

antazoey: Maybe `isolate()` or something? I want to also convey the existing project is copied into this temporary environment.

fubuloubu: `isolate` is good!

Was looking for something more commonly understood than `sandbox`

Something that denotes the action a bit better:
```py
with project.isolated_clone() as project:
    ...
```

antazoey: Here is the commonly understood concept of a sandbox: https://en.m.wikipedia.org/wiki/Sandbox_(software_development)

The reason I agree with you is because "sandbox" may imply some sort of testing or staging environment, whereas in our case it is supposed to be an isolated compilation environment, and those are not necessarily the same thing.

I like sandboxes too though because "they wash away" in the metaphor, like a temp folder.

antazoey: Come to think of it, however... the user could use this to test project changes without committing them..
But that isnt the reason it was made.

Hmmmmm

docs/userguides/compile.md
```diff
@@ -149,3 +143,17 @@ owner = accounts.test_accounts[0]
...
+
+## Output Extra
+
+Sometimes, there are extra output styles you may want.
+For example, to output minified ABI JSONs, use the following config:
+
+```yaml
+compile:
+  output_extra:
+     - ABI
```
fubuloubu, line 155: how does this work?

antazoey: I didn't want to put back the split contract types in the .build folder. The ones who wanted that just wanted smaller files for hosting. So they want minified ABIs. Every compile, it dumps the manifest's ABIs to a directory when you enable this; for hosting

docs/userguides/dependencies.md
antazoey: Done! Storage keys of dependencies are way better now.

src/ape/api/accounts.py
```diff
@@ -401,7 +402,7 @@ class AccountContainerAPI(BaseInterfaceModel):
     instances.
     """
 
-    data_folder: Path
+    name: str
```
antazoey, line 405: pulled out this change: https://github.com/ApeWorX/ape/pull/2025

src/ape/_cli.py
```diff
@@ -86,7 +87,8 @@ def invoke(self, ctx) -> Any:
         except click.UsageError as err:
             self._suggest_cmd(err)
         except ApeException as err:
-            if handle_ape_exception(err, [ctx.obj.project_manager.path]):
+            path = ctx.obj.project_manager.path
+            if isinstance(path, Path) and handle_ape_exception(err, (path,)):
```
antazoey, line 91: hoisted: https://github.com/ApeWorX/ape/pull/2028

src/ape/api/compiler.py
```diff
@@ -204,32 +220,41 @@ def enrich_error(self, err: ContractLogicError) -> ContractLogicError:
 
     @raises_not_implemented
     def trace_source(  # type: ignore[empty-body]
-        self, contract_type: ContractType, trace: TraceAPI, calldata: HexBytes
+        self, contract_source: ContractSource, trace: TraceAPI, calldata: HexBytes
```
antazoey, line 223: hoisted this change, will rebase: https://github.com/ApeWorX/ape/pull/2029

docs/userguides/compile.md
fubuloubu: so there's a list of accepted keys documented?

fubuloubu:
> ```python
> ...
> # Dependencies are better now.
> # Install directly from Python if you care to (CLI + a config is still much better).
> project.dependencies.install({"name": "apetrace", "github": "antazoey/ApeTrace", "ref": "main"})
> ```

I think this might feel a little nicer using kwargs e.g.:

```python
project.dependencies.install(name="apetrace", github="antazoey/ApeTrace", ref="main")
```

What do you think?

src/ape/utils/os.py
```diff
@@ -221,3 +222,36 @@ def clean_path(path: Path) -> str:
         return f"$HOME{os.path.sep}{path.relative_to(home)}"
 
     return f"{path}"
+
+
+def path_match(path: Union[str, Path], *exclusions: str) -> bool:
```
antazoey, line 227: https://github.com/ApeWorX/ape/pull/2081

src/ape/api/projects.py
```diff
@@ -1,667 +1,134 @@
...
-)
+from ape.api.config import ApeConfig
+from ape.utils import BaseInterfaceModel, abstractmethod
 
-if TYPE_CHECKING:
-    from ape.contracts import ContractContainer
 
-
-class ProjectAPI(BaseInterfaceModel):
+class DependencyAPI(BaseInterfaceModel):
```
fubuloubu, line 11: wonder if this having an "is immutable?" boolean might be helpful to basically describe whether it should be "trusted" as a dependency not to change, or always check to see if there is some integrity checksum that has been modified to drive re-acquiring the dependency (and recompiling it with the context of the old version to gain a performance boost from the old project's cache)

For example, github releases essentially shouldn't change (caveat here that they actually can, but typically don't), ipfs links will not change, http links definitely can change, so can git branches etc.

fubuloubu: Or maybe a property "needs resync?" that returns whether the integrity sum is still valid, or (in the case of IPFS) just is always true

For git repos, can store the commit hash of that commit for the branch/tag

src/ape/utils/basemodel.py
```diff
@@ -121,16 +121,14 @@ class ManagerAccessMixin:
...
 
-    dependency_manager: ClassVar["DependencyManager"] = cast(
-        "DependencyManager", injected_before_use()
-    )
-
     network_manager: ClassVar["NetworkManager"] = cast("NetworkManager", injected_before_use())
 
     plugin_manager: ClassVar["PluginManager"] = cast("PluginManager", injected_before_use())
 
     project_manager: ClassVar["ProjectManager"] = cast("ProjectManager", injected_before_use())
```
fubuloubu, line 128: as mentioned previously, maybe makes more sense to rename to `.local_project` or just `.project`

docs/userguides/compile.md
```diff
@@ -158,3 +143,19 @@ owner = accounts.test_accounts[0]
...
+```
+
+Then, after compiling, you should notice minified ABI json files in your `.build/abi` folder.
+This is useful if hosting these files on a web-server.
+
+**Note**: Currently, ABI is the only supported option for `output_extra`.
```
fubuloubu, line 161: Can you link to the generated autodoc for this enum instead?

docs/userguides/dependencies.md
```diff
@@ -1,6 +1,6 @@
 # Dependencies
 
-Ape downloads and caches dependencies in the `.ape/packages/<name>/<version-id>` directory where `<name>` refers to the name of the dependency and `<version-id>` refers to the version or branch of the package.
+Ape downloads and caches dependencies in the `.ape/packages/projects/<name>/<version-id>` directory where `<name>` refers to the full-name of the dependency (e.g. `"OpenZeppelin/openzeppelin-contracts"`) and `<version-id>` refers to the version or branch of the package.
```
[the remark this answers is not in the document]
antazoey: Because there are now 3 elements to maintain for dependencies (3 sub-folders):

* `projects/` - the raw project files, needed for basic local project compiling where it pulls in these sources to a temporary `.cache` folder when compiling.
* `manifests/` - contain the on-goingly built manifests, similar to `__local__.json` works, except it is less of a big deal of these ever can compile. However, when you do compile dependencies, their contract types go here (just like in `.build/__local__.json`.
* `apis/` - these are the data fields you populate in your `dependencies:` key or feed into `ape pm install` commands. This is needed because it allows you to manage dependencies anywhere, like if navigate away from the folder containing their source `ape-config.yaml`. As I have been using this feature, i find it also to be a good place to retain functioning config overrides.

- [ ] document all this more

src/ape/managers/__init__.py
```diff
@@ -9,38 +9,18 @@
...
-_data_folder = _Path.home().joinpath(".ape")
-_project_folder = _Path.cwd()
-
 ManagerAccessMixin.plugin_manager = PluginManager()
-
-ManagerAccessMixin.dependency_manager = DependencyManager(data_folder=_data_folder)
-
 ManagerAccessMixin.config_manager = ConfigManager(
-    # Store all globally-cached files
-    DATA_FOLDER=_data_folder,
```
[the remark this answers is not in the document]
antazoey: It now defaults to `Path.home() / ".ape"` but can be changed; the tests still use a temp dir.

src/ape/api/config.py
```diff
@@ -93,3 +122,330 @@ class GenericConfig(ConfigDict):
...
+        if uri_str.startswith("https://"):
+            # Assume GitHub dependency
+            version = uri_str.split("/")[-1]
+            dependency["github"] = uri_str.replace(f"/releases/tag/{version}", "")
+            dependency["github"] = dependency["github"].replace("https://github.com/", "")
+
+            # NOTE: If version fails, the dependency system will automatically try `ref`.
+            dependency["version"] = version
+
+        elif uri_str.startswith("file://"):
```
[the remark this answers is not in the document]
antazoey: probably a:

```python
        else:
            logger.error(f"Manifest URI {uri_str} not a supported dependency.")
            continue
```

so that ape can still attempt to do its best with weird manifests it may come across

src/ape/api/projects.py
antazoey: > For example, github releases essentially shouldn't change 

They can if you are using `ref: main` or something.

Re: resyncing, I think we can make an API out of this somehow but I would consider it extra, I think we could do it though. I think it is extra because when in doubt, `ape pm install --force` again, as a workaround. Also it is seems a bit complicated but maybe we can make a ticket for this.

src/ape/managers/__init__.py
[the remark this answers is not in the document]
fubuloubu: Where is this specifically though? Couldn't find where it moved to

src/ape/api/projects.py
fubuloubu: > > For example, github releases essentially shouldn't change
> 
> They can if you are using `ref: main` or something.

Yeah that was sort of my point. Anyways, my second suggestion is actually better if you can just use the commit hash as an integrity hash (e.g. you can quickly see if there are any changes to try and recompile with by checking to see if the upstream ref mismatches the stored commit hash you have for that ref)

> Re: resyncing, I think we can make an API out of this somehow but I would consider it extra, I think we could do it though. I think it is extra because when in doubt, `ape pm install --force` again, as a workaround. Also it is seems a bit complicated but maybe we can make a ticket for this.

I would think this would live under `DependencyAPI`, but unsure of where this resync logic lives now with the refactor

antazoey: > I would think this would live under DependencyAPI, but unsure of where this resync logic lives now with the refactor

There is no re-sync feature as of now, as far as I know

antazoey:
> I think this might feel a little nicer using kwargs e.g.:

I think i got this to work
8d4501b

src/ape/managers/__init__.py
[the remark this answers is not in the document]
antazoey: In `ape/managers/config.py:ConfigManager.__init__.py`:

```python
def __init__(self, data_folder: Optional[Path] = None, request_header: Optional[dict] = None):
        if not data_folder and "APE_DATA_FOLDER" in os.environ:
            self.DATA_FOLDER = Path(os.environ["APE_DATA_FOLDER"])
        else:
            self.DATA_FOLDER = data_folder or Path.home() / ".ape"

        self.REQUEST_HEADER = request_header or {}
```

I made it also support this env var so that tests that use subprocess-runners can use the temporary data folder as well.

src/ape/api/projects.py
antazoey: But yes, we would add it to DependencyAPI and then DependencyManager would also have a method probably

src/ape_pm/dependency.py
```diff
@@ -0,0 +1,321 @@
...
+            destination.parent.mkdir(parents=True, exist_ok=True)
+            destination.write_text(self.local.read_text())
+
+
+class GithubDependency(DependencyAPI):
+    """
+    A dependency from Github. Use the ``github`` key in your ``dependencies:``
+    section of your ``ape-config.yaml`` file to declare a dependency from GitHub.
+
+    Config example::
```
fubuloubu, line 84: does this do anything special? how does this example render?

src/ape_pm/projects.py
fubuloubu, line 1: Random thought (and not something I'm looking to do now) but seems like a lot of the project detection logic could actually live in ethpm-types, since it's mostly about figuring out how to read compiler and dependency settings to put into a standard structure.

Also if we could eventually negotiate this with crytic-compile (which fulfills a similar function but has much broader support of different project types) we would be able to share that code and more easily interop with slither

fubuloubu: Note that I do still think the different dependency options make sense here, as it's typically just a nice contextual wrapper around fetching a project to compile into a manifest (and it would make more sense just to standardize around manifests as the target of keys in `dependencies`)

src/ape/api/config.py
```diff
@@ -93,3 +122,330 @@ class GenericConfig(ConfigDict):
...
+
+    @property
+    def _plugin_configs(self) -> dict:
+        # NOTE: Ensure a dict exists so we can add to it
+        #  and have it persist.
+        self.__pydantic_extra__ = self.__pydantic_extra__ or {}
+        return self.__pydantic_extra__
+
+    def __repr__(self):
+        return "<ape-config.yaml>"
```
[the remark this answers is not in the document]
fubuloubu: nvm that isn't available

fubuloubu (approved):

src/ape/api/projects.py
fubuloubu: > But yes, we would add it to DependencyAPI and then DependencyManager would also have a method probably. And when requesting a dependency that the manager detects needs a refresh, we could have that happen.
> 
> Still though - this is a bit extra! Does pip even do that with `github` based installs?

Usually if you install a git dependency, it just always fetches and recompiles it. Unsure if any further caching is going on, but that could also just be the way we handle it (although I think it might be more annoying for us to always go and fetch dependency sources)

fubuloubu: There is definitely a difference in how pip installs packages, and how we get the files to compile dependencies

Should the dependency sources just live on disk until you manually refresh? that could also be fine, worth some thought

src/ape_pm/dependency.py
antazoey: Two colons mean custom header, you can put whatever you want for the title. We do this for our `Usage example::` everywhere as well. Here is the render (on a laptop web browser):

Screenshot 2024-05-22 at 08 40 10

src/ape_pm/projects.py
antazoey: But yes - I think I see what you are saying and we could do it.

1. Have method like `PackageManifest.initialize_from_project()` that sets up all settings
2. Get rid of ProjectAPI
3. Limit Ape to working with manifest for all project-types (which it already can do, so this moreorless is deleting code)

src/ape/api/projects.py
antazoey: > Should the dependency sources just live on disk until you manually refresh? that could also be fine, worth some thought

This is how it works now, and it is pretty simple i think so i like that.

docs/index.md
```diff
@@ -59,3 +59,13 @@
    methoddocs/types.md
    methoddocs/utils.md
 ```
+
+```{eval-rst}
+.. toctree::
+   :caption: Plugin Python Reference
+   :maxdepth: 1
+
+   methoddocs/ape_accounts.md
```
antazoey, line 68: @dtdang See how I am taking `ape-accounts` out of above and making a new section here for `Plugin Python Reference`? I want us to try and do this for core right now with all the core plugins. I had to make some changes in the core plugings to get it to work though,  had to add `__all__` = in the init, also some other docs files for the auto-modding

fubuloubu (approved):
image

PR #1963, opened 2024-03-14
fix: allow kwargs to flatten_contract()

mikeshultz:
### What I did

Adds missing (?) kwargs to `CompilerManager.flatten_contract()` to match `CompilerAPI.flatten_contract()`.

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -337,7 +337,7 @@ class CompilerManager(BaseManager):
         compiler = self.registered_compilers[ext]
         return compiler.enrich_error(err)
 
-    def flatten_contract(self, path: Path) -> Content:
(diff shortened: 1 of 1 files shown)


src/ape/managers/compilers.py
```diff
@@ -355,7 +355,7 @@ def flatten_contract(self, path: Path) -> Content:
             )
 
         compiler = self.registered_compilers[path.suffix]
-        return compiler.flatten_contract(path)
+        return compiler.flatten_contract(path, **kwargs)
```
mikeshultz, line 358: I'd also be happy to add `base_path: Optional[Path]` to `CompilerAPI` if that makes more sense.  But this way doesn't require changes to `ape_solidity`.

antazoey (approved):

Issue #1962, opened 2024-03-12
Allow signing raw 32 byte message hash directly


fubuloubu:
### Overview

Sometimes, it is required to be able to sign a raw hash. While this is an extremely dangerous, some very important legacy operations (such as setting a Delegate on the Safe API) require it, so it would be helpful to allow doing that

### Specification

If `msg` is of type `bytes` (works for `HexBytes` too) AND is *exactly* 32 bytes in length, use the legacy `eth_account.Account.signHash` method (or whatever supported method exists for each account plugin, if any) to sign that has directly.

Further, it should *always* raise a warning about exactly how dangerous it is to sign using this method prior to confirming the signing operation (except for `ape_test.TestAccount`, as that is irrelvant) to ensure that users are made aware never to do this unless absolutely necessary and only if they know exactly what they're signing, since signing the wrong hash could lead to a transfer of all their money away from them.

This method should be implemented into `ape_accounts.KeyfileAccount` at least, and if Ledger/Trezor support it then it should be added as well (might not be able to be supported). Pretty sure there might be no way of doing this for `ape_frame`. If there is no way to support this operation, it is preferrable to raise under this specific condition (32 byte hash) rather than skip signing, since it might not be clear enough that it is an unsupported operation for that account plugin.

### Dependencies

no dependencies

fubuloubu:
suggested UX:

```py
In [1]: acct = accounts.load("acct")

In [2]: acct.sign_message(b"\x00" * 32)
Sign using EIP-191? (Select 'y' if unsure) [y/N]: y
Raw bytes to sign: '0000000000000000000000000000000000000000000000000000000000000000'

Sign:  [y/N]: y
Enter passphrase to unlock 'acct' []: 
Leave 'acct' unlocked? [y/N]: y
Out[2]: <MessageSignature v=27 r=0x59de...95cd s=0x1150...b36b>

In [3]: acct.sign_message(b"\x00" * 32)
Sign using EIP-191? (Select 'y' if unsure) [y/N]: n
WARNING: Signing a raw hash directly is a dangerous action which could risk substantial losses! Only confirm if you are 100% sure of the origin!
Please confirm you wish to sign using `EthAccount.signHash` [y/N]: y
Out[3]: <MessageSignature v=28 r=0xaefc...90b3 s=0x5cd8...f683>
```

PR #1966, opened 2024-03-14
feat(AccountAPI): sign raw hash32

fubuloubu:
### What I did

Add a new path to `AccountAPI` called `.sign_raw_msghash` that allows signing a raw 32 byte hash without treating it as an EIP191 `SignableMessage` type

fixes: #1962

### How I did it

I needed this for work I was doing on ape safe, but it is a really dangerous thing to allow, so I added a warning message and additionally disallowed autosigning for it. For `AccountAPI.recover_message` I added an additional kwarg `recover_using_eip191=True` that can be set false if you want to opt-in for that behavior for verifying a signature.

### How to verify it

Try the experience locally, I have added tests for it. Ultimately, when installled alongside the delegate feature for ape safe (https://github.com/ApeWorX/ape-safe/pull/41) you should now be able to use it to directly sign something with the API

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -58,6 +58,18 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         return None
 
     def sign_raw_msghash(self, msghash: HexBytes) -> Optional[MessageSignature]:
+        """
+        Sign a raw message hash.
+
+        Args:
+          msghash (:class:`~eth_pydantic_types.HexBytes`):
+            The message hash to sign. Plugins may or may not support this operation.
+            Default implementation is to raise ``NotImplementedError``.
+
+        Returns:
+          :class:`~ape.types.signatures.MessageSignature` (optional):
+            The signature corresponding to the message.
+        """
         raise NotImplementedError(
             f"Raw message signing is not supported by '{self.__class__.__name__}'"
         )
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -57,6 +57,11 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         """
         return None
 
+    def sign_raw_msghash(self, msghash: HexBytes) -> Optional[MessageSignature]:
+        raise NotImplementedError(
+            f"Raw message signing is not supported by '{self.__class__.__name__}'"
+        )
+
     @abstractmethod
     def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]:
         """
@@ -295,8 +300,9 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
 
     def check_signature(
         self,
-        data: Union[SignableMessage, TransactionAPI, str, EIP712Message, int],
+        data: Union[SignableMessage, TransactionAPI, str, EIP712Message, int, bytes],
         signature: Optional[MessageSignature] = None,  # TransactionAPI doesn't need it
+        recover_using_eip191: bool = True,
     ) -> bool:
         """
         Verify a message or transaction was signed by this account.
@@ -307,6 +313,10 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
             signature (Optional[:class:`~ape.types.signatures.MessageSignature`]):
               The signature to check. Defaults to ``None`` and is not needed when the first
               argument is a transaction class.
+            recover_using_eip191 (bool):
+              Perform recovery using EIP-191 signed message check. If set False, then will attempt
+              recovery as raw hash. `data`` must be a 32 byte hash if this is set False.
+              Defaults to ``True``.
 
         Returns:
             bool: ``True`` if the data was signed by this account. ``False`` otherwise.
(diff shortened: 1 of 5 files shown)


src/ape_accounts/accounts.py
```diff
@@ -162,15 +162,31 @@ def delete(self):
...
+            display_msg = f"Signing raw integer: {msg}"
             msg = encode_defunct(hexstr=HexBytes(msg).hex())
+
         elif isinstance(msg, bytes):
-            user_approves = self.__autosign or click.confirm(f"Message: {msg.hex()}\n\nSign: ")
-            msg = encode_defunct(primitive=msg)
+            display_msg = f"Signing raw bytes: '{msg.hex()}'"
+
+            if sign_raw_hash := (
+                len(msg) == 32 and not click.confirm("Sign using EIP-191? (Select 'y' if unsure)")
```
antazoey, line 179: Using `click`-isms outside of a CLI is a bit strange, I know we are already doing it elsewhere in this calss, but it'd be nice to keep prompting in the CLI components.

src/ape_accounts/accounts.py
```diff
@@ -193,20 +209,29 @@ def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]
...
             logger.warning("Unsupported message type, (type=%r, msg=%r)", type(msg), msg)
             return None
 
-        if not user_approves:
+        if sign_raw_hash:
+            # NOTE: Signing a raw hash is so dangerous, we don't want to allow autosigning it
+            if not click.confirm("Please confirm you wish to sign using `EthAccount.signHash`"):
+                return None
+
+            signed_msg = EthAccount.signHash(msg, self.__key)
```
antazoey, line 227: **strong recommendation.** this should not happen within this method. There is a reason `EthAccount` kept them separate. the extra prompting, size checking, **leading zeroes ambiguity**

src/ape_accounts/accounts.py
```diff
@@ -193,20 +209,29 @@ def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]
...
+            display_msg = str(msg)
+
         else:
             logger.warning("Unsupported message type, (type=%r, msg=%r)", type(msg), msg)
             return None
 
-        if not user_approves:
+        if sign_raw_hash:
+            # NOTE: Signing a raw hash is so dangerous, we don't want to allow autosigning it
+            if not click.confirm("Please confirm you wish to sign using `EthAccount.signHash`"):
```
antazoey, line 224: this is a bit of a breaking change... imagine you have some script that is signing messages. now suddenly you are being prompted whenever a certain size is noticed.

not to mention issues with leading zeroes...

fubuloubu:
> **Another reason a separate method might be better:** Accounts that don't support this will be more obvious. Otherwise by the same rationale, we'd have to warn or raise exceptions in accounts trying to sign a bytes value of a certain length, which assumes stuff about the signer's intent.
> 
> By having separate methods, the intent is **clear**, and when the account does not support such thing, it can raise a `NotImplementedError` which is more intuitive than it silently working or warning "hey in case you are trying to size a hash directly, i can't do that, but if not, carry on). It'd be different if signing bytes always worked this way, but still not as good as separate methods imo.
> 
> Just something I was thinking about ...
> 
> 

I'm on board with this change, also makes it easier to roll out

src/ape_accounts/accounts.py
fubuloubu: This class needs to prompt the user directly to unlock or approve actions, so seems okay to me

src/ape_accounts/accounts.py
fubuloubu: Refactored in [fc4b250](https://github.com/ApeWorX/ape/pull/1966/commits/fc4b25035c4672f2e6461a21a7d09fd032b66cbc)

src/ape_accounts/accounts.py
fubuloubu: Refactored in [fc4b250](https://github.com/ApeWorX/ape/pull/1966/commits/fc4b25035c4672f2e6461a21a7d09fd032b66cbc)

src/ape/api/accounts.py
```diff
@@ -57,6 +57,23 @@ def alias(self) -> Optional[str]:
...
+        Args:
+          msghash (:class:`~eth_pydantic_types.HexBytes`):
+            The message hash to sign. Plugins may or may not support this operation.
+            Default implementation is to raise ``NotImplementedError``.
+
+        Returns:
+          :class:`~ape.types.signatures.MessageSignature` (optional):
+            The signature corresponding to the message.
+        """
+        raise NotImplementedError(
```
antazoey, line 73: suggestion: The `@raises_not_implemented` decorator does exactly this! same messaging.

src/ape/api/accounts.py
```diff
@@ -307,6 +325,10 @@ def check_signature(
             signature (Optional[:class:`~ape.types.signatures.MessageSignature`]):
               The signature to check. Defaults to ``None`` and is not needed when the first
               argument is a transaction class.
+            recover_using_eip191 (bool):
+              Perform recovery using EIP-191 signed message check. If set False, then will attempt
```
antazoey, line 329:
```suggestion
              Perform recovery using EIP-191 signed message check. If False, then will attempt
```

src/ape/api/accounts.py
```diff
@@ -315,6 +337,8 @@ def check_signature(
             data = encode_defunct(text=data)
         elif isinstance(data, int):
             data = encode_defunct(hexstr=HexBytes(data).hex())
+        elif isinstance(data, bytes) and (len(data) != 32 or recover_using_eip191):
```
antazoey, line 340: small perf, maybe, for short-circuit eval on the `or` stmt

```suggestion
        elif isinstance(data, bytes) and (recover_using_eip191 or len(data) != 32):
```

as checking a bool is prolly faster than a length check

antazoey (approved):
Looks good!
Some small feedback if you'll indulge, but LGTM either way

Issue #1968, opened 2024-03-22
Expose automining via `TestProviderAPI`


derekpierre:
### Overview

It would be great to utilize  [EthereumTester.disable_auto_mine_transactions](https://github.com/ethereum/eth-tester?tab=readme-ov-file#ethereumtesterdisable_auto_mine_transactions) through ape without needing to do something a tad ugly like

```python
w3 = networks.provider.web3
w3.provider.ethereum_tester.disable_auto_mine_transactions()
```

Ideally, disabling auto mining would be enabled/disabled via ape's `TestProviderAPI`.

fubuloubu:
Actually think re-opening this makes sense to eventually add it as a feature to `TestProviderAPI` like originally suggested

antazoey:
You mean `ProviderAPI`.
None of the dev-related methods are in `TestProviderAPI`, as a couple years we had a lengthy discussion saying resulting in us not doing that.

Issue #1969, opened 2024-03-22
Explicitly calling `chain.mine(1)` does not work after disabling auto mining.


derekpierre:
### Environment information

- OS: macOS
- Python Version: 3.12
- `ape`: 0.6.27 

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):
https://github.com/nucypher/ATxM/blob/main/tests/ape-config.yaml

### What went wrong?

Related to #1968 .

After disabling auto-mining (see #1968), it seems that the subsequent use of `chain.mine(1)` (using the `chain` fixture) does not work, and instead I have to specifically call `w3.provider.ethereum_tester.mine_block()`  for the mine to actually occur.

Ideally using the `chain` fixture should be sufficient to force the mining of blocks.

antazoey:
are you using the eth-tester provider that comes with Ape?
I haven't been able to reproduce this one, even after implementing the auto mine functionality.

derekpierre:
> are you using the eth-tester provider that comes with Ape?

Pretty sure we are. We don't have an explicit dependency on `eth-tester`.

derekpierre:
That being said we use an older version of eth-ape - `eth-ape==0.6.27`. Is that issue possibly fixed in a newer version.

We haven't updated to a newer version of `eth-ape` because of #1933

antazoey:
> We haven't updated to a newer version of `eth-ape` because of #1933

workaround for #1933 https://github.com/ApeWorX/ape/issues/1933#issuecomment-2025623526
btw, this isn't a bug on 0.8 (which isn't released yet but is in the works)

antazoey:
still unable to reproduce this.

antazoey:
please re-open , i dont think this is a problem!

PR #1972, opened 2024-03-27
test: auto mine functionality to eth-tester

antazoey:
### What I did

fixes: #1968

### How I did it

does the same as https://github.com/ApeWorX/ape-foundry/pull/59/files but for ape-test

thought: maybe need to squeeze this into the api class to keep things uniform, but i implemented the same as ape-foundry's just in case.

### How to verify it

connect to ape-test provider which uses eth-tester, and then you can do this

```python
from ape import chain

chain.provider.automine = True
...
chain.provider.automine = False
```

by default, it is on.

### Checklist

### DIFF
--- a/src/ape_test/provider.py
+++ b/src/ape_test/provider.py
@@ -2,7 +2,7 @@ import re
 from ast import literal_eval
 from functools import cached_property
 from re import Pattern
-from typing import Dict, Optional, cast
+from typing import Any, Dict, Optional, cast
 
 from eth.exceptions import HeaderNotFound
 from eth_pydantic_types import HexBytes
@@ -32,6 +32,7 @@ from ape_ethereum.provider import Web3Provider
 
 class EthTesterProviderConfig(PluginConfig):
     chain_id: int = DEFAULT_TEST_CHAIN_ID
+    auto_mine: bool = True
 
 
 class LocalProvider(TestProviderAPI, Web3Provider):
@@ -50,7 +51,8 @@ class LocalProvider(TestProviderAPI, Web3Provider):
 
         return self._evm_backend
 
-    def connect(self):
+    @cached_property
+    def tester(self):
         chain_id = self.settings.chain_id
         if self._web3 is not None:
             connected_chain_id = self._make_request("eth_chainId")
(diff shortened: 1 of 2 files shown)


tests/functional/test_provider.py
```diff
@@ -397,3 +397,17 @@ def test_create_access_list(eth_tester_provider, vyper_contract_instance, owner)
...
     with pytest.raises(APINotImplementedError):
         eth_tester_provider.create_access_list(tx)
+
+
+def test_auto_mine(eth_tester_provider):
+    eth_tester_provider.auto_mine = False
+    assert not eth_tester_provider.auto_mine
+
+    # Ensure can still manually mine.
+    block = eth_tester_provider.get_block("latest").number
```
antazoey, line 407: attempted to repro https://github.com/ApeWorX/ape/issues/1969 but mining still seems to work, at least at the provider level.

src/ape_test/provider.py
```diff
@@ -66,8 +68,32 @@ def connect(self):
...
+        # NOTE: Need to do this until https://github.com/pydantic/pydantic/pull/2625 is figured out
+        if attr == "auto_mine":
+            if value is True:
+                self.tester.ethereum_tester.enable_auto_mine_transactions()
+            elif value is False:
+                self.tester.ethereum_tester.disable_auto_mine_transactions()
+            else:
+                raise TypeError("Expecting bool-value for auto_mine setter.")
+        else:
+            super().__setattr__(attr, value)
```
fubuloubu, line 87: This should be solved in latest versions of pydantic: https://github.com/pydantic/pydantic/pull/5502

```suggestion
    @auto_mine.setter
    def set_auto_mine(self, value: Any) -> None:
        if value is True:
            self.tester.ethereum_tester.enable_auto_mine_transactions()
        elif value is False:
            self.tester.ethereum_tester.disable_auto_mine_transactions()
        else:
            raise TypeError("Expecting bool-value for auto_mine setter.")
```

antazoey: yes you're right, however i had to make it:

```
    @auto_mine.setter
    def auto_mine(self, value: Any) -> None:
```

else it was failing...
PyCharm gives me the warning:

```
Names of function and decorator don't match; property accessor is not created 
```
so that is how I knew to change the name.

but after doing so, it works!

tests/functional/test_provider.py
fubuloubu: Is this a blocker?

fubuloubu (approved):

tests/functional/test_provider.py
antazoey: not for this feature particularly. if we can get them to upgrade we can see if that helps them else will have to look for more info.

PR #1973, opened 2024-03-28
fix: None check in SharedBlobReceipt

fp-crypto:
### What I did

fixes: added a None check to hex_to_int in SharedBlobReceipt

### How I did it

added a simple None check with a 0 return

### How to verify it

Try hitting RPCs (such as tenderly's) that have not updated their returns post-Dencun.

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -453,4 +453,6 @@ class SharedBlobReceipt(Receipt):
     @field_validator("blob_gas_used", "blob_gas_price", mode="before")
     @classmethod
     def hex_to_int(cls, value):
+        if value == None:
+            return 0
         return value if isinstance(value, int) else int(HexBytes(value).hex(), 16)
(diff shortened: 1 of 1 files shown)


fubuloubu (dismissed):

fubuloubu:
although @banteg noted that this actually might be a legitimate value for it to have per EC spec

src/ape_ethereum/transactions.py
```diff
@@ -453,4 +453,6 @@ class SharedBlobReceipt(Receipt):
     @field_validator("blob_gas_used", "blob_gas_price", mode="before")
     @classmethod
     def hex_to_int(cls, value):
+        if value == None:
```
antazoey, line 456:
```suggestion
        if value is None:
```

antazoey (commented):
have to use `is` for None check

antazoey:
> have to use `is` for None check

so is this the case? we can adjust to allow None values

antazoey:
re-open if still a problem or create an issue?

PR #1974, opened 2024-03-28
feat!: add automine abstract property to TestProviderAPI

antazoey:
fixes: #1968

can't merge because this is a breaking change.

### DIFF
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -252,7 +252,7 @@ ape plugins install hardhat
 ### Mining
 
 Test providers allow you to control mining.
-For example, mine an empty block using the `mine` method:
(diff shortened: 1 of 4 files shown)


fubuloubu (approved):
Can't it just raise NotImplementedError for now until 0.8?

antazoey:
> Can't it just raise NotImplementedError for now until 0.8?

It could I guess. Might as well wait though. Functionality wise, you can use this feature on the providers that support it (ape-test and ape-foundry) so no one should be blocked.

fubuloubu (approved):
Approved for v0.8

Question: does `ape-test` provider not support?

antazoey:
> Question: does `ape-test` provider not support?

It does, it was added here: https://github.com/ApeWorX/ape/pull/1972

fubuloubu (approved):

PR #1975, opened 2024-03-29
fix: multicall allow failure

banteg:
### What I did

`allowFailure` was broken in multicall when a call actually failed

```
AttributeError: 'list' object has no attribute 'returnData'
```

### How I did it

- massively simplified the handler by moving to using multicall 3 only. this allows us to have one handler for the result value.
- fixed handling of partially failed calls. such calls would return `None`.
- allow partial failures by default, which is probably what the user want when reading a lot of data.
- (edit: reverted this one) rename `allowFailure` to `allow_failure` when adding individual calls.

### How to verify it

```python
$ ape console --network blast:mainnet:geth

In [1]: pos_manager = Contract('0x434575EaEa081b735C985FA9bf63CD7b87e227F9')

In [2]: from ape_ethereum.multicall import Call

In [3]: call = Call()
   ...: for i in range(5):
   ...:     call.add(pos_manager.ownerOf, i, allow_failure=True)
   ...:

In [4]: list(call())
Out[4]:
[None,
 '0xd6b64E44aae0938118aD0dAE251b859D85351c22',
 '0x121B5ac4De4a3E6F4171956BC26ceda40Cb61a56',
 '0xd6b64E44aae0938118aD0dAE251b859D85351c22',
 '0x0B78e02dB3626BF29a291549e875EfEE929Ea85e']

In [5]: call = Call()
   ...: for i in range(5):
   ...:     call.add(pos_manager.ownerOf, i, allow_failure=False)
   ...:

In [6]: list(call())
ERROR: (ContractLogicError) call failed
```

### Checklist

### DIFF
--- a/src/ape_ethereum/multicall/handlers.py
+++ b/src/ape_ethereum/multicall/handlers.py
@@ -93,7 +93,7 @@ class BaseMulticall(ManagerAccessMixin):
         self,
         call: ContractMethodHandler,
         *args,
-        allow_failure: bool = True,
+        allowFailure: bool = True,
         value: int = 0,
     ) -> "BaseMulticall":
         """
@@ -107,7 +107,7 @@ class BaseMulticall(ManagerAccessMixin):
             call (:class:`~ape_ethereum.multicall.handlers.ContractMethodHandler`):
               The method to call.
             *args: The arguments to invoke the method with.
-            allow_failure (bool): Whether the call is allowed to fail.
+            allowFailure (bool): Whether the call is allowed to fail.
             value (int): The amount of ether to forward with the call.
 
         Returns:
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/multicall/handlers.py
```diff
@@ -87,17 +87,13 @@ def handler(self) -> ContractTransactionHandler:
...
-
-        else:
-            return self.contract.aggregate
+        return self.contract.aggregate3
 
     def add(
         self,
         call: ContractMethodHandler,
         *args,
-        allowFailure: bool = False,
```
fubuloubu, line 100: I had it this way to match the semantics of the contract (similar to `.returnData`), but either way changing it turns this into a breaking change that will have to wait for 0.8

banteg: reverted this back

fubuloubu (approved):

PR #1980, opened 2024-04-02
fix!: handle missing size with computed lookup

antazoey:
### What I did

fixes: #1935 

Handles missing block size more gracefully. if it is there, we use it, else we have to look it up. also adds uncles field (would be needed if we had to calculate this thing), also handles a weird case where couldnt init a block from web3 data ( https://github.com/ethereum/web3.py/issues/3326 )

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -137,7 +137,7 @@ class BlockAPI(BaseInterfaceModel):
         """
         The size of the block in gas. Most of the time,
         this field is passed to the model at validation time,
-        but occassionally it is missing (like in Infura::newHeads),
+        but occassionally it is missing (like in `eth_subscribe:newHeads`),
         in which case it gets calculated if and only if the user
         requests it (or during serialization of this model to disk).
         """


tests/functional/test_query.py
```diff
@@ -29,9 +29,9 @@ def test_basic_query(chain, eth_tester_provider):
         "num_transactions",
         "number",
         "parent_hash",
-        "size",
```
antazoey, line 32: removed because potentially this is an expensive lookup and based on the comment, basic queries we want to avoid expensive lookups (transactions missing, etc.)

src/ape/api/providers.py
```diff
@@ -60,17 +60,40 @@ class BlockAPI(BaseInterfaceModel):
...
+    """
     parent_hash: Any = Field(
         EMPTY_BYTES32, alias="parentHash"
     )  # NOTE: genesis block has no parent hash
-    size: int
+
+    """
+    The timestamp the block was produced.
+    """
     timestamp: int
```
fubuloubu, line 88: Believe this note makes sense (double check me)

```suggestion
    timestamp: int  #  NOTE: pending block uses current timestamp
```

src/ape/api/providers.py
```diff
@@ -60,17 +60,40 @@ class BlockAPI(BaseInterfaceModel):
...
 
+    _size: Optional[int] = None
```
fubuloubu, line 90: Think this makes sense to do? Might replace some of the logic used to do pop and stuff
```suggestion
    _size: Optional[int] = Field(None, alias="size")
```

src/ape/api/providers.py
```diff
@@ -80,12 +103,54 @@ def convert_parent_hash(cls, data):
...
         return cast(List[TransactionAPI], list(self.query_manager.query(query)))
 
+    @computed_field()  # type: ignore[misc]
+    @cached_property
+    def size(self) -> int:
+        """
+        The size of the block in gas. Most of the time,
+        this field is passed to the model at validation time,
+        but occassionally it is missing (like in Infura::newHeads),
+        in which case it gets calculated.
```
fubuloubu, line 143:
```suggestion
        in which case it gets calculated if and only if the user
        requests it (or during serialization of this model to disk).
```

src/ape/api/providers.py
```diff
@@ -80,12 +103,54 @@ def convert_parent_hash(cls, data):
...
+        """
+
+        if self._size is not None:
+            # The size was provided with the rest of the model
+            # (normal).
+            return self._size
+
+        # NOTE: See `ape-ethereum.ecosystem.Block` for a real implementation.
+        return -1
```
fubuloubu, line 152: Should this.... raise `NotImplementedError` instead? (otherwise the block would get serialized as having `-1` size)

src/ape_ethereum/ecosystem.py
```diff
@@ -317,6 +318,33 @@ class Block(BlockAPI):
...
+    @computed_field()  # type: ignore[misc]
+    @property
+    def size(self) -> int:
+        """
+        The size of the block in gas. Most of the time,
+        this field is passed to the model at validation time,
+        but occasionally it is missing (like in Infura::newHeads),
+        in which case it gets calculated.
+        """
+
```
fubuloubu, line 330: Can skip docstrings on inherited methods w/ docstrings
```suggestion
```

src/ape_ethereum/ecosystem.py
```diff
@@ -317,6 +318,33 @@ class Block(BlockAPI):
...
+        if provider := self.network_manager.active_provider:
+            block = provider.get_block(self.number)
+            size = block._size
+            if size is not None and size > -1:
+                self._size = size
+                return size
+
+        # Don't cache.
+        # TODO: Could try to manually calculate here.
+        return -1
```
fubuloubu, line 346: Same as prior note, raise `NotImplementedError` to prevent serialization/caching?

src/ape/api/providers.py
antazoey: this might make more sense in the doc str so users can see it from the doc-site as well

src/ape/api/providers.py
antazoey: I don't think that would work, as the sub-class has a computed field of the same name.

src/ape/api/providers.py
antazoey: another thing: this is at the API level- so technically speaking, a Provider that implements ProviderAPI could handle this differently, especially on different ecosystems.

src/ape/api/providers.py
antazoey: You actually get another problem anyway:

```
NameError: Fields must not use names with leading underscores; e.g., use 'size' instead of '_size'.
```

As mentioned on the PR for trying to dundarize some fields in events, **you cannot have fields that start with underscores in Pydantic v2**. So this is a no-go.

src/ape/api/providers.py
antazoey: This _is_ at the abstract API layer. Technically, who's to say how downstream providers will handle this. Yes, ape-ethereum does this but it is kind of an assumption otherwise.

src/ape/api/providers.py
```diff
@@ -77,12 +101,54 @@ def convert_parent_hash(cls, data):
...
         query = BlockTransactionQuery(columns=["*"], block_id=self.hash)
         return cast(List[TransactionAPI], list(self.query_manager.query(query)))
 
+    @computed_field()  # type: ignore[misc]
+    @cached_property
+    def size(self) -> int:
+        """
+        The size of the block in gas. Most of the time,
+        this field is passed to the model at validation time,
+        but occassionally it is missing (like in Infura::newHeads),
```
fubuloubu, line 140:
```suggestion
        but occassionally it is missing (like in `eth_subscribe:newHeads`),
```

fubuloubu (approved):

src/ape/api/providers.py
antazoey: i thought it was only Infura you were seeing this? Do actual nodes have this problem too?

antazoey: ah, i thought i remembered you saying alchemy didnt have this problem

fubuloubu (approved):

PR #1988, opened 2024-04-03
fix: supress signHash deprecation warning from eth-account

antazoey:
### What I did

suppresses a deprecation warning when calling `signHash`.

Reasons:

1. We are already extensively warning out user.
2. We are warning our user is a more accurate way (not saying it is deprecated but instead saying it is dangerous)
3. the method is not actually deprecated imo, more info: https://github.com/ethereum/eth-account/issues/85

### Checklist

### DIFF
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -1,4 +1,5 @@
 import json
+import warnings
 from os import environ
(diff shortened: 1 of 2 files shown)


antazoey:
will just wait till fixed in eth-account and upgrade then

PR #1989, opened 2024-04-03
fix: issue preventing docs CI workflow from updating `gh-pages` branch

antazoey:
### What I did

looks a classic case of permissions suddenly being required.
we usually see this in the drafter, so i added it there too just to be safe.

### Checklist


antazoey:
should fix the failure we are seeing when merging to main

fubuloubu (approved):

PR #1991, opened 2024-04-04
feat: detect uniswap v3 pools

banteg:
### What I did

unfinished, the goal is:
- automatically detect uniswap v3 pools and instantiate them with a proper abi

im not sure where to put it, it's not really a proxy type and the ecosystem doesn't provide a hook for loading contracts. up to ape maintainers to decide how to best integrate that.

### How I did it

- vendor uniswap v3 factory and pool abis
- add a helper to check that bytecode contains all method selectors of the pool but none of the selectors of the factory (otherwise factory would be falsely detected as a pool since it contains the pool deployment code)

### How to verify it

`Contract('0x88e6A0c2dDD26FEEb64F039a2c41296FcB3f5640')` should work the same as the first verified pool `Contract('0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801')`

### Checklist


fubuloubu:
Had an idea that an evmole ExplorerAPI plugin might be cool, or at least after the major refactor we want to do with the contract system where it becomes a QueryAPI type instead (so we can merge multiple sources of contract data)

Have some sort of independent confidence evaluation that triggers whether it can do the query or not

PR #1994, opened 2024-04-09
feat: add Python 3.12 support

antazoey:
### What I did

requires released versions with:

https://github.com/ApeWorX/eth-pydantic-types/pull/8
https://github.com/ApeWorX/ethpm-types/pull/119
https://github.com/ApeWorX/eip712/pull/46
https://github.com/ApeWorX/evm-trace/pull/60

### Checklist


fubuloubu (approved):

fubuloubu:
Blocked by: https://github.com/ApeWorX/ape/issues/1995

antazoey:
> Blocked by: #1995

Unblocked now

antazoey:
looks like `pyevm` uses pkg_resources.

antazoey:
so ya... Pyevm uses pkg_resources up until the 0.9 range. I believe we would have update quite a bit of eth dependencies for this, which some people are argue is breaking and thus we may want to hold off until 0.8. Open to suggestions.

fubuloubu:
> so ya... Pyevm uses pkg_resources up until the 0.9 range. I believe we would have update quite a bit of eth dependencies for this, which some people are argue is breaking and thus we may want to hold off until 0.8. Open to suggestions.

Would it change how we use eth dependencies to such an extent it would break our own?

V0.8 is not so far away where I'm okay punting to that just to be extra cautious. At least we got our lower level dependencies upgraded already

antazoey:
> Would it change how we use eth dependencies to such an extent it would break our own?

I think not really, but the hexbytes change is rather major. I remember users complaining in the past when I bumped web3.py too early.

antazoey:
I created an issue because I noticed py-evm does not really claim to be fully Python 3.12 compatible: https://github.com/ethereum/py-evm/issues/2169

antazoey:
waiting on an eth-tester / web3.py release and this will be good

fubuloubu (approved):

fubuloubu (dismissed):

fubuloubu (approved):

Issue #1995, opened 2024-04-09
Replace `pkg_resources` with `importlib.resources`/`importlib.metadata`


fubuloubu:
`pkg_resources` apparently doesn't ship with Python 3.12 anymore, we use it in one place:

https://github.com/ApeWorX/ape/blob/f5b6c7a91defeee74cc329479a8f76db24e35eab/src/ape/plugins/_utils.py#L131

It should be possible to replace this with some functionality from `importlib` (or maybe `packaging`)

Either way, looks like some better methods we can use vs. `pkg_resources` (which is slow)

antazoey:
I wonder if we need to make an entry point specifically for ape plugins.

fubuloubu:
> I wonder if we need to make an entry point specifically for ape plugins.

I think entrypoints are just for cli extensions (which we do use)

Plugins themselves need dynamic imports, or something to trigger registration from `<plugin-package>/__init__.py`

PR #1996, opened 2024-04-10
fix: replace usage of `pkg_resources`

antazoey:
### What I did

fixes: #1995

### How I did it

use importlib distributions

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,6 @@ setup(
     install_requires=[
         "click>=8.1.6,<9",
         "ijson>=3.1.4,<4",
-        "importlib-metadata",
         "ipython>=8.5.0,<9",
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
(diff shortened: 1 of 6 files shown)


antazoey:
found a better way! in coming, actually way better

setup.py
```diff
@@ -97,7 +97,6 @@
     install_requires=[
         "click>=8.1.6,<9",
         "ijson>=3.1.4,<4",
-        "importlib-metadata",
```
fubuloubu, line 100: this is bundled python 3.8+?

fubuloubu (approved):

setup.py
antazoey: Yes, in 3.8+, you can now use `importlib.metadata` directly, so we could have done this a while ago

antazoey: Ugh - I guess in 3.8 you can't select by group.
Whyyy Python whyy.

antazoey:
I made sure plugins installed from remote, editable mode, and regular pip look good in `ape plugins list`

Issue #636, opened 2022-04-06
File extensions should glob the entire extension


fubuloubu:
### Environment information

* `ape` and plugin versions:

```
$ ape --version
0.1.4

$ ape plugins list
None
```

* Python Version: x.x.x
* OS: linux

### What went wrong?

Certain dependencies like dapptools and foundry us `*.t.sol` as their file extensions to denote testing files. However, we can't compile them currently for other reasons, nor do we need to since they're just testing files.

### How can it be fixed?

Currently, we will see these as just `*.sol` files, but if the file extension glob instead worked with the largest extension in can find on a file, we can differentiate between these types of files and avoid compiling them, which is the job of other tools.

unparalleled-js:
Extensions are weird! Here is more information on what this change means:

* The first `.` found in a contract source file name will note the start of the chain of extensions.
* You cannot have a `.` in the file that is not part of the chain of extensions.
* By including multiple `.` **anywhere** in your source file name, it will cause it not to compiled by `ape` **today**.

I think this is all okay, but we may want to include a way for users to register file-extension (chains) to their `solidity` or `vyper` configs. This way, we can support projects that do want to compile these files.

antazoey:
This is causing issues with users trying to publish contracts with names like `Contract.1.2.3.sol`

PR #1998, opened 2024-04-11
fix: issues where "full extension" was not always honored

antazoey:
### What I did

fixes some misses from this old PR: https://github.com/ApeWorX/ape/issues/636
this is really helpful for vyper 0.4.0 support - will use here https://github.com/ApeWorX/ape-vyper/pull/110/files

### Checklist

### DIFF
--- a/src/ape/utils/os.py
+++ b/src/ape/utils/os.py
@@ -139,3 +139,18 @@ class use_temp_sys_path:
         for path in self.exclude:
             if path not in sys.path:
                 sys.path.append(path)
+
+
+def get_full_extension(path: Path) -> str:
+    """
+    For a path like ``Path("Contract.t.sol")``,
+    returns ``.t.sol``, unlike the regular Path
+    property ``.suffix`` which returns ``.sol``.
+    """
+    if path.is_dir():
+        return ""
+
+    parts = path.name.split(".")
+    start_idx = 2 if path.name.startswith(".") else 1
+    suffix = ".".join(parts[start_idx:])
+    return f".{suffix}" if suffix and f".{suffix}" != f"{path.name}" else ""
--- a/src/ape/utils/os.py
+++ b/src/ape/utils/os.py
@@ -152,5 +152,9 @@ def get_full_extension(path: Path) -> str:
 
     parts = path.name.split(".")
     start_idx = 2 if path.name.startswith(".") else 1
+
+    # NOTE: Handles when given just `.hiddenFile` since slice indices
+    #   may exceed their bounds.
     suffix = ".".join(parts[start_idx:])
+
     return f".{suffix}" if suffix and f".{suffix}" != f"{path.name}" else ""
--- a/tests/functional/utils/test_os.py
+++ b/tests/functional/utils/test_os.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 import pytest
 
-from ape.utils.os import get_all_files_in_directory, get_relative_path
+from ape.utils.os import get_all_files_in_directory, get_full_extension, get_relative_path
 
 _TEST_DIRECTORY_PATH = Path("/This/is/a/test/")
 _TEST_FILE_PATH = _TEST_DIRECTORY_PATH / "scripts" / "script.py"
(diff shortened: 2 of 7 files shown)


tests/functional/utils/test_os.py
```diff
@@ -70,3 +70,18 @@ def test_get_all_files_in_directory():
...
+
+
+@pytest.mark.parametrize(
+    "path,expected",
+    [
+        ("test.json", ".json"),
+        ("Contract.t.sol", ".t.sol"),
+        ("this.is.a.lot.of.dots", ".is.a.lot.of.dots"),
+        ("directory", ""),
+    ],
```
fubuloubu, line 82: Add `".privateFolder"` to the list

antazoey: done and fixed

tests/functional/utils/test_os.py
```diff
@@ -70,3 +70,20 @@ def test_get_all_files_in_directory():
...
+
+@pytest.mark.parametrize(
+    "path,expected",
+    [
+        ("test.json", ".json"),
+        ("Contract.t.sol", ".t.sol"),
+        ("this.is.a.lot.of.dots", ".is.a.lot.of.dots"),
+        ("directory", ""),
+        (".privateDirectory", ""),
+        ("path/to/.gitkeep", ""),
```
fubuloubu, line 83: also what would something like `.config.json` or something like that do?

probably thinking super nonstandard at this point, as long as we avoid `contracts/.cache` think it's good

src/ape/utils/os.py
```diff
@@ -139,3 +139,14 @@ def __exit__(self, *exc):
         for path in self.exclude:
             if path not in sys.path:
                 sys.path.append(path)
+
+
+def get_full_extension(path: Path) -> str:
```
fubuloubu, line 144: Think if `path` is a directory, can just bail early w/ `""`

tests/functional/utils/test_os.py
antazoey: .cache is ignored.

src/ape/utils/os.py
antazoey: We probably can. However, just as a note: ChainLink contains directories with `.sol` extensions for whatever reason.

tests/functional/utils/test_os.py
antazoey: Starting to wish we never did https://github.com/ApeWorX/ape/issues/636

antazoey: Ok made path names like `.cache.json` do `.json`, I think that works... Kinda making up our own path rules at this point.

src/ape/utils/os.py
```diff
@@ -139,3 +139,18 @@ def __exit__(self, *exc):
...
+    For a path like ``Path("Contract.t.sol")``,
+    returns ``.t.sol``, unlike the regular Path
+    property ``.suffix`` which returns ``.sol``.
+    """
+    if path.is_dir():
+        return ""
+
+    parts = path.name.split(".")
+    start_idx = 2 if path.name.startswith(".") else 1
+    suffix = ".".join(parts[start_idx:])
```
fubuloubu, line 155: this doesn't raise when it's just `.something`?

fubuloubu (approved):

src/ape/utils/os.py
antazoey: It would fail the test I added if it did! But just for funzies, I will execute each line individually:

```
In [2]: path = Path(".something")

In [3]: parts = path.name.split(".")

In [4]: parts
Out[4]: ['', 'something']

In [5]: start_idx = 2 if path.name.startswith(".") else 1

In [6]: start_idx
Out[6]: 2

In [7]: suffix = ".".join(parts[start_idx:])

In [8]: suffix
Out[8]: ''

In [9]: f".{suffix}" if suffix and f".{suffix}" != f"{path.name}" else ""
Out[9]: ''
```

**Note**: In python, when using slices, the indices are allowed to exceed the last index in a sequence. For example:

```python
In [13]: my_lst[:9999]
Out[13]: [1, 2, 3]

In [14]: my_lst[9999:]
Out[14]: []
```

It is the design of slices, to make them more convenient.

antazoey: All that being said, I am happy to put in an index check for the sake of readability.

fubuloubu: > **Note**: In python, when using slices, the indices are allowed to exceed the last index in a sequence. For example:
> 
> ```python
> In [13]: my_lst[:9999]
> Out[13]: [1, 2, 3]
> 
> In [14]: my_lst[9999:]
> Out[14]: []
> ```
> 
> It is the design of slices, to make them more convenient.

TIL

fubuloubu: > All that being said, I am happy to put in an index check for the sake of readability. Shall I?

No we are good! I just didn't understand that bit

Maybe add a comment?

fubuloubu (approved):

Issue #1999, opened 2024-04-11
plugins not recognized when pip isn't installed


wakamex:
### Environment information

- OS: linux
- Python Version: Python 3.11.8
- `ape` and plugin versions:
- ape 0.7.13
- ape-alchemy 0.7.1

```sh
$ ape --version
0.7.13

$ ape plugins list
  alchemy      0.7.1.dev11+gad25ef2.d20240323
  etherscan    0.7.2
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
cat: ape-config.yaml: No such file or directory
```

### What went wrong?

couldn't use alchemy in my script. upon further inspection, it wasn't showing up in plugins or networks. i realized this is because of my non-standard installation with `uv`. since I don't have pip installed, it fails trying to query which plugins are installed.

here are some terminal printouts:
```
>> ape networks list              
ethereum  (default)
├── goerli
│   └── geth  (default)
├── local  (default)
│   ├── geth
│   └── test  (default)
├── mainnet
│   └── geth  (default)
└── sepolia
    └── geth  (default)
>> uv pip install pip          
Resolved 1 package in 25ms
Installed 1 package in 104ms
 + pip==24.0
>> ape networks list 
ethereum  (default)
├── goerli
│   ├── alchemy
│   └── geth  (default)
├── local  (default)
│   ├── geth
│   └── test  (default)
├── mainnet
│   ├── alchemy
│   └── geth  (default)
└── sepolia
    ├── alchemy
    └── geth  (default)
```

trying to use a script that uses alchemy without pip installed fails:
```
>> uv pip uninstall pip       
Uninstalled 1 package in 29ms
 - pip==24.0
>> python steth.py        
Traceback (most recent call last):
  File "/code/testape/steth.py", line 19, in <module>
    context = networks.parse_network_choice('ethereum:mainnet:alchemy')
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/testape/.venv/lib/python3.11/site-packages/ape/managers/networks.py", line 528, in parse_network_choice
    provider = self.get_provider_from_choice(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/testape/.venv/lib/python3.11/site-packages/ape/managers/networks.py", line 488, in get_provider_from_choice
[...]

PR #2000, opened 2024-04-11
feat: use `uv` if installed for `ape plugins` install / uninstall commands

wakamex:
### What I did

add setuptools and pip to base requirements so that `uv` can install everything correctly.

fixes: #1999

fixes: #

### How I did it

edited the setup.cfg

### How to verify it

>> uv venv .venv -p 3.10
>> source .venv/bin/activate
>> uv pip install -e ../newape
>> uv pip install -e ../ape-alchemy
>> ape plugins list
Installed Plugins
  alchemy    0.7.1.dev11+gad25ef2.d20240411
>> ape --version
0.7.14
>> ape networks list
ethereum  (default)
├── goerli
│   ├── alchemy
│   └── geth  (default)
├── local  (default)
│   ├── geth
│   └── test  (default)
├── mainnet
│   ├── alchemy
│   └── geth  (default)
└── sepolia
    ├── alchemy
    └── geth  (default)
>> python ../testape/steth.py
alchemy through ape took 1.13 seconds for 1076 events (events per second: 953.32)

### Checklist

- [ ] Consider eliminating pip dependency in favor of pkg_resources


setup.py
```diff
@@ -111,6 +111,7 @@
         "PyYAML>=5.0,<7",
         "requests>=2.28.1,<3",
         "rich>=12.5.1,<14",
+        "setuptools",  # Installation tool
```
fubuloubu, line 114: are we using setuptools? or just means as a downstream dependency of trying to install a plugin?

fubuloubu: tbh can probably add a TODO to remove when migrating to pyproject.toml workflow for all plugins

fubuloubu (dismissed):

setup.py
```diff
@@ -102,6 +101,7 @@
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
         "pandas>=1.3.0,<2",
+        "pip",
```
antazoey, line 104: I would rather fail early if we detect `pip` is not in your PATH, since we are not using pip as a library but only via `subprocess.call`. This way, `pip` does no have to ship with Ape.

fubuloubu: `pip` or `uv pip` detected, then we can proceed (since `uv pip` is cli-compatible with `pip`)

setup.py
wakamex: without it `ape plugins list` crashes, which seems pretty key (due to pkg_resources)

```
>> ape plugins list 
  File "/code/newape/.venv/bin/ape", line 5, in <module>
    from ape._cli import cli
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/__init__.py", line 10, in <module>
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/managers/__init__.py", line 23, in <module>
    ManagerAccessMixin.config_manager = ConfigManager(
  File "/code/newape/.venv/lib/python3.10/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/managers/config.py", line 134, in load_configs
    return cm.load()
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/managers/config.py", line 277, in load
    _ = self._plugin_configs
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/managers/config.py", line 205, in _plugin_configs
    valid_ecosystems = dict(self.plugin_manager.ecosystems)
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/plugins/__init__.py", line 138, in __getattr__
    self._register_plugins()
  File "/code/newape/.venv/lib/python3.10/site-packages/ape/plugins/__init__.py", line 193, in _register_plugins
    module = importlib.import_module(module_name)
  File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/code/newape/.venv/lib/python3.10/site-packages/ape_test/__init__.py", line 10, in <module>
    from ape_test.provider import EthTesterProviderConfig, LocalProvider
  File "/code/newape/.venv/lib/python3.10/site-packages/ape_test/provider.py", line 7, in <module>
    from eth.exceptions import HeaderNotFound
  File "/code/newape/.venv/lib/python3.10/site-packages/eth/__init__.py", line 1, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

not sure what the TODO should look like, please advise (or code suggest 😉)

setup.py
wakamex: okay so we could make `uv pip` stand-in fallback for `pip`....

wakamex:
I think I made uv a fallback replacement for pip

saw some warnings along the way:
```
.venv/lib/python3.10/site-packages/eth/__init__.py:1
  /code/newape/.venv/lib/python3.10/site-packages/eth/__init__.py:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources

.venv/lib/python3.10/site-packages/_pytest/config/__init__.py:1373
  /code/newape/.venv/lib/python3.10/site-packages/_pytest/config/__init__.py:1373: PytestConfigWarning: Unknown config option: timeout
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")
```

wakamex:
evm-trace has a capped dependency on py-evm versions which use pkg_resources. I remove it here: https://github.com/ApeWorX/evm-trace/pull/62

wakamex:
uncapping python-dateutil, eth-account, eth-utils, web3[tester], and eip712 points to the culprit:

```
uv pip install -e . --prerelease=allow
   Built file:///code/newape                                                                                          Built 1 editable in 676ms
Updating https://github.com/wakamex/evm-trace.git (HEAD)
⠙ rich==13.7.1                                                                                                          × No solution found when resolving dependencies:
  ╰─▶ Because eip712>=0.2.3 depends on eth-account>=0.10.0,<0.11 and only the following versions of eip712 are
      available:
          eip712<=0.2.3
          eip712==0.2.4
          eip712==0.2.5
      we can conclude that eip712>=0.2.3 depends on eth-account>=0.10.0,<0.11.
      And because eth-ape==0.7.15.dev5+gdbe67d70.d20240415 depends on eth-account>=0.11.2, we can conclude that
      eip712>=0.2.3 and eth-ape==0.7.15.dev5+gdbe67d70.d20240415 are incompatible.
      And because eth-ape==0.7.15.dev5+gdbe67d70.d20240415 depends on eip712>=0.2.3 and you require
```

uv has really nice dependency errors :)

wakamex:
uncapping and bumping pandas to >=2 lets uv pick 2.2.2 instead of 1.5.3 (why that one? i dunno)

1.5.3 results in a hecking weird error:
```
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

wakamex:
got it installed with eth_account==0.11.2 by uncapping eth-account in ape and eip712. also `ape plugins list` doesn't error on missing pkg_resources.

src/ape/cli/__init__.py
```diff
@@ -28,6 +31,8 @@
 )
 from ape.cli.paramtype import AllFilePaths, Path
 
+PIP_COMMAND = ["uv", "pip"] if which("uv") else [sys.executable, "-m", "pip"]
```
antazoey, line 34: This could probably be in `ape/plugins/_utils.py`

setup.py
```diff
@@ -101,34 +100,35 @@
...
         "packaging>=23.0,<24",
-        "pandas>=1.3.0,<2",
+        "pandas>=2",
         "pluggy>=1.3,<2",
         "pydantic>=2.5.2,<3",
         "pydantic-settings>=2.0.3,<3",
         "PyGithub>=1.59,<2",
         "pytest>=6.0,<8.0",
-        "python-dateutil>=2.8.2,<3",
+        "python-dateutil>=2.8.2",
```
antazoey, line 109: - [ ] put back all upper bounds as you can

setup.py
```diff
@@ -101,34 +100,35 @@
...
+        "eth-tester @ git+https://github.com/wakamex/eth-tester.git",
         "eth-typing>=3.5.2,<4",
-        "eth-utils>=2.3.1,<3",
+        "eth-utils>=2.3.1",
         "py-geth>=4.4.0,<5",
-        "web3[tester]>=6.16.0,<7",
+        "web3[tester]>=6.16.0",
         # ** Dependencies maintained by ApeWorX **
-        "eip712>=0.2.3,<0.4",
+        "eip712 @ git+https://github.com/wakamex/eip712.git",
```
antazoey, line 127: - [ ] replace with released versions

.github/workflows/test.yaml
```diff
@@ -23,20 +23,30 @@ jobs:
 
         - name: Install Dependencies
           run: |
-            python -m pip install --upgrade pip
-            pip install .[lint]
+            python -m pip install --upgrade uv
+            uv venv .venv
```
antazoey, line 27: Why is a virtual environment required? I wouldn't expect that.

src/ape/plugins/_utils.py
```diff
@@ -361,7 +361,7 @@ def _prepare_install(
             logger.warning(f"Plugin '{self.name}' is not an trusted plugin.")
 
         result_handler = ModifyPluginResultHandler(self)
-        pip_arguments = [sys.executable, "-m", "pip", "install"]
+        pip_arguments = PIP_COMMAND + ["install"]
```
antazoey, line 364: style nit

```suggestion
        pip_arguments = [*PIP_COMMAND, "install"]
```

antazoey (commented):
Thanks

The pkg_resources problem is annoying. Thankfully it is going away anyway as part of adding Python 3.12 support.

src/ape/plugins/_utils.py
wakamex: applied same style in 2 other places in bd28b327ef62e28a40a526ae4a31240c30c1ddc4

setup.py
wakamex: done by resetting dependencies (moved to another PR in https://github.com/ApeWorX/ape/pull/2003) in f04d70289737d69453a356398544c917cb547be5

setup.py
wakamex: done by resetting dependencies (moved to another PR in https://github.com/ApeWorX/ape/pull/2003) in f04d70289737d69453a356398544c917cb547be5

.github/workflows/test.yaml
wakamex: reset to main in d0acae8a67ea7c2ab02f504b3a5ad1888e35a6b2

fubuloubu (approved):

src/ape/cli/__init__.py
```diff
@@ -27,6 +27,7 @@
     verbosity_option,
 )
 from ape.cli.paramtype import AllFilePaths, Path
+from ape.plugins._utils import PIP_COMMAND
```
antazoey, line 30: I wasn't super clear I guess before! I meant remove this from this module entirely and import only from _utils everywhere. Why does this need to be defined here and in `__all__` ?

Not saying am against it, just mostly trying to follow. If we do decide to keep `PIP_COMMAND `in ape.cli` namespace, you can put it back, as that would make sense then.

antazoey: it looks like nothing is importing from here too, maybe it was leftover accidentally? does the circular go away if deleted?

wakamex: I have no idea why it's here

wakamex: removed in 125a616d89db1be3e0c2993078ef920d1829f6b2

setup.py
wakamex: reset dependencies to how they were in f04d70289737d69453a356398544c917cb547be5

fubuloubu (approved):
Re-ran failed test case for Python 3.9

I'm okay with the move of the `PluginManager` but curious what @antazoey thinks

antazoey:
> Re-ran failed test case for Python 3.9
> 
> I'm okay with the move of the `PluginManager` but curious what @antazoey thinks

I like the change but am worried if it's breaking.

fubuloubu:
> > Re-ran failed test case for Python 3.9
> > I'm okay with the move of the `PluginManager` but curious what @antazoey thinks
> 
> I like the change but am worried if it's breaking.

It lifts `PluginManager` from `ape.plugins` (which tbh should have been private e.g. `ape._plugins`) to the `ape.managers` namespace, which is also still not typically widely used except for typing maybe

The expectation is that singletons from `ape.managers` end up in top level namespace, and also injected via `ManagerAccessMixin`

wakamex:
She may have been referring to it failing 1 test. It seemed really random. Only MacOS Python 3.9 failed. Looks like after that someone may have re-run it and it now passes.

wakamex:
rebased. tests pass. gonna do another commit to update dependencies from main.

antazoey:
> (which tbh should have been private e.g. `ape._plugins`)

The registration function needs to be public for plugins: https://github.com/ApeWorX/ape-solidity/blob/main/ape_solidity/__init__.py#L1

wakamex:
looks like github runners only support darwin-arm64 down to python 3.10.11 according to the CI error

```
   Error: The version '3.8' with architecture 'arm64' was not found for macOS 14.4.1.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

fubuloubu:
> looks like github runners only support darwin-arm64 down to python 3.10.11 according to the CI error
> 
> ```
>    Error: The version '3.8' with architecture 'arm64' was not found for macOS 14.4.1.
>   The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
> ```

Yes we are tracking to fix with #2020

wakamex:
awesome. let me know if this PR is missing anything. maybe we can revisit after #2020.

antazoey:
during testing I ran into an issue but fixed it here: https://github.com/ApeWorX/ape/pull/2021

src/ape/managers/plugins.py
```diff
@@ -0,0 +1,178 @@
...
+                    if validated_plugin:
+                        yield validated_plugin
+
+    @property
+    def registered_plugins(self) -> Set[str]:
+        self._register_plugins()
+        return {x[0] for x in pluggy_manager.list_name_plugin()}
+
+    @functools.cached_property
+    def _plugin_modules(self) -> Tuple[str, ...]:
```
antazoey, line 84: merge miss: https://github.com/ApeWorX/ape/pull/2004

^ missing from the diff. (no longer do we need to parse pip freeze)

- [ ] - i can plop it in

antazoey (dismissed):

antazoey:
**issues**: tests fail when `uv` is installed. it seems like we need to test that case more anyway.

antazoey:
**issue**: `pkg_resources` shows up on trie 2.2.0, which is "installable" in current conditions (ran into this). Lucky, I think we can bump it to 3.0 and it will still work.

- [ ] bump peer dependency `trie` to `>=3.0.0,<4`/

fubuloubu (approved):

wakamex:
just tested installation speeds between uv and pip, on latest 0.7.23 26bec8b75eb4d6b489c8585458697d97052f24e9:

| tool | no cache | cached | uncached unpinned pandas |
|-|-|-|-|
| uv | 2:09.43 | 1.411 | 5.202 |
| pip | 2:29.40 | 8.867 | 9.192 |
| speedup (s) | 19.97 | 7.456 | 3.99 |
| speedup (%) | 13.37%  | 84.07% | 43% |

no fucking way
image

also, big speedup comes from loosening the pandas pin, which requires building numpy on my machine (Fedora 40)

fubuloubu:
Yeah, we actually want to remove the pandas pin eventually, it's not quite state of the art and end users might have a couple of different options they may want to choose from

PR #2001, opened 2024-04-12
feat!: contract creation metadata

banteg:
### What I did

implemented contract creation metadata. this improves upon `get_creation_receipt` by offloading the logic to query engine and also exposing deploy `block`, `deployer` and `factory` besides the `receipt`.

this pull request implements the needed logic with two query providers:
- otterscan (currently under geth query provider), which takes around 300ms
- default query engine, which takes around 5.5s https://github.com/ApeWorX/ape/pull/2001/commits/0460d718e0f7da97b166e4d2c0188919495ba234

the default query implementation is also significantly improved, as it now can detect deployments from factories.

the `ContractCreationQuery` is demoted from block query to base query, as now doesn't accept the block range arguments. its return type has also changed to `ContractCreation`. we are forced to return an iterator by `QueryAPI.perform_query`, but maybe we should relax this requirement where an iterator doesn't make sense and a single value is returned.

as this query provides superior information and accuracy, it supplants `get_creation_receipt`, so i have removed duplicate methods related to them, namely:
- `chain.provider.get_contract_creation_receipts`
- `chain.provider.ots_get_contract_creator`
- `chain.provider._get_contract_creation_receipt`
- `chain.contracts.get_creation_receipt`
- `ContractInstance.receipt` - available dynamically as `ContractInstance.creation.receipt`

instead, this pull request exposes the expanded deployment metadata is placed under `ContractInstance.creation` property, so you can do `Contract(addr).creation.block` and so on.

### How to verify it

```python
from ape.api.query import ContractCreationQuery

query = ContractCreationQuery(columns=['*'], contract='0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801')

# implemented in 3 query engines

next(chain.query_manager.query(query, engine_to_use='geth'))
# ContractCreation(txn_hash='0x37d8f4b1b371fde9e4b1942588d16a1cbf424b7c66e731ec915aca785ca2efcf', block=12369739, deployer='0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa', factory='0x1F98431c8aD98523631AE4a59f267346ea31F984')

next(chain.query_manager.query(query, engine_to_use='__default__'))
# ContractCreation(txn_hash='0x37d8f4b1b371fde9e4b1942588d16a1cbf424b7c66e731ec915aca785ca2efcf', block=12369739, deployer='0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa', factory='0x1F98431c8aD98523631AE4a59f267346ea31F984')

# NOTE etherscan query engine doesn't return factory address yet
# i reached out to them about this
next(chain.query_manager.query(query, engine_to_use='etherscan'))
# ContractCreation(txn_hash='0x37d8f4b1b371fde9e4b1942588d16a1cbf424b7c66e731ec915aca785ca2efcf', block=12369739, deployer='0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa', factory=None)

# easily access the creation metadata in ContractInstance.
# this one is cached in retrieved almost instantly on later invocations.
Contract('0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801').creation
# ContractCreation(txn_hash='0x37d8f4b1b371fde9e4b1942588d16a1cbf424b7c66e731ec915aca785ca2efcf', block=12369739, deployer='0x11E4857Bb9993a50c685A79AFad4E6F65D518DDa', factory='0x1F98431c8aD98523631AE4a59f267346ea31F984')
```

### Checklist

- [x] update etherscan query provider api https://github.com/ApeWorX/ape-etherscan/pull/124


src/ape_geth/query.py
```diff
@@ -29,7 +39,18 @@ def estimate_contract_creation_query(self, query: ContractCreationQuery) -> Opti
...
-    def get_contract_creation_receipt(self, query: ContractCreationQuery) -> Iterator[ReceiptAPI]:
+    def get_contract_creation_receipt(
+        self, query: ContractCreationQuery
+    ) -> Iterator[ContractCreation]:
         if self.network_manager.active_provider and isinstance(self.provider, EthereumNodeProvider):
-            if receipt := self.provider._get_contract_creation_receipt(query.contract):
-                yield receipt
+            ots = self.provider._make_request("ots_getContractCreator", [query.contract])
+            if ots is None:
+                return None
```
fubuloubu, line 48: Estimate query returning None is how we say that data is not available upstream of the query system

This rate can just do the call and see if it works, kind of a shortfall of the method (I am not sure of this query makes as much sense using the query system anymore)

banteg: i think it makes more sense to skip on estimate is when the provider doesn't support ots namespace. we can check the provider once via ots_getApiLevel once we refactor into a separate otterscan provider.

here None is a final answer, as in "this is not a contract". it can still be expensive to put into estimate.

banteg: addressed in https://github.com/ApeWorX/ape/pull/2001/commits/a070592625a287fc1abbbd8b7a4a74e295763059

src/ape/contracts/base.py
```diff
@@ -899,10 +900,18 @@ def receipt(self) -> ReceiptAPI:
                 return receipt
 
         # Brute force find the receipt.
-        receipt = self.chain_manager.contracts.get_creation_receipt(self.address)
+        receipt = self.meta.receipt
         self._cached_receipt = receipt
         return receipt
 
+    @cached_property
+    def meta(self):
```
antazoey, line 908: What is the return type here?

src/ape/contracts/base.py
```diff
@@ -899,10 +900,18 @@ def receipt(self) -> ReceiptAPI:
...
+        """
+        Contract creation details: deployer, factory, block, receipt.
+        """
+        query = ContractCreationQuery(columns=["*"], contract=self.address)
+        return next(self.query_manager.query(query))
```
antazoey, line 913: Even if unlikely or impossible on EVM, can we avoid the `StopIteration` here and raise a different error or return None?

src/ape_geth/query.py
```diff
@@ -1,12 +1,22 @@
...
+from pydantic import BaseModel
+
 from ape.api import ReceiptAPI
 from ape.api.query import ContractCreationQuery, QueryAPI, QueryType
 from ape.exceptions import QueryEngineError
+from ape.types.address import AddressType
 from ape_ethereum.provider import EthereumNodeProvider
 
 
+class ContractCreation(BaseModel):
```
antazoey, line 13: Should this be defined here? This seems like something that should live in the API level, since other plugins like `ape-etherscan` also implement this.

antazoey (commented):
Cool thanks!

src/ape/contracts/base.py
banteg: `ContractCreation`, for some reason adding an annotation causes a circular import

src/ape/contracts/base.py
banteg: not possible to hit that branch for an initialized contract instance

src/ape/contracts/base.py
banteg: fixed in https://github.com/ApeWorX/ape/pull/2001/commits/65446305c255d94c7a416ec8913262b08f5d8f48

src/ape_geth/query.py
banteg: resolved in https://github.com/ApeWorX/ape/pull/2001/commits/65446305c255d94c7a416ec8913262b08f5d8f48

src/ape/contracts/base.py
antazoey: I was imagining some weird ecosystem that didn't use `EthereumNodeProvider`, even if it was like some sort of Mock-EcosysemAPI, but maybe you're right. No big deal. I just really hate getting random `StopIteration` in tests or anything, it seems like it always happens to me 😅

src/ape/managers/chain.py
```diff
@@ -835,6 +841,37 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
...
+        Get contract creation metadata containing txn_hash, deployer, factory, block.
+
+        Args:
+            address (AddressType): The address of the contract.
+
+        Returns:
+            Optional[:class:`~pae.api.query.ContractCreation`]
+        """
+        creation = self._local_contract_creation.get(address)
+        if creation is not None:
```
fubuloubu, line 855:
```suggestion
        if creation := self._local_contract_creation.get(address):
```

src/ape/managers/chain.py
```diff
@@ -835,6 +841,37 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
...
+            return creation
+        # read from disk
+        creation = self._get_contract_creation_from_disk(address)
+        if creation is not None:
```
fubuloubu, line 859:
```suggestion
        if creation := self._get_contract_creation_from_disk(address):
```

src/ape/managers/chain.py
```diff
@@ -835,6 +841,37 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
...
+        try:
+            creation = next(self.query_manager.query(query))
+        except StopIteration:
+            creation = None
+        if creation is not None:
+            self._cache_contract_creation_to_disk(address, creation)
+            self._local_contract_creation[address] = creation
+            return creation
+
+        return None
```
fubuloubu, line 873:
```suggestion
        self._cache_contract_creation_to_disk(address, creation)
        self._local_contract_creation[address] = creation
        return creation
```

src/ape/managers/query.py
```diff
@@ -76,12 +78,47 @@ def perform_block_transaction_query(
...
+                    hi = mid
+
+            if self.provider.get_code(query.contract, block_id=hi):
+                return hi
+
+        # if it doesn't have code at head, skip the search
+        if not self.provider.get_code(query.contract):
+            return None
+
+        block = find_creation_block(0, self.chain_manager.blocks.height)
```
fubuloubu, line 102: block may be None

banteg: when?

src/ape/managers/chain.py
```diff
@@ -835,6 +841,33 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
...
 
+    def get_creation_metadata(self, address: AddressType) -> Optional[ContractCreation]:
+        """
+        Get contract creation metadata containing txn_hash, deployer, factory, block.
+
+        Args:
+            address (AddressType): The address of the contract.
+
+        Returns:
+            Optional[:class:`~pae.api.query.ContractCreation`]
```
fubuloubu, line 852:
```suggestion
            Optional[:class:`~ape.api.query.ContractCreation`]
```

src/ape/managers/chain.py
```diff
@@ -835,6 +841,33 @@ def get_proxy_info(self, address: AddressType) -> Optional[ProxyInfoAPI]:
...
+        # read from disk
+        if creation := self._get_contract_creation_from_disk(address):
+            self._local_contract_creation[address] = creation
+            return creation
+        # query and cache
+        query = ContractCreationQuery(columns=["*"], contract=address)
+        try:
+            creation = next(self.query_manager.query(query))
+        except StopIteration:
+            return None
```
fubuloubu, line 865: Think it's just better to raise a normal exception that tells you the algo is broken and to raise an issue in ape repo

Do we have one of those "we don't expect it to fail, but if it did please create an issue" type of exceptions?

src/ape/managers/query.py
```diff
@@ -76,12 +78,49 @@ def perform_block_transaction_query(
...
+                code = self.provider.get_code(query.contract, block_id=mid)
+                if not code:
+                    lo = mid
+                else:
+                    hi = mid
+
+            if self.provider.get_code(query.contract, block_id=hi):
+                return hi
+
+            return None
```
fubuloubu, line 98: Hmm, actually I think if your intention is that this should not be reachable, than this should probably just raise an exception

src/ape/managers/query.py
fubuloubu: If this raised an exception it would not longer return None in some rare future scenario

src/ape/managers/chain.py
antazoey: a `ValueError` works for me, as this is unlikely. Anything is better than `StopIteration`.

fubuloubu: Vyper has this special exception, it wouldn't subclass `ApeException` but would be a special exception you just put anywhere you think you're safe but just to be sure (or protect future maintainers from stupidity)

Maybe `ape.exceptions.Unreachable`?

antazoey: I really don't like the name `Unreachable`. Everything is reachable pretty much. You may think it is not, but damn it, I will somehow reach it!!

antazoey: Just raise  an exception saying "For some unknown reason, there is no creation block."

antazoey: I mean - I could create a contract after deploying, then revert to some older block, delete stuff from the cache or something, and then query a creation block and get nowhere.

antazoey: Could even be the `ape.exceptions.QueryError`

antazoey: actually, is `None` so bad?

fubuloubu: > Just raise an exception saying "For some unknown reason, there is no creation block."

Well the algo is basically not ever supposed to return None, if the algo is correct. If the algo is wrong, it may. But if it's wrong, that's a guaranteed bug

fubuloubu: "the algo" meaning the binary search function

antazoey: How does the algo handle time travel though?

fubuloubu: not sure what that means. the assumption here is that a contract is only created one time, and that you have the block available where it is created (otherwise it (should) raise)

antazoey: time travel refers to snapshotting and reverting. You can call a method anytime, You can clear a cache anytime.

src/ape/managers/query.py
```diff
@@ -76,12 +78,49 @@ def perform_block_transaction_query(
...
+
+        # if it doesn't have code at head, skip the search
+        if not self.provider.get_code(query.contract):
+            return None
+
+        block = find_creation_block(0, self.chain_manager.blocks.height)
+        traces = self.provider._make_request("trace_replayBlockTransactions", [block, ["trace"]])
```
antazoey, line 105: Before, nodes were **not** required to have `trace_replayBlockTransactions` RPC but now they are.
Is there a fallback we can do we here?

antazoey: @fubuloubu @banteg

banteg: hmm, i think for geth we can try using `debug_traceBlockByNumber` with a custom tracer? i'll research this.

banteg: working as before was incomplete as it couldn't find contracts deployed from factories. i guess this could be one of several fallbacks.

antazoey: Here is one course of action:

1. Have old way as a last-resort
2. Add a timeout to the last resort so it never feels too much like hanging
3. Log when using last resort about the factories (maybe as a DEBUG?)
4. Add a TODO + new issue for properly adding geth support that can get contracts from factories

banteg: added a geth variant that implements essentially the same logic as the parity variant.

will add the last resort a bit later.

src/ape/managers/query.py
```diff
@@ -101,14 +106,25 @@ def find_creation_block(lo, hi):
...
-        # iterate over transaction traces in a block to find the deploy transaction
+        # iterate over transaction traces in a block to find the deployment transaction
         # this method also supports contract factories
-        if "geth" in self.provider.client_version.lower():
-            yield from self._find_creation_in_block_via_geth(block, query.contract)
-        else:
-            yield from self._find_creation_in_block_via_parity(block, query.contract)
+        try:
+            if "geth" in self.provider.client_version.lower():
+                yield from self._find_creation_in_block_via_geth(block, query.contract)
```
antazoey, line 120: Some of this seems like it'd make more sense in `ape-ethereum`, since it uses EVM-based RPCs.

banteg: should i join all methods and move them under `ape_geth.query`?

banteg: did a bit of a refactor:
- move the binary search query from the default engine into ethereum engine
- move otterscan query from geth to otterscan engine

src/ape/managers/chain.py
banteg: i think returning None here is consistent with other similar methods like `get_proxy_info`. the query would return None for non-contracts.

src/ape/contracts/base.py
```diff
@@ -875,33 +877,17 @@ def from_receipt(cls, receipt: ReceiptAPI, contract_type: ContractType) -> "Cont
...
-            else:
-                self._cached_receipt = receipt
-                return receipt
-
-        # Brute force find the receipt.
-        receipt = self.chain_manager.contracts.get_creation_receipt(self.address)
-        self._cached_receipt = receipt
-        return receipt
+        data = self.chain_manager.contracts.get_creation_metadata(self.address)
+        assert data, "unreachable"
```
antazoey, line 889: Hitting this in tests. Thus it is reachable!

antazoey: does this work on EthTester?

banteg: i was under impression that `ContractInstance` is guaranteed to point to a live contract. but apparently it's not enforced. in this case we can just propagate None from the query. fixed this in https://github.com/ApeWorX/ape/pull/2001/commits/639c511f457bea5d002585c1aabe3b282f975fcc

antazoey (approved):

src/ape/contracts/base.py
```diff
@@ -875,33 +879,19 @@ def from_receipt(cls, receipt: ReceiptAPI, contract_type: ContractType) -> "Cont
...
+
+        # Cache creation.
+        creation = ContractCreation.from_receipt(receipt)
+        cls.chain_manager.contracts._local_contract_creation[address] = creation
+
         return instance
 
     @property
-    def receipt(self) -> ReceiptAPI:
+    def creation(self) -> Optional[ContractCreation]:
```
fubuloubu, line 890: suggestion: I still think that this slightly more verbose property would be less likely to cause a conflict
```suggestion
    def creation_metadata(self) -> Optional[ContractCreation]:
```

fubuloubu (approved):
This is looking really, really good now

antazoey:
I sense something unusual with the mac OS runners right now. I am re-running an Ubuntu to see if that is successful as I investigate.

fubuloubu (approved):

antazoey (approved):

PR #2003, opened 2024-04-15
feat: remove setuptools dependence

wakamex:
### What I did

spun out from #2000, remove setuptools dependency

### How I did it

two ways:
1. remove dependency caps. recent versions of some dependencies have removed usage of pkg_resources.
2. for those dependencies that haven't, fork them to remove usage of pkg_resources, and point to that fork.

### How to verify it

install with `uv` then run `ape plugins list` and see if it errors out complaining about pkg_resources.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ extras_require["dev"] = (
 with open("./README.md") as readme:
     long_description = readme.read()
 
+
 setup(
     name="eth-ape",
     use_scm_version=True,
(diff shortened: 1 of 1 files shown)


wakamex:
PRs to remove setuptools dependency from dependencies:
1. eth-tester https://github.com/ethereum/eth-tester/pull/288
2. eip712 https://github.com/ApeWorX/eip712/pull/48
3. evm-trace https://github.com/ApeWorX/evm-trace/pull/62

after this, we can remove forks from setup.py

wakamex:
PRs are in for all 3 dependencies, but eth-tester hasn't bumped their version, so we'd have to pin a hash if we want to use it right now, or wait for them to bump it.

wakamex:
eth-tester just put out a new beta which includes the PR to move away from setuptools.
I just updated to it, which also includes updating eth-account to >=0.11.2. will that break anything @antazoey?

wakamex:
so the only change in this PR that isn't merged on main is adding the eth-tester requirement. let me even see if that's still required.

wakamex:
both `uv pip` and `pip` install eth-tester 0.11.0b2, so nothing needs to be done here.

this is due to web3[tester] having the following dependence: "eth-tester[py-evm]>=0.11.0b1,<0.12.0b1", which picks up 0.11.0b2 😀

PR #2004, opened 2024-04-15
perf: replace pip freeze with importlib.metadata

antazoey:
### What I did

another place we should replace pip freeze

### Checklist

### DIFF
--- a/src/ape/plugins/__init__.py
+++ b/src/ape/plugins/__init__.py
@@ -1,7 +1,6 @@
 import functools
 import importlib
-import pkgutil
-import subprocess
+from importlib.metadata import distributions
 from typing import Any, Callable, Generator, Iterator, List, Optional, Set, Tuple, Type
 
 from ape.__modules__ import __modules__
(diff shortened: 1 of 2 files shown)


src/ape/plugins/__init__.py
```diff
@@ -167,22 +166,9 @@ def registered_plugins(self) -> Set[str]:
...
-        }
-        core_plugin_module_names = {
-            n for _, n, ispkg in pkgutil.iter_modules() if n.startswith("ape_")
-        }
-
-        # NOTE: Returns tuple because this shouldn't change.
-        return tuple(installed_plugin_module_names.union(core_plugin_module_names))
+        plugins = [x.name.replace("-", "_") for x in distributions() if x.name.startswith("ape-")]
+        locals = [p for p in __modules__ if p != "ape"]
+        return tuple([*plugins, *locals])
```
fubuloubu, line 171: I think this works
```suggestion
        return tuple(*plugins, *locals)
```

fubuloubu (approved):

src/ape/plugins/__init__.py
antazoey: looks like this is not correct: 

```
    return tuple(*plugins, *locals)
E   TypeError: tuple expected at most 1 argument, got 12
```

src/ape/plugins/__init__.py
```diff
@@ -168,7 +168,11 @@ def _register_plugins(self):
         if self.__registered:
             return
 
-        plugins = [x.name.replace("-", "_") for x in distributions() if x.name.startswith("ape-")]
+        plugins = [
+            getattr(x, "name", "").replace("-", "_")
```
fubuloubu, line 172: might be a little clearer, since we know it won't execute this unless `x` has `.name`
```suggestion
            x.name.replace("-", "_")
```

antazoey: Made the change, but PyCharm still highlights it for some reason, throws me off.

Screenshot 2024-04-16 at 11 14 24

fubuloubu (approved):

PR #2006, opened 2024-04-15
fix: issue with kwargs for get_logs using eth-tester

antazoey:
### What I did

somehow the latest web3.py causes intermittent failures with eth-tester get_logs kwargs.
this should take care of it

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ setup(
         "click>=8.1.6,<9",
         "ijson>=3.1.4,<4",
         "importlib-metadata",  # NOTE: Needed on 3.8 for entry_points `group=` kwarg.
-        "ipython>=8.23.0,<9",
+        "ipython>=8.5.0,<9",
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
         "pandas>=1.3.0,<2",
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ setup(
         "click>=8.1.6,<9",
         "ijson>=3.1.4,<4",
         "importlib-metadata",  # NOTE: Needed on 3.8 for entry_points `group=` kwarg.
-        "ipython>=8.5.0,<8.23.0",  # TODO: Fix 8.23.0
+        "ipython>=8.23.0,<9",
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
         "pandas>=1.3.0,<2",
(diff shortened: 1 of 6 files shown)


antazoey:
Conflicting dependency with eth-abi and eip712, introduced here https://github.com/ApeWorX/ape/pull/2002
I have no idea how Ape was able install in CI? 😕

antazoey:
required https://github.com/ApeWorX/eip712/pull/49 i guess

antazoey:
what is going on?!?! cmon

setup.py
fubuloubu, line 1: Update eip712 pin?

src/ape_ethereum/ecosystem.py
```diff
@@ -935,14 +935,14 @@ def get_abi(_topic: HexStr) -> Optional[LogInputABICollection]:
                     converted_arguments[key] = value
 
             yield ContractLog(
-                block_hash=log["blockHash"],
-                block_number=log["blockNumber"],
+                block_hash=log.get("blockHash", log.get("block_hash")),
+                block_number=log.get("blockNumber", log.get("block_number")),
```
fubuloubu, line 939: To be a little lazier
```suggestion
                block_hash=log.get("blockHash") or log.get("block_hash"),
                block_number=log.get("blockNumber") or log.get("block_number"),
```

fubuloubu (dismissed):

src/ape_ethereum/ecosystem.py
antazoey: i like some of this but maybe not for the ints, as it would do `0 or None` and then be `None` instead of `0`.

antazoey: or could do  a final `or 0`

antazoey: and transaction index can be None so, also will keep that in mind

fubuloubu: what scenario can it be none? when it's pending block?

antazoey: oh, I don't know, I was only looking at the type

antazoey: yes, pending block, according to the doc str: https://github.com/ApeWorX/ape/blob/main/src/ape/types/__init__.py#L279

fubuloubu (approved):

fubuloubu (approved):

fubuloubu:
Looks like test failure is not detecting scripts anymore

antazoey:
somehow there is a weird package called `scripts` in site packages that has now broken our scripting integration

antazoey:
yet, `sys.modules.get("scripts")` is `None`.

antazoey:
This also happens to me on `main` in a fresh env.

antazoey:
Pretty sure i found the culprit: https://github.com/ethereum/web3.py/tree/main/web3/scripts

Issue #2007, opened 2024-04-15
Embeded structures can't be encoded


pandadefi:
### Environment information

- OS: macOS
- Python Version: 3.10.12
- `ape` and plugin versions: 0.7.13

### What went wrong?

Here is a simple contract that contains a structure within a structure the inside structure contains arrays.

```solidity
contract SwapWrapper {
    enum Version {
        V1,
        V2,
        V2_1
    }

    struct Path {
        uint256[] pairBinSteps;
        Version[] versions;
        address[] tokenPath;
    }
    struct SwapData {
        address tokenIn;
        uint256 amountIn;
        uint256 minAmountOut;
        Path path;
    }

    function check(SwapData memory data) external view returns (bool) {
        return true;
    }
}
```

When running this relatively simple script:

```python
from ape import Contract, accounts, chain, project

def test_swap():
    user = accounts["0x1100000000000000000000000000000000000011"]
    swaper = user.deploy(project.SwapWrapper)
    wavax = "0xB31f66AA3C1e785363F0875A1B74E27b85FD66c7"
    usdc = "0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E"
    path = {
        "pairBinSteps": [0],
        "versions": [0],
        "tokenPath": [wavax, usdc]
    }
    swap_data = {
        "tokenIn": wavax,
        "amountIn": 10**18,
        "minAmountOut": 10**7,
        "path": path
    }
    swaper.check(swap_data)
```

The python execution will stop with the following error:

```
ERROR: (ConversionError) Value '{'pairBinSteps': [0], 'versions': [0], 'tokenPath': ['0xB31f66AA3C1e785363F0875A1B74E27b85FD66c7', '0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E']}' must be a list or tuple when given multiple types.
``` 
### How can it be fixed?

Encoding should not crash the execution.

antazoey:
Notes so far:

* It seems like the network is `polygon-fork`, based on the contents of the test.
* `swaper = user.deploy(project.SwapWrapper)` is failing for me out-the-gate with a `ContractNotFoundError` which is extremely weird.

pandadefi:
The network is avalanche mainnet fork.

antazoey:
Another issue:

* User exceeds balance.

[...]

PR #2009, opened 2024-04-16
fix: issue encoding nested structs in calldata

antazoey:
### What I did

fixes: #2007 

### How I did it

Missing recursion

### How to verify it

The test in the linked issue works now.

### Checklist

### DIFF
--- a/src/ape/utils/abi.py
+++ b/src/ape/utils/abi.py
@@ -91,7 +91,14 @@ class StructParser:
             and not isinstance(value, tuple)
         ):
             if isinstance(value, dict):
-                return tuple([value[m.name] for m in _type.components])
+                return tuple(
+                    (
+                        self._encode(m, value[m.name])
+                        if isinstance(value[m.name], dict)
+                        else value[m.name]
+                    )
+                    for m in _type.components
+                )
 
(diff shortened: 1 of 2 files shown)


src/ape/utils/abi.py
```diff
@@ -91,7 +91,14 @@ def _encode(self, _type: ABIType, value: Any):
             and not isinstance(value, tuple)
         ):
             if isinstance(value, dict):
-                return tuple([value[m.name] for m in _type.components])
+                return tuple(
```
antazoey, line 94: the fix is here

src/ape/managers/converters.py
```diff
@@ -284,42 +288,42 @@ def convert(self, value: Any, type: Union[Type, Tuple, List]) -> Any:
 
         Args:
             value (any): The value to convert.
-            type (type): The type to convert the value to.
```
antazoey, line 287: incredibly hard to debug when builtin `type` is overridden with something else

antazoey: ugh -- i guess will save for 0.8, but no need to use kwarg= on args :(

fubuloubu (approved):

PR #2020, opened 2024-04-22
fix: use macos-12 instead of macos-latest in CI

antazoey:
### What I did

The following issue seems to block python3.8-3.9 for macos arm: https://github.com/actions/setup-python/issues/808

NOTE: Anticipating things may still fail because of gacts/install-geth action

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -62,7 +62,9 @@ jobs:
 
         strategy:
             matrix:
-                os: [ubuntu-latest, macos-latest]   # eventually add `windows-latest`
+                # TODO: Replace with macos-latest when works again.
+                #   https://github.com/actions/setup-python/issues/808
+                os: [ubuntu-latest, macos-12]   # eventually add `windows-latest`
                 python-version: [3.8, 3.9, "3.10", "3.11"]
 
         env:


fubuloubu (approved):

wakamex:
this user suggests getting rid of the setup-python action, in favor of manually installing python and pip (though i'd suggest `uv pip`). it seems to work for macos-13 as well. https://github.com/actions/setup-python/issues/696#issuecomment-1994554432

antazoey:
> actions/setup-python#696 (comment)

We would have to separate `mac-os` from `ubuntu` more if we went this route. We have test.yml able to run on both, with the OS being in a matrix.

antazoey:
> this user suggests getting rid of the setup-python action, in favor of manually installing python and pip (though i'd suggest `uv pip`). it seems to work for macos-13 as well. [actions/setup-python#696 (comment)](https://github.com/actions/setup-python/issues/696#issuecomment-1994554432)

because of the refactor required and my optimism 808 will be resolved, I want to stick to having macos in the matrix and using setup-pyhon... but will wait until for more thoughts before merging this.

wakamex:
you could get rid of setup-python entirely with a step like this (not saying this is the way to go, just an option, maybe if the action keeps giving you problems):

```
      - name: Install dependencies
        run: |
          if [ "$RUNNER_OS" == "Linux" ]; then
            apt update
            apt install -y python${{ matrix.python-version }}
          elif [ "$RUNNER_OS" == "macOS" ]; then
            brew update
            brew install python@${{ matrix.python-version }}
          fi
```

antazoey:
> run: |
>           if [ "$RUNNER_OS" == "Linux" ]; then
>             apt update
>             apt install -y python${{ matrix.python-version }}
>           elif [ "$RUNNER_OS" == "macOS" ]; then
>             brew update
>             brew install python@${{ matrix.python-version }}
>           fi

eventually would also add windows here.

antazoey:
it feels like we are re-inventing wha setup-python is supposed to do... setup-python should be fixed, that seems like fixing it in the right spot.

PR #2021, opened 2024-04-22
fix: issues with local installs

antazoey:
### What I did

issue where plugin install checking failed because of local dists w/o name attribute

i think related to local or editable installs of packages...

### Checklist

### DIFF
--- a/src/ape/plugins/_utils.py
+++ b/src/ape/plugins/_utils.py
@@ -348,13 +348,7 @@ class PluginMetadata(BaseInterfaceModel):
         if not use_cache:
             _get_distributions.cache_clear()
 
-        for dist in _get_distributions():
-            if not (name := getattr(dist, "name", "")):
-                continue
-            elif self.package_name == name:
-                return True
-
-        return False
+        return any(self.package_name == getattr(d, "name", "") for d in _get_distributions())
 
     def _prepare_install(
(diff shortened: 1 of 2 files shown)


src/ape/plugins/_utils.py
```diff
@@ -344,12 +344,17 @@ def __str__(self):
...
+    def check_installed(self, use_cache: bool = True) -> bool:
         if not use_cache:
             _get_distributions.cache_clear()
 
-        ape_packages = [n.name for n in _get_distributions()]
-        return self.package_name in ape_packages
+        for dist in _get_distributions():
+            if not (name := getattr(dist, "name", "")):
+                continue
+            elif self.package_name == name:
```
fubuloubu, line 354: Could we reduce this one leg?
```suggestion
            if self.package_name == getattr(dist, "name", ""):
```

src/ape/plugins/_utils.py
```diff
@@ -344,12 +344,17 @@ def __str__(self):
...
+                return True
+
+        return False
```
fubuloubu, line 357: Or even just 
```suggestion
        return any(getattr(dist, "name", "") == self.package_name for dist in _get_distributions())
```
and delete the above

fubuloubu (dismissed):

PR #2024, opened 2024-04-25
docs: add doctesting to userguides

dtdang:
### What I did
Add doctesting to the documentation userguides. 

fixes: #657 

### Checklist


docs/conf.py
```diff
@@ -84,7 +85,23 @@
...
+doctest_default_flags = (
+    0
+    | doctest.DONT_ACCEPT_TRUE_FOR_1
+    | doctest.ELLIPSIS
+    | doctest.IGNORE_EXCEPTION_DETAIL
+    | doctest.NORMALIZE_WHITESPACE
+)
+
+doctest_global_setup = '''
+from ape import accounts
```
antazoey, line 101: we likely want to import all the ape top-level stuff here.

docs/conf.py
```diff
@@ -84,7 +85,23 @@
...
+
+my_account_alias = accounts.test_accounts[0]
```
antazoey, line 103: question: are we able to have individual test setups? something like this doesnt need to be global
issue: i commented elsewhere, but this is supposed to be an alias, not an account, and i think itll likely always fail, ideally we ignore it because it is like a "task failed successfully" kinda moment. but otherwise, you could try `my_account_alias = accounts.test_accounts[0].alias`

docs/userguides/projects.md
```diff
@@ -46,19 +46,22 @@ For more information on compiling your project, see [this guide](./compile.html)
...
+...     account.deploy(project.MyContract)
 ```
 
 **NOTE**: You can also deploy contracts from the container itself:
 
 ```python
 
-account = accounts.load("my_account_alias")
+account = accounts.load(f"{my_account_alias}")
```
antazoey, line 64: this will likely always fail in a doc-test i think that is ok. is there a way to ignore all ApeExceptions?

docs/conf.py
dtdang: Yes we can do individual test setups. I was trying this out to test if it would be nicer for this to have in a global setup but I'll switch it to individual since it's currently not being used for more than one test.

docs/userguides/projects.md
dtdang: Yes currently this is failing but if `my_account_alias = accounts.test_accounts[0].alias` works in the testsetup then I can try and use that instead. I'll look into ignoring ApeExceptions since that would be also useful in a few other userguides.

docs/conf.py
fubuloubu: `from ape import *`?

fubuloubu:
side note: makes it a little harder to review since github does not render the markdowns as well, would be helpful to see a generated html file

docs/conf.py
antazoey: that works!

antazoey:
> side note: makes it a little harder to review since github does not render the markdowns as well, would be helpful to see a generated html file

you just have to build the docs locally and look at them! we could have docs get build and deployed on PRs, maybe by special command, or maybe if it detect changes to any docs file... that is cool but a bit over-the-top imo, just build them `python build_docs.py` and then `pythom -m something something` it is in the contrib.

fubuloubu:
> > side note: makes it a little harder to review since github does not render the markdowns as well, would be helpful to see a generated html file
> 
> you just have to build the docs locally and look at them! we could have docs get build and deployed on PRs, maybe by special command, or maybe if it detect changes to any docs file... that is cool but a bit over-the-top imo, just build them `python build_docs.py` and then `pythom -m something something` it is in the contrib.

punt this down the road, but would be nice to have a temporary render of each PR's docs updates to see it. can just deploy to a bucket

fubuloubu (approved):

docs/conf.py
```diff
@@ -101,8 +101,15 @@
...
 from ape import *
-"""
+from pathlib import Path
+from ethpm_types import ContractType
 
+contract_path = Path("tests/functional/data/contracts/ethereum/local/VyperContract.json")
+contract_json = contract_path.read_text()
+
+contract = ContractType.model_validate_json(contract_json)
+# project._cached_projects["MyContract"] = contract
```
antazoey, line 111: is this not needed?

```suggestion
```

docs/userguides/contracts.md
```diff
@@ -95,54 +95,64 @@ More information on publishing contracts can be found in [this guide](./publishi
...
-```python
-from ape import Contract
+```{eval-rst}
+
+.. doctest::
 
-contract = Contract("v2.registry.ychad.eth")
+    >>> from ape import Contract
+
+    >>> contract = Contract("v2.registry.ychad.eth")
```
antazoey, line 128: this will fail if `ape-ens` and `ape-etherscan` plugins are no installed, which i dont think they are in CI

docs/userguides/networks.md
```diff
@@ -474,12 +474,15 @@ with ape.networks.arbitrum.mainnet.use_default_provider():
...
-# Same as doing `networks.ethereum.local.use_provider("test")`.
-with networks.parse_network_choice("ethereum:local:test") as provider:
-    print(provider)
+.. doctest::
+>>> from ape import networks
+
+>>> # Same as doing `networks.ethereum.local.use_provider("test")`.
+>>> with networks.parse_network_choice("ethereum:local:test") as provider:
+...     print(provider.name)
+    test
```
antazoey, line 485: what is this final `test` here? (the word)

docs/conf.py
dtdang: It is necessary but I'm still working on storing the contract as a cached project to be used in other examples.

docs/userguides/networks.md
dtdang: The final `test` here is the output string that the doctest is expecting to compare to the results from the test above.

docs/userguides/contracts.md
dtdang: I will remove this doctest then.

PR #2025, opened 2024-04-25
refactor!: use `name` instead of `data_folder` for accounts

antazoey:
### What I did

Every account plugin does the same thing for data folder - it puts a name after the regular ape data folder.
This PR changes (0.8) the account API to require you to specify the name of the container rather than a data folder location. 
I think this is a lot more intuitive!

### Checklist


src/ape/api/accounts.py
```diff
@@ -401,15 +401,24 @@ class AccountContainerAPI(BaseInterfaceModel):
...
-    data_folder: Path
+    """
+    The name of the account container.
+    For example, the ``ape-ledger`` plugin
+    uses ``"ledger"`` as its name.
+    """
+    name: str
 
+    """
+    The type of account in this container.
```
antazoey, line 412: migrated some of the docs changes so we can get it in the next 0.7 release as well: https://github.com/ApeWorX/ape/pull/2026/files

src/ape/managers/accounts.py
```diff
@@ -46,8 +46,8 @@ def containers(self) -> dict[str, TestAccountContainerAPI]:
             t for t in self.plugin_manager.account_types if issubclass(t[1][1], TestAccountAPI)
         ]
         for plugin_name, (container_type, account_type) in account_types:
-            # Pydantic validation won't allow passing None for data_folder/required attr
-            containers[plugin_name] = container_type(data_folder="", account_type=account_type)
+            # Pydantic validation won't allow passing None for name/required attr
+            containers[plugin_name] = container_type(name="", account_type=account_type)
```
fubuloubu, line 50: Can this be like `name=f"<{plugin_name}>"` or something just to make sure if never overwrites anything?

fubuloubu (approved):

src/ape/managers/accounts.py
antazoey: This made me realize this needs change. Here is my plan, let me know what you think:

1. Make this `name=plugin_name` and always provide a container name. Make it _actually_ required.
2. Allow containers to change the data folder again but don't make it be required; have it default to `.ape/self.name`. But allowing it still to change is helpful in the case a plugin has multiple API classes; it could nest the accounts under `.ape/plugin_name/accounts` or something alike.

Doing it now and will link the commit, might be easier to see what I mean.

antazoey: OK tad different but: https://github.com/ApeWorX/ape/pull/2025/commits/0965c50a3a9a04a5f2568ba012e5214463b9d318

1. Uses plugin_name for name always, event test containers
2. for test containers, point data_folder at /dev/null... It is not used anywhere and this achieves the same idea and avoid typing AND pydantic concerns because it is still a path at least.

antazoey: and if any plugin can change this via overriding the property

antazoey: I think this works great - still allows overriding, better null-value for the type, no actual changes to any plugin besides the field change.

Let me know if have any other ideas and ill edit-rebase them in!

PR #2027, opened 2024-04-25
docs: fix guide example formatting

antazoey:
### What I did

mdformat seems to not like this (sometimes?). Idk it keeps changing in my PRs so I am fixing it a way that is more like the other examples so it is better.

### Checklist


antazoey:
note: i realize you may changes these again during doc tests ; this is maybe a temp change beforehand

dtdang (approved):

dtdang:
It's a good change to add doctest to afterwards otherwise I wouldn't be able to add doctest with the previous format.

PR #2029, opened 2024-04-25
refactor!: use `ContractSource` type requirement

antazoey:
### What I did

a `ContractSource` is required anyway, so this changes the type to match. So it is better because it would exit early if for some reason missing some source info and also helps with the project-refactor.

basically, you can't trace a source without a source...

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -243,13 +243,14 @@ class Receipt(ReceiptAPI):
     @cached_property
     def source_traceback(self) -> SourceTraceback:
         if contract_type := self.contract_type:
-            try:
-                return SourceTraceback.create(contract_type, self.trace, HexBytes(self.data))
-            except Exception as err:
-                # Failing to get a traceback should not halt an Ape application.
-                # Sometimes, a node crashes and we are left with nothing.
-                logger.error(f"Problem retrieving traceback: {err}")
-                pass
+            if contract_src := self.project_manager._create_contract_source(contract_type):
+                try:
+                    return SourceTraceback.create(contract_src, self.trace, HexBytes(self.data))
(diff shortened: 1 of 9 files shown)


antazoey:
> ### What I did
> a `ContractSource` is required anyway, so this changes the type to match. So it is better because it would exit early if for some reason missing some source info and also helps with the project-refactor.

basically, you can't trace a source without a source... 
> 
> ### How I did it
> ### How to verify it
> ### Checklist
> * [ ]  All changes are completed
> * [ ]  New test cases have been added
> * [ ]  Documentation has been updated

fubuloubu:
Is this based off sone other large set of changes (or wrong base branch)?

antazoey:
> Is this based off sone other large set of changes (or wrong base branch)?

dang rebasing messed it up

src/ape_node/query.py
```diff
@@ -1,5 +1,6 @@
+from collections.abc import Iterator
```
fubuloubu, line 1: is this not in typing?

src/ape_ethereum/transactions.py
```diff
@@ -243,13 +243,14 @@ def method_called(self) -> Optional[MethodABI]:
     @cached_property
     def source_traceback(self) -> SourceTraceback:
```
fubuloubu, line 244: should this instead be optional, and return None if it can't?

fubuloubu (approved):

src/ape_ethereum/transactions.py
antazoey: I think it always can, for the most part. If you don't have tracing, you still get some top-level data.
However, it does fail now and then for some random node-related reasons on providers _with_ tracing support, hence the handling

PR #2030, opened 2024-04-26
refactor: use manager._make_request instead of provider

mikeshultz:
### What I did

Refactored to use `_make_request()` on the web3.py manager rather than `make_request()` on the web3.py provider.  This makes sure middleware are run.

I'm curious what anyone else thinks about this.  It's a private method I'm leveraging (not ideal) but skipping middleware *probably* isn't the play either.

Another option might be to just implement the internals of this function which applies middleware and calls `request_func()`.  I don't have any strong feelings, this PR is mainly the most minimal change to do what I'm looking for.  Maybe we don't even want to leverage any of the middleware and I'm off base here anyway.

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -68,7 +68,7 @@ from ape.types import (
     SourceTraceback,
     TraceFrame,
 )
-from ape.utils import gas_estimation_error_message, run_until_complete, to_int
+from ape.utils import gas_estimation_error_message, to_int
 from ape.utils.misc import DEFAULT_MAX_RETRIES_TX
 from ape_ethereum._print import CONSOLE_CONTRACT_ID, console_contract
 from ape_ethereum.transactions import AccessList, AccessListTransaction
@@ -230,7 +230,7 @@ class Web3Provider(ProviderAPI, ABC):
         if self._web3 is None:
             return False
 
-        return run_until_complete(self._web3.is_connected())
+        return self._web3.is_connected()
 
     @property
     def max_gas(self) -> int:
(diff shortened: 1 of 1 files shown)


mikeshultz:
Well, the failing tests suggest there's a bit more to this than a 1:1 change...

antazoey:
idea: could we keep it as provider and use our middleware in two spots: one in web3 middleware and one here in our own `make_request()` ?

mikeshultz:
> idea: could we keep it as provider and use our middleware in two spots: one in web3 middleware and one here in our own `make_request()` ?

Yeah, I think we can re-implement [essentially what `web3.provider._make_request()` is doing](https://github.com/ethereum/web3.py/blob/3229a5e42cfd63ad8136a470ad4534ff8b0a7460/web3/manager.py#L286-L291) in our ape provider.  It's essentially what this PR is doing anyway.

src/ape_ethereum/provider.py
```diff
@@ -1048,8 +1048,7 @@ def _create_trace_frame(self, evm_frame: EvmTraceFrame) -> TraceFrame:
 
     def _make_request(self, endpoint: str, parameters: Optional[List] = None) -> Any:
         parameters = parameters or []
-        coroutine = self.web3.provider.make_request(RPCEndpoint(endpoint), parameters)
-        result = run_until_complete(coroutine)
+        result = self.web3.provider.make_request(RPCEndpoint(endpoint), parameters)
```
mikeshultz, line 1051: thought: this is why I hate rebasing.  Always fucks up history.

mikeshultz:
Blergh, I don't know if I want to continue down this path.  So many tests mock out make_request which is apparently no longer leveraged with this change.  And every plugin provider probably dabbles in this code a bit too requiring a refactor there as well and breaking changes all around. 

I'm already spending too much time on this so I'm going to just bail and see if I can just put the last touches on that other PR in ape-ethereum.

PR #2037, opened 2024-04-29
fix: issue with selecting first network from an ecosystem as default

antazoey:
### What I did

When an ecosystem has no default network, the logic says to pick the first network. However, you get a `KeyError` in that case because it mistakenly tries `0` instead of the the first key.

### How I did it

Use `next` to get the key and use that instead of `0`.

### How to verify it

Install an ecosystem plugin with no `local` network and have the default be unset in the config.
Try it out.
Notice KeyError.

Try with this PR
Notice no KeyError and it actually selects the correct one.

Also added tests.

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -320,7 +320,7 @@ class EcosystemAPI(ExtraAttributesMixin, BaseInterfaceModel):
             str
         """
         if network := self._default_network:
-            # Was set programatically.
+            # Was set programmatically.
             return network
 
         elif network := self.config.get("default_network"):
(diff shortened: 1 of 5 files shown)


src/ape/api/networks.py
```diff
@@ -333,7 +333,8 @@ def default_network_name(self) -> str:
 
         elif len(self.networks) >= 1:
             # Use the first network.
-            return self.networks[0]
+            key = next(iter(self.networks.keys()))
```
fubuloubu, line 336: What python version made dict key order not random?

fubuloubu (dismissed):

src/ape/api/networks.py
antazoey: 3.6

fubuloubu (dismissed):

PR #2039, opened 2024-04-29
feat: add `--config-override` option for `ape pm install`

antazoey:
### What I did

fixes: #

### How I did it

feat: adds new `config_override_option` and `JSON` CLI param type
fix: when decoding a dependency, checks config override for contracts folder instead of using default
(only a problem when contracts/ exists but changing it to something else anyway)
chore: move `@ape_cli_conext` to top of every cli impl usage list of decors, as a better standard for its placement.
feat: allow package callback in `ape pm` command to work if given a Path (only happens during testing)
test(refactor): for all usage of `runner.invoke` use either `str` or `tuple` if possible for arguments. `str` works for single commands and `tuple` works if you are not building up commands... less chance of mutability accidents and is just better practice.

### Checklist

### DIFF
--- a/src/ape_pm/_cli.py
+++ b/src/ape_pm/_cli.py
@@ -1,10 +1,10 @@
 import json
 from pathlib import Path
-from typing import Tuple
+from typing import Dict, Tuple
 
 import click
 
-from ape.cli import ape_cli_context
+from ape.cli import ape_cli_context, config_override_option
 
 
 @click.group()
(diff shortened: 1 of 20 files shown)


src/ape_pm/_cli.py
```diff
@@ -161,14 +170,19 @@ def _package_callback(ctx, param, value):
...
+def install(cli_ctx, package, name, version, ref, force, config_override):
     """
     Download and cache packages
     """
 
     log_name = None
 
     if not package or package == ".":
+        if config_override:
+            # NOTE: Handled correctly on feat/0.8
```
fubuloubu, line 183: Should be a TODO to remove?

fubuloubu (dismissed):

PR #2042, opened 2024-05-01
chore: update dockerfile to python 3.11

mikeshultz:
### What I did

Updates the `Dockerfile` python version to 3.11 because features.

Any reason not to do this?  Will this roll the docker hub versions automagically?

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # See LICENSE in the project root for license information.
 #---------------------------------------------------------------------------------------------
 
-ARG PYTHON_VERSION="3.10"
+ARG PYTHON_VERSION="3.11"
 ARG PLUGINS_FILE="./recommended-plugins.txt"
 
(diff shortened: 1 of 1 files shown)


fubuloubu:
> Will this roll the docker hub versions automagically?

Dockerhub currently updates `latest` tag to whatever `main` is, and then makes a tagged that mirrors any github release made (e.g. `v0.7.11`) and slaps `stable` on whatever is the last release

I have been considering to migrate to GHCR because of the recent changes with Dockerhub, would also be nice to have it all in one place as well

fubuloubu (approved):

fubuloubu:
WIll note that I do not believe that the docker images build until they hit `main`, so just verify it locally and we can merge

mikeshultz:
> WIll note that I do not believe that the docker images build until they hit `main`, so just verify it locally and we can merge

I did verify that it builds and runs locally.

PR #2044, opened 2024-05-01
fix: `ape plugins update` command type error

antazoey:
### What I did

fixes a bug introduced here: https://github.com/ApeWorX/ape/commit/5eeec421af32ed5a34e65e731f4401bdbc370205#diff-0b86928e81aa95cc356070c25d73b92777562a9465597bf82b802e00736a1f96

where `ape plugins update` (for upgrading full ape suite) was failing because it was looking through dists instead of plugin names.

### How I did it

use `_filter_plugins_from_dists`

### Checklist

### DIFF
--- a/src/ape_plugins/_cli.py
+++ b/src/ape_plugins/_cli.py
@@ -15,6 +15,7 @@ from ape.plugins._utils import (
     PluginMetadata,
     PluginMetadataList,
     PluginType,
+    _filter_plugins_from_dists,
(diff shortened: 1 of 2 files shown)


antazoey:
a TODO here: have `update` command only try upgrading if it realizes the next version range has an available. Else have it use the max version in the current range and still upgrade all the plugins.

dtdang (approved):

Issue #2035, opened 2024-04-27
Catch and re-throw w/ `CustomError` if `ErrorABI` match detected


fubuloubu:
### Overview

If an error with a contract is encoded using Solidity's custom error feature, currently ape displays this message:
```
ERROR: (ContractLogicError) custom error e0b4aa5a:00000000000000000000000000000000…00000000000000000000000000030931 (64 bytes)
```

This is not enough to actually be able to decode into a `CustomError` object (available from `ContractInstance._errors_`)

Further, if the data is available, there is not an easy way to decode that exception from bytes back into the custom error type

### Specification

Add method to catch a contract logic error and instead generate a `CustomError` type that gets re-thrown

### Dependencies

This may impact/simplify the custom error reverts checking logic that the testing-only reverts manager employs

antazoey:
We should be doing this alredady. What provider are you using?

fubuloubu:
@wavey0x or @fp-crypto had this issue I think

antazoey:
> @wavey0x or @fp-crypto had this issue I think

I try to repro tomorrow with providers i know they use but otherwise that is the most desired info I need as each provider shows the hex different and we have to make sure we are passing expected-styles to `compiler.enrich_error` for it to work, lots of players in this game!

antazoey:
I have tried on:

* foundry
* eth tester
* geth (dev)
* hardhat

using solidity v0.8.24... and I am getting custom errors just fine.

more info please :))) Ill keep trying to repro.

banteg:
here we encode `WRAP_ETH(dev, 1 ether)`. the first one should work, the second should fail with `InsufficientETH()`.

```python
from ape import Contract, accounts

def main():
    universal_router = Contract('0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD')
    dev = accounts.test_accounts[0]
    for value in [10**18, 0]:
        receipt = universal_router.execute('0x0b', ['0x000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb922660000000000000000000000000000000000000000000000000de0b6b3a7640000'], sender=dev, value=value)
        receipt.show_trace()
```

```shell
ape run bug --network :mainnet-fork
```

PR #2047, opened 2024-05-01
fix: issue when contracts w/o source IDs would not enrich

antazoey:
### What I did

fixes: #2035 

Enrichment was skipped for contracts missing source IDs. This adds it.

bonuses that are not needed but were still part of the fun:

* https://github.com/ApeWorX/ape-foundry/pull/100
* https://github.com/ApeWorX/ape-solidity/pull/141

TODO: Refactor CompilerAPI plugins to use new helper method (instead of re-implementing n-places) `compiler_manager.get_custom_error()`

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -614,7 +614,7 @@ class EcosystemAPI(ExtraAttributesMixin, BaseInterfaceModel):
     def decode_custom_error(
         self,
         data: HexBytes,
-        contract_type: ContractType,
+        address: AddressType,
         **kwargs,
     ) -> Optional[CustomError]:
         """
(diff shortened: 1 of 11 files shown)


src/ape/managers/compilers.py
```diff
@@ -321,28 +322,72 @@ def enrich_error(self, err: ContractLogicError) -> ContractLogicError:
         Returns:
             :class:`~ape.exceptions.ContractLogicError`: The enriched exception.
         """
+        # First, try enriching using their ABI.
+        if custom_err := self.get_custom_error(err):
+            return custom_err
```
fubuloubu, line 327: should this immediately return? or should it do `err = custom_err`?

fubuloubu: note that `CustomError` is a subclass of `ContractLogicError`

fubuloubu (approved):

banteg:
so in my initial tests it shows the correct error at the end

image

but in the trace it's still `reverted without message`

image

moreover, if an error encodes some additional details, it doesn't get caught at all. here is an updated test. 

```python
def test_custom_error_2(accounts):
    universal_router = Contract("0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD")
    for value in [10**18, 0]:
        receipt = universal_router.execute(
            "0x0b00",
            [
                "0x0000000000000000000000003fc91a3afd70395cd496c647d5a6cc9d4b2b7fad0000000000000000000000000000000000000000000000000de0b6b3a7640000",
                "0x000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb922660000000000000000000000000000000000000000000000000de0b6b3a7640000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002bc02aaa39b223fe8d0a0e5c4f27ead9083c756cc20001f4a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000000000000000000000",
            ],
            sender=accounts[0],
            value=value,
        )
        receipt.show_trace()
```

expected `AllowanceExpired(uint256)` that encodes `0`

actual
```python
E   ape.exceptions.ContractLogicError: 0xd81b2f2e0000000000000000000000000000000000000000000000000000000000000000
```

antazoey:
inputs should work! 

but it looks like it is struggling to find the `AllowanceNotAccepted` ABI - it is not on the contract for some reason.

src/ape_ethereum/provider.py
```diff
@@ -130,8 +130,8 @@ def send_tx_wrapper(self, txn: TransactionAPI) -> ReceiptAPI:
         return super().__new__(cls)  # pydantic v2 doesn't want args
 
     def __init__(self, *args, **kwargs):
-        logger.create_logger("web3.RequestManager", handlers=(_sanitize_web3_url,))
-        logger.create_logger("web3.providers.HTTPProvider", handlers=(_sanitize_web3_url,))
+        # logger.create_logger("web3.RequestManager", handlers=(_sanitize_web3_url,))
+        # logger.create_logger("web3.providers.HTTPProvider", handlers=(_sanitize_web3_url,))
```
fubuloubu, line 134: trying something here?

fubuloubu (approved):

antazoey:
OK - custom errors appear in traces now..
The only thing left is figuring out why some contract's are missing ABIs, will fix that on another PR though

src/ape_ethereum/transactions.py
```diff
@@ -276,17 +282,35 @@ def show_trace(self, verbose: bool = False, file: IO[str] = sys.stdout):
...
-            else:
+                # Extra revert-message
                 decoded_result = decode(("string",), returndata[4:])
                 if len(decoded_result) == 1:
                     revert_message = f'reverted with message: "{decoded_result[0]}"'
                 else:
                     revert_message = default_message
 
+            else:
+                # Try to enrich revert error using ABI.
```
antazoey, line 296: note: this will be much easier and more sensical in 0.8 with the trace API

src/ape/api/networks.py
```diff
@@ -610,6 +611,65 @@ def get_python_types(  # type: ignore[empty-body]
...
+            **kwargs: Additional init kwargs for the custom error class.
+
+        Returns:
+            Optional[CustomError]: If it able to decode one, else ``None``.
+        """
+
+        # Use an instance (required for proper error caching).
+        contract = self.chain_manager.contracts.instance_at(address)
+
+        selector = data[:4]
```
fubuloubu, line 637: Probably specific to ethereum, but fine for now until someone else needs a different mechanism to handle it (just add a comment)

fubuloubu (dismissed):

src/ape/api/networks.py
antazoey: I was thinking of implementing this there

fubuloubu: Don't think it's a huge deal now, probably all EVM networks handle it this way, but a note to come back and refactor into EcosystemAPI would be nice

antazoey: i just went ahead and moved it! wasnt too bad

PR #2048, opened 2024-05-02
fix: issue where could not have dependencies with `contracts_folder` pointed at the root project path

antazoey:
### What I did

fixed a bug where could not compile dependencies like:

```yaml
dependencies:
  - name: b64
    github: Brechtpd/base64
    ref: main
    contracts_folder: .
```

issue: it got turned to "" via a bad replace during a config manifestation.

### How I did it

if the paths re equal, make the temp config value for contracts_folder be "."

**NOTE**: This all makes so much more sense on the project refactor for 0.8 .... just sayin!  But this is a solid fix for the 0.7 range I think.

**ALSO**: I purposely made the test an integration test to ensure this carries over to 0.8 because any functional test would likely get erased (so I didn't add any of those).

### How to verify it

Use this config:

```yaml
dependencies:
  - name: b64
    github: Brechtpd/base64
    ref: main
    contracts_folder: .
```

and run:

```sh
ape pm install .
```

### Checklist


src/ape/managers/project/types.py
```diff
@@ -150,9 +150,17 @@ def process_config_file(self, **kwargs) -> bool:
             config_data["version"] = self.version
 
         contracts_folder = kwargs.get("contracts_folder") or self.contracts_folder
-        contracts_folder_config_item = (
-            str(contracts_folder).replace(str(self.path), "").strip(os.path.sep)
```
antazoey, line 154: The bug was this part

fubuloubu (approved):

PR #2051, opened 2024-05-03
fix: handle `None` blob-receipt gas values

antazoey:
### What I did

Running into this problem here: https://github.com/ApeWorX/ape-foundry/actions/runs/8939795074/job/24556545952?pr=101

It seems like `None` should be allowed.

Looking at this PR: https://github.com/foundry-rs/foundry/pull/7242/files, they also seem to allow `nil` here.

### Checklist

### DIFF
--- a/src/ape_ethereum/transactions.py
+++ b/src/ape_ethereum/transactions.py
@@ -466,12 +466,12 @@ class SharedBlobReceipt(Receipt):
     blob transaction.
     """
 
-    blob_gas_used: int
+    blob_gas_used: Optional[int] = None
     """
     The total amount of blob gas consumed by the transactions within the block.
     """
 
-    blob_gas_price: int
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/transactions.py
```diff
@@ -461,10 +461,26 @@ def _decode_ds_note(self, log: Dict) -> Optional[ContractLog]:
 
 
 class SharedBlobReceipt(Receipt):
+    """
+    An `EIP-4844 <https://eips.ethereum.org/EIPS/eip-4844#blob-transaction>`__"
+    blob transaction.
+    """
+
     blob_gas_used: int
```
antazoey, line 469: alternatively i could make one or both of these optional.
i really dont know, ive been reading through the EIP but just really know

mikeshultz (approved):
LGTM

PR #2063, opened 2024-05-07
fix: raise error when env var WEB3_PROVIDER_URI set

antazoey:
### What I did

Env var WEB3_PROVIDER_URL is a special env var in web3.py basically saying _use this RPC URL no matter what_.
So, even though Ape passes in a new one, Web3.py tried to use the one from the env var.

This caused quite a bit of confusion for a good chunk of yesterday morning!
this PR attempts to address this problem by ....

### How I did it

... raising a ProviderError 

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -96,6 +96,34 @@ def _sanitize_web3_url(msg: str) -> str:
     return f"{prefix} URI: {sanitized_url} {' '.join(rest[1:])}"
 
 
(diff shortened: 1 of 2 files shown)


antazoey:
@milkyklim

antazoey:
I was considering trying to allow this to work when the networks do match, and maybe we could do that later? Open to ideas... something has to happen though...

fubuloubu (approved):

PR #2073, opened 2024-05-08
refactor!: In CompilerAPI, use Iterable and Iterator instead of Sequence and Dict

antazoey:
### What I did

fixes: #660 

Part of the project refactor but wanted it separate for more thorough review

### Checklist

### DIFF
--- a/src/ape/managers/project/types.py
+++ b/src/ape/managers/project/types.py
@@ -265,7 +265,13 @@ class BaseProject(ProjectAPI):
                 self._cached_manifest.contract_types = filtered_contract_types
                 self._contracts = filtered_contract_types
 
-            return {ct.name: ct for ct in self.compiler_manager.compile(srcs_to_compile) if ct.name}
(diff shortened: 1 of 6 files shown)


src/ape/managers/project/types.py
```diff
@@ -265,7 +265,7 @@ def _compile_sources(proj_srcs: _ProjectSources) -> dict[str, ContractType]:
                 self._cached_manifest.contract_types = filtered_contract_types
                 self._contracts = filtered_contract_types
 
-            return self.compiler_manager.compile(srcs_to_compile)
+            return {ct.name: ct for ct in self.compiler_manager.compile(srcs_to_compile) if ct.name}
```
fubuloubu, line 268: If there isn't a name defined, it will skip?

Also, would it be better to do `ct.name is not None` (I think it can be `None`)

fubuloubu (approved):

src/ape/managers/project/types.py
antazoey: 81ef891

PR #2079, opened 2024-05-08
fix: issue where `err.address` would be zero address sometimes

antazoey:
### What I did

fixes: #

### Checklist

### DIFF
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -204,6 +204,17 @@ class TransactionError(ApeException):
 
     @property
     def address(self) -> Optional["AddressType"]:
+        if addr := self.contract_address:
+            return addr
+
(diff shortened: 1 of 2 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -1190,7 +1191,10 @@ def decode_custom_error(
         **kwargs,
     ) -> Optional[CustomError]:
         # Use an instance (required for proper error caching).
-        contract = self.chain_manager.contracts.instance_at(address)
+        try:
+            contract = self.chain_manager.contracts.instance_at(address)
+        except Exception:
+            return None
```
mikeshultz, line 1197: suggestion: since it's a blanket catch, would it be worth loging the error?

Issue #2075, opened 2024-05-08
post_send_transaction hook crashes on base sepolia


banteg:
### Environment information

- OS: macOS / linux / win
- Python Version: x.x.x
- `ape` and plugin versions:

```sh
$ ape --version
# ...copy and paste result of above command here...

$ ape plugins list
# ...copy and paste result of above command here...
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
# ...copy and paste result of above command here...
```

### What went wrong?

```python
from ape import project, accounts, networks, Contract

with networks.parse_network_choice('base:sepolia'):
    dev = accounts.load('sepolia')
    token = Contract("0x5E59Ff9f8F33bA898906aEB73E8c206072929760")
    tx = token.mint(dev, '100 ether', sender=dev)
```

after the tx is confirmed, i get this error. i tracked it down to here https://github.com/ApeWorX/ape/blob/aa1f16d2ab7a99ff542031c83aedaf842b017386/src/ape_ethereum/provider.py#L153

<img width="779" alt="Screenshot 2024-05-08 at 22 07 45" src="https://github.com/ApeWorX/ape/assets/4562643/f46a5e9a-2002-4b34-b8ea-bcc22978347a">

```python
requests.exceptions.HTTPError: 405 Client Error: Method Not Allowed for url: https://api.developer.coinbase.com/rpc/v1/base-sepolia/7FGRqMtFpg_aMODEEPxhetYLxZhgozP8
```

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

PR #2080, opened 2024-05-09
fix: handle HTTPError-based not-implemented RPCs and parity-trace detection error handling easement

antazoey:
### What I did

fixes: https://github.com/ApeWorX/ape/issues/2075

### How I did it

fixes in 2 spots:

1. detect APINotImplemented, which is the proper fix
2. when an unknown exception happens while  testing for parity, mark it as non-parity and go for geth. this should prevent future problems like this more.

### Checklist


antazoey:
while this my fix the issue, id like to fully investigate more. tbd

fubuloubu (approved):

PR #2083, opened 2024-05-09
refactor: utilize super().config

antazoey:
### What I did

simple code sharing refactor and add a missing cfg doc

### Checklist

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -88,6 +88,10 @@ BLUEPRINT_HEADER = HexBytes("0xfe71")
 
 class NetworkConfig(PluginConfig):
     required_confirmations: int = 0
+    """
+    The amount of blocks to wait before
+    considering a transaction 'confirmed'.
+    """
 
     default_provider: Optional[str] = "geth"
     """
@@ -353,7 +357,7 @@ class Ethereum(EcosystemAPI):
 
     @property
     def config(self) -> EthereumConfig:
-        return cast(EthereumConfig, self.config_manager.get_config(self.name))
+        return cast(EthereumConfig, super().config)
 
     @property
     def default_transaction_type(self) -> TransactionType:


dtdang (approved):

src/ape_ethereum/ecosystem.py
```diff
@@ -353,7 +357,7 @@ class Ethereum(EcosystemAPI):
 
     @property
     def config(self) -> EthereumConfig:
-        return cast(EthereumConfig, self.config_manager.get_config(self.name))
+        return cast(EthereumConfig, super().config)
```
fubuloubu, line 360: Won't this not work unless overrided somewhere in the class heirarchy? (it will have attribute error if not subclassed)

antazoey: it seems to work:

```python
(ape310) ➜  ape git:(chore/network-cfg-clean) ape console

In [1]: networks.ethereum.config.default_network
Out[1]: 'local'

In [2]: type(networks.ethereum.config)
Out[2]: ape_ethereum.ecosystem.EthereumConfig

In [3]: networks.polygon.config.default_network
Out[3]: 'local'

In [4]: type(networks.polygon.config)
Out[4]: ape_polygon.ecosystem.PolygonConfig
```

What `super().config` does as far as I understand is call `EcoystemAPI.config` which _is_ for sure a base class to `Ethereum`. The implementation for `EcosystemAPI.config` is:

```
    @property
    def config(self) -> PluginConfig:
        """
        The configuration of the ecosystem. See :class:`ape.managers.config.ConfigManager`
        for more information on plugin configurations.

        Returns:
            :class:`ape.api.config.PluginConfig`
        """

        return self.config_manager.get_config(self.name)
```

so ultimately we are sharing code by calling up to `super()`.

___

But I am curious why you say what you say here in case I don't understand because ngl, some of the class stuff i Python can be gray.

fubuloubu: As long as `EcosystemAPI.config` doesn't keep calling up (which it doesn't appear to do)

fubuloubu (approved):

src/ape_ethereum/ecosystem.py
antazoey: It would also have to call super() and so on

PR #2092, opened 2024-05-15
feat: add doctesting with github workflow

dtdang:
### What I did
Add github workflow for doctesting and some basic testing on the accounts userguide.

fixes: #

### Checklist

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -37,17 +37,6 @@ jobs:
               name: DocumentationHTML
               path: docs/_build/ape
 
-        - name: Doctesting
-          run: |
-            sphinx-build -b doctest docs docs/_build/doctest
-            if grep -q '0 failed' docs/_build/doctest/output.txt; then
-              echo "All doctests passed!"
-            else
-              echo "Some doctests failed. See the output below."
-              cat docs/_build/doctest/output.txt
-              exit 1
-            fi
-
         - name: Commit and publish documentation changes to gh-pages branch
           run: |
               if [[ "${GITHUB_EVENT_NAME}" =~ "pull_request" ]]; then
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -38,10 +38,8 @@ jobs:
               path: docs/_build/ape
 
         - name: Doctesting
-          run: sphinx-build -b doctest docs docs/_build/doctest
-        
-        - name: Check doctest results
           run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/docs.yaml
```diff
@@ -37,6 +37,19 @@ jobs:
               name: DocumentationHTML
               path: docs/_build/ape
 
+        - name: Doctesting
+          run: sphinx-build -b doctest docs docs/_build/doctest
+        
+        - name: Check doctest results
+          run: |
```
fubuloubu, line 44: Think you need to combine these steps for it to work
```suggestion
          run: |
            sphinx-build -b doctest docs docs/_build/doctest
```

.github/workflows/docs.yaml
fubuloubu, line 1: Should doctest be run here? It runs on every PR right?

docs/userguides/accounts.md
fubuloubu, line 1: Good way to start, see if it works

fubuloubu (dismissed):

.github/workflows/docs.yaml
dtdang: This should be running on every PR since its part of the workflow checks. I was also debating if it should be under testing but thought that docs was the more fitting for where I had it.

docs/userguides/accounts.md
dtdang: Failing tests will show as failing in the docs workflow but PR still shows as passing because it only checks to see if lint and tests are passing so it might be better to move it under tests rather than docs.

.github/workflows/docs.yaml
fubuloubu: hmm, big difference is that docs only runs w/ 1 python version, vs. our tests run on a matrix across our supported versions

I could go either/or on which might be better, these seems like less setup though for sure

dtdang: Moving it to `test.yaml` works but seems to fail on github actions multiple times before it gets a passing check so I'm going to move the doctest back to `docs.yaml`

docs/userguides/accounts.md
fubuloubu: We can (and should) make the docs workflow required instead

fubuloubu (approved):

docs/userguides/accounts.md
fubuloubu: Based on what happened here: https://github.com/ApeWorX/ape/actions/runs/9113855127/job/25056403447

I think it looks pretty good. I just made the docs workflow required

PR #2096, opened 2024-05-17
fix: datetime utcnow bug

antazoey:
### What I did

`get_current_timestamp_ms` claimed to return UTC now but it wasn't actually. Also, `utcnow()` is deprecated.

### Checklist

### DIFF
--- a/tests/functional/utils/test_misc.py
+++ b/tests/functional/utils/test_misc.py
@@ -1,4 +1,4 @@
-from datetime import timezone, datetime
+from datetime import datetime
 from zoneinfo import ZoneInfo
 
 import pytest
@@ -12,6 +12,7 @@ from ape.utils.misc import (
     _dict_overlay,
     add_padding_to_strings,
     extract_nested_value,
+    get_current_timestamp_ms,
     get_package_version,
     is_evm_precompile,
     is_zero_hex,
@@ -19,7 +20,6 @@ from ape.utils.misc import (
     pragma_str_to_specifier_set,
     raises_not_implemented,
     run_until_complete,
-    get_current_timestamp_ms,
     to_int,
 )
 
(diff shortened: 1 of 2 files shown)


tests/functional/utils/test_misc.py
```diff
@@ -169,3 +173,15 @@ def my_method():
...
+def test_get_current_timestamp_ms():
+    """
+    Proving the current timestamp is UTC.
+    """
+    actual = get_current_timestamp_ms()  # UTC
+    zone = ZoneInfo("America/Chicago")
+    dt = datetime(2020, 10, 31, 12, tzinfo=zone)
+    unexpected = round(datetime.now(tz=dt.tzinfo).timestamp() * 1000)
+    diff = actual - unexpected
+    assert diff == pytest.approx(18_000_000)  # Difference of CST to UTC.
```
antazoey, line 187:
```suggestion
    assert diff == pytest.approx(18_000_000) or diff == pytest.approx(19_000_000) # Difference of CST to UTC.
```

to account for daylight savings i guess

antazoey (commented):
ci failure is a hmmm

johnson2427 (dismissed):
LGTM

Historically, I've used datetime.UTC, looks like dateutil.tz.UTC is available as well, but they suggest using your method, so I'm good with it

johnson2427:
Just realized the test is failing, that test failure makes no sense

antazoey:
> LGTM
> 
> Historically, I've used datetime.UTC, looks like dateutil.tz.UTC is available as well, but they suggest using your method, so I'm good with it

Ya, datetime.UTC isnt available until 3.11 i think or else i would have used it, and we already had dateutil dep so i figured why not.

antazoey:
> Just realized the test is failing, that test failure makes no sense

I am also baffled! I know the runners are already on utc but i fail to see why that fudge up the test

johnson2427:
Never mind my last comment, you'd be better off dividing the first value than you would multiplying the second value by 1000.

fubuloubu (approved):

johnson2427 (approved):

PR #2099, opened 2024-05-29
feat: core plugin docs [APE-1750]

dtdang:
### What I did
Add methoddocs for core plugins to Plugin Python Reference section for further documentation.

fixes: #

### Checklist


fubuloubu:
Would say that the cache provider doesn't work very well, so probably should fix it first before documenting it

PR #2106, opened 2024-06-01
fix: issue where trusted plugins showed up as 3rd party

antazoey:
### What I did

Regression from the GitHub refactor - turns out GH only gives you 30 repos at a time, so now we are paging and getting them all.

### How to verify it

install a bunch of plugins and type:

```
ape plugins list
```

### Checklist


dtdang (approved):

fubuloubu:
Believe this resolves the docker build issues

PR #2117, opened 2024-06-05
fix: issues with remapping and dependency detection in foundry projects

antazoey:
### What I did

all the massaging needed to get foundry projects really working.

Goes with https://github.com/ApeWorX/ape-solidity/pull/147

### How I did it

ugh.

### How to verify it

this beast is now compiling without any ape-config.yaml required: https://github.com/yearn/tokenized-strategy
all from reading the foundry's project stuff.
the trickiest part was handling weird stuff with gitsubmodules and older foundry libs with so many different wways of doing remappings and names and i cant even explain it. Ape as a side-effect is way smarter now.

The attached ape-solidity PR will attempt to correct bad import remappings so things just work now, this also helps compile older Ape projects.

### Checklist

### DIFF
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -626,7 +626,7 @@ remappings = [
             assert actual["contracts_folder"] == "src"
             assert "solidity" in actual, "Solidity failed to migrate"
             actual_sol = actual["solidity"]
-            assert actual_sol["import_remappings"] == [
+            assert actual_sol["import_remapping"] == [
                 "forge-std/=forge-std/src/",
                 "@openzeppelin/=openzeppelin-contracts/",
             ]
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -622,13 +622,15 @@ remappings = [
             assert isinstance(api, FoundryProject)
 
             # Ensure solidity config migrated.
-            actual = temp_project.config  # Is result of ``api.extract_config()``.
+            actual = temp_project.config.model_dump(
+                by_alias=True
+            )  # Is result of ``api.extract_config()``.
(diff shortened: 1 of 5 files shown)


tests/functional/test_project.py
```diff
@@ -622,13 +622,15 @@ def test_extract_config(self, toml, gitmodules, mock_github):
...
+            )  # Is result of ``api.extract_config()``.
             assert actual["contracts_folder"] == "src"
             assert "solidity" in actual, "Solidity failed to migrate"
             actual_sol = actual["solidity"]
-            assert actual_sol["import_remappings"] == [
-                "forge-std/=forge-std/src/",
-                "@openzeppelin/=openzeppelin-contracts/",
+            assert actual_sol["import_remapping"] == [
+                "@openzeppelin=src/.cache/openzeppelin/v4.9.5/",
+                "forge-std=src/.cache/forge-std/v1.5.2/src",
```
fubuloubu, line 633: Are `.cache`s back?

fubuloubu (dismissed):

tests/functional/test_project.py
antazoey: .cache has never left, I am not sure where everyone got that idea. I can try to explain again now.
`.cache` was removed from the user's contracts folder and solidity now compiles in a sandbox. .cache folder is still needed in the sandbox but it is consistently recreated whenever a new compilation sandbox is happening so there are no more issues with it. Also, the user does not have to .gitignore it or worry about it interfering with other tools this way. The concept of the .cache remains however as the dependencies MUST be in the contracts folder for Solidity to work with input JSON

fubuloubu (approved):

PR #2122, opened 2024-06-10
fix: eip-712 hex data display

banteg:
### What I did

better display of bytes data in eip-712 message when signing.

fixes: #

### Checklist

### DIFF
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -190,6 +190,8 @@ class KeyfileAccount(AccountAPI):
             # Message Data
             display_msg += "Message\n"
             for field, value in msg._body_["message"].items():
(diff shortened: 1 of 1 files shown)


src/ape_accounts/accounts.py
```diff
@@ -190,6 +190,8 @@ def sign_message(self, msg: Any, **signer_options) -> Optional[MessageSignature]
             # Message Data
             display_msg += "Message\n"
             for field, value in msg._body_["message"].items():
+                if isinstance(value, bytes):
+                    value = HexBytes(value).hex()
```
antazoey, line 194: if we care about 0x prefix, this may no longer be the way to do this once web3.py 7 is out

fubuloubu (approved):

PR #2123, opened 2024-06-10
fix: `project.unpack()` to include full contracts directory

antazoey:
### What I did

sources with missing extensions were not copied over when doing `unpack()` which is bad for a few reasons.

1. problems when you are trying to compile a non-standard file via imports or something such as `.t.sol`
2. problem if copying a project somewhere before installing plugins
3. problem with performance (shutil copy tree way way way faster than looping through all sources)

this is needed for some ape-solidity work

### Checklist


fubuloubu (dismissed):

antazoey:
going absolutely bonkers trying to either reproduce the failing CI issues locally or fix them, like what the ducking duck

PR #2130, opened 2024-06-12
feat: better CLI message when compiling a dependency produces no contracts

antazoey:
### What I did

was confused for a bit because i had uninstalled `ape-solidity` but `ape pm compile . --force` was not telling me it was successful but it nothing was really happening... 

### How I did it

detect if no result and warn.
additionally, if you notice the dependency has .vy or .sol contracts and you are missing those from 

### Checklist

### DIFF
--- a/src/ape_pm/_cli.py
+++ b/src/ape_pm/_cli.py
@@ -7,7 +7,8 @@ import click
 from ape.cli.options import ape_cli_context, config_override_option
 from ape.exceptions import ProjectError
 from ape.logging import logger
-from ape.managers.project import Dependency
+from ape.managers.project import Dependency, LocalProject
+from ape.utils import get_all_files_in_directory, get_full_extension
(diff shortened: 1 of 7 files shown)


src/ape_pm/_cli.py
```diff
@@ -321,10 +314,35 @@ def compile(cli_ctx, name, version, force, config_override):
...
+        _compile_dependency(cli_ctx, dependency, force)
+
+
+def _compile_dependency(cli_ctx, dependency: Dependency, force: bool):
+    try:
+        result = dependency.compile(use_cache=not force)
+    except Exception as err:
+        cli_ctx.logger.error(str(err))
+    else:
+        if result:
```
antazoey, line 326: - [ ] Move this check to `dependency.compile()` logic so you get the warning in regular python code as well (?)

dtdang (approved):

PR #2136, opened 2024-06-13
chore: upgrade pandas

antazoey:
### What I did

see what it do

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ setup(
         "ipython>=8.18.1,<9",
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
-        "pandas>=1.3.0,<2",
+        "pandas>=2.2.2,<3",
         "pluggy>=1.3,<2",
         "pydantic>=2.6.4,<3",
         "pydantic-settings>=2.0.3,<3",


setup.py
```diff
@@ -103,7 +103,7 @@
         "ipython>=8.18.1,<9",
         "lazyasd>=0.1.4",
         "packaging>=23.0,<24",
-        "pandas>=1.3.0,<2",
+        "pandas>=2.2.2,<3",
```
antazoey, line 106: the install only take 1 minute now on Python 3.12 so i think this fixes it

johnson2427: We have python3.11 failures, I also worry about API changes from 1.x.x to 2.x.x. There's quite a bit of deprecating that's happened between the versions.

antazoey: The failures are unrelated flaky tests I have been seeing all day. My attempt at de-flaking is [here](https://github.com/ApeWorX/ape/pull/2137).

 Re: the deprecating pandas stuff, maybe a quick audit would help! I am not sure how much we are using any pandas features. I know we use some dataframes but for the most part we create them and return them to the user, I think.

fubuloubu (approved):

Issue #324, opened 2021-12-16
Testing: Equality handling between currency units and int


NotPeopling2day:
### Overview

Conversion of `"1 gwei"` to `int` would be valuable for things like specifying gas prices, as well as comparisons to account balances:
```python 
assert contract.balance == "1 ether"
```

Related to ticket #261 and #323

fubuloubu:
temporary workaround:
```py
assert contract.balance == convert("1 ether", int)
```

PR #2149, opened 2024-06-17
feat: allow comparing currency values to currency strings

antazoey:
### What I did

```python
def test_balance(accounts):
    acct = accounts[0]
    # Before
    assert acct.balance == convert("1000000 ETH", int)
    # After
    assert acct.balance == "1000000 ETH"
```

fixes: #324 

### How I did it

super simple int wrapper in a couple cases

### Checklist

### DIFF
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Any
 from eth_pydantic_types import HexBytes
 
 from ape.exceptions import ConversionError
-from ape.types import AddressType, ContractCode
+from ape.types import AddressType, ContractCode, CurrencyValue
 from ape.utils import BaseInterface, abstractmethod, cached_property, log_instead_of_fail
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


fubuloubu (approved):
:eyes:

banteg:
do i read it right it only works with `account.balance`? what would it take to make it also work with contract call return values like `token.balanceOf`?

antazoey:
> do i read it right it only works with `account.balance`? what would it take to make it also work with contract call return values like `token.balanceOf`?

We would have to detect `balanceOf` name in the ABI and wrap it `CurrencyValue` during the result.
Alternatively, can make the name of `CurrencyValue` more generic like `CurrencyComparable` and just wrap all return data ints in that.

banteg:
yeah i think it's the best to just make every int support this

antazoey:
> yeah i think it's the best to just make every int support this

I don't think we can make every int support this though, without modifying python or something?
We can make every int returned from `decode_returndata()` handle this tho

fubuloubu:
I think adding comparison of return values to values decoded by the conversion manager for types like `int` will satisfy 90% of the use cases

We already have custom `__eq__` methods for things like structs, I think only `Decimal` would be the other python type to handle, maybe `AddressType` might also be useful e.g. `assert c.owner() == "ychad.eth"`
https://github.com/ApeWorX/ape/blob/main/src/ape/managers/converters.py#L260-L275

banteg:
> We can make every int returned from decode_returndata() handle this tho

yes, sure, i meant exactly this

PR #2150, opened 2024-06-17
fix: issue where plugin config loading error was not shown

antazoey:
### What I did

I had my config set up with some environment variables that did not exist, and i noticed that instead of Ape crashing where it was supposed to, my error s were dropped and the configs were turned to regular `PluginConfig` and i got attribute errors within the plugin instead, kinda weird... 

### How I did it

Removed duplicate `get_config` method that hid the errors for no reason.
instead, let `ConfigManager` run `ApeConfig.get_config` and the error shows up correctly.

### How to verify it

```yaml
foundry:
  fork:
    shibarium:
      mainnet:
        block_number: {asdf}

```

try to use foundry

### Checklist

### DIFF
--- a/src/ape/managers/config.py
+++ b/src/ape/managers/config.py
@@ -7,7 +7,6 @@ from typing import Any, Optional
 
 from ethpm_types import PackageManifest
 
-from ape.api import PluginConfig
 from ape.api.config import ApeConfig
 from ape.managers.base import BaseManager
 from ape.utils import create_tempdir, in_tempdir, log_instead_of_fail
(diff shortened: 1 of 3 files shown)


src/ape/managers/config.py
```diff
@@ -106,22 +104,6 @@ def extract_config(cls, manifest: PackageManifest, **overrides) -> ApeConfig:
         """
         return ApeConfig.from_manifest(manifest, **overrides)
 
-    def get_config(self, plugin_name: str) -> PluginConfig:
```
antazoey, line 109: bug was fixed by deleting this method.
`get_config` still exists on `ApeConfig` class, and thus is still available here, only without swallowing every single error

fubuloubu (approved):

Issue #2154, opened 2024-06-19
Ape command fails to execute


ShubhamChndrvnshi:
### Environment information

- OS: macOS (m1)
- Python Version: 3.9.5
- `ape` and plugin versions:
eth-ape==0.8.2
ape-solidity==0.8.0
ape-hardhat==0.8.0
ape-etherscan==0.8.0
ape-infura==0.8.0

```sh
$ ape --version
# ...copy and paste result of above command here...

$ ape plugins list
# ...copy and paste result of above command here...
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
name: olas-opinion-market
contracts_folder: src
plugins:
  - name: solidity
  - name: hardhat
dependencies:
  - name: OpenZeppelin
    github: OpenZeppelin/openzeppelin-contracts
    version: 5.0.2
solidity:
  version: 0.8.24
  via_ir: True
hardhat:
  host: 127.0.0.1:8545
test:
  mnemonic: test test test test test test test test test test test junk
  number_of_accounts: 100
deployments:
  ethereum:
    sepolia:
      - contract_type: PrivateVote
        address: "0x6b12a13dbe488fd4b2e01eef819aadc245bb8ee6"
default_ecosystem: ethereum

```

### What went wrong?

Please include information like:

- what command you ran
Github actions were failing for ape test. Turns out ape command in not able to run.
Ran: ape --version, ape test, etc.
- the code that caused the failure (see [this link](https://help.github.com/articles/basic-writing-and-formatting-syntax/) for help with formatting code)
- full output of the error you received
ape --version
Traceback (most recent call last):
  File "/home/harambe/venv/bin/ape", line 5, in <module>
    from ape._cli import cli
  File "/home/harambe/venv/lib/python3.9/site-packages/ape/__init__.py", line 10, in <module>
    from ape.managers.project import ProjectManager as Project
  File "/home/harambe/venv/lib/python3.9/site-packages/ape/managers/__init__.py", line 6, in <module>
    from .chain import ChainManager
  File "/home/harambe/venv/lib/python3.9/site-packages/ape/managers/chain.py", line 10, in <module>
    import pandas as pd
  File "/home/harambe/venv/lib/python3.9/site-packages/pandas/__init__.py", line 22, in <module>
    from pandas.compat import is_numpy_dev as _is_numpy_dev  # pyright: ignore # noqa:F401
[...]

PR #2155, opened 2024-06-20
chore: pin numpy i guess

antazoey:
fixes: #2154

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,10 @@ setup(
         "ijson>=3.1.4,<4",
         "ipython>=8.18.1,<9",
         "lazyasd>=0.1.4",
+        # Numpy is a peer-dependency (pandas) but apparently
+        # numpy broke pandas and now it MY responsibility to
+        # pin it in Ape (•̀o•́)ง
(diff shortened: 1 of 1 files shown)


antazoey:
there isnt much evidence this is needed... Core was passing tests before down-pinning numpy and it was successfully installing numpy==2.0.0.

i am guessing there is more to this issue than pinning.

setup.py
```diff
@@ -102,6 +102,10 @@
         "ijson>=3.1.4,<4",
         "ipython>=8.18.1,<9",
         "lazyasd>=0.1.4",
+        # Numpy is a peer-dependency (pandas) but apparently
+        # numpy broke pandas and now it MY responsibility to
+        # pin it in Ape (•̀o•́)ง
```
fubuloubu, line 107:
```suggestion
        # Numpy is a peer-dependency of Pandas but they don't have upper pin
```

fubuloubu (dismissed):

setup.py
antazoey: pandas works with numpy 2.0 though.

PR #2163, opened 2024-06-27
feat: easily create projects from site-packages

antazoey:
### What I did

inspired by trying to resolve https://github.com/ApeWorX/ape-vyper/issues/119 and conversations with @fubuloubu a while back.

Also allows you to specify dependencies this way using the "python:" key.
In ape-vype, I want to have this work automatically when it detects you are using (importing) something that is not marked anywhere but does exist as a python dependency. then it will just work.

### How I did it

check site-packages for the given name and then return a local project pointing at it.

### How to verify it

```sh
pip install snekmate
```

```python
from ape import Project

snekmate = Project.from_python_library("snekmate")
```

### Checklist


docs/userguides/projects.md
```diff
@@ -55,6 +55,16 @@ project = Project.from_manifest("path/to/manifest.json")
...
 ```
 
+## Installed Python Projects
+
+If you have installed a project using `pip` or alike and you wish to reference its project, use the `Project.from_python_library()` class method.
+
+```python
+from ape import Project
+
+snekmate = Project.from_python_library("snekmate")
```
antazoey, line 65: potentially could call this `from_site_package` instead?

src/ape/managers/project.py
```diff
@@ -1581,14 +1626,6 @@ def is_compiled(self) -> bool:
         """
         return (self._manifest.contract_types or None) is not None
 
-    @classmethod
```
antazoey, line 1584: note: just a duplicate method, isnt needed - no changes, it still works

docs/userguides/projects.md
antazoey, line 65: I actually don't recommend this though... snekmate does not publish any configuration so it is practically impossible to use the package as a project on its own.

src/ape/managers/project.py
```diff
@@ -1515,7 +1516,51 @@ def from_manifest(
         Returns:
             :class:`~ape.managers.project.ProjectManifest`
         """
-        return Project.from_manifest(manifest, config_override=config_override)
+        config_override = config_override or {}
+        manifest = _load_manifest(manifest) if isinstance(manifest, (Path, str)) else manifest
+        return Project(manifest, config_override=config_override)
+
+    @classmethod
+    def from_python_library(
```
antazoey, line 1524: in addition to this, we could have a `PythonDependency(DepdendencyAPI)` where you config like:

```
dependencies:
  python: snekmate
```

somehow need this to happen automatically though....

src/ape/managers/project.py
```diff
@@ -1515,7 +1516,43 @@ def from_manifest(
...
+        Args:
+            package_name (str): The name of the package's folder that would
+              appear in site-packages.
+            config_override (dict | None): Optionally override the configuration
+              for this project.
+
+        Returns:
+            :class:`~ape.managers.project.LocalProject`
+        """
+        for site_packages_path_str in getsitepackages():
```
fubuloubu, line 1541: I feel like `importlib.resources` may have better API for this

antazoey: OK it has been refactored to using `importlib.resources` instead.

fubuloubu (approved):

Issue #2180, opened 2024-07-19
feat: Add option to explicitly ignore reverts 


Aviksaikat:
### Overview

Provide a simple overview of what you wish to see added. Please include:

- Add an option like brownie to explicitly ignore reverts like brownie does with `allow_revert=True` option
-  Would be a great feature idea, and something to consider further for "smart receipts" we want to add to v1 (e.g. blocking or non-blocking receipt handling based on how you call a transaction and work with the receipt) (- @fubuloubu via discord)

### Specification

N/A

### Dependencies

N/A

fubuloubu:
We should add `revert_on_fail=False` as a new kwarg to contract transactions, which in the case of a failure would log a warning instead of raise an error, and ensure that `ReceiptAPI` has `.error` property so it's there in case a failure did occur and you want to access it or do instance checks against it (e.g. `if isinstance(tx.error, contract.CustomError): ...`).

antazoey:
I think `revert_on_fail` may imply something inaccurate... `revert_on_fail` makes me sound like somehow we are able to make the VM not revert at all and let it crash out some other way.

I think `raise_on_revert` or something alike is more accurate

antazoey:
I like `allow_revert` as well because it mimics `click`'s `allow_failure` flag in their test-runner invoke.

PR #2181, opened 2024-07-22
feat: `raise_on_revert=False` kwarg for allowing failures on calls and transactions

antazoey:
### What I did

fixes: #2180 
Fixes: Ape-1788

### Checklist

### DIFF
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -258,7 +258,7 @@ assert receipt.return_value == 123
 
 Transactions may also fail, known as a "revert".
 When a transaction reverts, Ape (by default) raises a subclass of `TransactionError`, which is a Python exception.
-To learn more reverts, see the [reverts guide](../reverts.html).
+To learn more reverts, see the [reverts guide](./reverts.html).
 
 For more general information on transactions in the Ape framework, see [this guide](./transactions.html).
 
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -256,6 +256,24 @@ If you need the `return_value` from a transaction, you have to either treat tran
 assert receipt.return_value == 123
 ```
 
+Transactions may also fail.
+When a transaction fails, Ape (by default) raises the virtual machine error which may crash your script.
+You will see exception trace data like this:
+
+```
+  File "$HOME/ApeProjects/ape-playground/scripts/fail.py", line 7, in main
+    contract.setNumber(5, sender=owner)
+
+ERROR: (ContractLogicError) Transaction failed.
+```
+
+If you know a transaction is going to revert and you wish to allow it, use the `fail_on_revert=False` flag:
+
+```python
+receipt = contract.setNumber(5, sender=owner, fail_on_revert=False)
+assert receipt.failed
+```
+
 For more general information on transactions in the Ape framework, see [this guide](./transactions.html).
 
 ### Calls
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -257,22 +257,9 @@ assert receipt.return_value == 123
 ```
 
 Transactions may also fail.
-When a transaction fails, Ape (by default) raises the virtual machine error which may crash your script.
-You will see exception trace data like this:
-
-```
-  File "$HOME/ApeProjects/ape-playground/scripts/fail.py", line 7, in main
-    contract.setNumber(5, sender=owner)
-
-ERROR: (ContractLogicError) Transaction failed.
-```
-
-If you know a transaction is going to revert and you wish to allow it, use the `raise_on_revert=False` flag:
-
-```python
-receipt = contract.setNumber(5, sender=owner, raise_on_revert=False)
-assert receipt.failed
-```
+This is called a `revert`!
+When a transaction reverts, Ape (by default) raises the virtual machine error which may crash your script.
+To learn more reverts, see the [reverts guide](../reverts.html).
 
 For more general information on transactions in the Ape framework, see [this guide](./transactions.html).
 
--- a/docs/userguides/contracts.md
+++ b/docs/userguides/contracts.md
@@ -256,9 +256,8 @@ If you need the `return_value` from a transaction, you have to either treat tran
 assert receipt.return_value == 123
 ```
 
-Transactions may also fail.
-This is called a `revert`!
-When a transaction reverts, Ape (by default) raises the virtual machine error which may crash your script.
+Transactions may also fail, known as a "revert".
+When a transaction reverts, Ape (by default) raises a subclass of `TransactionError`, which is a Python exception.
 To learn more reverts, see the [reverts guide](../reverts.html).
 
 For more general information on transactions in the Ape framework, see [this guide](./transactions.html).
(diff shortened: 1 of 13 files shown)


docs/userguides/contracts.md
```diff
@@ -256,6 +256,24 @@ If you need the `return_value` from a transaction, you have to either treat tran
...
+When a transaction fails, Ape (by default) raises the virtual machine error which may crash your script.
+You will see exception trace data like this:
+
+```
+  File "$HOME/ApeProjects/ape-playground/scripts/fail.py", line 7, in main
+    contract.setNumber(5, sender=owner)
+
+ERROR: (ContractLogicError) Transaction failed.
+```
+
```
fubuloubu, line 269: Is there a section somewhere that talks about decoding custom errors? this might be a good place for it.

docs/userguides/contracts.md
```diff
@@ -256,6 +256,24 @@ If you need the `return_value` from a transaction, you have to either treat tran
...
+If you know a transaction is going to revert and you wish to allow it, use the `fail_on_revert=False` flag:
+
+```python
+receipt = contract.setNumber(5, sender=owner, fail_on_revert=False)
+assert receipt.failed
+```
```
fubuloubu, line 275: Might be worth showing what happens when you try to use certain properties of the receipt that aren't available when it fails e.g. `.return_value`

src/ape_test/provider.py
```diff
@@ -240,11 +249,18 @@ def send_transaction(self, txn: TransactionAPI) -> ReceiptAPI:
...
+                else:
+                    receipt.error = vm_err
+
+            if txn.fail_on_revert:
+                # If we get here, for some reason the tx-replay did not produce
+                # a VM error.
+                receipt.raise_for_status()
+
+        if receipt.error:
+            logger.error(receipt.error)
```
fubuloubu, line 263: Wonder if a nicer refactor might be to always set `.error` and then use that within `.raise_for_status()`

fubuloubu (dismissed):
Overall, I think I do like `raise_on_revert` better for the kwarg name, as it shows the python action (e.g. `raise`ing an error) is tied to the EVM result (e.g. the transaction `revert`s)

docs/userguides/contracts.md
antazoey: There is one in testing: https://docs.apeworx.io/ape/stable/userguides/testing.html#custom-errors
Perhaps we need a separate `reverts.md` guide and reference it from both spots.

src/ape_test/provider.py
antazoey: that works for some cases, I can do that, but other cases we revert before getting a receipt

docs/userguides/contracts.md
fubuloubu: Yeah, it might be a great candidate for a separate user guide

docs/userguides/reverts.md
```diff
@@ -0,0 +1,135 @@
+# Reverts
+
+Reverts occur when a transaction or call fails for any reason.
+In the case of EVM networks, reverts result in funds being returned to the sender (besides network-fees) and contract state changes rewinding.
```
fubuloubu, line 4:
```suggestion
In the case of EVM networks, reverts result in funds being returned to the sender (besides network-fees) and contract state changes unwinding.
```

docs/userguides/reverts.md
```diff
@@ -0,0 +1,135 @@
...
+# continue on!
+```
+
+If you wish to allow reverts without having Ape raise exceptions, use the `raise_on_revert=False` flag:
+
+```shell
+receipt = contract.setNumber(123, sender=not_owner, raise_on_revert=False)
+print(receipt.failed)
+# Also, access the `error` on the receipt:
+print(receipt.error)
```
fubuloubu, line 63: show the values?

docs/userguides/contracts.md
antazoey: you know what it is weird (or cool), i think `.return_value` is the revert message or something:

```
>>> receipt.return_vaue
'0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b21617574686f72697a6564000000000000000000000000000000000000000000'
```

antazoey: i think in that case, the receipt data was available though.

antazoey: that was when i used a geth node though. Using `eth-tester`, i get an empty tuple!

antazoey: i guess what im saying is that the properties will always be available in one way or another

docs/userguides/contracts.md
antazoey: done

docs/userguides/reverts.md
antazoey: done

src/ape_test/provider.py
antazoey: we'd lose the `from err` part potentially

antazoey: i made it at least alway set `.error` so when using debuggers, the receipt is more useful

antazoey:
- [ ] Support `ape-hardhat`
- [ ] Support `ape-foundry`

docs/userguides/contracts.md
fubuloubu: I think we can either document it as "the value of `.return_value` when `.failed` is true is undocumented behavior", or standardizing what `.return_value` is for all the different providers (1 sounds better)

antazoey: oh that probably only has to do with tracing, since eth-tester doesnt have tracing and the receipt data doesnt contain the return value (I think Charles has a EIP for this)

fubuloubu (dismissed):
This looks great!

One thing I am considering is what happens when partial failure(s) exist in a receipt? (Reverts but handled by upstream logic e.g. SafeTx execution)

This is probably a tracing feature (out or scope for this PR), something like `ReceiptAPI.ignored_reverts -> list[ContractLogicError]`, but might be good to think of how they might work together before merging this

antazoey:
> but might be good to think of how they might work together before merging this

We could put all reverts in a list and have `.error` refer to the last one?

antazoey:
tbd im not sure what the experience is like with partially failed txs. Is there a way using local developer networks to simulate this?

fubuloubu:
> tbd im not sure what the experience is like with partially failed txs. Is there a way using local developer networks to simulate this?

You'd need a contract that does a call to another contract, and then handle the result. In vyper you can do this with `raw_call`, there is no native exception handling yet. Solidity does allow exception handling with the new custom errors stuff I think using try/catch: https://docs.soliditylang.org/en/v0.8.26/control-structures.html#try-catch

fubuloubu:
> > but might be good to think of how they might work together before merging this
> 
> We could put all reverts in a list and have `.error` refer to the last one?

I think in this scenario if it doesn't revert then `.error` will be `None`, but then separately `.ignored_reverts` may not be empty regardless of whether ultimately the transaction reverted or not (as an error may have been ignored earlier in the trace)

fubuloubu (approved):

PR #2184, opened 2024-07-22
chore: handle 1.11 mypy upgrades

antazoey:
### What I did

fixes new mypy complaints and upgrades

### Checklist

### DIFF
--- a/src/ape/utils/basemodel.py
+++ b/src/ape/utils/basemodel.py
@@ -201,7 +201,7 @@ class _AttrLookup(dict):
     def __getitem__(self, item):
         return self._callback(item)
 
-    def get(self, item):
+    def get(self, item):  # type: ignore[override]
         return self._callback(item)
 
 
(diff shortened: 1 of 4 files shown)


src/ape/utils/basemodel.py
```diff
@@ -201,7 +201,7 @@ def __getattr__(self, item):
     def __getitem__(self, item):
         return self._callback(item)
 
-    def get(self, item):
+    def get(self, item):  # type: ignore[override]
```
antazoey, line 204: basically consider this an interface and we are never using any other kwargs in the context it gets called

src/ape_run/_cli.py
```diff
@@ -213,7 +213,7 @@ def get_command(self, ctx, name):
 
         # NOTE: don't return anything so Click displays proper error
 
-    def result_callback(self, result, interactive):
+    def result_callback(self, result, interactive: bool):  # type: ignore[override]
```
antazoey, line 216: we just be hacking click a bit here, dont mind us, mypy

fubuloubu: the hack is annotating `interactive`?

antazoey: the hack is having `interactive` at all, it isnt in the original method. There are no real changes here.

fubuloubu (approved):

Issue #1294, opened 2023-02-09
Allow bypassing caching behavior for 2nd/3rd party QueryAPI plugins


fubuloubu:
### Elevator pitch:

Query plugins may implement caching behavior by overriding the `QueryAPI.update_cache` method (which is empty by default), but sometimes this behavior may be buggy or otherwise undesirable.

### Design approach:

Add a key to `ape-config.yaml`:

```yaml
query:
  skip_caching:
    - query-plugin-1
    - ...
```

Filter out plugins to cache using this setting so that users can bypass this behavior if desired:

https://github.com/ApeWorX/ape/blob/main/src/ape/managers/query.py#L146

### Task list:

- [ ] Add config item for `query.skip_caching`
- [ ] Add handling in `QueryManager` for bypassing calls to `engine.update_cache`

PR #2186, opened 2024-07-23
Allow bypassing caching behavior for 2nd/3rd party QueryAPI plugins

rafael-abuawad:
### What I did

This is related to this issue. [https://github.com/ApeWorX/ape/issues/1294](https://github.com/ApeWorX/ape/issues/1294)

1. Added a new method `_get_skip_caching_plugins()` that retrieves the list of plugins to skip caching from the configuration. If the configuration doesn't exist, it returns an empty list.
2. In the `query()` method, fetch the list of plugins to skip caching.
3. When iterating through the engines to update caches, check if the engine's name is in the skip list. If it's not in the list, proceed with the cache update as before. If it is in the list, skip the cache update and log a debug message.

### How I did it

Updated [class QueryManager](https://github.com/ApeWorX/ape/blob/ffd9ef2357a7354aca5f462844143aa262d35f3a/src/ape/managers/query.py#L99C13-L99C14)

### How to verify it

...

### Checklist
- [x] Add config item for `query.skip_caching`
- [x]  Add handling in `QueryManager` for bypassing calls to `engine.update_cache`

### DIFF
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -203,13 +203,18 @@ class QueryManager(ManagerAccessMixin):
                 f" executed in {exec_time} ms (expected: {est_time} ms)"
             )
 
+        skip_caching_plugins = self._get_skip_caching_plugins()
+
         # Update any caches
-        for engine in self.engines.values():
+        for engine_name, engine in self.engines.items():
             if not isinstance(engine, sel_engine.__class__):
-                result, cache_data = tee(result)
-                try:
-                    engine.update_cache(query, cache_data)
-                except QueryEngineError as err:
-                    logger.error(str(err))
+                if engine_name not in skip_caching_plugins:
+                    result, cache_data = tee(result)
+                    try:
+                        engine.update_cache(query, cache_data)
+                    except QueryEngineError as err:
+                        logger.error(str(err))
+                else:
+                    logger.debug(f"Skipping cache update for plugin: {engine_name}")
 
         return result
--- a/src/ape/managers/query.py
+++ b/src/ape/managers/query.py
@@ -128,6 +128,15 @@ class QueryManager(ManagerAccessMixin):
 
     def _suggest_engines(self, engine_selection):
         return difflib.get_close_matches(engine_selection, list(self.engines), cutoff=0.6)
+    
+    def _get_skip_caching_plugins(self):
+        """
+        Get the list of query plugins to skip caching for from the configuration.
+        """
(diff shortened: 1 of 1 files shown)


src/ape/managers/query.py
```diff
@@ -128,6 +128,15 @@ def engines(self) -> dict[str, QueryAPI]:
 
     def _suggest_engines(self, engine_selection):
         return difflib.get_close_matches(engine_selection, list(self.engines), cutoff=0.6)
+    
+    def _get_skip_caching_plugins(self):
+        """
+        Get the list of query plugins to skip caching for from the configuration.
+        """
+        try:
+            return self.config_manager.query.skip_caching
```
fubuloubu, line 137: Adding this to the config would make it clearer how this feature works. Might also be nice to have the engine class override selectable via config, so it will always use that engine if set

fubuloubu (commented):
I think a better name would be `skip-cache` e.g.

```yaml
query:
  skip-cache:
    - engine1
    - ...
```

src/ape/managers/query.py
```diff
@@ -194,13 +203,18 @@ def query(
...
         # Update any caches
-        for engine in self.engines.values():
+        for engine_name, engine in self.engines.items():
             if not isinstance(engine, sel_engine.__class__):
-                result, cache_data = tee(result)
-                try:
-                    engine.update_cache(query, cache_data)
-                except QueryEngineError as err:
-                    logger.error(str(err))
+                if engine_name not in skip_caching_plugins:
```
antazoey, line 211: refactor nit (feel free to ignore):

```python
if engine_name in skip_caching_plugins:
    logger.debug(f"Skipping cache update for plugin: {engine_name}")
    continue

result, cache_data = tee(result)
try:
    engine.update_cache(query, cache_data)
except QueryEngineError as err:
    logger.error(str(err))
```

is less indenting and less words types

src/ape/managers/query.py
```diff
@@ -128,6 +128,15 @@ def engines(self) -> dict[str, QueryAPI]:
 
     def _suggest_engines(self, engine_selection):
         return difflib.get_close_matches(engine_selection, list(self.engines), cutoff=0.6)
+    
+    def _get_skip_caching_plugins(self):
```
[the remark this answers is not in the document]
fubuloubu: I think naming it something like `_engines_to_skip_caching` may help more with readability

PR #2188, opened 2024-07-24
fix: issue where Vyper source-lines were not displayed in call-stacktraces

antazoey:
### What I did

We show the lines of Vyper files in the Python stack traces well for Transactions.
However, I noticed it was not working for call.

This PR fixes that.

### How I did it

Mostly just remove the unnecessary reliance on the txn= kwarg when setting the traceback

### How to verify it

make a call that fails, im doing:

```
contract.setNumber.call(5) 
```

in a script, then you should see the trace showing the vyper lines:

```
  File "$HOME/ApeProjects/ape-playground/contracts/VyperContract.vy", line 98, in 
setNumber
    assert msg.sender == self.owner, "!authorized"
```

### Checklist

### DIFF
--- a/src/ape/exceptions.py
+++ b/src/ape/exceptions.py
@@ -24,6 +24,7 @@ if TYPE_CHECKING:
     from ape.api.providers import SubprocessProvider
     from ape.api.trace import TraceAPI
     from ape.api.transactions import ReceiptAPI, TransactionAPI
+    from ape.managers.project import ProjectManager
     from ape.types import AddressType, BlockID, SnapshotID, SourceTraceback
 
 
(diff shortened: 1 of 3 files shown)


src/ape/exceptions.py
```diff
@@ -836,19 +839,27 @@ def _get_ape_traceback_from_tx(txn: FailedTxn) -> Optional["SourceTraceback"]:
 
 
 def _get_custom_python_traceback(
```
antazoey, line 841: @bitwise-constructs if you are ever curious how to hack Python's stack-frames to inject made-up frames pointing a lines in your Vyper / Solidity files, we do it here (only works on Vyper currently)

dtdang (approved):

PR #2193, opened 2024-07-30
perf: RPC / provider level speed optimizations

antazoey:
### What I did

A collection of perfs to make Ape faster... ongoing research and implementation..

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -173,7 +173,6 @@ class EcosystemAPI(ExtraAttributesMixin, BaseInterfaceModel):
         Returns:
             bytes
         """
-
         if not self.signature:
             raise SignatureError("The transaction is not signed.")
 
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -598,7 +598,6 @@ class Ethereum(EcosystemAPI):
         error = receipt_kwargs.pop("error", None)
         receipt = receipt_cls.model_validate(receipt_kwargs)
         receipt.error = error
-
         return receipt
 
     def decode_block(self, data: dict) -> BlockAPI:
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -550,7 +550,6 @@ class Ethereum(EcosystemAPI):
             status = self.conversion_manager.convert(status, int)
             status = TransactionStatusEnum(status)
 
-        txn_hash = None
         hash_key_choices = (
             "hash",
             "txHash",
@@ -559,18 +558,13 @@ class Ethereum(EcosystemAPI):
             "transactionHash",
             "transaction_hash",
         )
-        for choice in hash_key_choices:
-            if choice in data:
-                txn_hash = data[choice]
-                break
+        txn_hash = next((data[choice] for choice in hash_key_choices if choice in data), None)
+        if txn_hash and isinstance(txn_hash, bytes):
+            txn_hash = txn_hash.hex()
 
-        if txn_hash:
-            txn_hash = txn_hash.hex() if isinstance(txn_hash, bytes) else txn_hash
-
-        data_bytes = data.get("data", b"")
+        data_bytes = data.get("data")
         if data_bytes and isinstance(data_bytes, str):
             data["data"] = HexBytes(data_bytes)
-
         elif "input" in data and isinstance(data["input"], str):
             data["input"] = HexBytes(data["input"])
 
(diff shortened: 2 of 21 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -559,36 +558,31 @@ def decode_receipt(self, data: dict) -> ReceiptAPI:
...
             data["data"] = HexBytes(data_bytes)
-
         elif "input" in data and isinstance(data["input"], str):
             data["input"] = HexBytes(data["input"])
 
         block_number = data.get("block_number", data.get("blockNumber"))
         if block_number is None:
             raise ValueError("Missing block number.")
 
-        receipt_kwargs = dict(
```
fubuloubu, line 581: I kinda like this method better, is there some tradeoff vs. Doing it as a dict literal?

fubuloubu (dismissed):

src/ape_ethereum/ecosystem.py
antazoey: Yes, it seems significant enough:

```
(ape08) ➜  ape git:(perf/pytest) ✗ python -m timeit "dict()"
5000000 loops, best of 5: 54.6 nsec per loop
(ape08) ➜  ape git:(perf/pytest) ✗ python -m timeit "{}"
10000000 loops, best of 5: 30.5 nsec per loop
```

antazoey: This is also kind of interesting:

```python
>>> from dis import dis
>>> dis("dict()")
  0           0 RESUME                   0

  1           2 PUSH_NULL
              4 LOAD_NAME                0 (dict)
              6 CALL                     0
             14 RETURN_VALUE
>>> dis("{}")
  0           0 RESUME                   0

  1           2 BUILD_MAP                0
              4 RETURN_VALUE
```

I using Python 3.12 btw, older versions I heard it was even worse.

fubuloubu: Wow

fubuloubu (approved):

antazoey:
before:

In [3]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 108 ms

In [4]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 96.2 ms

In [5]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 93.9 ms


after

In [2]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 100 ms

In [3]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 87.5 ms

In [4]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 87.8 ms


Ape 0.6.22

In [3]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 372 ms

In [3]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 103 ms

In [4]: %time contract = owner.deploy(project.VyperContract, 123)
Wall time: 125 ms

src/ape/api/networks.py
```diff
@@ -177,7 +177,7 @@ def serialize_transaction(self) -> bytes:
         if not self.signature:
             raise SignatureError("The transaction is not signed.")
 
-        txn_data = self.model_dump(exclude={"sender"})
+        txn_data = self.model_dump(exclude=("sender",))
```
fubuloubu, line 180: Is there a performance benefit to using tuple literal instead of set literal? Doing this somewhere else

antazoey: Tuples are the fastest collection-based data type in Python! You should only use sets if doing set stuff.

```
(ape08) ➜  ape git:(perf/pytest) python -m timeit "('foo',)"
50000000 loops, best of 5: 6.28 nsec per loop
(ape08) ➜  ape git:(perf/pytest) python -m timeit "{'foo'}" 
5000000 loops, best of 5: 59.4 nsec per loop
```

fubuloubu: btw I don't think `include=`/`exclude=` actually works with a tuple:
https://github.com/pydantic/pydantic/blob/a2f7271fd394e9d2a5b19cf41f531d7fb0a6b6f0/pydantic/deprecated/copy_internals.py#L24

antazoey: Odd:

```
>>> from pydantic import BaseModel
>>> class MyModel(BaseModel):
...    x: int = 3
...    y: int = 4
... 
>>> model = MyModel()
>>> model.model_dump(exclude=("x",))
{'y': 4}
```

I suppose since it is undocumented, I will undo it though, but I am going to open an issue in Pydantic.

antazoey: I reverted back to sets since tuple is undocumented, but we will see what happens with this: https://github.com/pydantic/pydantic/issues/10010

antazoey: There is some interesting discussion there, but in conclusions, yeah tuples are not yet properly supported, even though it works in all the cases we are using it. It sounds like they may add better support for any sequence types here in the future. Resolving this for now and sticking with sets!

antazoey:
More metrics:

```
Name (time in ms)          Min       Max      Mean  StdDev    Median     IQR

before:
test_showt            463.9104  479.3264  471.3744  6.1150  471.2911  9.7154


after
test_showt            361.3695  461.4727  383.7982  43.4572  365.1572  25.610
```

fubuloubu (dismissed):

PR #2203, opened 2024-08-02
refactor: share hex-related validators

antazoey:
### What I did

noticed bugs and also lack of decent use of annotated-validators for hex-to-int situations, so made `HexInt` for that - likely will end up in `eth-pydantic-types` after @bitwise-constructs work is complete there.
also had to delete some unnecessary validators for hex-str and hexbytes as it happens automatically when using those types.

note, did this originally because there was a regression i think with a recent merge validating BlobReceipts, somehow it occasionally go a hex-str for an int there and I was going to add the validator but I am getting tired of repeating this validator everywhere, hence HexInt's birthday

### How to verify it

things still work (only better in the case of Blob receipts)!

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -4,8 +4,8 @@ from collections.abc import Iterator
 from datetime import datetime
 from typing import IO, TYPE_CHECKING, Any, NoReturn, Optional, Union
 
-from eth_pydantic_types import HexBytes
-from eth_utils import is_0x_prefixed, is_hex, to_hex, to_int
+from eth_pydantic_types import HexBytes, HexStr
+from eth_utils import is_hex, to_hex, to_int
 from ethpm_types.abi import EventABI, MethodABI
 from pydantic import ConfigDict, field_validator
 from pydantic.fields import Field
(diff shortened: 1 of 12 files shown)


src/ape/api/transactions.py
```diff
@@ -51,19 +52,19 @@ class TransactionAPI(BaseInterfaceModel):
...
     """
 
-    chain_id: Optional[int] = Field(default=0, alias="chainId")
+    chain_id: Optional[HexInt] = Field(default=0, alias="chainId")
     receiver: Optional[AddressType] = Field(default=None, alias="to")
     sender: Optional[AddressType] = Field(default=None, alias="from")
-    gas_limit: Optional[int] = Field(default=None, alias="gas")
-    nonce: Optional[int] = None  # NOTE: `Optional` only to denote using default behavior
-    value: int = 0
+    gas_limit: Optional[HexInt] = Field(default=None, alias="gas")
```
antazoey, line 58: note, these are not breaking changes (I added a test to make sure).
`HexInt` is only an annotated int for Pydantic, doesn't  change any serialization types or even anything the API is going to compare about.

I added a test just in case in `test_transaction` to prove this!

fubuloubu (dismissed):

fubuloubu (approved):
This is interesting:
https://peps.python.org/pep-0593/#aliases-concerns-over-verbosity

PR #2204, opened 2024-08-02
fix: readme note

dtdang:
### What I did
Note in README wasn't rendering properly.

fixes: #

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,10 +95,9 @@ Ape's modular plugin system allows users to have an interoperable experience wit
 
 - Learn more about **developing** your own plugins from this [developing user guide](https://docs.apeworx.io/ape/stable/userguides/developing_plugins.html).
 
-> [!NOTE]
+> \[!NOTE\]
 > If a plugin does not originate from the [ApeWorX GitHub Organization](https://github.com/ApeWorX?q=ape&type=all), you will get a warning about installing 3rd-party plugins.
-Install 3rd party plugins at your own risk.
-
+> Install 3rd party plugins at your own risk.
 
 [accounts-guide]: https://docs.apeworx.io/ape/stable/userguides/accounts.html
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,10 +95,9 @@ Ape's modular plugin system allows users to have an interoperable experience wit
 
 - Learn more about **developing** your own plugins from this [developing user guide](https://docs.apeworx.io/ape/stable/userguides/developing_plugins.html).
 
-```{note}
-If a plugin does not originate from the [ApeWorX GitHub Organization](https://github.com/ApeWorX?q=ape&type=all), you will get a warning about installing 3rd-party plugins.
-Install 3rd party plugins at your own risk.
-```
+> \[!NOTE\]
```
fubuloubu, line 98: Oh nice! Is this specific to Github markdown, or does it also work with sphinx-markdown?

fubuloubu (approved):

README.md
dtdang: This one is specific for Github markdown. I don't think it works the same with sphinx-markdown.

fubuloubu: Remember that this is the quickstart guide for our docs as well

dtdang: image

This will render in the docs like this which I think is still a good way of doing this since it makes it more clear on github about the note and is still clear enough when rendered with sphinx.

fubuloubu: Personally I love the note highlighting we have in the docs now, I would love to have it render in github better but I think the docs rendering takes precedence over github

fubuloubu:
As a side note, I'm very in favor of this type of change with the addition of an extension or plugin that converts the Github alert syntax (using markdown-it) to the equivalent HTML or directive syntax for Sphinx-flavored markdown (which uses CommonMark)

As a reference
https://github.com/antfu/markdown-it-github-alerts

PR #2208, opened 2024-08-06
perf: avoid enriching entire trace when only requesting `return_value`

antazoey:
### What I did

metrics from running a test in a real project with lots of stuff going on

before:
`107.63s`

after:
`28.28s`

### How I did it

Use raw trace frames carefully if have not already enriched.

### Checklist

### DIFF
--- a/tests/functional/geth/test_trace.py
+++ b/tests/functional/geth/test_trace.py
@@ -33,16 +33,16 @@ ContractA\.methodWithoutArguments\(\) -> 0x[A-Fa-f0-9]{2}..[A-Fa-f0-9]{4} \[\d+ 
 │   └── ContractC\.methodC1\(
 │         windows95="simpler",
 │         jamaica=345457847457457458457457457,
-│         cardinal=ContractA
+│         cardinal=Contract[A|C]
 │       \) \[\d+ gas\]
 ├── SYMBOL\.callMe\(blue=tx\.origin\) -> tx\.origin \[\d+ gas\]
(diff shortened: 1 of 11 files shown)


tests/functional/geth/test_trace.py
```diff
@@ -307,3 +307,20 @@ def test_call_trace_supports_debug_trace_call(geth_contract, geth_account):
...
+    trace = receipt.trace
+    expected = [1, 2, 3]  # Hardcoded in contract
+    assert receipt.return_value == expected
+
+    # In `trace.return_value`, it is still a tuple.
+    # (unlike receipt.return_value)
+    actual = trace.return_value[0]
+    assert actual == expected
+
+    # NOTE: This is very important from a performance perspective!
```
antazoey, line 324: i wasnt able to do benchmarking because the performance problems only arise in realistic projects with tons of stuff going on.

so i added this assertion to ensure we can get the return value w/o enrichment, that should do it

fubuloubu (approved):

tests/functional/geth/test_trace.py
```diff
@@ -307,3 +307,20 @@ def test_call_trace_supports_debug_trace_call(geth_contract, geth_account):
...
+    # (VERY IMPORTANT). We should need to enrich anything.
```
fubuloubu, line 325:
```suggestion
    # (VERY IMPORTANT). We shouldn't need to enrich anything.
```

fubuloubu (approved):

PR #2212, opened 2024-08-08
feat: lower the default LogLevel for `ape test` command

antazoey:
### What I did

The default log level for `ape test` is too  verbose imo,
the amount of logs of deploying txns and stuff is in the way of the stacktraces 

### Checklist

### DIFF
--- a/src/ape_test/_cli.py
+++ b/src/ape_test/_cli.py
@@ -12,6 +12,7 @@ from watchdog import events
 from watchdog.observers import Observer
 
 from ape.cli import ape_cli_context
+from ape.logging import LogLevel
 from ape.utils import ManagerAccessMixin, cached_property
 
 # Copied from https://github.com/olzhasar/pytest-watcher/blob/master/pytest_watcher/watcher.py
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
Making sure I understand this, this only increases the default log level for `ape test`?

antazoey:
> Making sure I understand this, this only increases the default log level for `ape test`?

Correct!

To test (which I just did!), make a script like:

```python

def main():
    logger.info("test")
    logger.error("test2")
```

Notice it still logs the info! (and the error of course)

Then, make a test like:

```python

def test_log():
    logger.info("test")
    logger.info("test")
    logger.warning("asdf")
    logger.error("ASDFASDFASDFASDFASDFASDF")
    logger.info("test")
    assert False
```

The captured logs look like:

```
----------------------------------- Captured log call -----------------------------------
WARNING  ape:test_showt.py:6 asdf
ERROR    ape:test_showt.py:7 ASDFASDFASDFASDFASDFASDF
```

antazoey:
and then `ape test -v info` will all the logs (from example above)

Issue #2215, opened 2024-08-12
RPC method `debug_setHead` being called during pytest execution, not working on Gnosis (chainId=100)


gabrielfior:
### Environment information

- OS: macOS
- Python Version: 3.11.9
- `ape` and plugin versions:
ape 0.8.11
ape-foundry 0.8.3

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml

name: prediction-market-agent-tooling

plugins:
  - name: foundry

default_ecosystem: ethereum

node:
  gnosis:
    mainnet:
      uri: MY-RPC


networks:
  custom:
     - name: mainnet                   # Required
       chain_id: 100                   # Required
       ecosystem: gnosis
       base_ecosystem_plugin: ethereum
       default_provider: node


foundry:
  host: http://localhost:8545/
  fork:
    gnosis:
      mainnet:
        upstream_provider: node

```

### What went wrong?

I'm trying to run pytest with a forked Gnosis chain as environment (using ape-foundry), so that `ape` handles `anvil` in a seamless way.
I wrote a pytest fixture like below:
```
@pytest.fixture(scope="class")
def local_web3(chain: ChainManager) -> t.Generator[ProviderAPI, None, None]:
    with chain.network_manager.fork(provider_name="foundry") as provider:
        w3 = Web3(Web3.HTTPProvider(provider.http_uri))
        yield w3

    # clean-up
    # chain.network_manager.active_provider.disconnect()
```
If I use the fixture above as is, it throws an error, see below

```
E               ape.exceptions.APINotImplementedError: RPC method 'debug_setHead' is not implemented by this node instance.

../../..//lib/python3.11/site-packages/ape_ethereum/provider.py:1121: APINotImplementedError
```

I tried adding the commented out line (`disconnect`) which solves the error - however, this makes the fixture become `per-function` based, since every time I call the fixture, it gets disconnected during clean-up. I wanted it to only disconnect in the very end.
Note that I'm aware (from the docs) that I should not call `disconnect` directly.

Please include information like:

### How can it be fixed?

I also tried overwriting the method `chain.restore = lambda x: do_nothing()`, but it's also a hacky solution.

Thank you for your guidance.

antazoey:
To fix, we need try / except the line in `ape-node` that is

```
self.make_request("debug_setHead", [block_number_hex_str])
```

[...]

PR #2217, opened 2024-08-13
fix: adds exception check for debug_setHead

bitwise-constructs:
### What I did
adds exception check for debug_setHead when dealing with nodes that don't support it

fixes: #2215
Fixes: APE-1797

### How I did it

added a cached property to provider for skipping debug_setHead restore if not supported

### How to verify it

run on a forked gnosis chain

### Checklist

### DIFF
--- a/src/ape_node/provider.py
+++ b/src/ape_node/provider.py
@@ -289,6 +289,7 @@ class NodeSoftwareNotInstalledError(ConnectionError):
 # NOTE: Using EthereumNodeProvider because of it's geth-derived default behavior.
 class GethDev(EthereumNodeProvider, TestProviderAPI, SubprocessProvider):
     _process: Optional[GethDevProcess] = None
+    _supports_debug_set_head: bool = True
     name: str = "node"
 
     @property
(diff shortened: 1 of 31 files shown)


src/ape_node/provider.py
```diff
@@ -405,7 +409,10 @@ def restore(self, snapshot_id: SnapshotID):
             logger.error("Unable to set head to future block.")
             return
 
-        self.make_request("debug_setHead", [block_number_hex_str])
+        try:
+            self.make_request("debug_setHead", [block_number_hex_str])
+        except Exception:
+            self._supports_debug_set_head = False
```
fubuloubu, line 415: Should this raise `NotImplementedError` instead at this point?

src/ape_node/provider.py
```diff
@@ -405,7 +409,10 @@ def restore(self, snapshot_id: SnapshotID):
             logger.error("Unable to set head to future block.")
             return
 
-        self.make_request("debug_setHead", [block_number_hex_str])
+        try:
+            self.make_request("debug_setHead", [block_number_hex_str])
+        except Exception:
```
antazoey, line 414:
```suggestion
        except NotImplementedError:
```

src/ape_node/provider.py
antazoey: that is actually the exception that was already be raised!

We could move the handling `chain.snapshot` though: https://github.com/ApeWorX/ape/blob/main/src/ape/managers/chain.py#L1607

src/ape_node/provider.py
antazoey: Perhaps we could re-raise the exception and also raise during `not self._supports_debug_set_head:` and then catch `NotImplementedError` in `chain.snapshot` method instead

antazoey: or perhaps the handling should just happen in he pytest integration

antazoey (approved):

Issue #2143, opened 2024-06-14
feat: create a bare ape image


johnson2427:
### Overview

Create a bare ape image w/o the recommended plugins, and refactor the silverack image to reference this new image as the base image.

### Specification

Either delete lines 35-39, or create a new dockerfile `Dockerfile.base` with those lines removed. Then add a new workflow to publish this docker image to the ghcr.

PR #2219, opened 2024-08-13
feat: adding slim dockerfile

johnson2427:
### What I did

Created Dockerfile.slim for a slim version of ape without recommended-plugins

fixes: #2143 

### How I did it

Needed to create the whl files with python:3.11, then switch to python:3.11-slim after the whl files are created. This install ends up being 500MB in size vs 2.24GB in the current Dockerfile build.

I then install the recommended plugins in the Dockerfile for the latest version. The size of the image reduced from 2.24GB to 571MB.

### How to verify it

```bash
git checkout feat/create-bare-ape-image
docker build -t ape:latest-slim -f Dockerfile.slim .
docker image ls
...
...      ...               ...   SIZE
ape   latest-slim                500MB
ape   latest                     571MB
```

### Checklist

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,9 +25,6 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Checkout
-        uses: actions/checkout@v3
-
       - name: Get tags
         run: git fetch --tags origin
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -59,8 +59,6 @@ jobs:
           file: ./Dockerfile.slim
           tags: ${{ steps.meta.outputs.tags }}-slim
           labels: ${{ steps.meta.outputs.labels }}
-          build-args: |
-            VERSION=${{ version }}
 
       - name: Build and push
         uses: docker/build-push-action@v4
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Get current date and time
         id: build_date
-        run: echo 'BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")' >> $GITHUB_ENV
+        run: echo "BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_ENV
 
       - name: Get commit sha
         id: commit_sha
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,6 +32,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
+            type=ref,event=pr
             type=ref,event=branch,branch=main,latest=true
             type=ref,event=branch,branch=main
             type=ref,event=branch,branch!=main
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-tags: 'true'
+          fetch-depth: 0
 
       - name: PRINT GIT TAG
         run: git tag
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,6 +53,8 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}-slim
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache,mode=max
 
       - name: Build and push
         uses: docker/build-push-action@v4
@@ -62,5 +64,7 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-
-
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache,mode=max
+          build-args: |
+            SLIM_IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.meta.outputs.tags }}-slim
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -41,7 +41,7 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}
 
       - name: PRINT METADATA TAGS
-        run: echo "TAGS: ${{ steps.meta.outputs.tags }}"
+        run: echo ${{ steps.meta.outputs.tags }}
 
       - name: Log into GitHub Container Registry
         if: github.event_name != 'pull_request'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,10 +32,16 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=branch
-            type=ref,event=pr
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
+            type=ref,event=branch,branch=main,latest=true
+            type=ref,event=branch,branch=main
+            type=ref,event=branch,branch!=main
+            type=semver,pattern={{version}},branch=main,latest=false
+            type=ref,event=tag
+
+      - name: Show tags
+        run: |
+          git tag
+          echo ${{ steps.meta.outputs.tags }}
 
       - name: Log into GitHub Container Registry
         if: github.event_name != 'pull_request'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,6 +37,9 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
 
+      - name: Set up Docker Buildkit
+        run: echo "DOCKER_BUILDKIT=1" >> $GITHUB_ENV
+
       - name: Cache Docker Layers
         uses: actions/cache@v4
         with:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -61,8 +61,6 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
             VERSION=${{ version }}
-            BUILD_DATE=${{ github.event.repository.updated_at }}
-            VCS_REF=${{ github.sha }}
 
       - name: Build and push
         uses: docker/build-push-action@v4
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,6 +25,12 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Get tags
+        run: git fetch --tags origin
+
       - name: Build and push slim
         uses: docker/build-push-action@v4
         with:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,17 +37,6 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
 
-      - name: Set up Docker Buildkit
-        run: echo "DOCKER_BUILDKIT=1" >> $GITHUB_ENV
-
-      - name: Cache Docker Layers
-        uses: actions/cache@v4
-        with:
-          path: /tmp/docker-cache
-          key: ${{ runner.os }}-docker-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-docker-
-
       - name: Log into GitHub Container Registry
         if: github.event_name != 'pull_request'
         uses: docker/login-action@v2
@@ -61,11 +50,8 @@ jobs:
         with:
           context: .
           file: ./Dockerfile.slim
-          push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}-slim
           labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=local,src=/tmp/docker-cache
-          cache-to: type=local,dest=/tmp/docker-cache
 
       - name: Build and push
         uses: docker/build-push-action@v4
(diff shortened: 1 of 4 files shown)


Dockerfile.slim
```diff
@@ -0,0 +1,42 @@
...
+
+COPY . .
+
+RUN pip wheel .
+
+FROM python:${PYTHON_VERSION}-slim
+
+# See http://label-schema.org for metadata schema
+# TODO: Add `build-date` and `version`
+LABEL maintainer="ApeWorX" \
```
antazoey, line 22: Question: Would there be a nice way to share Dockerfile-logic between this one and the other one? I noticed they do some of the same things, and I don't want to update one and forget to update the other one.

antazoey (approved):
One question, with an idea for potential improvement, but otherwise LGTM

Dockerfile.slim
fubuloubu: I think as long as `-slim` was built first, that this one could use `-slim` as it's base to install `recommended-plugins` over

johnson2427: @fubuloubu you are correct, I'm just working on the implementation locally to ensure I'm not missing anything

johnson2427: I'm having issues keeping this in the python:3.11-slim though, which is the hold up

johnson2427:
@fubuloubu make sure this works the way it did before. I checked the pip list, and for some reason this is installing eth-ape in two places, so I'm still trying to figure that out.

Dockerfile.slim
```diff
@@ -0,0 +1,49 @@
...
+
+RUN pip wheel .[recommended-plugins] --wheel-dir=/wheels/recommended_plugins
+
+FROM python:${PYTHON_VERSION}-slim
+
+# See http://label-schema.org for metadata schema
+# TODO: Add `build-date` and `version`
+LABEL maintainer="ApeWorX" \
+      org.label-schema.schema-version="2.0" \
+      org.label-schema.name="ape-slim" \
```
antazoey, line 30: this may need to be a parameter, as it will be the same otherwise for both slim and fat

johnson2427: I think it should be the same for both

antazoey: Ya that makes sense. Alternatively, you can exclude from the slim

Dockerfile.slim
```diff
@@ -0,0 +1,49 @@
...
+RUN mkdir /wheels/recommended_plugins
+
+COPY ./recommended-plugins.txt ./recommended-plugins.txt
+
+RUN pip wheel .[recommended-plugins] --wheel-dir=/wheels/recommended_plugins
+
+FROM python:${PYTHON_VERSION}-slim
+
+# See http://label-schema.org for metadata schema
+# TODO: Add `build-date` and `version`
```
fubuloubu, line 27: Wonder if OCI is better:
https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/

Dockerfile.slim
```diff
@@ -0,0 +1,49 @@
...
+LABEL maintainer="ApeWorX" \
+      org.label-schema.schema-version="2.0" \
+      org.label-schema.name="ape" \
+      org.label-schema.description="Ape Ethereum Framework." \
+      org.label-schema.url="https://docs.apeworx.io/ape/stable/" \
+      org.label-schema.usage="https://docs.apeworx.io/ape/stable/userguides/quickstart.html#via-docker" \
+      org.label-schema.vcs-url="https://github.com/ApeWorX/ape" \
+      org.label-schema.docker.cmd="docker run --volume $HOME/.ape:/home/harambe/.ape --volume $HOME/.vvm:/home/harambe/.vvm --volume $HOME/.solcx:/home/harambe/.solcx --volume $PWD:/home/harambe/project --workdir /home/harambe/project apeworx/ape compile"
```
fubuloubu, line 35: This is just a general guide? Think this can be shortened
```suggestion
      org.label-schema.docker.cmd="docker run --volume $PWD:/home/harambe/project --workdir /home/harambe/project apeworx/ape run my_script.py"
```

Or something like that since we are not bundling any compilers here

Dockerfile.slim
```diff
@@ -0,0 +1,49 @@
...
+
+RUN useradd --create-home --shell /bin/bash harambe
+COPY --from=builder /wheels/*.whl /wheels/
+
+RUN pip install --upgrade pip
+RUN pip install /wheels/*.whl
+
+COPY --from=builder /wheels/recommended_plugins/*.whl /wheels/
```
fubuloubu, line 43:
```suggestion
```

Dockerfile
fubuloubu, line 1: This one should keep the stuff particular to `recommended-plugins`

fubuloubu (commented):
Keep coming back to our [github-action](https://github.com/ApeWorX/github-action) and wondering if there is a good way to make the `recommended-plugins` image work with the action to speed that up

Sorry about all the comments, didn't realize just how out of date the old dockerfile was

.github/workflows/build.yaml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Get tags
+        run: git fetch --tags origin
```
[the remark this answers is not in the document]
antazoey: There is a `fetch-tags` parameter on the checkout action we could use.
Tags are needed for setuptools-scm to install the correct version.

Dockerfile.slim
johnson2427: I'll try to find a way to do this outside of slim, I'm currently creating the wheels for recommended plugins, but not installing them.

We install them in the Dockerfile. It isn't ideal, I'll fix this

.github/workflows/build.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Build and push slim
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile.slim
+          push: true
```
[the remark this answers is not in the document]
johnson2427: absolutely, will add when I get the build and push to work

.github/workflows/build.yaml
```diff
@@ -0,0 +1,50 @@
...
+          tags: ${{ env.REGISTRY }}/apeworx/ape:latest-slim
```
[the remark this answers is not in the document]
johnson2427: we're going to create 2 tags here. One will be latest-slim and then a second for the version release (0.8.13-slim). That way, you'll reach out to latest for this release, when the next release comes out, the `latest-slim` tag will be replaced, but the 0.8.13-slim will remain in place.

Dockerfile.slim
johnson2427: This has been updated. Slim builds with `Dockerfile.slim`, recommended plugins is built in the `Dockerfile`. The logic is tidied up here now.

.github/workflows/build.yaml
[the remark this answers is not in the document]
johnson2427: added this, checking that it works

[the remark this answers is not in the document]
johnson2427: This worked!

.github/workflows/build.yaml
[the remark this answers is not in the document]
fubuloubu: oh, usually I have:
- `latest` = whatever is head on `main` branch
- `stable` = last release
- `vX.Y.Z` = specific release

and then to that we do `{tag}-slim` for the slim build

.github/workflows/build.yaml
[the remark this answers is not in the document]
johnson2427: @fubuloubu I **think** this is what we want to do. Letting this build now. The only issue is, I'll need the fat image to pull the slim image. I have to look at letting it reference the build rather than pulling from ghcr

.github/workflows/build.yaml
```diff
@@ -0,0 +1,65 @@
...
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+
```
fubuloubu, line 39: Add a debug step to see the tags and labels

.github/workflows/build.yaml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push slim
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile.slim
```
fubuloubu, line 52: Forgot this?
```suggestion
          file: ./Dockerfile.slim
          push: ${{ github.event_name != 'pull_request' }}
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,65 @@
...
+          tags: ${{ steps.meta.outputs.tags }}-slim
```
fubuloubu, line 53: Does this produce two tags when published? (`v0.X.y` and `stable`)

.github/workflows/build.yaml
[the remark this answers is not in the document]
johnson2427: There is no way to use the docker engine with cache in github workflows. So I have to push the slim. I don't have to push a stable. Slim will push with a pr-xxxx label

.github/workflows/build.yaml
johnson2427: I can't build the fat image without the slim being built with our current setup. If we want to do that, the Dockerfile will have to duplicate the code from Dockerfile.slim

.github/workflows/build.yaml
johnson2427: Hard to know at the moment. What I did should work, but I'm not certain until we push.

Dockerfile.slim
johnson2427: Let me know what you want to do here

antazoey: Yes, OCI is more standard now. Also http://label-schema.org is does not work on https which is sus

fubuloubu: It also just says it is deprecated and use OCI

johnson2427: OCI added

johnson2427:
This is what we end up with
```bash
            "build-arg:BUILD_DATE": "2024-08-16T18:48:15Z",
            "build-arg:VCS_REF": "c3eada713fcf78b7fba8d82571298869883ffb14",
            "build-arg:VERSION": "",
            "label:org.opencontainers.image.created": "2024-08-16T19:13:35.036Z",
            "label:org.opencontainers.image.description": "The smart contract development tool for Pythonistas, Data Scientists, and Security Professionals",
            "label:org.opencontainers.image.licenses": "Apache-2.0",
            "label:org.opencontainers.image.revision": "c3eada713fcf78b7fba8d82571298869883ffb14",
            "label:org.opencontainers.image.source": "https://github.com/ApeWorX/ape",
            "label:org.opencontainers.image.title": "ape",
            "label:org.opencontainers.image.url": "https://github.com/ApeWorX/ape",
            "label:org.opencontainers.image.version": "pr-2219"
```

johnson2427:
I need to fix the documentation for the `Intalling with 'docker'` section, but I'm having a hard time doing so without publishing the images. I think we should update docs in a separate PR, but I'll leave that up to you two

fubuloubu:
> I need to fix the documentation for the `Intalling with 'docker'` section, but I'm having a hard time doing so without publishing the images. I think we should update docs in a separate PR, but I'll leave that up to you two

How do you think it will be different?

johnson2427:
> > I need to fix the documentation for the `Intalling with 'docker'` section, but I'm having a hard time doing so without publishing the images. I think we should update docs in a separate PR, but I'll leave that up to you two
> 
> How do you think it will be different?

We are no longer going to be pushing to dockerhub from what I remember in our conversation from earlier this week. I'd like to give the `ghcr` location for users.

fubuloubu:
> > > I need to fix the documentation for the `Intalling with 'docker'` section, but I'm having a hard time doing so without publishing the images. I think we should update docs in a separate PR, but I'll leave that up to you two
> > 
> > How do you think it will be different?
> 
> We are no longer going to be pushing to dockerhub from what I remember in our conversation from earlier this week. I'd like to give the `ghcr` location for users.

Ah yes, for sure. I think what we can do with this PR is stop publishing to dockerhub and start publishing to ghcr, and then update the instructions before we create a release containing this update. Once merged, I will stop automated publishing on dockerhub, and we should also port some of the older images to ghcr as well to ensure continuity for any users, and then make a public announcement for that as well

johnson2427:
Currently, from the automated metadata creation, this is where our image will go `ghcr.io/apeworx/ape`. I believe you want this to go to `ghcr.io/apeframework/ape`, correct @fubuloubu?

fubuloubu:
> Currently, from the automated metadata creation, this is where our image will go `ghcr.io/apeworx/ape`. I believe you want this to go to `ghcr.io/apeframework/ape`, correct @fubuloubu?

No! That place is good, wanted it @ ghcr.io/apeworx/ape

README.md
```diff
@@ -59,7 +59,31 @@ There are three ways to install ape: `pipx`, `pip`, or `Docker`.
 
 Ape can also run in a docker container.
 
-Please visit our [Dockerhub](https://hub.docker.com/repository/docker/apeworx/ape) for more details on using Ape with Docker.
+You can pull our images from [ghcr](https://ghcr.io/apeworx/ape). We currently have a `slim` docker image as well as an image that is built with the recommended-plugins.
```
fubuloubu, line 62:
```suggestion
You can pull our images from [ghcr](https://ghcr.io/apeworx/ape).
We currently have a `slim` docker image that is built without any installed plugins, as well as an image that is built with all plugins included from our `recommended-plugins` extra.
```

README.md
```diff
@@ -59,7 +59,31 @@ There are three ways to install ape: `pipx`, `pip`, or `Docker`.
...
+
+You can pull the image:
+
+```bash
+docker pull ghcr.io/apeworx/ape:latest-slim  # installs ape with required packages
+```
+
+or pull with the recommended-plugins:
```
fubuloubu, line 70: I think put the other image first, and not recommended to install w/ `slim` unless you need to

johnson2427: lmk if that's what you're looking for

README.md
```diff
@@ -59,7 +59,29 @@ There are three ways to install ape: `pipx`, `pip`, or `Docker`.
 
 Ape can also run in a docker container.
 
-Please visit our [Dockerhub](https://hub.docker.com/repository/docker/apeworx/ape) for more details on using Ape with Docker.
+You can pull our images from [ghcr](https://ghcr.io/apeworx/ape).
+We currently have a `slim` docker image that is built without any installed plugins, as well as an image that is built with all plugins included from our `recommended-plugins` extra.
```
fubuloubu, line 63:
```suggestion
This image is built using our `recommended-plugins` extra, so it is a great starting point for running ape in a containerized environment.

We also have a `slim` docker image that is built without any installed plugins.
This image is meant for production support and must be further configured if any plugins are in use.
```

README.md
```diff
@@ -70,6 +92,9 @@ docker run \
   apeworx/ape compile
 ```
 
+> **Note:**
+> The above command requires the full install with the recommended-plugins.
```
fubuloubu, line 96:
```suggestion
> The above command requires the full install whicy includes `recommended-plugins` installation extra.
```

fubuloubu (approved):
If you feel confident in the workflow, it is ready to go. Would recommend just doing some local testing with a repo under your personal if you've not sure

PR #2221, opened 2024-08-15
fix: issues with `CurrencyComparableValue` appearing `ContractLog.event_arguments` & on Pydantic models

antazoey:
### What I did

The following bug fixes are in this PR:

1. Can now use `CurrencyComparableValue` as a type annotation (And still using comparison-feature!)
2. Can validate CurrencyComparableValue into int declared type
3. ContractLog serialized event_arguments out of this type because of https://github.com/pydantic/pydantic/issues/10152

### How I did it

Implement a custom serializer in the type that treats it like a normal int

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -19,7 +19,7 @@ from ethpm_types import (
 )
 from ethpm_types.abi import EventABI
 from ethpm_types.source import Closure
-from pydantic import BaseModel, BeforeValidator, field_validator, model_validator
+from pydantic import BaseModel, BeforeValidator, field_serializer, field_validator, model_validator
 from pydantic_core.core_schema import (
     CoreSchema,
     ValidationInfo,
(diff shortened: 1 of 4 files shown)


mikeshultz:
Sounds like the pydantic issue is going to take a while. Can we get a model or field serializer on `BaseContractLog` that handles the original situation?

This is what I tested that works, but it's probably not really what we want:

```python
    @model_serializer
    def ser_model(self) -> dict[str, Any]:
        def ser_arg(v):
            if isinstance(v, int):
                return int(v)
            return v

        return {
            "event_name": self.event_name,
            "contract_address": self.contract_address,
            "event_arguments": {k: ser_arg(v) for k, v in self.event_arguments.items()},
        }
```

Bet you have a better idea.

antazoey:
I was having issues with bytes types (even HexBytes) when using `mode="json"` is the `model_dump()`. No matter what I did (e.g. make them hex-strings), I still get invalid codec issues... which is very annoying because it's like wtf is the point of a serializer if I can't handle that, but I am guessing `dict[str, Any]` is just too open-ended for the Pydantic framework. UGh, so annoying... In any case, this PR fixes like 5 bugs.

antazoey:
@mikeshultz  let me know if this resolves your issue

fubuloubu (approved):

mikeshultz:
> @mikeshultz let me know if this resolves your issue

Looks good. Thanks for getting the temporary fix in.

PR #2228, opened 2024-08-21
fix: incorrect image release tags in build

johnson2427:
### What I did

The tag that is created when merged to `main` is `main`. We want this to be `latest` and then on version release, we want this to be `stable`

fixes: APE-1769

### How I did it

Added the tag to the build yaml

### How to verify it

After merge, check the release

### Checklist

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=pr,tag=test
+            type=ref,event=pr,pattern=test
             type=ref,event=branch,branch=main,tag=latest
             type=ref,event=tag,tag=stable
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,10 +34,10 @@ jobs:
           tags: |
             type=ref,event=pr
             type=ref,event=branch,branch=main,latest=true
-            type=ref,event=branch,branch=main
+            type=ref,event=branch,branch=main,tag=latest
             type=ref,event=branch,branch!=main
-            type=semver,pattern={{version}},branch=main,latest=false
-            type=ref,event=tag
+            type=semver,pattern={{version}},branch=main,latest=false,tag=stable
+            type=ref,event=tag,tag={{ tag }}
 
       - name: Show tags
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -61,6 +61,7 @@ jobs:
         with:
           context: .
           file: ./Dockerfile.slim
+          push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}-slim
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,10 +32,9 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
+            type=raw,value=latest
             type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
             type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/') }}
-            type=ref,event=tag
-            type=ref,event=pr
 
       - name: Show tags
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=pr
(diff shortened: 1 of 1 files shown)


antazoey:
I am guessing this is container-related?

johnson2427:
> I am guessing this is container-related?

yes it is, not getting the right tags on merge to main

johnson2427:
These changes **should** do what we're looking for. I'm not familiar with tagging in github, so I'm riding on documentation, but still a little unclear on this

.github/workflows/build.yaml
```diff
@@ -34,10 +34,10 @@ jobs:
           tags: |
             type=ref,event=pr
             type=ref,event=branch,branch=main,latest=true
-            type=ref,event=branch,branch=main
+            type=ref,event=branch,branch=main,tag=latest
             type=ref,event=branch,branch!=main
-            type=semver,pattern={{version}},branch=main,latest=false
-            type=ref,event=tag
+            type=semver,pattern={{version}},branch=main,latest=false,tag=stable
+            type=ref,event=tag,tag={{ tag }}
```
fubuloubu, line 40: Not quite familiar with how this works

fubuloubu (dismissed):

fubuloubu:
Noticed that the untagged versions were getting, did https://github.com/ApeWorX/ape/pull/2228/commits/a597ba9d13b9e946615df747fb5789501ad8254d resolve that?

fubuloubu:
Also can you delete this?
https://github.com/orgs/ApeWorX/packages/container/ape/261405854?tag=main

johnson2427:
> Noticed that the untagged versions were getting pushed from this PR, did [a597ba9](https://github.com/ApeWorX/ape/commit/a597ba9d13b9e946615df747fb5789501ad8254d) resolve that?
> 
> ref: https://github.com/ApeWorX/ape/pkgs/container/ape/261370690

I'm reviewing docs right now, this isn't supposed to push unless it's merged to main

fubuloubu:
Lastly, would love it if the untagged versions didn't get published at all, feel like that bloats things

.github/workflows/build.yaml
fubuloubu, line 1: I think what you could do for releases is run an extra optional step (after both `-slim` and the normal) to make sure whatever image hash that `stable` gets built as is also tagged with the corresponding `v*` tag as well

.github/workflows/build.yaml
```diff
@@ -32,12 +32,9 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=pr
-            type=ref,event=branch,branch=main,latest=true
-            type=ref,event=branch,branch=main
-            type=ref,event=branch,branch!=main
-            type=semver,pattern={{version}},branch=main,latest=false
-            type=ref,event=tag
+            type=ref,event=pr,pattern=test
```
fubuloubu, line 35: what does this do? just a random non-committed tag for use inside the workflow?

fubuloubu (approved):

.github/workflows/build.yaml
johnson2427: just trying to see what results I get when I change things so I know what it'll look like when we merge

johnson2427:
> Lastly, would love it if the untagged versions didn't get published at all, feel like that bloats things

you don't want to release unless it's a tagged version? So don't release on merge to main?

fubuloubu:
> > Lastly, would love it if the untagged versions didn't get published at all, feel like that bloats things
> 
> you don't want to release unless it's a tagged version? So don't release on merge to main?

No no, those random untagged releases which are image hashes. Don't know where those come from

johnson2427:
> > > Lastly, would love it if the untagged versions didn't get published at all, feel like that bloats things
> > 
> > 
> > you don't want to release unless it's a tagged version? So don't release on merge to main?
> 
> No no, those random untagged releases which are image hashes. Don't know where those come from

I *think* this is resolved. I don't see it in the package releases. And I'm no longer pushing until we merge to main. We should be okay here

.github/workflows/build.yaml
johnson2427: To be clear, we'd want a `stable` and `vX.X.X` release on a version tag release

fubuloubu:
> > > > Lastly, would love it if the untagged versions didn't get published at all, feel like that bloats things
> > > 
> > > 
> > > you don't want to release unless it's a tagged version? So don't release on merge to main?
> > 
> > No no, those random untagged releases which are image hashes. Don't know where those come from
> 
> I *think* this is resolved. I don't see it in the package releases. And I'm no longer pushing until we merge to main. We should be okay here

So we can get this merged to see?

fubuloubu (approved):

.github/workflows/build.yaml
fubuloubu: Yes that's correct. The stable always points at the same image as the latest tagged release, but if you use a release version then it never changes

johnson2427: np, let me add this

johnson2427:
If this does not work, I'm removing the metadata step and forcefully injecting the tags

fubuloubu (approved):

PR #2230, opened 2024-08-21
fix: contract retrieval from ABI fails from fetch-attempt

bitwise-constructs:
### What I did

improved error checking when using just an abi to retrieve contract

### Checklist


tests/functional/test_chain.py
```diff
@@ -159,3 +159,12 @@ def test_set_pending_timestamp_failure(chain):
 def test_set_balance(chain, owner):
     with pytest.raises(APINotImplementedError):
         chain.set_balance(owner, "1000 ETH")
+
+
+def test_get_contract_by_abi(chain, vyper_contract_instance):
```
antazoey, line 164: so we have basically this exact test right here: https://github.com/ApeWorX/ape/blob/main/tests/functional/test_contract.py#L13

and also here:
https://github.com/ApeWorX/ape/blob/main/tests/functional/test_contracts_cache.py#L91

i think we need to still re-create, even if mocking required, the new condition you are adding

Also,

antazoey (commented):
the fix makes sense but the test doesn't

antazoey:
Oh I finally see the problem. We are loading ABI from the str but never initializing a ContractType and then we reach the condition causing the failure, probably a result of a bugfix

PR #2233, opened 2024-08-22
feat: make the dependency manager accessible on `ManagerAccessMixin`

antazoey:
### What I did

I noticed when trying to manager global dependencies in python, it can be awkward referencing the local project to do this.
this makes it less awkward.

### Checklist


fubuloubu (approved):
Are there other places where this would be useful?

antazoey:
> Are there other places where this would be useful?

I don't think so, most everything uses the ManagerAccessMixin, so this basically fixes it all in 1 spot!

PR #2235, opened 2024-08-22
test: show we can have sets of projects

antazoey:
### What I did

I had in my notes to fix this but it doesn't seem like a problem. In any case, there were no tests for this so i am just making sure...

### Checklist

### DIFF
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -627,6 +627,22 @@ class TestProject:
         assert "site-packages" in str(web3.path)
         assert web3.path.name == "web3"
 
+    def test_hash(self, with_dependencies_project_path, project_from_manifest):
+        """
+        Show we can use projects in sets.
+        """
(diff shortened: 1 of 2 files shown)


tests/functional/test_project.py
```diff
@@ -627,6 +627,22 @@ def test_from_python_library(self):
         assert "site-packages" in str(web3.path)
         assert web3.path.name == "web3"
 
+    def test_hash(self, with_dependencies_project_path, project_from_manifest):
+        """
+        Show we can use projects in sets.
+        """
+        project_0 = Project(with_dependencies_project_path)
+        project_1 = Project.from_python_library("web3")
```
fubuloubu, line 635: does web3 provide manifests?

fubuloubu (approved):

tests/functional/test_project.py
antazoey: no, but you can create a "project" from any python package this way, even if empty.

PR #2236, opened 2024-08-22
test: showing dependencies works in `project.extract_manifest()`

antazoey:
### What I did

i heard there was a bug here but not sure what it is, here is a test proving dependencies works in package manifest.
lemme know how to add more conditions to reach a failing state

### Checklist

### DIFF
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -261,11 +261,41 @@ def test_extract_manifest(tmp_project, mock_sepolia, vyper_contract_instance):
     assert PackageName("manifest-dependency") in (manifest.dependencies or {})
     bip122_chain_id = to_hex(tmp_project.provider.get_block(0).hash)
     expected_uri = f"blockchain://{bip122_chain_id[2:]}"
-    for key in manifest.deployments or {}:
-        if key.startswith(expected_uri):
-            return
 
-    assert False, "Failed to find expected deployment URI"
+    # Prove that deployments are available on an extract-manifest.
+    found = False
+    for key, val in (manifest.deployments or {}).items():
+        if key.startswith(expected_uri):
+            assert "VyperContract.json" in val
+            actual_deployment = val["VyperContract.json"]
+            assert actual_deployment.contract_type == "contracts/VyperContract.vy:VyperContract"
+            assert actual_deployment.address == vyper_contract_instance.address
+            found = True
+            break
+
+    assert found, "Failed to find expected deployment URI"
+
+    # Prove that dependencies works.
+    # NOTE: tmp_project has several dependencies!
+    expected_dependencies = (
+        "default",
+        "renamed-contracts-folder",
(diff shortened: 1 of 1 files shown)


tests/functional/test_project.py
```diff
@@ -261,11 +261,41 @@ def test_extract_manifest(tmp_project, mock_sepolia, vyper_contract_instance):
...
+            assert "VyperContract.json" in val
+            actual_deployment = val["VyperContract.json"]
+            assert actual_deployment.contract_type == "contracts/VyperContract.vy:VyperContract"
+            assert actual_deployment.address == vyper_contract_instance.address
+            found = True
+            break
+
+    assert found, "Failed to find expected deployment URI"
+
+    # Prove that dependencies works.
```
fubuloubu, line 278: These are all local package dependencies?

antazoey: yes, but chatting with The God of Ninjas, I have learned to try stuff with snekmate, so going to experiment there.

fubuloubu: It may be the case we just don't have a good way to share non-local dependencies using EthPM, and so we can just ask for additional dependencies when compiling in a more ape-friendly config format to use in his service

antazoey: As long as it's in your config, it should work, it just turns to a URL format.

fubuloubu: But it has to be the URL of an EthPM Manifest (which is not very common)

antazoey: Ok, but it probably wont be a compiled one. most dependencies, including openzeppelin and snekmate, seemingly arent meant to compile on their own but instead offer sources to projects that do compile

antazoey: Why just be a link to a directory containing the files? That is how it works now, and ape puts them in a manifest

fubuloubu: This isn't really how it's meant to work, which is why EthPM needs to be fundamentally re-thought anyways. Dependencies are best as like links to a zip file, ipfs directory, or other filesystem-like resource that is globally reference-able and contains sources to use (and potentially compiler settings, compiled artifacts, and other reference metadata that may be useful too) so a downstream project can describe its build manifest in full

For the `hosted-compiler` we can just use a list of ape-compatible dependency objects which we can push into an ape-config to use to compile with

antazoey: FWIW, Ape is getting better at deducing settings from a project structure, such as Foundry projects.
So if you have a URL, you can fetch it and ape can do `Project(path).extract_manifest()`

antazoey: snekmate includes absolutely nothing besides contracts in their published PyPI package, which is frustrating. I told them that and they didn't care.

fubuloubu: Technically, a Python wheel is a zip file at some URI on the pypi server

PR #2240, opened 2024-08-23
docs: add exclusions and clarify convert examples

dtdang:
### What I did
Noticed that a few methods that should be excluded in the docs weren't being excluded properly. 

fixes: #

### Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -85,7 +85,7 @@ autodoc_default_options = {
     "exclude-members": (
         "__repr__, __weakref__, __metaclass__, __init__, __format__,  __new__, __str__, __dir__,"
         "model_config, model_dump, model_fields, model_post_init, model_computed_fields,"
-        "__ape_extra_attributes__, __get_item__, __contains__, __iter__, __len__, __eq__, __hash__,"
+        "__ape_extra_attributes__, __get_item__, __contains__, __hash__,"
     )
 }
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -85,7 +85,7 @@ autodoc_default_options = {
     "exclude-members": (
         "__repr__, __weakref__, __metaclass__, __init__, __format__,  __new__, __str__, __dir__,"
         "model_config, model_dump, model_fields, model_post_init, model_computed_fields,"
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -84,8 +84,8 @@
 autodoc_default_options = {
     "exclude-members": (
         "__repr__, __weakref__, __metaclass__, __init__, __format__,  __new__, __str__, __dir__,"
-        "model_config, model_fields, model_post_init, model_computed_fields,"
-        "__ape_extra_attributes__,"
+        "model_config, model_dump, model_fields, model_post_init, model_computed_fields,"
+        "__ape_extra_attributes__, __get_item__, __contains__, __iter__, __len__, __eq__, __hash__,"
```
antazoey, line 88: Occasionally we override `__iter__` and `__len__` and `__eq__` and provider docs for these methods, so we don't always want to exclude them

docs/conf.py
```diff
@@ -84,8 +84,8 @@
 autodoc_default_options = {
     "exclude-members": (
         "__repr__, __weakref__, __metaclass__, __init__, __format__,  __new__, __str__, __dir__,"
-        "model_config, model_fields, model_post_init, model_computed_fields,"
-        "__ape_extra_attributes__,"
+        "model_config, model_dump, model_fields, model_post_init, model_computed_fields,"
+        "__ape_extra_attributes__, __get_item__, __contains__, __hash__,"
```
antazoey, line 88: `__contains__` and `__getitem__` may also occasionally be overriden and custom

antazoey: If there are some that are black, for any dunderized methods, you can ignore them per construct directly, just ends up being more verbose, but that is ok

dtdang: If that's the case then just the `model_dump` should be excluded since that's a pydantic model dump and shouldn't be overridden by anything that we do.

antazoey: correct, model_dump should be excluded
but there are maybe some other models where we should exclude `__iter__` and `__contains__` etc. we just cannot do it globally.

antazoey (approved):
LGTM as it is right now

PR #2244, opened 2024-08-23
fix: `.return_value` bugs with tuples, structs, and length

antazoey:
### What I did

Always use call-tree approach since debug_tt approach for complex values seems bugged (for now) and super hard to do correctly as the rhymthe and reason disappears when using hardhat.

### How I did it

rely on calltree

### How to verify it

ugh, see my test, had to copy in a bunch of raw data because was hard to get into to be bad in a simple way.

### Checklist


bitwise-constructs (approved):
looking solid to me

fubuloubu (approved):

fubuloubu (approved):

PR #2246, opened 2024-08-26
feat: expose 'data' (query-manager) as a top-level object

antazoey:
### What I did

Allows

```
from ape import queries
```

### Checklist

### DIFF
--- a/src/ape/__init__.py
+++ b/src/ape/__init__.py
@@ -46,16 +46,13 @@ Contract = chain.contracts.instance_at
 convert = _ManagerAccessMixin.conversion_manager.convert
 """Conversion utility function. See :class:`ape.managers.converters.ConversionManager`."""
 
+data = _ManagerAccessMixin.query_manager
+"""Query chain data."""
+
 reverts = RevertsContextManager
 """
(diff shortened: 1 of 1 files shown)


antazoey:
maybe `querier` makes more sense but not an easy word to remember how to type

src/ape/__init__.py
```diff
@@ -51,6 +51,11 @@
 Catch and expect contract logic reverts. Resembles ``pytest.raises()``.
 """
 
+queries = _ManagerAccessMixin.query_manager
```
fubuloubu, line 54: Random suggestion

```suggestion
data = _ManagerAccessMixin.query_manager
```

fubuloubu (commented):
When I originally added this, I meant it as an internal-only manager to use

What would be the use case for exposing this?

src/ape/__init__.py
fubuloubu: Because then it looks like this:

```py
from ape import data


print(data.engines)

for item in data.query(query):
    ...
```

antazoey:
> What would be the use case for exposing this?

Users just trying to work with query manager more directly, like especially when you are trying to use a specific engine.
One example is even testing a query-engine plugin directly.

PR #2247, opened 2024-08-26
feat: adding retention policy

johnson2427:
### What I did

Adding a retention policy to remove old versions of containers from ape's ghcr

fixes: #

### How I did it

Used [snok](https://github.com/snok/container-retention-policy)

### How to verify it

This is a setup for future, have to build out previous versions and tag them to see if this will work. Currently set to dry-run

### Checklist


.github/workflows/retention.yaml
```diff
@@ -0,0 +1,23 @@
...
+  clean:
+    name: Delete package versions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: snok/container-retention-policy@v3
+        with: 
+          account: ApeWorX
+          token: ${{ secrets.GITHUB_TOKEN }}
+          image-names: "ape"
+          image-tags: "!stable* !latest* !v0.7* !v0.8*"
```
fubuloubu, line 21: our policy would basically be "current and last minor version"?

.github/workflows/build-historical.yaml
fubuloubu, line 1: actually, can this just be done manually?

johnson2427: this is a manual trigger, I don't have permissions to do this from my machine, otherwise, yes, yes it can

johnson2427:
```bash
#!/bin/bash

# Array of tags you want to build and push
tags=(
  "v0.7.0"
  "v0.7.1"
  "v0.7.2"
  # Add other versions here up to v0.8.13
  "v0.8.12"
  "v0.8.13"
)

# GitHub Container Registry info
REPO="ghcr.io/ApeWorX/ape"  # Replace with your actual username and repo
DOCKERFILE_PATH="."  # Path to your Dockerfile, usually the root of the repo

for tag in "${tags[@]}"; do
  echo "Building and pushing $tag..."
  
  # Check out the tag
  git checkout tags/$tag
  
  # Build the Docker image
  docker build -t $REPO:$tag -f $DOCKERFILE_PATH/Dockerfile .
  
  # Tag the image as latest for the most recent tag
  if [[ "$tag" == "v0.8.13" ]]; then
    docker tag $REPO:$tag $REPO:latest
  fi

  # Push the Docker image
  docker push $REPO:$tag
  
  # Push the 'latest' tag if applicable
  if [[ "$tag" == "v0.8.13" ]]; then
    docker push $REPO:latest
  fi
  
  echo "$tag pushed to $REPO"
done

# Clean up
git checkout main
```

Something like this

.github/workflows/retention.yaml
johnson2427: okay, in this instance we'd keep 0.7 and 0.8 as well as stable and latest, correct?

.github/workflows/build.yaml
```diff
@@ -77,3 +76,51 @@ jobs:
...
+
+          # Extract the minor versions (X.Y) from tags in the format vX.X.X
+          latest_two_minors=$(echo "$all_tags" | tr ' ' '\n' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sed -E 's/^v([0-9]+\.[0-9]+)\.[0-9]+$/\1/' | uniq | tail -n 2)
+          echo "Last two minor versions:"
+          echo "$latest_two_minors"
+
+          keep_tags=""
+          for minor in $latest_two_minors; do
+            echo "Processing minor version: $minor"
+            latest_patch=$(echo "$all_tags" | tr ' ' '\n' | grep "^v$minor\." | sort -V | tail -n 1)
```
fubuloubu, line 106: I think we can probably just do wildcard on all patches in the last two minors

johnson2427: I'm not sure what you mean

johnson2427: If I do a wildcard on the last two minors, I'd get all of the patches for those minors. Is that what we want?

fubuloubu: Yes, someone might be expecting a slightly older patch than whatever is latest there

johnson2427: ah, I asked earlier, but inferred incorrectly, I just didn't know if we were trying to reduce space or not. I'll add a wildcard

fubuloubu: last two minors reduces space quite a bit. I think it's a decent policy, I'm not particularly concerned about space yet but wanted to limit the amount of uploads you had to do from dockerhub

fubuloubu (approved):

johnson2427:
@fubuloubu are we holding off on anything to merge this? It won't affect anything, but just want your thoughts

fubuloubu:
> @fubuloubu are we holding off on anything to merge this? It won't affect anything, but just want your thoughts

Seems like we can merge it. Also contains that fix

PR #2248, opened 2024-08-26
fix: make `.extract_manifest()` also compile if needed

antazoey:
### What I did

pre 0.8. this happened, and I don't see why it should not happen, so making it happen again.
(here you go @Ninjagod1251 )

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -2460,9 +2460,7 @@ class LocalProject(Project):
 
         sources = dict(self.sources)
         contract_types = {
-            n: ct
-            for n, ct in self.load_contracts().items()
-            if ct.source_id in sources
+            n: ct for n, ct in self.load_contracts().items() if ct.source_id in sources
         }
 
         # Add any remaining data to the manifest here.
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -2460,7 +2460,9 @@ class LocalProject(Project):
 
         sources = dict(self.sources)
         contract_types = {
-            n: ct for n, ct in self.load_contracts().items() if ct.source_id in sources
+            n: c.contract_type
+            for n, c in self.load_contracts().items()
+            if c.contract_type.source_id in sources
         }
 
         # Add any remaining data to the manifest here.
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -2461,7 +2461,7 @@ class LocalProject(Project):
         sources = dict(self.sources)
         contract_types = {
             n: ct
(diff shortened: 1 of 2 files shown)


antazoey:
Some manual testing showing it working as well:

```python
In [7]: project.manifest.contract_types = {}

In [8]: mani_pedi = project.extract_manifest()
INFO (ape-vyper): Compiling using Vyper compiler '0.4.0'.
Input:
        contracts/Token.vy

In [9]: mani_pedi = project.extract_manifest()

In [10]: mani_pedi = project.extract_manifest()
```

antazoey:
- [ ] Make also happen in manifest projects

src/ape/managers/project.py
```diff
@@ -1891,6 +1891,12 @@ def reconfigure(self, **overrides):
         self.account_manager.test_accounts.reset()
 
     def extract_manifest(self) -> PackageManifest:
+        # Ensure it compiled (at least attempt).
+        try:
+            self.load_contracts()
+        except CompilerError as err:
+            logger.debug(err)
```
fubuloubu, line 1898: Should this be higher? Maybe warning

fubuloubu (dismissed):

src/ape/managers/project.py
antazoey: perhaps! I am just worried about dependencies trying to extract manifests from that aren't meant to be compiled.

antazoey: i suppose they could just `.manifest` in that case.

antazoey: I changed it to warning but added a comment. let me know if that makes sense, along with my concerns.

Ninjagod1251 (approved):

src/ape/managers/project.py
```diff
@@ -2461,7 +2461,7 @@ def extract_manifest(self) -> PackageManifest:
         sources = dict(self.sources)
         contract_types = {
             n: ct
-            for n, ct in (self.manifest.contract_types or {}).items()
+            for n, ct in self.load_contracts().items()
```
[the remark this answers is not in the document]
antazoey: you can delete the `.load_contracts()` line, yes (once 0.8.14 is out).
However, it also doesn't hurt to have it.

src/ape/managers/project.py
fubuloubu: > perhaps! I am just worried about dependencies trying to extract manifests from that aren't meant to be compiled.

Wonder if it makes more sense as like `.enrich_manifest` or `.fully_compiled_manifest`, or something

PR #2249, opened 2024-08-27
fix: `ValueError: Unknown level: 'LOGLEVEL.WARNING'`

antazoey:
### What I did

Fix the bug in 3 spots because am mad:

* Handle stringified `LogLevel` values with LogLevel prefix everywhere
* Use `.value` in `ape test` because is more performant and makes more sense and you don't run into weird issues in some CI/CD environments.
* Validate pytest args by excluding Ape's usage of `-v` (in case it is ending up in there)

### How to verify it

https://github.com/ApeWorX/ApePay/pull/99

### Checklist

### DIFF
--- a/src/ape/cli/options.py
+++ b/src/ape/cli/options.py
@@ -73,10 +73,8 @@ def _create_verbosity_kwargs(
 
     def set_level(ctx, param, value):
         if isinstance(value, str):
-            if "." in value:
-                # Handle "LogLevel" prefix that exists in some environments.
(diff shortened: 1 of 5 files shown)


fubuloubu (approved):

antazoey:
so this didn't fix it yet! but no worries, trying something else

antazoey:
Wow it is still happening.

antazoey:
OK verified the bug is fixed. stomped out of existence.

fubuloubu (approved):

PR #2251, opened 2024-08-27
fix: if only single compiler error, avoid wrapping in a new error

antazoey:
### What I did

Ape shows you all compiler errors across compilers.
That is great, except when there was only a single error, thus causing a pointless exception being thrown on the stack.

### How I did it

if len(errors) is just 1, just raise that error instead.

### How to verify it

Compile with 1 compiler failing - should just see the custom
Compile with 2 compilers failing - should see them amassed

I added tests for both of these cases.

### Checklist

### DIFF
--- a/src/ape/managers/compilers.py
+++ b/src/ape/managers/compilers.py
@@ -157,10 +157,17 @@ class CompilerManager(BaseManager, ExtraAttributesMixin):
                 errors.append(err)
                 continue
 
-        if errors:
+        if len(errors) == 1:
+            # If only 1 error, just raise that.
+            raise errors[0]
+
+        elif len(errors) > 1:
+            # Raise a combined error.
             formatted_errors = [f"{e}" for e in errors]
             error_message = "\n\n".join(formatted_errors)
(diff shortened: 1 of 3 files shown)


src/ape/managers/compilers.py
```diff
@@ -157,11 +157,18 @@ def compile(
...
-        if errors:
+        if len(errors) == 1:
+            # If only 1 error, just raise that.
+            raise errors[0]
+
+        elif len(errors) > 1:
+            # Raise a combined error.
             formatted_errors = [f"{e}" for e in errors]
             error_message = "\n\n".join(formatted_errors)
             raise CompilerError(error_message)
```
fubuloubu, line 168: Wonder if python 3.11+ exception groups make sense to use when it becomes available to us

fubuloubu (dismissed):

src/ape/managers/compilers.py
antazoey: Wow this is cool!

Idea: could try to work this in for the Python versions that support it and have it the dumb way now

PR #2252, opened 2024-08-27
feat: support `request_header` driven from configs

antazoey:
### What I did

* Allow you to configure request headers across the board, via ecosystems, via networks, and via providers.
* Actually use said request_header in web3.py (was missing!)

### How to verify it

```yaml
request_headers:
  Top-Level: "UseThisOnEveryRequest"

ethereum:
  request_headers:
    Ecosystem-Level: "UseThisOnEveryEthereumRequest"
  
  mainnet:
    request_headers:
      Network-Level: "UseThisOnAllRequestsToEthereumMainnet"

node:
  request_headers:
    Provider-Level: "UseThisOnAllRequestsUsingNodeProvider"
```

then connect:

```sh
ape console --network ethereum:local:node
```

Then verify headers:

```python
In [12]: http = provider.web3.provider._potential_providers[2]()

In [13]: http._request_kwargs
Out[13]: 
{'headers': {'User-Agent': 'Ape/0.8.14.dev0+g228ab19a1.d20240827 (Python/3.12.2 final)',
  'Top-Level': 'UseThisOnEveryRequest',
  'Ecosystem-Level': 'UseThisOnEveryEthereumRequest',
  'Provider-Level': 'UseThisOnAllRequestsUsingNodeProvider'},
 'timeout': 1800}
```

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -184,7 +184,7 @@ ape plugins install .
 For Ape's HTTP usage, such as requests made via `web3.py`, optionally specify extra request headers.
 
 ```yaml
-request_header:
+request_headers:
   # NOTE: Only using Content-Type as an example; can be any header key/value.
   Content-Type: application/json
 ```
@@ -209,6 +209,8 @@ node:
   request_headers:
     Content-Type: application/json
 ```
+
+To learn more about how request headers work in Ape, see [this section of the Networking guide](./networks.html#request-headers).
 
 ## Testing
 
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -179,6 +179,16 @@ Install these plugins by running command:
 ape plugins install .
 ```
 
+## Request Headers
+
+For Ape's HTTP usage, such as requests made via `web3.py`, optionally specify extra request headers.
+
+```yaml
+request_header:
+  # NOTE: Only using Content-Type as an example; can be any header key/value.
+  Content-Type: application/json
+```
+
 ## Testing
 
 Configure your test accounts:
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -189,6 +189,27 @@ request_header:
   Content-Type: application/json
 ```
 
+You can also specify request headers at the ecosystem, network, and provider levels:
+
+```yaml
+# NOTE: All the headers are the same only for demo purposes.
+# You can use headers you want for any of these config locations.
+ethereum:
+  # Apply to all requests made to ethereum networks.
+  request_headers:
+    Content-Type: application/json
+  
+  mainnet:
+    # Apply to all requests made to ethereum:mainnet (using any provider)
+    request_headers:
+      Content-Type: application/json
+  
+node:
+  # Apply to any request using the `node` provider.
+  request_headers:
+    Content-Type: application/json
+```
+
 ## Testing
 
 Configure your test accounts:
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -384,9 +384,6 @@ class Ethereum(EcosystemAPI):
 
     fee_token_symbol: str = "ETH"
 
-    # NOTE: This gets appended to Ape's root User-Agent string.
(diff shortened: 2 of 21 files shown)


antazoey:
note, may have to update alchemy and infura

docs/userguides/config.md
```diff
@@ -179,6 +179,37 @@ Install these plugins by running command:
 ape plugins install .
 ```
 
+## Request Headers
```
fubuloubu, line 182: Describe how they merge

fubuloubu (commented):
I feel like `User-Agent` should not be legal to configure outside of ape internals (but kinda ambivalent about that), but it might be a good use case to insert plugin names and versions into a user agent string for 2nd/3rd party ecosystem or provider plugins so we know what people are using (if this information is even searchable)

antazoey:
> I feel like `User-Agent` should not be legal to configure outside of ape internals (but kinda ambivalent about that), but it might be a good use case to insert plugin names and versions into a user agent string for 2nd/3rd party ecosystem or provider plugins so we know what people are using (if this information is even searchable)

We could detect User-Agent and do an append of some kind instead of replacing it

fubuloubu:
> > I feel like `User-Agent` should not be legal to configure outside of ape internals (but kinda ambivalent about that), but it might be a good use case to insert plugin names and versions into a user agent string for 2nd/3rd party ecosystem or provider plugins so we know what people are using (if this information is even searchable)
> 
> We could detect User-Agent and do an append of some kind instead of replacing it

Yeah, I think that's the only heqder I would handle special. We don't really do much else

src/ape_ethereum/ecosystem.py
fubuloubu, line 1: It would be cool to add `ape-ethereum` to User-Agent here

src/ape_ethereum/provider.py
fubuloubu, line 1: It would be cool to add `web3/vX.X.X` to User-Agent here (if Web3py isn't already)

antazoey: I agree! And we can. I also made this issue for snake-charmers to make it easier on us: https://github.com/ethereum/web3.py/issues/3466

src/ape_ethereum/ecosystem.py
```diff
@@ -382,6 +384,9 @@ class Ethereum(EcosystemAPI):
 
     fee_token_symbol: str = "ETH"
 
+    # NOTE: This gets appended to Ape's root User-Agent string.
+    request_header: dict = {"User-Agent": "ape-ethereum"}
```
fubuloubu, line 388: hmm, wonder if could do some magic with a field validator to sneak out the module that was subclassing this module and insert it's name, something like `cls.__package__.__name__` so that the right package name appears (`ape-ethereum`, `ape-polygon`, etc.)

Might be too fancy though

src/ape_ethereum/ecosystem.py
antazoey: done!

src/ape_ethereum/provider.py
antazoey: done

src/ape_ethereum/ecosystem.py
antazoey: I think that works nicely for these plugin-level ones, I will give it a shot.
(Note: it does not work as great for the config-level ones, but not going to get into that).

antazoey: done!

tests/functional/test_ecosystem.py
```diff
@@ -80,6 +80,22 @@ def test_name(ethereum):
...
+
+
+def test_request_header_subclass():
+    class L2(Ethereum):
+        name: str = "l2"
+
+    l2 = L2()
+    actual = l2.request_header
+    expected = {"User-Agent": "ape-l2"}
+    assert actual == expected
```
fubuloubu, line 96: Nice

fubuloubu (approved):

antazoey:
- [ ] Sanity check doesn't break plugins (running some test suites)

PR #2255, opened 2024-08-30
fix: support bytes array argument in solidity and vyper contracts

bitwise-constructs:
### What I did

Added support for bytes array argument in solidity and vyper contracts

### How I did it

Added a value converter for a list of hex values to a single concatenated hex value

### How to verify it

tx = contract_instance.functionWithCalldata(["0x123456", "0xabcd"], sender=owner)
for
functionWithCalldata(bytes calldata data)

### Checklist


src/ape/managers/converters.py
```diff
@@ -70,6 +70,18 @@ def convert(self, value: Any) -> int:
...
+class HexListConverter(ConverterAPI):
+    """
+    Convert list of hex values to single concatenated hex value.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, list) and all(is_hex(v) or isinstance(v, bytes) for v in value)
+
+    def convert(self, value: Any) -> bytes:
+        return HexBytes(b"".join([HexBytes(v) for v in value]))
```
fubuloubu, line 82: Technically don't need a list comprehension since `str.join` takes an iterable (saves you a list obj initialization)

```suggestion
        return HexBytes(b"".join(HexBytes(v) for v in value))
```

fubuloubu (approved):

src/ape/managers/converters.py
```diff
@@ -70,6 +70,18 @@ def convert(self, value: Any) -> int:
         return to_int(HexBytes(value))
 
 
+class HexListConverter(ConverterAPI):
+    """
+    Convert list of hex values to single concatenated hex value.
```
antazoey, line 75: > Convert list of hex values to single concatenated hex value.

Should be `Convert list of hex values to single concatenated HexBytes value` so it is not confused with a hex-str

(not sure why GH not letting use the suggest-code feature on this PR)

antazoey (dismissed):
Lemme know if you need help with failing tests - those should probably be less affected by changes if you want to genericize them more

src/ape/managers/converters.py
```diff
@@ -70,6 +70,18 @@ def convert(self, value: Any) -> int:
         return to_int(HexBytes(value))
 
 
+class HexListConverter(ConverterAPI):
+    """
+    Convert list of hex values to single concatenated ``HexBytes`` value.
+    """
+
+    def is_convertible(self, value: Any) -> bool:
+        return isinstance(value, list) and all(is_hex(v) or isinstance(v, bytes) for v in value)
```
antazoey, line 79: it might also be nice to support tuples here

```suggestion
        return isinstance(value,(list, tuple)) and all(is_hex(v) or isinstance(v, bytes) for v in value)
```

tests/functional/test_contract_instance.py
```diff
@@ -990,3 +990,9 @@ def test_sending_funds_to_non_payable_constructor_by_accountDeploy(
 def test_as_transaction(tx_type, vyper_contract_instance, owner, eth_tester_provider):
     tx = vyper_contract_instance.setNumber.as_transaction(987, sender=owner, type=tx_type.value)
     assert tx.gas_limit == eth_tester_provider.max_gas
+
+
+@pytest.mark.parametrize("calldata", ("0x123456", ["0x123456", "0xabcd"]))
+def test_calldata_arg(calldata, contract_instance, owner):
+    tx = contract_instance.functionWithCalldata(calldata, sender=owner)
+    assert not tx.failed
```
antazoey, line 998: a better assertion may be something like

```python
assert tx.transaction.data == EXPECTED_DATA
```

antazoey (dismissed):
a tuple of nits but looks good

(joke by @bitwise-constructs )

src/ape/managers/converters.py
fubuloubu: I wonder if there's a way to just be like, generally, this supports the iterable protocol.

antazoey: there is!

```python
>>> from collections.abc import Iterable
>>> isinstance([1,2,3], Iterable)
True
>>> isinstance((1,2,3), Iterable)
True
```

antazoey (approved):

antazoey (approved):
LGTM

Issue #2271, opened 2024-09-09
Contracts not re-compiling during `ape test` after changes.


Schlagonia:
### Environment information

- OS:  linux 
- Python Version: 3.9.5
- `ape` and plugin versions:

ape-version: 0.8.14

Installed Plugins
  alchemy      0.8.2
  etherscan    0.8.3
  hardhat      0.8.1
  infura       0.8.1
  polygon      0.8.0
  solidity     0.8.3

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
name: tokenized-strategy-aave-v3

plugins:
  - name: solidity
  - name: etherscan
  - name: hardhat
  - name: infura
  - name: alchemy
  - name: polygon
  
default_ecosystem: ethereum

dependencies:
  - name: openzeppelin
    github: OpenZeppelin/openzeppelin-contracts
    ref: 4.9.5
  - name: tokenized-strategy
    github: yearn/tokenized-strategy
    ref: v3.0.2
    config_override:
      contracts_folder: src
    exclude:
      - test/**/*
  - name: periphery
    github: yearn/tokenized-strategy-periphery
    ref: master
    config_override:
      contracts_folder: src
    exclude: 
     - test/**/*

ethereum:
  default_network: mainnet-fork
  mainnet_fork:
    default_provider: hardhat

hardhat:
  fork:
    ethereum:
      mainnet:
        upstream_provider: infura
```

### What went wrong?

After making changes to the main solidity contracts being developed under the local contracts/ folder and running `ape test` the contracts did not recompile and instead used the previously cached versions for the tests.

The contracts are deployed in the conftest and reused during every test.

I have previously experienced this issue with Vyper contracts on older versions of Ape (pre 0.7)

### How can it be fixed?

Fill this in if you have ideas on how the bug could be fixed.

antazoey:
Contracts are re-compiling after changes in-between tests (for me and all my projects / tests). Likely something else is the culprit and this is just where the error is ending up.

antazoey:
We did have a user experiencing this and they resolved it by deleting their `.pytest_cache` folder which was stuck in a weird state. You could maybe try that.

Otherwise, I am going to need more info! Using a solidity in a project, contract changes are reflected in the tests. I will note that the output from compiling is not part of the output anymore without using `-v info`.

antazoey:
The only time I have been able to reproduce is by doing the following steps:

1. Install ape-solidity
[...]

PR #2272, opened 2024-09-09
fix: mistakenly changed sources to in-memory manifest

antazoey:
### What I did

fixes: #2271

To reproduce problem:

1. Have a contract, working good, with a test.
2. Run ape test, all is good
3. Make a change to the contract that causes a compiler error
4. Run ape test again, test acts like it did before, and there is no compiler error.

**They key is that the change must be problematic and cause errors**

### How I did it

Frustratingly find spot where source Id is being overridden
Restore it.
Also, adjust error message

Test tbd

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -2359,16 +2359,11 @@ class LocalProject(Project):
             yield self
 
         else:
-            # Add sources to manifest memory, in case they are missing.
-            original_sources = self._manifest.sources
-            self._manifest.sources = dict(self.sources.items())
-            try:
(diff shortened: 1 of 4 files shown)


bitwise-constructs (dismissed):

bitwise-constructs (dismissed):

antazoey:
- [ ] Also fix bug where when you uninstall a compiler and then make changes to a change
- [ ] Get feedback from bug reporter
- [ ] Possibly have a more integration-y test

fubuloubu (approved):

Issue #2262, opened 2024-09-02
Ape-compiled Contracts not being recognized in test suite


davoice-321:
### Environment information

- OS: macOS / linux / win
- Python Version: 3.11.8
- `ape` and plugin versions:

```sh
$ ape --version
# 0.8.13

$ ape plugins list
# alchemy      0.8.0
  ens          0.8.0
  etherscan    0.8.2
  foundry      0.8.4
  hardhat      0.8.1
  infura       0.8.0
  solidity     0.8.3
  template     0.8.0
  tokens       0.8.0
  vyper        0.8.4
```

- Contents of your `ape-config.yaml` (NOTE: do not post anything private like RPC urls or secrets!):

```sh
$ cat ape-config.yaml
# name: Project

plugins:
  - name: solidity
    version: ">=0.6.0 <0.9.0"
  - name: hardhat
    version: "0.8.1"
  - name: etherscan
    version: "0.8.2"
  - name: foundry
    version: "0.8.4"

dependencies:
  # - name: openzeppelin
  #   github: OpenZeppelin/openzeppelin-contracts
  #   version: 4.8.3
  - name: openzeppelin-3.4.2
    github: OpenZeppelin/openzeppelin-contracts
    version: 3.4.2
  - name: openzeppelin-4.5.0
    github: OpenZeppelin/openzeppelin-contracts
    version: 4.5.0
  - name: openzeppelin-4.3.0
    github: OpenZeppelin/openzeppelin-contracts
    version: 4.3.0
  - name: setprotocol
    github: SetProtocol/set-protocol-v2
    version: 0.11.0
```

### What went wrong?

ABI and type of contract-deployed contracts are not being read during testing
Situation

-I'm using conftest.py to set up a deployment for a series of contracts. 

2 of the contracts are deployed using owner.deploy()
2 of the contracts are deployed through a contract that was compiled using Ape

Using the contract-deployed contracts works in the following instances: 

When the contracts are deployed and used in a script (ape run)
When the contracts are deployed and used in a conftest.py function (i'm assuming because running this function is like running a script)

The contracts are being outputted from the conftest.py function. Expected behavior is that the ABI and type of contract-deployed contracts are read properly during testing. 

Here is the output from the conftest.py function re: the contracts (printing the contracts in the test function). 

<Identified Contract1 0x68b9...>, <Unnamed contract 0xe94b0...>, <Unnamed contract 0xCcd7A...>, <Identified Contract2 0x539a46F...>)

[...]

Issue #1287, opened 2023-02-06
Test function isolation isn't working when using fixture with parameters


pandadefi:
### Environment information

- OS: macOS
- Python Version: 3.9.16
- `ape` and plugin versions: 0.6.2

### What went wrong?

When running tests, the different tests should run in isolation, including the tests function with @pytest.fixture params. With the latest version of ape, the blockchain isn't reverted to the snapshot between several tests using the different values provided by @pytest.fixture params.

here is an example: https://github.com/yearn/veYFI/pull/220/files#diff-dd59d96f9ab6202cc08431b4f984bd2f6934b6d6bf9c37e8697ee68ad4f6058eR220
Here the first test will pass with no tokens looked in the contract, but this assertion will fail on the second run with the second parameter.

### Previous version
I have downgraded to ape version 0.5.9. The issue is the same.

unparalleled-js:
We discovered that parametrized fixtures are not always working correctly, especially in complex setups. However we found that if you change the parametrization to be on the test using `pytest.mark.paramtrize()`, things work as expected. We are going to look into the issue with parametrized fixture and isolation! Thank you for reporting.

PR #2277, opened 2024-09-11
fix: update isolation as new fixtures of certain scopes arrive

antazoey:
### What I did

Make isolation "work" on session, package, module, and class scopes.
Basically, the bug was that as new fixtures of certain scopes come in after a snapshot was taken in that scope, they wouldn't really persist as the scope below left, like if the module changes or something

fixes: #2262 
fixes: #1287 

### How I did it

Awful. Basically the only way to fix this is to rebase the blockchain as new upper scopes up.
A new session comes in later, we have to undo to the module (or closest lowest) snapshot invalidate those fixture, inject those fixtures in the correct order, and re-set all the snapshots.

### How to verify it

Can technically uses scopes / parametrized fixtures in any way now... and it should also "work"... it is just is slow if using higher scoped fixtures later on in the test suite, can be really bad actually. I definitely don't recommend using parametrized fixtures unless I can figure out a more optimal way with those

### Checklist

### DIFF
--- a/docs/userguides/testing.md
+++ b/docs/userguides/testing.md
@@ -70,11 +70,6 @@ def test_authorization(my_contract, owner, not_owner):
         my_contract.authorized_method(sender=not_owner)
 ```
 
-```{note}
-Ape has built-in test and fixture isolation for all pytest scopes.
-To disable isolation add the `--disable-isolation` flag when running `ape test`
-```
-
 ## Fixtures
 
 Now that we have discussed the full flow of a test, let's dive deeper into the specific parts, starting with `pytest.fixtures`.
@@ -200,11 +195,9 @@ You also have access to the `project` you are testing. You will need this to dep
 ```python
 import pytest
 
-
 @pytest.fixture
 def owner(accounts):
     return accounts[0]
-
 
 @pytest.fixture
 def my_contract(project, owner):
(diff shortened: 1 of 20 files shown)


fubuloubu (approved):

bitwise-constructs (dismissed):

antazoey:
suspecting something with the runner now...

src/ape/pytest/runners.py
```diff
@@ -137,34 +137,10 @@ def pytest_runtest_setup(self, item):
...
-            if name in item.fixturenames
-            for definition in definitions
-        ]
-
-        for scope in ["session", "package", "module", "class"]:
-            # iterate through scope levels and insert the isolation fixture
-            # prior to the first fixture with that scope
-            try:
-                idx = scopes.index(scope)  # will raise ValueError if `scope` not found
-                item.fixturenames.insert(idx, f"_{scope}_isolation")
```
antazoey, line 156: I am not sure if this was the problem or not, but using the `idx` of the scope when inserting the fixturename seems wrong. A simple `append` does the trick (fixed below)

src/ape/pytest/runners.py
```diff
@@ -272,3 +248,37 @@ def pytest_unconfigure(self):
...
+        # prior to the first fixture with that scope
+        try:
+            idx = scopes.index(scope)  # will raise ValueError if `scope` not found
+        except ValueError:
+            continue
+
+        name = f"_{scope}_isolation"
+
+        # Don't let isolation fixtures get added more than once!
+        if name not in item.fixturenames:
```
antazoey, line 274: this is a new check , I wonder if this resolved the culprit. Basically just don't allow isolation fixture from getting added more than once.

I am going to check if this resolves https://github.com/ApeWorX/ape/issues/1287

antazoey:
- [ ] Ensure isolation fixtures ran _last_ (this has to be the problem!)

src/ape/pytest/runners.py
antazoey: Yes - doing more research, this seems like it was a problem (though still not sure if it is _the_ problem).
But basically you can end up in a case where the fixtures were not added in the right spot because it uses the first index of the scope and not the last.

I think having the isolation fixtures all at the end of the list makes sense, but going to experiment more to see.

antazoey:
Update, mentioned on Telegram but I am pretty sure I actually  found the bug; I am still working on the full fix this PR is not ready for testing yet). Standby for updates.

fubuloubu (dismissed):

antazoey:
If you want to see the most magical part(s), look at the changes made to the `runner.py` module.

antazoey:
This is "working" but may have performance issues if users don't use `autouse=True` when they should or use parametrized fixturs.

src/ape/pytest/utils.py
```diff
@@ -10,3 +10,20 @@ class Scope(str, Enum):
 
     def __str__(self) -> str:
         return self.value
```
fubuloubu: technically, you could change this to an int enum and get relative ordering (e.g. `Scope.CLASS < Scope.FUNCTION`) and then just use `self.name.lower()` here to return the pytest scope when you do `str(scope)`

antazoey:
- [ ] Either do or consider only reinitializing the autouse=True fixtures (rest should still be invalidated but maybe dont replay them
- [ ] Ignore fixtures that don't change chain head from invalidation consideration

docs/userguides/testing.md
```diff
@@ -226,6 +219,58 @@ def my_contract(Contract):
...
+By default, tests run with chain-isolation.
+This means, at the start of each test, a snapshot is taken.
+After each test completes, the chain reverts to that snapshot from the beginning of the test.
+
+By default, every `pytest` fixture is `function` scoped, meaning it will be replayed each time it is requested (no result-caching).
+For example, if you deploy a contract in a function-scoped fixture, it will be re-deployed each time the fixture gets used in your tests.
+To only deploy once, you can use different scopes, such as `"session"`, `"package"`, `"module"`, and `"class"`.
+
+Isolation also works on greater fixture scopes, `session`, `package`, `module`, and `class`.
+For example, if you define a session scope that deploys a contract and makes transactions, the state changes from those transactions remain in subsequent tests that request that fixture.
```
fubuloubu: Something relating back to snapshots (unsure if correct)
```suggestion
For example, if you define a session scope that deploys a contract and makes transactions, the state changes from those transactions remain in subsequent tests that request that fixture thanks to a snapshot that is taken when that fixture is used.
```

docs/userguides/testing.md
```diff
@@ -226,6 +219,58 @@ def my_contract(Contract):
...
+def test_my_contract_0(my_contract):
+    my_contract.myMethod()
+
+def test_my_contract_1(my_contract):
+    my_contract.myMethod()
+```
+
+To disable isolation, run `ape test` with the `--disable-isolation` flag.
+When isolation is disabled, the blockchain's state persist as the tests run.
+This will be more performant and less complicated, but will also cause you to have to state-cognizant in your tests.
```
fubuloubu:
```suggestion
This will be more performant and less complicated, but will also cause non-deterministic results in your tests as each test inherits the state of whatever was run before it.
This may be further complicated when running with other pytest plugins such as `pytest-xdist` or `pytest-split` which re-arranges the order that tests are executed in (not recommended to use these plugins together with ape).
```

fubuloubu:
>   **Questions**: How does one know if a fixture has caused the chain height to increase? There is a way to know if an entire has or if a test has, but there doesn't seem to be a way to know if an individual fixture has, that I can think of.

Can we store extra info in the cache? Such as block height of last access? If auto-mining is used, that would roughly correspond to 1 block = 1 transaction, so this is a good approximation (If auto-mining is turned off, this would not work)

docs/userguides/testing.md
```diff
@@ -226,6 +219,58 @@ def my_contract(Contract):
...
 
+## Isolation
+
+By default, tests run with chain-isolation.
+This means, at the start of each test, a snapshot is taken.
+After each test completes, the chain reverts to that snapshot from the beginning of the test.
+
+By default, every `pytest` fixture is `function` scoped, meaning it will be replayed each time it is requested (no result-caching).
+For example, if you deploy a contract in a function-scoped fixture, it will be re-deployed each time the fixture gets used in your tests.
+To only deploy once, you can use different scopes, such as `"session"`, `"package"`, `"module"`, and `"class"`.
```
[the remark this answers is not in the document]
antazoey: From pytest docs:

```
Fixture scopes
Fixtures are created when first requested by a test, and are destroyed based on their scope:

function: the default scope, the fixture is destroyed at the end of the test.
class: the fixture is destroyed during teardown of the last test in the class.
module: the fixture is destroyed during teardown of the last test in the module.
package: the fixture is destroyed during teardown of the last test in the package.
session: the fixture is destroyed at the end of the test session.
```

Imagine running pytest in a repo with multiple packages I guess.

antazoey:
> Can we store extra info in the cache? Such as block height of last access? If auto-mining is used, that would roughly correspond to 1 block = 1 transaction, so this is a good approximation (If auto-mining is turned off, this would not work)

Yes, this is how we would do it. I just can't figure out how to know which fixture is responsible. It seems really complicated. For example, if you were using multiple module-scoped fixtures at the same time and some caused height increase and some didn't, the chain height would be higher the next usage for all of them, so how does one know which fixture caused the bump?

antazoey:
- [ ] This causes ApePay to fail in tests - figure out why and fix.

antazoey:
> * [ ]  Fix create_stream min time goes to zero issues that has arrived in ApePay

Notes:

* Figured out there is a problem when using multiple parametrized fixtures at once.

docs/userguides/testing.md
```diff
@@ -226,6 +219,58 @@ def my_contract(Contract):
...
+
+def test_my_contract_1(my_contract):
+    my_contract.myMethod()
+```
+
+To disable isolation, run `ape test` with the `--disable-isolation` flag.
+When isolation is disabled, the blockchain's state persists as the tests run.
+This will be more performant and less complex, but will also cause non-deterministic results in your tests as each test inherits the state of whatever was run before it.
+
+This may be further complicated when running with other pytest plugins such as `pytest-xdist` or `pytest-split` which re-arranges the order that tests are executed in (not recommended to use these plugins together with ape).
```
fubuloubu, line 262:
```suggestion
This may be further complicated when running with other pytest plugins such as `pytest-xdist` or `pytest-split` which re-arranges the order that tests are executed in (not recommended to use these plugins together with ape until specific support is added).
```

antazoey:
- [ ] Only trigger invalidation on stateful fixtures of higher scope
- [ ] Better handling of parametrized fixtures - right now, always treats them as "new", meaning it is wayy overly invalidating...

fubuloubu (dismissed):

fubuloubu (dismissed):

antazoey:
So before I merge this, I am going to add a few more projects to the integration tester thing.
I want to really make sure I am not wrecking all of tests by doing this.

antazoey:
ApePay

this branch
32 passed in 102.98s (0:01:42)

main
32 passed in 106.16s (0:01:46)

antazoey:
Once I get https://github.com/ApeWorX/ape/pull/2327 to where I like, I want to also compare running Ape's tests w/ and w/o this feature.

fubuloubu (approved):

PR #2280, opened 2024-09-12
feat: allow private transaction receipts to bail out before transaction acceptance

bitwise-constructs:
### What I did

Made it so private transactions won't time out awaiting confirmations that will never show.

fixes: ApeWorX/ape-alchemy#82
Fixes: APE-1805

### How I did it

In the provider method for getting a receipt, moved around the logic to pass through transaction info to receipt data earlier so that private transactions can early return a partial receipt.

### How to verify it

Add the `private=True` kwarg to `provider.get_receipt()`

### Checklist


src/ape_ethereum/provider.py
```diff
@@ -624,6 +624,25 @@ def get_receipt(
...
+        if transaction := kwargs.get("transaction"):
+            # perf: If called `send_transaction()`, we should already have the data!
+            txn = (
+                transaction
+                if isinstance(transaction, dict)
+                else transaction.model_dump(by_alias=True, mode="json")
+            )
+
+        if kwargs.get("private"):
+            # Bail before confirmation because it won't be on chain yet.
```
antazoey, line 637:
```suggestion
            # We don't auto-wait for acceptance or confirmations for private transactions.
            # TODO: Update comment with instructions for how to manually wait
            #   once the feature is ready in core Ape.
```

src/ape_ethereum/provider.py
```diff
@@ -624,6 +624,25 @@ def get_receipt(
...
+            data = {
+                "required_confirmations": 0,
```
antazoey, line 639: **problem**: this still needs to be set to the correct value so that when the user _does_ call `wait()`, it will wait for the correct amount of confs.

**Note**: I do see why you are doing this waiting for the nonce bump but I added a different  comment below re: that

src/ape_ethereum/provider.py
```diff
@@ -624,6 +624,25 @@ def get_receipt(
...
+                "block_number": -1,
+                **txn,
```
antazoey, line 641: **problem**: we are going to need access to `timeout` for properly timing out during the potential manual-call to wait for tx acceptance, so maybe plop that in the kwargs here? I am not sure if it is being used anywhere but we will have to figure that out in the core refactor.

if it causes problems, we should add a TODO to remember that because knowing the calculated timeout is Alchemy-specific and the receipt should have that.

tests/functional/test_provider.py
```diff
@@ -128,6 +128,18 @@ def test_get_receipt_exists_with_timeout(eth_tester_provider, vyper_contract_ins
     assert receipt_from_provider.receiver == vyper_contract_instance.address
 
 
+def test_get_receipt_for_private_bypasses_confirmations(
```
antazoey, line 131:
```suggestion
def test_get_receipt_does_not_auto_wait(
```

I don't want to ignore confs, it is still important. We are just not going to automatically wait in this case but instead treat it more like a promise-based receipt, which we are working on adding.

antazoey (dismissed):
After this review, I am thinking we really need the promise-based receipt functionality (As much as we can in the 0.8 range and for 0.9 make it a lot more usable with any breaking changes necessary).

src/ape_ethereum/provider.py
```diff
@@ -624,11 +624,34 @@ def get_receipt(
...
+            )
+
+        private = kwargs.get("private")
+
         try:
             receipt_data = dict(
                 self.web3.eth.wait_for_transaction_receipt(hex_hash, timeout=timeout)
             )
         except TimeExhausted as err:
+            # We don't auto-wait for acceptance or confirmations for private transactions.
```
antazoey, line 643: since we are going with set-everything-to-zero plan, im not sure this comment still holds true

tests/functional/test_provider.py
```diff
@@ -128,6 +128,28 @@ def test_get_receipt_exists_with_timeout(eth_tester_provider, vyper_contract_ins
     assert receipt_from_provider.receiver == vyper_contract_instance.address
 
 
+def test_get_receipt_ignores_timeout_when_private(
+    eth_tester_provider, mock_web3, vyper_contract_instance, owner
+):
+    receipt_from_invoke = vyper_contract_instance.setNumber(888, sender=owner)
+
+    real_web3 = eth_tester_provider._web3
+    # mock_web3.eth = real_web3.eth
```
antazoey, line 137:
```suggestion
```

antazoey (commented):
commented out code left over?

antazoey (approved):

antazoey (approved):

Issue #2282, opened 2024-09-16
Long-name not working with ape pm uninstall command


bitwise-constructs:
### Environment information

- OS: macOS / linux / win
- Python Version: 3.11.9
- `ape` and plugin versions:

```sh
$ ape --version
0.8.14

$ ape plugins list
Installed Plugins
  solidity    0.8.4
```

### What went wrong?

current ape pm list:

```
NAME                                 VERSION  COMPILED
Uniswap/v3-core                      1.0.0    -
Uniswap/v3-periphery                 1.3.0    -
```

ran: 
`ape pm uninstall Uniswap/v3-core`

and received error:
`ERROR (ape-depend-import-problem-main): Package(s) 'Uniswap/v3-core' not installed.`

### How can it be fixed?

The uninstall command works with the short name, but need to make it work with the long name that gets displayed in `ape pm list` as well.

PR #2287, opened 2024-09-17
fix: makes uninstall work when using the long name displayed in ape pm list

bitwise-constructs:
### What I did

Added a check and test for also checking the package_id of a dependency when uninstalling by name. package_id is the long name displayed in `ape pm list`

fixes: #2282
Fixes: APE-1816

### How to verify it

run `ape pm uninstall OpenZeppelin/openzeppelin-contracts` on an OpenZeppelin dependency

### Checklist


tests/functional/test_dependencies.py
```diff
@@ -137,9 +137,9 @@ def test_decode_dependency_with_config_override(project):
 def test_uri_map(project_with_dependency_config):
     actual = project_with_dependency_config.dependencies.uri_map
     here = Path(__file__).parent
-    expected = f"file://{here}/data/projects/LongContractsFolder"
+    expected = Path(f"file://{here}/data/projects/LongContractsFolder")
```
antazoey, line 140: how come this is wrapped in Path? Windows? Maybe add a comment.
Note: this the str is a file URI (starting with `file://`) so it is not typical to wrap, but I think I see why you did it

bitwise-constructs: Yes, it was failing for me on Windows because the path separator character. Will add a comment

antazoey (approved):
LGTM! Just update the userguide a bit and we are good to go

antazoey (approved):

PR #2306, opened 2024-10-02
feat: automatically calculate GitHub dependency name when not provided

antazoey:
### What I did

allows

```yaml
dependencies:
  - github: ApeWorX/ApePay
    ref: main
```

to work, where `name:` is not configured.
In this case, it uses he name `apepay` automatically from the GH input.

### How I did it

model validator

### How to verify it

use config i just said

### Checklist

### DIFF
--- a/src/ape_pm/dependency.py
+++ b/src/ape_pm/dependency.py
@@ -138,10 +138,15 @@ class GithubDependency(DependencyAPI):
 
     @model_validator(mode="before")
     @classmethod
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
Nice

Would this be able to be used the vyper plugin e.g. `snekmate`?

antazoey:
> Would this be able to be used the vyper plugin e.g. `snekmate`?

pypi / python dependencies already had this handling so it should work!

PR #2308, opened 2024-10-03
fix: serialized `HexBytes` for `Any` types in `ContractLog`

antazoey:
### What I did

When annotating with `Any` and validating with `HexBytes` types, the serializers will fail.
This is because Pydantic hasn't a way of knowing.
The fix is custom serializers in `ContractLog`.
Models annotating with `HexBytes` do no have this problem (what it was made for).

fixes: https://github.com/ApeWorX/eth-pydantic-types/issues/14

### How I did it

Custom serializers on contract log models

### How to verify it

@fubuloubu knows
also see test

### Checklist

### DIFF
--- a/src/ape/types/__init__.py
+++ b/src/ape/types/__init__.py
@@ -265,15 +265,15 @@ class BaseContractLog(BaseInterfaceModel):
         (https://github.com/pydantic/pydantic/issues/10152)
         we have to ensure these are regular ints.
         """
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
Works for me!

antazoey:
Ah, failing test shows we need to check the mode and only do the str during JSON

fubuloubu (approved):

PR #2311, opened 2024-10-04
fix: prevent running tests on mainnet if configured as the default network

antazoey:
### What I did

Prevents running tests on mainnet network

1. If you default network is a mainnet, use Ape's regular default for ape test (eth:local:test or make them specify).
2. If the user passed in --network mainnet type network, exit out right away.
3. dont ask about 3

### How to verify it

to test 1:

```yaml
ethereum:
   default_network: mainnet
```

to test 2:

```sh
ape test --network ethereum:mainnet:alchemy
```

### Checklist

### DIFF
--- a/src/ape/pytest/plugin.py
+++ b/src/ape/pytest/plugin.py
@@ -10,6 +10,16 @@ from ape.pytest.runners import PytestApeRunner
 from ape.utils.basemodel import ManagerAccessMixin
 
 
+def _get_default_network() -> str:
+    default_ecosystem = ManagerAccessMixin.network_manager.default_ecosystem
+    if default_ecosystem.default_network.is_mainnet:
+        # Don't use mainnet for tests, even if it configured as
+        # the default.
+        return "ethereum:local:test"
+
+    return default_ecosystem.name
+
+
 def pytest_addoption(parser):
     def add_option(*names, **kwargs):
         try:
@@ -29,7 +39,7 @@ def pytest_addoption(parser):
     add_option(
         "--network",
         action="store",
-        default=ManagerAccessMixin.network_manager.default_ecosystem.name,
+        default=_get_default_network(),
         help="Override the default network and provider (see ``ape networks list`` for options).",
     )
     add_option(
--- a/src/ape/pytest/plugin.py
+++ b/src/ape/pytest/plugin.py
@@ -1,6 +1,8 @@
 import sys
 from pathlib import Path
+from typing import Optional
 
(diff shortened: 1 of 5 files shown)


src/ape/pytest/plugin.py
```diff
@@ -10,6 +10,16 @@
 from ape.utils.basemodel import ManagerAccessMixin
 
 
+def _get_default_network() -> str:
+    default_ecosystem = ManagerAccessMixin.network_manager.default_ecosystem
+    if default_ecosystem.default_network.is_mainnet:
+        # Don't use mainnet for tests, even if it configured as
+        # the default.
+        return "ethereum:local:test"
```
fubuloubu, line 18: I think if this is true, we should just raise to tell the user of a configuration error instead of proceeding

fubuloubu (approved):
one comment above, otherwise approved

antazoey:
I tested it when running both `ape test` and `pytest`

antazoey:
> one comment above, otherwise approved

Resolved!

src/ape/pytest/runners.py
```diff
@@ -36,7 +36,11 @@ def __init__(
 
     @property
     def _provider_context(self) -> ProviderContextManager:
-        return self.network_manager.parse_network_choice(self.config_wrapper.network)
```
fubuloubu, line 39: To play devil's advocate here, since you have the ConfigError already raising in that other place, then it doesn't necessarily also need to be here in case somebody needs to run on mainnet for whatever reason. They just have to explicitly call it, it cannot be the default

fubuloubu (dismissed):
Just one suggestion.

src/ape/pytest/runners.py
antazoey: I was imagining someone automating a shell command poorly such that it resulted in mainnet being passed in accidentally

fubuloubu: I'd want to block where you set mainnet as the default network in config, but if you explictly set it by accident that's kind of on you. I don't think that would be common anyways, and I'm about 50-50 if it would be intentional (and thus annoying to be unable to do it by any means), so I am a bit biased against raising for explicitly setting it to mainnet

antazoey: OK, I will change it to allow running tests on mainnet.

fubuloubu (approved):

antazoey:
mypy fixed here: https://github.com/ApeWorX/ape/pull/2318/files

PR #2315, opened 2024-10-12
build(eth-typing): pin version

0xthedance:
ContractName is depecrated from version 4.2.0

### What I did
I pinned eth-typing version 4.1.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,7 @@ setup(
         # All version pins dependent on web3[tester]
         "eth-abi",
         "eth-account",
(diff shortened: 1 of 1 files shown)


fubuloubu:
Would love it if we can just do the upgrade to v5

Also, thanks so much for helping with these types of small updates!

antazoey:
Thanks for your contrib! I think they fixed it in 5.0.1 so we can bump to that: https://github.com/ApeWorX/ape/pull/2319

PR #2316, opened 2024-10-14
feat: properly support project-level settings

antazoey:
### What I did

This was an undocumented feature as of 0.8.0 but I actually want it to be a real feature because I am relying on it

### How I did it

* adds a test to make sure it continue working
* adds documentation for it

### How to verify it

I am using it here: https://github.com/ApeWorX/ApeProjectTester/blob/main/ape-config.yaml

### Checklist

### DIFF
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -258,7 +258,10 @@ solidity:
 
 Projects can use their own settings.
 Meaning, you can put whatever data you want in an `ape-config.yaml` file and read it in Ape.
-**NOTE**: These types of settings lack sophisticated Pydantic validation and are limited in that respect.
+
+```{note}
+These types of settings lack sophisticated Pydantic validation and are limited in that respect.
+```
 
 ```yaml
 my_project_key:
--- a/docs/userguides/config.md
+++ b/docs/userguides/config.md
@@ -273,5 +273,5 @@ Then, to access it (or any setting for that matter):
 ```python
 from ape import project
 
-setting = project.config.my_project_key.my_project_setting  #  my_value
+setting = project.config.my_project_key.my_project_setting  #  "my_value"
 ```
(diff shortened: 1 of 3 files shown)


docs/userguides/config.md
```diff
@@ -242,3 +243,35 @@ test:
...
+my_project_key:
+  my_project_setting: "123"
+```
+
+Then, to access it (or any setting for that matter):
+
+```python
+from ape import project
+
+setting = project.config.my_project_key.my_project_setting  # "123"
```
fubuloubu, line 276: Maybe might want to show how you have to do post-validation?
```suggestion
setting = int(project.config.my_project_key.my_project_setting)  # 123
```

fubuloubu (dismissed):

antazoey:
mypy fixed here: https://github.com/ApeWorX/ape/pull/2318/files

docs/userguides/config.md
antazoey: it is a string though and I was trying to show that. I can use a different value though

antazoey: i changed it to `"my_value"` in both spots, maybe clearer.

fubuloubu: Question: if you put an integer, will it parse as an integer?

antazoey: In a config file, no, there is no way to know. That is what I meant in the my note I put in the docs that no validation will happen on project-level config. However if using `temp_config(value=123)`, it will be whatever type you are using, so you can do types programmatically.

antazoey: I could a weird typing mechanism to the yaml like:

```yaml
my_weird_thing:
  foo: 123  # int
```

but tbh seems overkill, but willing to do it if you think it's worth it.

antazoey: I could also just force types, like if `isnumeric()`, turn to int or if it is a bool-string name turn to a bool. I could be down with that.

antazoey: oh wait i was mistaken, it seems simple validation _does_ already happen

bitwise-constructs (approved):

fubuloubu (approved):

PR #2324, opened 2024-10-17
fix: always serialize block size to int

antazoey:
### What I did

* handle hex values in block validation for size
* handle hex values when injected in _size
* always serialize to int for size

### How I did it

field serializer on size to force to be an int even when it is not

### Checklist


mikeshultz (approved):
Testing appears to be working well.  LGTM

PR #2328, opened 2024-10-19
feat: lookup network in evmchains; plugin-less networks, adhoc networks w/ correct name

antazoey:
### What I did

* Make it so if you connect to a network Ape hasn't an installed plugin for, it will look it up in evmchains and "just work"
* Make it so if you use a custom network e.g. connecting using a URL only, and that network is found in evmchains, it will set the correct names for the network

### How I did it

Use `evmchains` library.

### How to verify it

First, I have no plugin for moonbeam installed, but yet i can do this:

```sh
ape console --network moonbeam
```

and it "just works"

Second, if I connect to a URL in the same fashion:

```sh
ape console --network "https://moonriver.unitedbloc.com"
```

Not only does it work (did before), but the network name and stuff is correct:

```
In [1]: provider.network.name
Out[1]: 'moonriver'

In [2]: provider.network.ecosystem.name
Out[2]: 'moonbeam'
```

taa daaaa

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -2,6 +2,8 @@ from collections.abc import Collection, Iterator
 from functools import cached_property
 from typing import TYPE_CHECKING, Optional, Union
 
+from evmchains import PUBLIC_CHAIN_META
+
 from ape.api.networks import EcosystemAPI, NetworkAPI, ProviderContextManager
 from ape.exceptions import EcosystemNotFoundError, NetworkError, NetworkNotFoundError
 from ape.managers.base import BaseManager
@@ -53,7 +55,6 @@ class NetworkManager(BaseManager, ExtraAttributesMixin):
         """
         The currently connected provider if one exists. Otherwise, returns ``None``.
         """
-
         return self._active_provider
 
     @active_provider.setter
@@ -164,7 +165,6 @@ class NetworkManager(BaseManager, ExtraAttributesMixin):
         """
         The set of all ecosystem names in ``ape``.
         """
-
         return set(self.ecosystems)
 
     @property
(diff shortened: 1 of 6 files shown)


antazoey:
Failing test only happens on Web3>=6.20.2 not on 6.20.1 (which is what i was using locally)

antazoey:
:<

fubuloubu (approved):
I feel like this could do with a little extra testing from  @bitwise-constructs

src/ape/managers/networks.py
```diff
@@ -437,10 +438,28 @@ def get_ecosystem(self, ecosystem_name: str) -> EcosystemAPI:
             :class:`~ape.api.networks.EcosystemAPI`
         """
 
-        if ecosystem_name not in self.ecosystem_names:
-            raise EcosystemNotFoundError(ecosystem_name, options=self.ecosystem_names)
+        if ecosystem_name in self.ecosystem_names:
+            return self.ecosystems[ecosystem_name]
 
-        return self.ecosystems[ecosystem_name]
+        elif ecosystem_name in PUBLIC_CHAIN_META:
```
bitwise-constructs, line 444:
```suggestion
        elif ecosystem_name.lower() in PUBLIC_CHAIN_META:
            ecosystem_name = ecosystem_name.lower()
```

The names in evmchains json use pascal case so this should flatten case if it falls back far enough to check evmchains so it will justwork(tm) for users

bitwise-constructs: although really it should slugify for chains like polygon-zkevm

bitwise-constructs (approved):
This looks good, I confirmed that custom network names still take precedence over names in evmchains

bitwise-constructs:
Although I was digging more into the custom network names and I'm not clear on how that works.

If I set up one in my yaml like this

```
networks:
  custom:
    - name: scroll                    
      chain_id: 109                   
      ecosystem: shibarium            
      base_ecosystem_plugin: ethereum  
      default_provider: node          
 ```
 
and then call `ape console --network scroll` then as it evaluates, first it will check if it exists in the ecosystems of the NetworkManager, where this custom network is present but it is under its ecosystem value `shibarium`, not its custom network name `scroll`
 
 ```
 (Pdb) self.ecosystem_names
{'ethereum', 'shibarium'}
 ```
 
but if I change the call to `ape console --network shibarium` to target the correct custom network, then the value in there doesn't have `chain_id` from the custom network config, it only has it in `_networks_from_evmchains`

bitwise-constructs (approved):

PR #2327, opened 2024-10-17
test: use `ape test` to test ape

antazoey:
### What I did

APE TEST | TEST APE

### How I did it

being meta

### How to verify it

tests pass

### Checklist

- [ ] All changes are completed
- [ ] New test cases have been added
- [ ] Documentation has been updated

PR #2332, opened 2024-10-21
fix: issue where snapshots lingered on disconnect

antazoey:
### What I did

this bug is preventing us from using `ape test` to test ape.
this may fix other oddities with multi-chain tests and isolation.

### How I did it

* use chain ID when keeping track of snaps
* when disconnecting, also clear snaps from that cache

probably 

### Checklist

### DIFF
--- a/src/ape/managers/chain.py
+++ b/src/ape/managers/chain.py
@@ -1,4 +1,5 @@
 import json
+from collections import defaultdict
 from collections.abc import Collection, Iterator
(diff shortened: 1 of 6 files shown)


fubuloubu (approved):

antazoey:
note: feature is even more tested when we merge in with https://github.com/ApeWorX/ape/pull/2327 branch

PR #2333, opened 2024-10-22
perf: make `ape --help` faster

antazoey:
Makes `ape --help` faster by:

1. Making all imports as lazy as possible, only importing what is needed everywhere
2. Use module-level `__getattr__` for exports (`__init__.py` files in root of all plugins) so things are loading as requested and not on initial module load
3. ^  similar thing with `_cli.py` files - make all the slows imports have locally and lazier

Before:
```
ape --help  4.04s user 0.62s system 124% cpu 3.747 total
```

After:
```
ape --help  1.45s user 0.17s system 99% cpu 1.629 tota
```

Some of the work was broken out in smaller PRs to keep this draft simpler:

https://github.com/ApeWorX/ape/pull/2334
https://github.com/ApeWorX/ape/pull/2335
https://github.com/ApeWorX/ape/pull/2336
https://github.com/ApeWorX/ape/pull/2337
https://github.com/ApeWorX/ape/pull/2338

and i can probably do more if still needed

**NOTE**: Any CLI plugin needs to also be updated. Installing `ape-tokens` with this branch renders the benefits null unless it also follows suite and making its `__init__.py` and `_cli.py` files load faster.

Plugin PRs:

https://github.com/ApeWorX/ape-tokens/pull/46
TBD vyper, solidity and whatever else


src/ape/_cli.py
```diff
@@ -28,17 +29,20 @@ def display_config(ctx, param, value):
 
     click.echo("# Current configuration")
 
+    access = import_module("ape.utils.basemodel").ManagerAccessMixin
```
antazoey, line 32: - [ ] lazy loading ManagerAccessMixin can be undone everywhere i think, it ends up getting loaded in anyway

bitwise-constructs (approved):

antazoey:
- [ ] Fix root methoddocs (is empty now)

fubuloubu:
I hate to say this, but with some of the "public" api moving around would it make sense to call this a breaking change?

antazoey:
> I hate to say this, but with some of the "public" api moving around would it make sense to call this a v0.9 breaking change?

I think all of the original imports still work though, but I was going to cross check all the plugins.

- [ ] Cross check all the plugins

antazoey:
> I hate to say this, but with some of the "public" api moving around would it make sense to call this a v0.9 breaking change?

Like for example, if something was in `ape.types.__init__` and was moved to a submodule, the import from `ape.types` should still work for users. That was intent everywhere anyway.

Internally, we are importing from more direct sources but that is just for making Ape faster and not a clear indication of public API usage.

I am still testing all the plugins just to be sure, and can test ApePay as well.

antazoey:
I did find something with ApePay but technically ApePay was already doing it wrong.
I can still re-expose the values from `ape.types` just in case but this never should have worked to begin with.
(albeit `ape.types` may be a better namespace for the base model stuff, maybe a 0.9-ism)

https://github.com/ApeWorX/ApePay/pull/115

antazoey:
@fubuloubu OK, I have tested a bunch of stuff to ensure nothing is breaking anywhere...
I am still down for punting this to 0.9 if you want, but I feel like it is pretty safe. You can see everything I checked in this comment: https://github.com/ApeWorX/ape/pull/2333#issuecomment-2436050272

Also - down to check more repos and such too.

Thoughts?

fubuloubu (approved):
> Thoughts?

LFG

PR #2347, opened 2024-10-26
perf: hardcoded trusted plugin during `--help` check

antazoey:
### What I did

when you have a trusted plugin installed, even if that plugin is also optimized like we did for the ones with core, it is still has a  small slowdown because of the trusted check, which makes 1 request to GitHub during `--help`

this PR hardcoded the trusted plugin to avoid the request altogether.

### How I did it

hardcode the trusted plugin

### How to verify it

first install ape-tokens==0.8.1 (to get the full benefit, will still work on 0.8.0)

then

```sh
ape --help
```

Before i was seeing like 2.23 seconds..
Now I am seeing like 2 seconds

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -91,7 +91,7 @@ class ApeCLI(click.MultiCommand):
 
             if plugin.in_core:
                 sections["Core"].append((cli_name, help))
-            elif plugin.is_installed and not plugin.is_third_party:
+            elif plugin.check_trusted(use_web=False):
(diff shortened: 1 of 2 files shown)


antazoey:
Other work needed still get `ape --help` faster:

1. make `ape_vyper` module load faster
2. make `ape_solidity` module load faster
3. make `ape_ledger` module load faster
4. make `ape_trezor` module load faster

I think that is the remaining plugins.

fubuloubu (dismissed):

Issue #2348, opened 2024-10-27
`ProviderContextManager` is not concurrency-aware


fubuloubu:
Ran into this with a Silverback bot using the new [`FORK_MODE`](https://github.com/ApeWorX/silverback/pull/157) PR in the Silverback SDK (currently solving with an async concurrency lock)

Not sure this can be resolved without full async support of Ape's network management stack

fubuloubu:
Not quite sure what the issue was, but it lead to weird "trying to fork network `mainnet-fork` is not possible" or "ecosystem 'base' has no network `mainnet-fork-fork`" bugs

antazoey:
> Not quite sure what the issue was, but it lead to weird "trying to fork network `mainnet-fork` is not possible" or "ecosystem 'base' has no network `mainnet-fork-fork`" bugs

i can at least improve the error

antazoey:
Maybe this helps a little? https://github.com/ApeWorX/ape/pull/2349

fubuloubu:
> Maybe this helps a little? #2349

Yes I believe that will prevent the error, but not totally ideal

I will learn if it works well enough or not when I try to implement backtesting to silverback (at some later date)

for reference, the backtesting mode of silverback takes a range of blocks and executes each block (and relevant event logs/transactions) in an isolated fork that only exists for the span of that one block

antazoey:
Wha about a fork-lock for each network so you can't fork until you hold the lock, waiting for other threads / processes to release the lock

fubuloubu:
> Wha about a fork-lock for each network so you can't fork until you hold the lock, waiting for other threads / processes to release the lock

I don't think we need that now with the synchronous version, I will test this fix to see if it works good enough for my use case and get back to you

fubuloubu:
This might have been a bit of a false positive, it seems like the provider context manager was not resetting the context stack item when a failure was occuring within the context:

```
INFO:     Starting 'anvil' process.                                                                                                                                
INFO:     Stopping 'anvil' process.                                                                                                                                
SUCCESS:  exec_event1[...] - 1.150s (8.8%)                 
INFO:     exec_event2[...] - Started                        
INFO:     Starting 'anvil' process.                                                                                                                                
INFO:     exec_block[...] - Started                                
ERROR:    exec_block[...] - 0.000s (0.0%)                          
INFO:     check_logs[...] - Started                                
ERROR:    check_logs[...] - 0.000s (0.0%)                          
[...]

PR #2349, opened 2024-10-27
feat: allow forking a forked network

antazoey:
### What I did

Instead of `mainnet-fork-fork`, you just get another `mainnet-fork` process with a different port

### How I did it

Fix name and use auto host

### How to verify it

(ape) ➜  ape git:(feat/allow-forking-a-fork) ✗ ape console --network ethereum:sepolia

```
In [1]: with networks.fork("foundry") as provider:
   ...:     print(provider.uri)
   ...:     with networks.fork("foundry") as prov2:
   ...:         print(prov2.uri)
   ...: 
INFO:     Starting 'anvil' process.
INFO:     Connecting to existing Geth node at https://eth-sepolia.g.alchemy.com/[hidden].
http://127.0.0.1:8545
INFO:     Starting 'anvil' process.
http://127.0.0.1:52015
INFO:     Stopping 'anvil' process.
INFO:     Stopping 'anvil' process.
```

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -780,7 +780,7 @@ class ProviderContextManager(ManagerAccessMixin):
         current_id = self.provider_stack.pop()
 
         # Disconnect the provider in same cases.
-        if self.disconnect_map[current_id]:
+        if self.disconnect_map.get(current_id):
             if provider := self.network_manager.active_provider:
                 provider.disconnect()
 
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -126,15 +126,19 @@ class NetworkManager(BaseManager, ExtraAttributesMixin):
             :class:`~ape.api.networks.ProviderContextManager`
         """
         network_name = self.network.name
-        forked_network_name = (
-            network_name if network_name.endswith("-fork") else f"{network_name}-fork"
-        )
+        is_fork_already = network_name.endswith("-fork")
+        forked_network_name = network_name if is_fork_already else f"{network_name}-fork"
         try:
             forked_network = self.ecosystem.get_network(forked_network_name)
         except NetworkNotFoundError as err:
             raise NetworkError(f"Unable to fork network '{network_name}'.") from err
 
         provider_settings = provider_settings or {}
+        if is_fork_already:
+            # Forking a fork- to ensure is using a different Port,
+            # use the "auto-port" feature.
+            provider_settings["uri"] = "auto"
+
         fork_settings = {}
         if block_number is not None:
             # Negative block_number means relative to HEAD
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -125,10 +125,14 @@ class NetworkManager(BaseManager, ExtraAttributesMixin):
         Returns:
             :class:`~ape.api.networks.ProviderContextManager`
         """
+        network_name = self.network.name
+        forked_network_name = (
+            network_name if network_name.endswith("-fork") else f"{network_name}-fork"
+        )
         try:
-            forked_network = self.ecosystem.get_network(f"{self.network.name}-fork")
+            forked_network = self.ecosystem.get_network(forked_network_name)
         except NetworkNotFoundError as err:
-            raise NetworkError(f"Unable to fork network '{self.network.name}'.") from err
+            raise NetworkError(f"Unable to fork network '{network_name}'.") from err
 
         provider_settings = provider_settings or {}
         fork_settings = {}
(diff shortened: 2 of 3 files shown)


bitwise-constructs (approved):
Fork yeah

fubuloubu (approved):
Maintainer note: Side-steps but does not resolve #2348 (prevents the bug where the un-isolated network context tries to fork the fork network, but does not properly isolate fork contexts from each other)

antazoey:
I think for this to really work (unrelated to #2348) it would have spin up another fork process using auto-port or something, not sure how to make that feel clean yet in the UX.

fubuloubu:
> I think for this to really work (unrelated to #2348) it would have spin up another fork process using auto-port or something, not sure how to make that feel clean yet in the UX.

Yeah I think at the end of the day, in an async ape this code would need some deeper thought (and probably changes to the provider plugin interface)

fubuloubu:
Confirmed this is working

antazoey:
> Confirmed this is working

I think you may run into issues on disconnect

fubuloubu:
> > Confirmed this is working
> 
> I think you may run into issues on disconnect

Hmm, it is stalling weirdly

antazoey:
> Hmm, it is stalling weirdly

One of them disconnecting is probably affecting that other, that is why I think we need this to work with auto port somehow so you can maintain all of your different forks more separately.

fubuloubu:
> > Hmm, it is stalling weirdly
> 
> One of them disconnecting is probably affecting that other, that is why I think we need this to work with auto port somehow so you can maintain all of your different forks more separately. 
> 

Its a good case study, I don't need this feature for now but it got me much further on my path towards backtesting silverback bots. Eventual plan is to multiprocess many forks across many blocks in parallel for speed. (I hope one day we get a forkable pure Python provider so it doesnt require double the processes)

Its also cool how we can see ape's query system becoming useful for this.

Anyways, a dog can dream

antazoey:
@fubuloubu Wondering if this will fix it for you now:

```
INFO:     Connecting to existing Geth node at https://eth-sepolia.g.alchemy.com/[hidden].
http://127.0.0.1:8545
INFO:     Starting 'anvil' process.
http://127.0.0.1:52015
INFO:     Stopping 'anvil' process.
INFO:     Stopping 'anvil' process.
```

You can see the second fork is a automatically a different port than the first one.

fubuloubu (approved):
Nice, this looks good. It doesn't require a plugin change?

antazoey:
> Nice, this looks good. It doesn't require a plugin change?

No, both foundry and hardhat should work this way, though I only tested Foundry because I am having issues with NPM right now

fubuloubu:
> > Nice, this looks good. It doesn't require a plugin change?
> 
> No, both foundry and hardhat should work this way, though I only tested Foundry because I am having issues with NPM right now

Sick

Tenderly was the only other fork provider, but it's not currently very maintained

I might try and fix that at some point, but not a blocker now

antazoey:
> Tenderly was the only other fork provider, but it's not currently very maintained

At worst, the host setting is just not handled and it will work like it did before the recent changes in this PR (won't do a fork-fork but also wont really make a new network.

Is there a way to have multiple forks of the same networks in tenderly? Maybe we'd have to handle this in the API somewhere

fubuloubu:
> > Tenderly was the only other fork provider, but it's not currently very maintained
> 
> At worst, the host setting is just not handled and it will work like it did before the recent changes in this PR (won't do a fork-fork but also wont really make a new network.
> 
> Is there a way to have multiple forks of the same networks in tenderly? Maybe we'd have to handle this in the API somewhere

anytime you started a new fork network context with the tenderly provider, it would call the API to create the fork (returning a random UUID for it). when the network exits context, it would delete the fork by UUID

antazoey:
> anytime you started a new fork network context with the tenderly provider, it would call the API to create the fork (returning a random UUID for it). when the network exits context, it would delete the fork by UUID

That may make testing multi-chains apps kinda difficult. If you have to shift between networks back and forth. But that is how `.fork()` behaves anyway so it would fine here.

fubuloubu:
> > anytime you started a new fork network context with the tenderly provider, it would call the API to create the fork (returning a random UUID for it). when the network exits context, it would delete the fork by UUID
> 
> That may make testing multi-chains apps kinda difficult. If you have to shift between networks back and forth. But that is how `.fork()` behaves anyway so it would fine here.
> 

Yeah I think for the multichain use case, especially with async, there would have to be some additional functionality/state that would allow it to work

For this use case, I basically just need a ton of forks that are completely independent of each other

fubuloubu (approved):

PR #2355, opened 2024-10-30
perf: make all commands faster, especially `ape init`, adds `--name` flag to `init` cmd

antazoey:
### What I did

this may be problematic, not sure.
but the reason commands are so slow, like `ape init` is because of the pre-execution config validation.
i believe that was added as part of showing config errors and such, otherwise you may  run into weird problems and not really know why. but oh well, better to have a faster  command. Hopefully we can come up with a better config validation system, if we have to.

also adds `--name` flag to init so you can automate

Before:

```
ape init --name poop  3.46s
```

Now:

```
ape init --name poop  0.29s
```

### How I did it

delete helpful validation

### How to verify it

things whip

### Checklist

### DIFF
--- a/src/ape/_cli.py
+++ b/src/ape/_cli.py
@@ -6,19 +6,15 @@ from functools import cached_property
 from gettext import gettext
 from importlib.metadata import entry_points
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Optional
+from typing import Any, Optional
 from warnings import catch_warnings, simplefilter
 
 import click
-import rich
 import yaml
 
 from ape.cli.options import ape_cli_context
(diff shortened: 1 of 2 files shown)


antazoey:
TESTS PASS YOLO

fubuloubu (approved):
I don't think I understand what that ex8sting code was doing. Can you summarize?

antazoey:
> I don't think I understand what that existing code was doing. Can you summarize?

The config determines some default values for click options. Now, those defaults are all callable / lazy so there shouldnt be a need to validate your config until a non-help command is being executed. This code did it early so you could see the config-line numbers / problems.

PR #2359, opened 2024-10-31
perf: able to load local project faster

antazoey:
### What I did

crushing bottle necks 1 by 1

Avoids config processing in the `project` init and move to `path` to property to make loading project faster

### How to verify it

easiest way:

ipython

before:

```
In [1]: %time from ape import project
CPU times: user 2.93 s, sys: 580 ms, total: 3.51 s
Wall time: 2.57 s
```

after

```
%time from ape import project
CPU times: user 1.97 s, sys: 554 ms, total: 2.52 s
Wall time: 1.6 s
```

### Checklist

### DIFF
--- a/src/ape/managers/project.py
+++ b/src/ape/managers/project.py
@@ -2217,7 +2217,7 @@ class LocalProject(Project):
             err.args = (message,)
             raise  # The same exception (keep the stack the same height).
 
-    @property
+    @cached_property
     def path(self) -> Path:
         """
         The path to the project's "base" (where contract source IDs are relative to).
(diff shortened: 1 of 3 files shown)


src/ape/managers/project.py
```diff
@@ -2219,6 +2217,13 @@ def __getattr__(self, item: str) -> Any:
             err.args = (message,)
             raise  # The same exception (keep the stack the same height).
 
+    @property
+    def path(self) -> Path:
```
fubuloubu, line 2221: Is this heavily used? Could be cached?

fubuloubu (dismissed):

src/ape/managers/project.py
antazoey: done!

PR #2360, opened 2024-10-31
perf: lazy load style on api, utils, and types

antazoey:
before:

```
In [1]: %time from ape.api import ProjectAPI
CPU times: user 1.04 s, sys: 154 ms, total: 1.19 s
Wall time: 1.19 s
```

after:

```
In [1]: %time from ape.api import ProjectAPI
CPU times: user 400 ms, sys: 51.6 ms, total: 452 ms
Wall time: 451 ms
```

- [ ] Fix docs


antazoey:
It does seem like tests in Ape are running overall faster from this. Maybe loading modules in a more sparsed way is less CPU load or something, IDK

fubuloubu (approved):

PR #2394, opened 2024-12-02
feat: support web3.py v7 (in addition to v6)

antazoey:
### What I did

work was already done in another branch but after discussing with @fubuloubu a couple weeks ago, we are going to push forward with it before 0.9

- [x] update any plugins importing stuff from web3 at the same time, such as ape-hardhat

- [x] ape-alchemy (https://github.com/ApeWorX/ape-alchemy/pull/89)
- [x] ape-infura (https://github.com/ApeWorX/ape-infura/pull/97) 
- [x] ape-hardhat (https://github.com/ApeWorX/ape-hardhat/pull/185)
- [x] ape-chainstack (https://github.com/ApeWorX/ape-chainstack/pull/12)
- [x] ape-ens (https://github.com/ApeWorX/ape-ens/pull/40)
- [x] ape-foundry (https://github.com/ApeWorX/ape-foundry/pull/118)
- [x] ape-frame (https://github.com/ApeWorX/ape-frame/pull/14)
- [x] ape-tenderly (https://github.com/ApeWorX/ape-tenderly/pull/17)

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -28,7 +28,6 @@ from web3.exceptions import (
     TransactionNotFound,
 )
 from web3.gas_strategies.rpc import rpc_gas_price_strategy
-from web3.middleware import ExtraDataToPOAMiddleware
 from web3.middleware.validation import MAX_EXTRADATA_LENGTH
 from web3.providers import AutoProvider
 from web3.providers.auto import load_provider_from_environment
@@ -57,6 +56,7 @@ from ape.logging import logger, sanitize_url
 from ape.types.events import ContractLog, LogFilter
 from ape.types.gas import AutoGasLimit
 from ape.types.trace import SourceTraceback
+from ape.utils._web3_compat import ExtraDataToPOAMiddleware
 from ape.utils.basemodel import ManagerAccessMixin
 from ape.utils.misc import DEFAULT_MAX_RETRIES_TX, gas_estimation_error_message, to_int
 from ape_ethereum._print import CONSOLE_ADDRESS, console_contract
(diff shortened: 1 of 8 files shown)


fubuloubu (approved):
Is there any way to make it work with both v6 and v7?

antazoey:
> Is there any way to make it work with both v6 and v7?

Yes we can, but the tests will probably only run on v7 unless we add another dimension to the matrix (which I try to avoid).

antazoey:
> Is there any way to make it work with both v6 and v7?

Another thing to note! Getting it to work on both is easy, but where we fudged up is the plugins: they have no pins for peer ape dependencies with comment (`# Use same version as eth-ape`) which seemed like a good idea at the time but actually wasn't... because now plugins may be importing stuff from web3 v6 even if you install web3 v7 and things won't fail until runtime... so I still need to fix allll the plugins here at the same time if we are to go through with this.

fubuloubu:
> > Is there any way to make it work with both v6 and v7?
> 
> Yes we can, but the tests will probably only run on v7 unless we add another dimension to the matrix (which I try to avoid).

that sounds fine, I just want to avoid unintentially rugging people on the older version, at least until v0.9 where we can drop v6

fubuloubu (approved):

fubuloubu (dismissed):

bitwise-constructs (dismissed):

bitwise-constructs (approved):

src/ape_ethereum/provider.py
```diff
@@ -1326,9 +1336,7 @@ class EthereumNodeProvider(Web3Provider, ABC):
     name: str = "node"
 
     # NOTE: Appends user-agent to base User-Agent string.
-    request_header: dict = {
-        "User-Agent": construct_user_agent(str(HTTPProvider)),
-    }
+    request_header: dict = {"User-Agent": f"EthereumNodeProvider/web3.py/{web3_version}"}
```
antazoey, line 1339: note to reviewer: this basically makes it the same as it was before, v7 changes this method to make it more web3-specific by including a required module arg

PR #2400, opened 2024-12-09
fix: issue where gnosis-safe proxy sometimes was not detected

0xthedance:
### What I did
- Added some proxy tests and fixed the ones already there.

- Changed `get_proxy_info` function for GnosisSafe Proxy verification. I removed the bytecode pattern verification.
 
partial fix of: #760 

### How to verify it

I've added the contracts to data/source and the compiled contracts( json) to contracts

### Checklist


antazoey:
@0xthedance I changed the PR title to be `fix: issue where gnosis-safe proxy sometimes was not detected #2400` in attempt to focus on the production-related changed here (for visibility to other users). Does the new title make sense? Can you describe here more about why that change is needed?

0xthedance:
When I compile the contract with the Ape compiler, Ape fails to identify the proxy. As I understand it, the ‘get_proxy_info’ function now searches for Safe-specific bytecode using PUSH32 keccak256("masterCopy()"), which translates to 0x7fa619486e000....
However, in the JSON compiled by Ape, this pattern doesn’t appear. I suspect this might be due optimizations, as when I compile the contract following the Safe repo instructions (and therefore hardhat), I obtain bytecode with this pattern.
 I have removed the part of the code where the bytecode pattern was being checked, since later the code calls ´masterCopy()´ and checks if the target matches, and I believe this is sufficient to identify the proxy.

tests/functional/data/sources/GnosisSafe.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
antazoey, line 1: Should the name of this file be `SafeProxy.sol` instead?

tests/functional/geth/test_proxy.py
```diff
@@ -0,0 +1,93 @@
...
+
+    actual = ethereum.get_proxy_info(contract_instance.address)
+    assert actual is not None
+    assert actual.type == ProxyType.Beacon
+    assert actual.target == target
+
+
+@geth_process_test
+def test_uups_proxy(get_contract_type, geth_contract, owner, ethereum):
+
```
antazoey, line 47: **nit**: can we avoid newlines at the start of the test?

antazoey (commented):
Looking close! Just pointing out something you may have missed, but nice work navigating the difficulties of our tests haha

tests/functional/data/sources/ERCProxy.sol
```diff
@@ -0,0 +1,20 @@
...
+
+
+  function implementation() public view returns (address) {
+    return target;
+  }
+
+  function proxyType() public pure returns (uint256 proxyTypeId){
+    return 1;
+  }
+}
```
antazoey, line 20:
```suggestion
}

```

antazoey (dismissed):

antazoey (approved):
This is great, thank you

antazoey:
I updated https://github.com/ApeWorX/ape/issues/760 to show what proxies have been tested and which remain.

PR #2411, opened 2024-12-11
feat: use preferred `.site_package` key instead of `.python` for less confusion

antazoey:
### What I did

Prep work for 0.9, deprecates a key and adds the other one in a non-breaking way.
In 0.9 we can fully remove now.

### Checklist

### DIFF
--- a/src/ape_pm/dependency.py
+++ b/src/ape_pm/dependency.py
@@ -432,12 +432,10 @@ class PythonDependency(DependencyAPI):
     @model_validator(mode="before")
     @classmethod
     def validate_model(cls, values):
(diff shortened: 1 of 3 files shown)


src/ape_pm/dependency.py
```diff
@@ -405,14 +405,12 @@ def _get_version_from_package_json(
     return data.get("version")
 
 
-# TODO: Rename to `PyPIDependency` in 0.9.
```
antazoey, line 408: decided not going to do this one either

bitwise-constructs (dismissed):

bitwise-constructs (approved):

bitwise-constructs (approved):

PR #2414, opened 2024-12-11
feat: allow creating `MockContractLog` when multiple events have the same name

antazoey:
### What I did

If I have a contract with multiple event ABIs in the contract with the same, it becomes awkward to try and create mock contract logs, as we doo

For example, this:

```
assert tx.events == [c.FooEvent(bar=123)]
```

becomes:

```
assert tx.events == [c.get_event_by_signature("FooEvent(uint256 bar)"(123)]
```

So with this PR, the top one will work. Before, the top one would error out ˙◠˙ x`x`

### Checklist

### DIFF
--- a/src/ape/contracts/base.py
+++ b/src/ape/contracts/base.py
@@ -550,6 +550,16 @@ class ContractEvent(BaseInterfaceModel):
         return sum(1 for _ in logs)
 
     def __call__(self, *args: Any, **kwargs: Any) -> MockContractLog:
+        """
+        Create a mock-instance of a log using this event ABI and contract-instance.
+
+        Args:
+            *args: Position arguments for the event.
+            **kwargs: Key-word arguments for the event.
+
+        Returns:
+            :class:`~ape.types.events.MockContractLog`
+        """
         # Create a dictionary from the positional arguments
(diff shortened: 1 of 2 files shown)


bitwise-constructs (approved):

fubuloubu:
> ### What I did
> 
> If I have a contract with multiple event ABIs in the contract with the same, it becomes awkward to try and create mock contract logs, as we doo
> 
> For example, this:
> 
> ```
> assert tx.events == [c.FooEvent(bar=123)]
> ```
> 
> becomes:
> 
> ```
> assert tx.events == [c.get_event_by_signature("FooEvent(uint256 bar)"(123)]
> ```

Just double checking, there is a typo here right?

```py
assert tx.events == [c.get_event_by_signature("FooEvent(uint256 bar)")(bar=123)]
```

antazoey:
> > ```
> > assert tx.events == [c.get_event_by_signature("FooEvent(uint256 bar)"(123)]
> > ```
> 
> Just double checking, there is a typo here right?
> 
> ```python
> assert tx.events == [c.get_event_by_signature("FooEvent(uint256 bar)")(bar=123)]
> ```

yess, my bad, but yeah the point is you would have to use `get_event_by_signature()` instead of normal `.__getatttr__()`

fubuloubu (approved):

PR #2416, opened 2024-12-13
refactor: create and use disk-cache utilities for contracts-cache items

antazoey:
### What I did

OK This is a big refactor for the contract-type caching related code. Nothing is breaking but there is a lot more public stuff added. This makes thing way easier to use both for us and for users. We can no more easily improve the caching features, which is what I was trying to do for 0.9.

### How I did it

Make a base model for writing and reading from Disk
Make some stuff to help the deployments JSON patrick added a while back.
Incorporate this stuff into the caches for chain.contracts, deleting lots of repetitious code along the way.
Move contract caching stuff to its own module for faster loading.
The list goes on!

### How to verify it

looking for everything working pretty much as it did, except there are some new toolings and methods as well, which is now used all over our tests.

### Checklist


src/ape/utils/os.py
```diff
@@ -369,3 +370,69 @@ def extract_archive(archive_file: Path, destination: Optional[Path] = None):
 
     else:
         raise ValueError(f"Unsupported zip format: '{archive_file.suffix}'.")
+
+
+class CacheDirectory:
+    """
+    A directory for caching data where each data item is named
+    ``<key>.json`` and is in the directory. You can access the
```
bitwise-constructs, line 378: should this have some prefix like `cache_<key>.json`?

bitwise-constructs (approved):

src/ape/utils/os.py
antazoey: maybe, or all in `/caches` sub-folder but i dont want to change the structure now as it will be too breaking

bitwise-constructs (approved):

PR #2428, opened 2024-12-19
feat: allow provider plugins to define NAME at class-level

antazoey:
### What I did

this allow plugins like `ape-titanoboa` use the name `boa` instead of `titanoboa`.

### How I did it

NAME
(also see the TODOS, itll be cleaner in 0.9)

### How to verify it

```
pip install path/to/ape-titanoboa
ape console --network ::boa
```

### Checklist

### DIFF
--- a/src/ape/api/providers.py
+++ b/src/ape/api/providers.py
@@ -14,7 +14,7 @@ from logging import FileHandler, Formatter, Logger, getLogger
 from pathlib import Path
 from signal import SIGINT, SIGTERM, signal
 from subprocess import DEVNULL, PIPE, Popen
-from typing import TYPE_CHECKING, Any, Optional, Union, cast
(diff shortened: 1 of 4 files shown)


src/ape/api/providers.py
```diff
@@ -179,8 +179,12 @@ class ProviderAPI(BaseInterfaceModel):
     plugin or the `ape-hardhat <https://github.com/ApeWorX/ape-hardhat>`__ plugin.
     """
 
+    # TODO: In 0.9, make not optional.
```
antazoey, line 182: an alt to this would be to hav `@property name(self)` which by default returns the `__class__.name__`

antazoey: little tricky as it needs to be known prior to instantiation

bitwise-constructs (approved):

PR #2430, opened 2024-12-19
fix: could not decode ABIs when missing leading zeroes

antazoey:
### What I did

boa gives us log data  that is not always padded with leading zeroes.
i dont see why i cant just pad it myself and move on...

NOTE: We won't need this if this merges instead: https://github.com/ethereum/eth-abi/pull/240

### How I did it

make a custom decoder for uint25

question: is this supposed to happen even when strict=False? ill ask the snake charmers...

### How to verify it

see my test

### Checklist

### DIFF
--- a/src/ape/utils/abi.py
+++ b/src/ape/utils/abi.py
@@ -3,9 +3,14 @@ from collections.abc import Sequence
 from dataclasses import make_dataclass
 from typing import Any, Optional, Union
 
-from eth_abi import decode, grammar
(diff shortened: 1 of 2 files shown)


bitwise-constructs (approved):

antazoey:
> Have you seen it with any other data types?

not with strict=False, that is why im hopin my eth-abi PR goes through

bitwise-constructs (approved):

PR #2431, opened 2024-12-20
feat: alias `tx.gas` for `tx.gas_limit` and `tx.hash` for `tx.txn_hash`

antazoey:
### What I did

THIS ALLOWS US TO USE OUR TRANSACTION MODEL IN THE EVM DIRECTLY.
sorry idk why i put that in all caps...

simple alias to make transactions work better in vm directly for performance, hoping to avoid having to use so many models in boa.

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -134,6 +134,13 @@ class TransactionAPI(BaseInterfaceModel):
         """
 
     @property
+    def hash(self) -> HexBytes:
+        """
+        Alias for ``self.txn_hash``.
+        """
+        return self.txn_hash
+
+    @property
     def receipt(self) -> Optional["ReceiptAPI"]:
         """
         This transaction's associated published receipt, if it exists.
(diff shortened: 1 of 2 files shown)


src/ape/api/transactions.py
```diff
@@ -122,6 +133,13 @@ def txn_hash(self) -> HexBytes:
         The calculated hash of the transaction.
         """
 
+    @property
+    def hash(self) -> HexBytes:
```
fubuloubu, line 137: Maybe a 0.9 thing, but happy to re-name this attribute fully to match this kind of use case. don't necessarily need it to match a specific naming convention

antazoey: tx.txn_hash does feel a bit redundant

antazoey: I added a TODO comment for 0.9, i still think the alias is now for now

bitwise-constructs (approved):

PR #2440, opened 2024-12-23
chore: fix 404 status URL

linchizhen:
### What I did

fix 404 status URL

### Checklist


antazoey:
Which URL? Having a trouble finding it in the diff..

These are compiled artifacts tied to specific versions of OpenZeppelin, a bad URL is probably meant to be if it is in a published contract anyway.

Can you tell me more what brings this change?

antazoey:
This PR made me realize though, we should be using Ape's dependency system here and not hardcoding massive OZ files.
Here is an issue for this: https://github.com/ApeWorX/ape/issues/2441

If you want to work on this, we will support you. Basically, delete these files and update the ape-config to use the different OpenZeppelin versions as dependency and update any tests accordingly. That would be extremely helpful!

PR #2448, opened 2024-12-24
fix: Fix mutable default arguments in plugin configurations

ursulabauer:
### What I did

I replaced mutable default arguments (e.g., lists and dicts) with `Optional` types and initialized them properly in `__post_init__` methods. This change prevents potential issues where mutable defaults could be shared across instances, leading to unintended side effects.  

### Changes  
1. Updated `exclude` and `reports` attributes in `GasConfig` and `CoverageConfig` to use `Optional[List]` and initialized them in `__post_init__`.  
2. Ensured all mutable attributes are properly instantiated per instance to avoid shared state issues.

This ensures the configurations behave correctly in multi-instance scenarios.

### Checklist

### DIFF
--- a/src/ape_test/config.py
+++ b/src/ape_test/config.py
@@ -1,6 +1,6 @@
-from typing import NewType, Optional, Union
+from typing import NewType, Optional, Union, List
 
-from pydantic import NonNegativeInt, field_validator
+from pydantic import NonNegativeInt, field_validator, BaseModel
 
 from ape.api.config import PluginConfig
 from ape.utils.basemodel import ManagerAccessMixin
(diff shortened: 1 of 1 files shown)


fubuloubu:
Thanks for your submission @ursulabauer! That is indeed typically a big problem in most Python code, so good eyes for seeing it!

Pydantic however has special handling for exactly this scenario by performing a deep copy on every instantiation of a Model:
https://docs.pydantic.dev/latest/concepts/fields/#validate-default-values

Closing this for now, but would consider switching to `Field(default_factory=...)` as an alternative approach, as it removes this nuance of understanding the differences between pydantic and vanilla python code for mutable defaults (and may be slightly more performant)

PR #2461, opened 2025-01-07
feat: config can load project name from `pyproject.toml` project setting

antazoey:
### What I did

allow ape-config to use:

```
[project]
name = "myproject"
```

and use that as the project name so you don't necessarily have to do both configs (tool.ape as well)

### How I did it

during config load from file, grab it.

### How to verify it

make a project with pyproject.toml like

```
[project]
name = "myproject"
```

```sh
ape console -c "print(project.name)"
myproject
```

### Checklist

### DIFF
--- a/src/ape/utils/misc.py
+++ b/src/ape/utils/misc.py
@@ -218,7 +218,14 @@ def load_config(path: Path, expand_envars=True, must_exist=False) -> dict:
             contents = expand_environment_variables(contents)
 
         if path.name == "pyproject.toml":
-            config = tomllib.loads(contents).get("tool", {}).get("ape", {})
+            pyproject_toml = tomllib.loads(contents)
+            config = pyproject_toml.get("tool", {}).get("ape", {})
+
+            # Utilize [project] for some settings.
+            if project_settings := pyproject_toml.get("project"):
+                if "name" not in config and "name" in project_settings:
+                    config["name"] = project_settings["name"]
+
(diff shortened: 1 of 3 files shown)


fubuloubu:
> ### What I did
> 
> allow ape-config to use:
> 
> ```
> [project]
> name = "myproject"
> ```

This is the pyproject native name key?

src/ape/utils/misc.py
```diff
@@ -218,7 +218,14 @@ def load_config(path: Path, expand_envars=True, must_exist=False) -> dict:
             contents = expand_environment_variables(contents)
 
         if path.name == "pyproject.toml":
-            config = tomllib.loads(contents).get("tool", {}).get("ape", {})
+            pyproject_toml = tomllib.loads(contents)
+            config = pyproject_toml.get("tool", {}).get("ape", {})
+
+            # Utilize [project] for some settings.
+            if project_settings := pyproject_toml.get("project"):
```
fubuloubu, line 225: Add a note that the Ape config value overrides whatever the root pyproject value is

fubuloubu (dismissed):

src/ape/utils/misc.py
fubuloubu: *in the docs

antazoey: done

fubuloubu (approved):

PR #2463, opened 2025-01-08
fix(ops): issue template did not show a textarea for config

fubuloubu:
### What I did
Not sure if this is the solution, but here is what it looks like today

image

### Checklist


antazoey (approved):

antazoey:
I thought we had one of these at the org level?

Issue #2472, opened 2025-01-17
support RPC batch requests


bitwise-constructs:
### Describe

Alchemy and other providers have a subset of their API that supports batched requests, so we need a clean way to batch and send these

PR #2475, opened 2025-01-21
feat: `provider.batch_requests()` low-level implementation

antazoey:
### What I did

implements batch requets at the provider level, not sure about a more top level API yet (maybe something like multicall?)

fixes: #2472

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -12,7 +12,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union, cast
 
 import ijson  # type: ignore
-import requests
+import requests as requests_lib
 from eth_pydantic_types import HexBytes
 from eth_typing import BlockNumber, HexStr
 from eth_utils import add_0x_prefix, is_hex, to_hex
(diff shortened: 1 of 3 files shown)


fubuloubu (commented):
@BobTheBuidler had some sort of batching middleware for Brownie, wonder if that might be useful here to look at

0xthedance:
I had been looking at this issue, and I'm curious. What is the reason for using JSON and the requests library instead of using batch requests from web3.py? https://web3py.readthedocs.io/en/v7.7.0/web3.main.html#batch-requests

antazoey:
> What is the reason for using JSON and the requests library instead of using batch requests from web3.py? https://web3py.readthedocs.io/en/v7.7.0/web3.main.html#batch-requests

Thanks for pointing this out, let me look into it.

src/ape_ethereum/provider.py
```diff
@@ -1375,14 +1383,35 @@ def stream_request(self, method: str, params: Iterable, iter_path: str = "result
...
+
+        response = requests_lib.post(uri, json=requests)
+        try:
+            response.raise_for_status()
+        except HTTPError as err:
+            raise ProviderError(str(err)) from err
+
+        return [
+            self._get_result_from_rpc_response(uri, r, raise_on_failure=False)
+            for r in response.json()
```
banteg, line 1412: i remember a bug where erigon returned results out of order, so to be safe we can sort by id

src/ape_ethereum/provider.py
```diff
@@ -1375,14 +1383,35 @@ def stream_request(self, method: str, params: Iterable, iter_path: str = "result
...
 
+    def batch_requests(self, requests: list[dict]) -> Any:
+        if not (uri := self.http_uri):
+            raise ProviderError("This provider has no HTTP URI and is unable to batch requests.")
+
+        for idx, request in enumerate(requests):
+            if "jsonrpc" not in request:
+                request["jsonrpc"] = "2.0"
+            if "id" not in request:
+                request["id"] = idx + 1
```
banteg, line 1402: - enumerate has a second arg for start index
- if we allow id overrides (any good reason for that?), then we need to check there are no duplicates

PR #2477, opened 2025-01-22
feat: validate vrs

NotPeopling2day:
### What I did

fixes: #

### Checklist


src/ape/types/signatures.py
```diff
@@ -84,6 +85,29 @@ class _Signature:
     The signature proof point (``s``) in an ECDSA signature.
     """
 
+    @field_validator("v", mode="before")
```
antazoey, line 88: you can use the `HexInt` type, it should do all this automatically when using that type

antazoey (commented):
look into `ape.types.basic.HexInt`

PR #2479, opened 2025-01-27
fix(bug): prefix config environment vars to avoid conflicts

ydm:
### What I did

The configuration classes (e.g. `ape_ethereum.NetworkConfig` or `ape_test.CoverageReportsConfig`) incorrectly interpret environment variables as part of their own settings.

For example there are `terminal` and `html` fields in `CoverageReportsConfig`:

```Python
class CoverageReportsConfig(PluginConfig):
    terminal: Union[bool, dict] = True
    xml: Union[bool, dict] = false
    html: Union[bool, dict] = False
```

Now, if you try running `ape` with one of those (somewhat common) names set as an environment variable, it will either crash or, even worse, run with one of its settings likely augmented without the user ever intending it.

```
$ TERMINAL=xterm ape compile

...

pydantic_core._pydantic_core.ValidationError: 2 validation errors for CoverageReportsConfig
terminal.bool
  Input should be a valid boolean, unable to interpret input [type=bool_parsing, input_value='xterm', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/bool_parsing
terminal.dict[any,any]
  Input should be a valid dictionary [type=dict_type, input_value='xterm', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
```

The reason for this behaviour is that the `pydantic-settings` module searches for environment variables of the same name as each setting: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#environment-variable-names

I couldn't find a way to disable the environment variable settings.

To solve the problem, this patch introduces the following prefixes based on their root module:
- `APE_`
- `APE_CACHE_`
- `APE_COMPILE_`
- `APE_CONSOLE_`
- `APE_ETHEREUM_`
- `APE_NETWORKS_`
- `APE_NODE_`
- `APE_TEST_`

The following config classes are affected:

```
| File                                 | Class                   | Prefix                    | Test |
|--------------------------------------+-------------------------+---------------------------+------|
| src/ape/api/config.py                | ApeConfig               | APE_                      | yes  |
| src/ape/api/config.py                | DeploymentConfig        | APE_                      | NO!  |
| src/ape/api/config.py                | PluginConfig            | APE_                      | NO!  |
| src/ape_cache/config.py              | CacheConfig             | APE_CACHE_                | yes  |
| src/ape_compile/config.py            | Config                  | APE_COMPILE_              | yes  |
| src/ape_console/config.py            | ConsoleConfig           | APE_CONSOLE_              | yes  |
| src/ape_ethereum/ecosystem.py        | BaseEthereumConfig      | APE_ETHEREUM_             | yes  |
| src/ape_ethereum/ecosystem.py        | ForkedNetworkConfig     | APE_ETHEREUM_ (inherited) | yes  |
| src/ape_ethereum/ecosystem.py        | NetworkConfig           | APE_ETHEREUM_             | yes  |
| src/ape_networks/config.py           | CustomNetwork           | APE_NETWORKS_             | yes  |
| src/ape_networks/config.py           | NetworksConfig          | APE_NETWORKS_             | NO!  |
| src/ape_node/provider.py             | EthereumNetworkConfig   | APE_NODE_                 | yes  |
| src/ape_node/provider.py             | EthereumNodeConfig      | APE_NODE_                 | yes  |
| src/ape_test/config.py               | ApeTestConfig           | APE_TEST_                 | yes  |
| src/ape_test/config.py               | CoverageConfig          | APE_TEST_                 | yes  |
| src/ape_test/config.py               | CoverageReportsConfig   | APE_TEST_                 | yes  |
| src/ape_test/config.py               | EthTesterProviderConfig | APE_TEST_                 | yes  |
| src/ape_test/config.py               | GasConfig               | APE_TEST_                 | yes  |
| src/ape_test/config.py               | GasExclusion            | APE_TEST_                 | yes  |
| src/ape_test/config.py               | IsolationConfig         | APE_TEST_                 | yes  |
| tests/functional/test_config.py      | CustomConfig            | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | CustomConfig            | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | MyConfig                | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | MyConfig                | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | MyConfig                | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | MyConfig                | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | SubConfig               | APE_API_ (inherited)      | -    |
| tests/functional/test_config.py      | SubConfig               | APE_API_ (inherited)      | -    |
| tests/functional/test_ecosystem.py   | L2NetworkConfig         | APE_ETHEREUM_ (inherited) | -    |
| tests/functional/test_network_api.py | MyConfig                | APE_ETHEREUM_ (inherited) | -    |
```

Thus the previously used as an example `TERMINAL` environment variable now becomes `APE_TEST_TERMINAL`. The same applies respectively to all other settings as well.

This patch carries the risk of invalidating the settings of users who rely on environment variables based configuration (even though this seems highly unlikely), so make sure to include it in the changelog as a breaking change, should this patch get approved.


src/ape/api/config.py
```diff
@@ -66,7 +66,7 @@ class PluginConfig(BaseSettings):
     a config API must register a subclass of this class.
     """
 
-    model_config = SettingsConfigDict(extra="allow")
+    model_config = SettingsConfigDict(extra="allow", env_prefix="APE_API_")
```
antazoey, line 69: Thoughts on making the default base-level prefix `APE_` instead of `APE_API_`?

ydm: That would create a common namespace for all settings across all config classes and make no distinction between different purposes, but I'd still prefer it over the option with prefix.

antazoey: > That would create a common namespace for all settings across all config classes

`APE_NODE_` and `APE_TEST_` etc still would be valid, which are plugins, so I am not sure how this is true.

antazoey:
So I tried to see if I could use `APE_API_CONTRACTS_FOLDER` but it didn't work.
I realize the problem was in the `ApeConfg.__init__`.

I had to change:

```
super(BaseSettings, self).__init__(*args, **kwargs)
```

to

```
super().__init__(*args, **kwargs)
```

At that point I can add a test:

```
def test_model_validate_handles_environment_variables():
    os.environ["APE_API_CONTRACTS_FOLDER"] = "contracts-env-var-test"
    cfg = ApeConfig()
    assert cfg.contracts_folder == "contracts-env-var-test"
```

ydm:
OK, let me do that, will get back to you shortly.

src/ape/api/config.py
ydm: I misunderstood you, will fix that as well.

antazoey:
> Maintainer note: Breaking change for 0.9

I am arguing against this for a couple reasons.
1, issues the user mentioned are bad enough to change now. Like a mere `TERMINAL` environment variable causes problems
2, the feature is un-documented; i was not aware we were supporting environment variables like TERMINAL, etc.

src/ape/api/config.py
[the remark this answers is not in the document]
antazoey: I had the same comment: https://github.com/ApeWorX/ape/pull/2479#discussion_r1931002642

fubuloubu:
> > Maintainer note: Breaking change for 0.9
> 
> I am arguing against this for a couple reasons. 1, issues the user mentioned are bad enough to change now. Like a mere `TERMINAL` environment variable causes problems 2, the feature is un-documented; i was not aware we were supporting environment variables like TERMINAL, etc.

Fair enough! Agree with releasing as a patch update

Documentation can be updated in a later PR

src/ape/api/config.py
fubuloubu: A further note, the idea with 2nd/3rd party plugins (e.g. those not bundled with the core ape installation) would have their own prefixes automatically injected via `PluginConfig`

Is there a way to dynamically source this based on which module a subclass is created in?

E.g. `ape-alchemy` should automatically apply `APE_ALCHEMY_` prefix to it's config

ydm: Perhaps there is a way to achieve something dynamic using the `inspect` module, but that sounds so risky and unreliable to me. I, even as a user, would prefer the simpler approach of specifying the `envprefix` manually.

fubuloubu: Let's say you have two provider plugins and they both have the same config key:

```yaml
alchemy:
  api-key: AbCd

infura:
  api-key: EfGh
```

Using this format they would both be called `APE_API_API_KEY` (because the namespacing of the plugin modules are processed separately currently, their names I don't think will show up in the env var), and thus they would overwrite each other

The ideal would be if they showed up like `APE_ALCHEMY_API_KEY` and `APE_INFURA_API_KEY` so that they both end up under separate namespaces

Not something we have to tackle now, but that was the goal we had in mind in adding support for env var loading. It is possible to do when loading the inherited plugin config through the plugin registration phase, we can just set `ConfigClass.model_config = SettingsConfigDict(...)` prior to loading their config (or apparently adding `_env_prefix=...` on class initialization, [see docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#validation-of-default-values))

antazoey: I was thinking that same thing, could be a feature of `PluginConfig` and plugins can still customize if they need. But at the same time, it isnt that hard to just have each plugin set the value for now.

ydm:
@antazoey , please check the latest version, it includes tests and docs.

src/ape_compile/main.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+
+print(__package__)
```
antazoey, line 3: what is going on here? is this file needed?

docs/userguides/config.md
```diff
@@ -37,6 +37,55 @@ plugin:
 
 This helps keep your secrets out of Ape!
 
+If a configuration is left unset (i.e., not included in the
```
antazoey, line 40: thanks for the docs! don't be alarmed if i push some nitty commits here before merge

docs/userguides/config.md
```diff
@@ -37,6 +37,55 @@ plugin:
...
+APE_CONTRACTS_FOLDER=src/contracts
+APE_TEST_NUMBER_OF_ACCOUNTS=3
+APE_TEST_SHOW_INTERNAL=true
+APE_COMPILE_EXCLUDE='["one", "two", "three"]'
+APE_COMPILE_INCLUDE_DEPENDENCIES=true
+```
+
+Here is the complete list of supported prefixes:
+
+| Module       | Prefix       |
```
antazoey, line 73: (i can fix these with my nitty commits), but just to clarify, these are not actually modules but are packages shipped along with ape! like `ape.ethereum` would be a module but `ape_ethereum` is a package.  that is the nature of ape's plugin architecture, we ship plugins with ape so it works out of the box.

tests/functional/test_config.py
```diff
@@ -618,3 +623,69 @@ def test_project_level_settings(project):
         assert project.config.my_string == "my_string"
         assert project.config.my_int == 123
         assert project.config.my_bool is True
+
+
+def test_model_validate_handles_environment_variables():
+    def f(cls: Callable, attr: str, name: str, value: str, expected: Any = None):
```
antazoey, line 629: nit: pep8 standards advocates against 1 word definition names

antazoey (commented):
Amazing! Thank you so much. I think we can take it from here, just a few nits I can address and then merge.

antazoey (dismissed):

antazoey (approved):

antazoey (approved):

Issue #1672, opened 2023-09-23
Allow passing additional flags to `ape networks run`


fubuloubu:
### Overview

It would be really nice to be able to pass additional flags to subprocess providers when running them via `ape networks run` (specifically `--block-time`)

### Specification

If the click command `ape networks run` used extra args (like `ape test`) and passed them through to the subprocess command, it would probably be the nicest way of adding this feature

### Dependencies
n/a

PR #2482, opened 2025-01-28
feat: Include additional flags to ape networks run

0xthedance:
### What I did

Allows passing additional flags to the command `ape networks run', such as: `--block-time`,`--base-fee-multiplier`, `--transaction-timeout`

fixes: #1672

### Checklist

### DIFF
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -996,6 +996,10 @@ class NetworkAPI(BaseInterfaceModel):
         """
         return self.config.get("base_fee_multiplier", 1.0)
 
+    @base_fee_multiplier.setter
+    def base_fee_multiplier(self, value: float):
+        self.config["base_fee_multiplier"] = value
+
     @property
     def chain_id(self) -> int:
         """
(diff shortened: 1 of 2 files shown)


fubuloubu (dismissed):
Love it! Looking good, what level of testing have you done with it so far?

I think if we can test with a few different providers, we would be good to go on this

antazoey:
I got the flakey test fixed on another PR.
Consider adding integration tests next to the others in `tests/integration/cli/test_networks.py`, they don't have to do much, besides ensure you can pass in values for these.

antazoey:
does setting `--block-time` actually set the value in `anvil` to? that's neat!

0xthedance:
Actually, it didn't do it. But I've fixed it, and now it does configure the block-time for Anvil.I'm not sure about the local Geth. I think it doesn't allow you to configure the block-time.

antazoey:
> I'm not sure about the local Geth. I think it doesn't allow you to configure the block-time.

I added a way here: https://github.com/ApeWorX/ape/pull/2487/files
So it will work after this

antazoey (approved):

PR #2488, opened 2025-01-31
perf: avoid all default web3 middleware

antazoey:
### What I did

You can avoid all the middleware completely from web3 if you do `middleware=[]` at the initialization sequence.

I didn't really see much of a performance gain btw

### How I did it

middleware=[] at the initialization sequence

### How to verify it

[x for x in web3.middleware_onion] is empty now, isn't before.
### Checklist


fubuloubu:
> You can avoid all the middleware completely from web3 if you do `middleware=[]` at the initialization sequence.

What are the default ones that get installled?

antazoey:
> What are the default ones that get installled?

```
(Pdb) w3 = Web3(provider)
(Pdb) middys = [x for x in w3.middleware_onion]
(Pdb) middys
[
    <class 'web3.middleware.gas_price_strategy.GasPriceStrategyMiddleware'>,
    <class 'web3.middleware.names.ENSNameToAddressMiddleware'>,
    <class 'web3.middleware.attrdict.AttributeDictMiddleware'>,
    <function FormattingMiddlewareBuilder.build at 0x1047fea20>,
    <class 'web3.middleware.buffered_gas_estimate.BufferedGasEstimateMiddleware'>
]

```

**GasPriceStrategyMiddleware** = Something for pre-1559 transactions but I think only does anything if you actually have set a gas price strategy. That being said, it is possibly for users to be using a web3 provider from ape and setting a gas price strategy.

**ENSNameToAddressMiddleware** = This basically does what we are already doing with `ape-ens`... I suppose it is nice to have on web3 because it allows users to use ens without the `ape-ens` plugin, in some contexts. Although for performance reasons, I wouldn't want to have ens names in requests. I would recommend resolving them once and using them there after. That is why `ape-ens` caches. Resolving ENS takes a long time. Because of the performance concerns and the existence of our plugin.

** AttributeDictMiddleware** - the serialization thing we are trying to get rid of.

** FormattingMiddlewareBuilder** - I am not sure what the really is, it seems to assemble the middlewares though? Removing it didn't seem to do anything.

** BufferedGasEstimateMiddleware ** - This includes gas estimates when they are missing, something that probably isn't needed if using Ape. Though, you can use web3 directly from Ape and you wouldn't have this anymore.

I think it is generally safe to remove these and allows users to opt in to ones they need.

fubuloubu (approved):
Looks good! Thank you for documenting reasoning behind removing the default middlewares

Issue #2259, opened 2024-08-30
Easier ways to exclude CompilerAPIs from `ape compile` command


antazoey:
### Overview

right now to compile only vyper contracts, i do this:

```sh
ape compile contracts/*.vy
```

but that approach is a bit limited.

### Specification

Have it automatically work w/o full path:

```sh
# only compile .vy contracts
ape compile *.vy
```

idea: `--exclude-compiler` flag?

```sh
ape compile --exclude-compiler solidity
```

exclude_compiler config:

```yaml
compile:
  exclude_compiler:
    - solidity
```

basically i dont want to uninstall `ape-solidity` but i dont want to compile solidity contracts in some projects.
this may also be nice for some dependencies

### Dependencies

Include links to any open issues that must be resolved before this feature can be implemented.

PR #2494, opened 2025-02-06
feat: add --exclude-compiler flag to ape compile command

0xthedance:
### What I did

Add a new flag to the `ape compile` command that allows to specify a compiler to be excluded from the compilation:

```
ape compile  --exclude-compiler
```

fixes: #2259

### How I did it

Before invoking compiler's `compile` method , the code checks if  `exclude-compiler` parameter matches the current compiler. If it does, the method skips that compiler and moves to the next one.

### How to verify it
If you think is correct I could proceed with the tests.


src/ape_compile/_cli.py
```diff
@@ -42,6 +42,11 @@ def _include_dependencies_callback(ctx, param, value):
     help="Also compile dependencies",
     callback=_include_dependencies_callback,
 )
+@click.option(
```
antazoey, line 45: This may also be useful in the `ape pm compile` command. We could potentially add this to the `ape.cli.options` module so it can be re-used.

src/ape_compile/_cli.py
```diff
@@ -42,6 +42,11 @@ def _include_dependencies_callback(ctx, param, value):
...
+    "--exclude-compiler",
+    type=click.Choice(["solidity", "vyper"], case_sensitive=False),
```
antazoey, line 47: This shouldn't be a click.Choice. There are more compilers besides `solidity` and `vyper`. With Ape's plugin system, there isn't a fixed number of compilers.

A normal `str` type is just find.

src/ape/managers/project.py
```diff
@@ -449,11 +449,15 @@ def _get_needs_compile(self, paths: Iterable[Union[Path, str]]) -> Iterable[Path
                 else:
                     yield path
 
-    def _compile_contracts(self, paths: Iterable[Union[Path, str]]):
+    def _compile_contracts(
+        self, paths: Iterable[Union[Path, str]], exclude_compiler: Optional[str] = None
```
antazoey, line 453: maybe this should be a list of compiler names

src/ape/cli/options.py
```diff
@@ -597,3 +597,12 @@ def _json_option(name, help, **kwargs):
 
 def config_override_option(**kwargs):
     return _json_option("--config-override", help="Config override mappings", **kwargs)
+
+
+def exclude_compiler_option(**kwargs):
+    return click.option(
+        "--exclude-compiler",
+        help="Exclude a specific compiler from the compilation process",
+        type=str,
```
antazoey, line 606: we should do `is_multiple=True` here so multiple compilers can be excluded

src/ape/managers/project.py
```diff
@@ -2592,7 +2601,10 @@ def update_manifest(self, **kwargs):
         self.manifest_path.write_text(manifest_text, encoding="utf8")
 
     def load_contracts(
-        self, *source_ids: Union[str, Path], use_cache: bool = True
+        self,
+        *source_ids: Union[str, Path],
+        use_cache: bool = True,
+        exclude_compiler: Optional[Union[str, list[str]]] = None,
```
fubuloubu, line 2607: Simplify typing here
```suggestion
        excluded_compilers: Optional[list[str]] = None,
```

fubuloubu (approved):
Cleaned up the tensing of this variable a bit

src/ape/cli/options.py
```diff
@@ -597,3 +597,13 @@ def _json_option(name, help, **kwargs):
 
 def config_override_option(**kwargs):
     return _json_option("--config-override", help="Config override mappings", **kwargs)
+
+
+def exclude_compiler_option(**kwargs):
```
fubuloubu, line 602:
```suggestion
def excluded_compilers_option(**kwargs):
```

src/ape_compile/_cli.py
```diff
@@ -5,7 +5,12 @@
 import click
 
 from ape.cli.arguments import contract_file_paths_argument
-from ape.cli.options import ape_cli_context, config_override_option, project_option
+from ape.cli.options import (
+    ape_cli_context,
+    config_override_option,
+    exclude_compiler_option,
```
fubuloubu, line 11:
```suggestion
    excluded_compilers_option,
```

src/ape_compile/_cli.py
```diff
@@ -42,6 +47,7 @@ def _include_dependencies_callback(ctx, param, value):
     help="Also compile dependencies",
     callback=_include_dependencies_callback,
 )
+@exclude_compiler_option()
```
fubuloubu, line 50:
```suggestion
@excluded_compilers_option()
```

fubuloubu (approved):
Just a few minor suggestions

fubuloubu (approved):

fubuloubu:
Looks good to me, @antazoey can merge when ready

src/ape/managers/compilers.py
```diff
@@ -137,6 +138,8 @@ def compile(
 
         for next_ext, path_set in files_by_ext.items():
             compiler = self.registered_compilers[next_ext]
+            if excluded_compilers and compiler.name.lower() in [c.lower() for c in excluded_compilers]:
```
fubuloubu, line 141: This is the line that needs ruff to reformat, however looking at it makes me think that we can move this logic into the callback of `excluded_compilers_option` so we don't have to do this here

Potentially by using `click.Choice` with the keys of `CompilerManager.registered_compilers`

0xthedance: I moved the for loop logic to the callback.did you mean that?

src/ape/cli/options.py
```diff
@@ -599,12 +599,27 @@ def config_override_option(**kwargs):
...
+    return [c.lower() for c in value]
+
+
 def excluded_compilers_option(**kwargs):
+    from ape.utils.basemodel import ManagerAccessMixin
+
+    registered_compilers_options = [
+        compiler.name
+        for compiler in ManagerAccessMixin.compiler_manager.registered_compilers.values()
+    ]
```
fubuloubu, line 615: They are already listed by name
```suggestion
    registered_compilers_options = list(
        ManagerAccessMixin.compiler_manager.registered_compilers
    )
```

fubuloubu (dismissed):

src/ape/managers/compilers.py
fubuloubu: yes, that looks perfect!

src/ape/cli/options.py
```diff
@@ -597,3 +597,28 @@ def _json_option(name, help, **kwargs):
...
+
+    return [c.lower() for c in value]
+
+
+def excluded_compilers_option(**kwargs):
+    from ape.utils.basemodel import ManagerAccessMixin
+
+    registered_compilers_options = list(
+        ManagerAccessMixin.compiler_manager.registered_compilers
+    )
```
fubuloubu, line 614: I have to stop editing from github lmao
```suggestion
    registered_compilers_options = list(ManagerAccessMixin.compiler_manager.registered_compilers)
```

fubuloubu (approved):

src/ape/cli/options.py
0xthedance: Now they are listed by file extension. Now to exclude a compiler, you should use the extension [.vy|.vyi|.sol|.json]. Isn't it more intuitive using  the compiler name: Vyper, Solidity?

fubuloubu: You are totally right, my apologies on that one

fubuloubu (approved):

0xthedance:
I don't know why the CI is failing. Locally, all tests pass with python 3.11

PR #2505, opened 2025-02-13
feat: adds `ContractLog.topics` property for calculating the topics from a log

antazoey:
### What I did

I need this for implementing `get_contract_logs()` in `ape-titanoboa`

### Checklist

### DIFF
--- a/src/ape/types/events.py
+++ b/src/ape/types/events.py
@@ -79,7 +79,7 @@ class LogFilter(BaseModel):
         Construct a log filter from an event topic query.
         """
         abi = getattr(event, "abi", event)
-        topic_filter = encode_topics(abi, search_topics)
+        topic_filter = encode_topics(abi, search_topics or {})
         return cls(
             addresses=addresses or [],
             events=[abi],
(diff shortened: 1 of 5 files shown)


src/ape/types/events.py
```diff
@@ -256,6 +219,16 @@ def abi(self) -> EventABI:
         self._abi = abi
         return abi
 
+    @cached_property
+    def topics(self) -> list[HexStr]:
```
antazoey, line 223: If we are worried about name clashing here, we could make this a method, e.g.:

```
def get_topics()`
```

or

```
@property
def encoded_topics
```
or whatever else

fubuloubu (approved):
Wonder if this could actually move a bit upstream into `ethpm_types.EventABI`, would be useful here: https://github.com/ApeWorX/silverback/blob/main/silverback/runner.py#L355

antazoey:
> Wonder if this could actually move a bit upstream into `ethpm_types.EventABI`, would be useful here: https://github.com/ApeWorX/silverback/blob/main/silverback/runner.py#L355 (used for creating streaming subscriptions)

Done! https://github.com/ApeWorX/ethpm-types/pull/148

antazoey:
This will allow us to actually get `ape-titanoboa` to implement `.get_contract_logs()` and honor the given topics.

bitwise-constructs (approved):

antazoey:
Now https://github.com/ApeWorX/ape-titanoboa/pull/17 is unblocked

PR #2509, opened 2025-02-16
fix: delete unused, misleading, and error-prone method

antazoey:
### What I did

this method is just... wrong... it is like it is supposed to be on TranssactionAPI but it is on EcosystemAPI. It is not used anywhere, wouldnt work even if you called it

### Checklist


bitwise-constructs (approved):

bitwise-constructs:
Chuck it into the sea

Issue #2511, opened 2025-02-17
bug: 'NoneType' object has no attribute 'model_dump' in `ContractCache.get`


fubuloubu:
Bug is in this code:
https://github.com/ApeWorX/ape/blob/ced89abfb0718959f4ca66ce0e6a4652d1385736/src/ape/managers/_contractscache.py#L574-L587

if `implementation_contract_type` is `None` (from `self.get`), then the logic will attempt to store a `None` value, which is not a valid `ContractType`.

Should probably raise here instead

antazoey:
Can you show more of the error? I think you're missing something as I can't reproduce:

```python
(Pdb) self.contract_types[address_key] = None
(Pdb) address_key in self.contract_types
```

antazoey:
The linked PR shows the efforts and addresses the problem you showed, but I don't think that actually is the problem.

fubuloubu:
> Can you show more of the error? I think you're missing something as I can't reproduce:
> 
> (Pdb) self.contract_types[address_key] = None
> (Pdb) address_key in self.contract_types
> False

My set up was not having an explorer installed and trying to fetch a proxy (USDC on Optimism I think)

PR #2517, opened 2025-02-18
fix: avoid attempting to cache None for a contract-type

antazoey:
### What I did

Attempted to fix 2511 but I can't reproduce. My tests still pass on the main branch.

fixes: #2511

### DIFF
--- a/src/ape/managers/_contractscache.py
+++ b/src/ape/managers/_contractscache.py
@@ -577,15 +577,18 @@ class ContractCache(BaseManager):
                     if fetch_from_explorer
                     else None
                 )
+                contract_type_to_cache = None
+
                 if proxy_contract_type:
                     contract_type_to_cache = _get_combined_contract_type(
                         proxy_contract_type, proxy_info, implementation_contract_type
                     )
-                else:
+                elif implementation_contract_type is not None:
                     contract_type_to_cache = implementation_contract_type
 
-                self.contract_types[address_key] = contract_type_to_cache
-                return contract_type_to_cache
+                if contract_type_to_cache is not None:
+                    self.contract_types[address_key] = contract_type_to_cache
+                    return contract_type_to_cache
 
             # Also gets cached to disk for faster lookup next time.
             if fetch_from_explorer:
--- a/src/ape/managers/_contractscache.py
+++ b/src/ape/managers/_contractscache.py
@@ -577,18 +577,21 @@ class ContractCache(BaseManager):
                     if fetch_from_explorer
                     else None
                 )
-                contract_type_to_cache = None
-
-                if proxy_contract_type:
-                    contract_type_to_cache = _get_combined_contract_type(
+                if proxy_contract_type is not None and implementation_contract_type is not None:
+                    combined_contract = _get_combined_contract_type(
                         proxy_contract_type, proxy_info, implementation_contract_type
                     )
+                    self.contract_types[address_key] = combined_contract
+                    return combined_contract
+
                 elif implementation_contract_type is not None:
                     contract_type_to_cache = implementation_contract_type
-
-                if contract_type_to_cache is not None:
-                    self.contract_types[address_key] = contract_type_to_cache
+                    self.contract_types[address_key] = implementation_contract_type
                     return contract_type_to_cache
+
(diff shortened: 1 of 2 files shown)


src/ape/managers/_contractscache.py
```diff
@@ -577,15 +577,18 @@ def get(
                     if fetch_from_explorer
                     else None
                 )
+                contract_type_to_cache = None
+
                 if proxy_contract_type:
                     contract_type_to_cache = _get_combined_contract_type(
                         proxy_contract_type, proxy_info, implementation_contract_type
```
fubuloubu, line 584: Here was also a problem, because it would try and merge `implementation_contract_type = None` w/ `proxy_contract_type`

src/ape/managers/_contractscache.py
```diff
@@ -577,15 +577,18 @@ def get(
...
                     )
-                else:
+                elif implementation_contract_type is not None:
                     contract_type_to_cache = implementation_contract_type
 
-                self.contract_types[address_key] = contract_type_to_cache
-                return contract_type_to_cache
+                if contract_type_to_cache is not None:
+                    self.contract_types[address_key] = contract_type_to_cache
+                    return contract_type_to_cache
```
fubuloubu, line 591: Honestly feel like it might be better to clean up this logic a bit by returning directly in the nested `if` statements e.g.

```py
if proxy_contract_type is not None:
    if not implementation_contract_type:
        # If we haven't detected an implementation but know it's a proxy, that's an issue
        raise  # some error
        # but also, should we try to fetch from explorer above this statement?

    contract_type_to_cache = _get_combined_contract_type(
        proxy_contract_type, proxy_info, implementation_contract_type
    )
    self.contract_types[address_key] = contract_type_to_cache
    return contract_type_to_cache

elif implementation_contract_type is not None:
    self.contract_types[address_key] = implementation_contract_type
    return implementation_contract_type

# else: pass-through to fetching from explorer (should we do this above?)
```

antazoey: done

src/ape/managers/_contractscache.py
antazoey: ok i think i got it fixed now

src/ape/managers/_contractscache.py
```diff
@@ -577,15 +577,21 @@ def get(
...
+                    return combined_contract
+
+                elif implementation_contract_type is not None:
                     contract_type_to_cache = implementation_contract_type
+                    self.contract_types[address_key] = implementation_contract_type
+                    return contract_type_to_cache
 
-                self.contract_types[address_key] = contract_type_to_cache
-                return contract_type_to_cache
+                elif proxy_contract_type is not None:
```
fubuloubu, line 592: so actually, this scenario is kind of a bug because if we know it's a proxy but don't know it's implementation ABI, you can't really do much with it

antazoey: You can have a proxy without it having an implementation set yet, say for example you deployed it yourself but have yet to set the implementation. I don't see how that is a bug.

antazoey: There are still ABIs available on the proxy contract itself, still things you can do etc.

fubuloubu: > You can have a proxy without it having an implementation set yet, say for example you deployed it yourself but have yet to set the implementation. I don't see how that is a bug.

that's actually a good point, okay I'm on board now

fubuloubu (approved):

src/ape/managers/_contractscache.py
antazoey: Oh wait, no you're right... In this case, the implementation contract _is_ set, but for some reason, we were unable to fetch its contact type or we don't know it. 🤔

antazoey: I suppose it could be the case that the user isn't using an explorer and purposely deleted the contract type from the cache, which is a valid set of moves.

antazoey: It gets confusing though, because it is even hard to know if an implementation is set or not, as it may be a placeholder address

Issue #2519, opened 2025-02-18
Show first 4 bytes of `data`, not just 3


fubuloubu:
### Have you ensured that all of these are up to date?

- [x] Ape
- [x] Ape plugins

### What version of Ape are you on?

0.8.26

### Operating System

Linux

### Description

Noticed that when submitting a contract transaction, it shows the first 3 bytes of `data` when I am signing, and it should show 4 (because most methods are 4 bytes method_id + args)

antazoey:
I'm guessing this is referring to the keyfile account's sign?

PR #2522, opened 2025-02-19
fix: show first 4 bytes of transaction data in signing string (instead of 3)

antazoey:
### What I did

Transaction showed data like (3 bytes)

```
data: 0x307833...303762
```

but is now (4 bytes):

```
data: 0x30783334...30303762
```

fixes: #2519 

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -190,13 +190,13 @@ class TransactionAPI(BaseInterfaceModel):
             if isinstance(data["data"], str):
                 data["data"] = (
                     "0x"
-                    + bytes(data["data"][:3], encoding="utf8").hex()
+                    + bytes(data["data"][:4], encoding="utf8").hex()
                     + "..."
-                    + bytes(data["data"][-3:], encoding="utf8").hex()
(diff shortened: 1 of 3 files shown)


bitwise-constructs (approved):

src/ape/api/transactions.py
fubuloubu, line 1: should we make this configurable? like displaying the full data vs. only displaying partial?

```yaml
accounts:
  show_full_calldata: true  # defaults `false`
```

fubuloubu (approved):
sweet. now I can go "What is method_id `0x30783334` in etherscan?" really quickly

src/ape/api/transactions.py
antazoey: I am always done for using config to change behavior like that

antazoey:
Question: what about also showing the actual method name? We should know it.

src/ape/api/transactions.py
antazoey: ill make another PR for this feature

antazoey: https://github.com/ApeWorX/ape/pull/2523

PR #2523, opened 2025-02-19
feat: allow configuring accounts to show full calldata when signing

antazoey:
### What I did

```yaml
calldata_repr: full  # defaults to "abridged", I plan on also adding "decoded" and maybe making the default "decoded"
```

### Checklist


src/ape_accounts/config.py
```diff
@@ -0,0 +1,14 @@
+from ape.api.config import PluginConfig
+
+
+class AccountsConfig(PluginConfig):
+    """
+    Config accounts generally.
+    """
+
+    show_full_calldata: bool = False
```
antazoey, line 9: I am wondering if we should make this like an enum instead so we can support other forms of calldata, such "decoded"

```
accounts:
  calldata_repr: "full"  #  defaults to "abridged"
```

src/ape_accounts/config.py
```diff
@@ -0,0 +1,21 @@
+from enum import Enum
+
+from ape.api.config import PluginConfig
+
+
+class CalldataRepr(str, Enum):
+    full = "full"
+    abridged = "abridged"
```
antazoey, line 8: I plan to add "decoded" here and it will attempt to decode the calldata if possible

fubuloubu: decoded would be amazing tbh

@BowTiedDevil was saying the same yesterday

src/ape/api/transactions.py
```diff
@@ -183,26 +183,47 @@ def __repr__(self) -> str:
...
-        # NOTE: Using JSON mode for style.
-        data = self.model_dump(mode="json")
-        if len(data["data"]) > 9:
-            # only want to specify encoding if data["data"] is a string
-            if isinstance(data["data"], str):
-                data["data"] = (
+        return self.to_string()
+
+    def to_string(
+        self, calldata_repr: Optional[Union[Literal["full"], Literal["abridged"]]] = None
```
fubuloubu, line 189: Can this use `CalldataRepr` enum?

src/ape_accounts/__init__.py
```diff
@@ -1,6 +1,13 @@
 from importlib import import_module
 
-from ape.plugins import AccountPlugin, register
+from ape.plugins import AccountPlugin, Config, register
+
+
+@register(Config)
+def config_class():
```
fubuloubu, line 7: Wonder if this instead should be a "core config" item, e.g.
```yaml
display:
  transactions: abridged
```

fubuloubu: Mostly since behavior isn't just limited to `ape-accounts`

src/ape/api/transactions.py
antazoey: It could, but I may just have to move the enum out of the core plugin then.

src/ape_accounts/__init__.py
antazoey: Good point, I like that

src/ape_accounts/config.py
antazoey: It'll be a follow up to this!

src/ape/api/transactions.py
fubuloubu: I think that makes sense if you turn this into a core display configuration instead of just a config for ape-accounts

antazoey: done

src/ape/api/transactions.py
```diff
@@ -183,26 +184,45 @@ def __repr__(self) -> str:
...
+              to see the full calldata. Defaults to the value from the config.
+
+        Returns:
+            str
+        """
+        data = self.model_dump(mode="json")  # JSON mode used for style purposes.
+
+        if calldata_repr is None:
+            # If was not specified, use the default value from the config.
+            calldata_repr = self.local_project.config.calldata_repr
```
fubuloubu, line 204: What do you think about namespacing it under `display: ` key? In case we add more core configuration for display purposes later?

antazoey: done

fubuloubu (approved):

PR #2524, opened 2025-02-19
fix: unnecessarily was using explorer when providing proxy info manually

antazoey:
### What I did

It seems to unnecessarily use the explorer when you provider the proxy manually and the contract type is already cached.

### Checklist

### DIFF
--- a/src/ape/managers/_contractscache.py
+++ b/src/ape/managers/_contractscache.py
@@ -556,9 +556,11 @@ class ContractCache(BaseManager):
         if contract_type := self.contract_types[address_key]:
             # The ContractType was previously cached.
             if default and default != contract_type:
-                # Replacing contract type
-                self.contract_types[address_key] = default
-                return default
+                # The given default ContractType is different than the cached one.
+                # Merge the two and cache the merged result.
+                combined_contract_type = _merge_contract_types(contract_type, default)
+                self.contract_types[address_key] = combined_contract_type
+                return combined_contract_type
 
             return contract_type
 
@@ -927,9 +929,7 @@ def _get_combined_contract_type(
     proxy_info: ProxyInfoAPI,
     implementation_contract_type: ContractType,
 ) -> ContractType:
-    proxy_abis = [
-        abi for abi in proxy_contract_type.abi if abi.type in ("error", "event", "function")
-    ]
+    proxy_abis = _get_relevant_additive_abis(proxy_contract_type)
 
     # Include "hidden" ABIs, such as Safe's `masterCopy()`.
     if proxy_info.abi and proxy_info.abi.signature not in [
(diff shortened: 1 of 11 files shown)


src/ape/managers/_contractscache.py
```diff
@@ -606,6 +591,38 @@ def get(
...
+            )
+            self.contract_types[address] = combined_contract
+            return combined_contract
+
+        elif implementation_contract_type is not None:
+            contract_type_to_cache = implementation_contract_type
+            self.contract_types[address] = implementation_contract_type
+            return contract_type_to_cache
+
+        elif proxy_contract_type is not None:
```
fubuloubu, line 620: Add a note here, something like "cache and return proxy's contract_type even if we didn't find implementation"

src/ape/managers/_contractscache.py
```diff
@@ -606,6 +591,38 @@ def get(
...
+            self.contract_types[address] = proxy_contract_type
+            return proxy_contract_type
+
+        return None
```
fubuloubu, line 624: Maybe consider a WARNING log here, something like "couldn't determine any contract_type for proxy {address}"

src/ape/managers/_contractscache.py
```diff
@@ -561,41 +561,26 @@ def get(
...
+        # Contract is not cached yet. Check broader sources, such as an explorer.
+        if not proxy_info and detect_proxy:
+            # Proxy info not provided. Attempt to detect.
+            if not (proxy_info := self.proxy_infos[address_key]):
+                if proxy_info := self.provider.network.ecosystem.get_proxy_info(address_key):
+                    self.proxy_infos[address_key] = proxy_info
+
+        if proxy_info:
+            # Contract is a proxy (either was detected or provided).
+            if impl_contract := self._get_proxy_contract_type(
```
fubuloubu, line 573: This can be one of 4 things:
1. combined proxy + impl abi
2. impl-only abi (for things like simple forwarder proxies)
3. proxy-only abi (for things like upgradable proxies that don't have impls, or unverified impl)
4. undeterminable contract_type (eitehr impl or proxy)

I think should add a note about this, and consider calling this variable like `proxy_contract_type` instead (maybe?)

fubuloubu (dismissed):

src/ape/managers/_contractscache.py
```diff
@@ -561,41 +561,26 @@ def get(
```
antazoey, line 561: @fubuloubu So I have been looking at this, and on lines 558 - 560, the logic is kinda weird and I am wondering your thoughts on changing it. Right now, if a contract type exists and you give it a default, it overrides the contract type and uses the default. IMO that is a strange way to update a contract type; there are better ways.

Here is my thought: What if instead, we combine the ABIs in this case too... You give it a default, there is a cached contract-type, the default is different than the cached, so combine them.

Also, I'd be down for removing that part, but I fear it would be a breaking change. By combing the ABIs it avoids being as break-y

src/ape/managers/_contractscache.py
antazoey: do you think this warrents a warning or a debug log?

src/ape/managers/_contractscache.py
fubuloubu: I think merging them might work, worst case they can just nuke the file from the cache folder.

A practical example is my `ape-tokens` plugin, which loads everything from the tokenlist as a generic `ERC20` contract type. I think right now it does not set `fetch_from_explorer=False` so it actually does take advantage of this behavior that loads from the explorer the more "full" type, but I could imagine where that kwarg was added to the section of the `ape-tokens` plugin so that it always loads _at least_ an ERC20 (with or without an explorer), *if* a cached type doesn't exist from the user previously calling `Contract` directly with `fetch_from_explorer=True`, where the type was merged according to this suggested new behavior.

src/ape/managers/_contractscache.py
fubuloubu: Maybe info. I think the user needs to be made aware of it, but it's not really a problem (like the note says)

fubuloubu (dismissed):
This looks great!

fubuloubu (approved):

PR #2532, opened 2025-02-26
feat: 2.5x faster docker build with uv

banteg:
## Summary
- Add a new Dockerfile.uv using the faster uv package manager
- Improve .dockerignore to significantly reduce build context size 
- Optimize the Docker build process for improved performance

## Performance Analysis

This PR introduces a Docker build optimization using the uv package manager and an improved .dockerignore file. The changes are presented for the consideration of Ape maintainers and are not yet integrated into any CI or build pipelines.

### Build Times Comparison
- **Standard Docker build (slim)**: 111.74 seconds
- **UV Docker build**: 44.86 seconds
- **Speed Improvement**: 2.49x faster with UV

### Detailed Step Analysis

1. **Build Context Transfer**:
   - Standard: `transferring context: 2.34GB 11.8s`
   - UV: `transferring context: 2.85MB 0.0s`
   - **Key Finding**: The .dockerignore file reduced the build context from 2.34GB to just 2.85MB (822x smaller!)

2. **File Copying Times**:
   - Standard: `COPY . . - DONE 13.3s`
   - UV: `COPY . . - DONE 0.1s`
   - **Key Finding**: Copying files is 133x faster with the optimized .dockerignore

3. **Package Build/Install**:
   - Standard: `RUN pip wheel . - DONE 50.1s`
   - UV: `RUN uv build --wheel . - DONE 4.5s`
   - **Key Finding**: UV's package building is ~11x faster than pip

4. **Image Size**:
   - Standard: 525MB
   - UV: 698MB
   - **Note**: The UV image is slightly larger due to additional development dependencies

### Key Performance Factors

1. **Build Context Optimization**:
   - The .dockerignore file provides enormous benefits by excluding the .git directory (1.5GB) and other unnecessary files.
   - This dramatically reduces the time spent transferring files to the Docker daemon.

2. **UV Performance**:
   - UV's package building is significantly faster than pip, reducing the wheel creation time by ~45 seconds.
   - UV's dependency resolution and installation is more efficient.

3. **Build Process Structure**:
   - The simplified build process in the UV Dockerfile has fewer steps and more efficient resource usage.

## Test plan
- The Docker build completes successfully
- The resulting image functions correctly
- Build performance is significantly improved

🤖 Generated with Claude Code

### DIFF
--- a/Dockerfile.uv
+++ b/Dockerfile.uv
@@ -1,0 +1,60 @@
+#---------------------------------------------------------------------------------------------
+# See LICENSE in the project root for license information.
+#---------------------------------------------------------------------------------------------
+
+# Use the official uv Docker image with Python 3.11
+FROM ghcr.io/astral-sh/uv:python3.11-bookworm-slim AS builder
+
+# Set working directory
+WORKDIR /build
+
+# Copy project files
+COPY . .
+
+# Build wheels using uv
+RUN uv build --wheel .
+
+# Create the final image
+FROM ghcr.io/astral-sh/uv:python3.11-bookworm-slim
+
+# See http://label-schema.org for metadata schema
+LABEL org.opencontainers.image.title="ape" \
+    org.opencontainers.image.description="Ape Framework" \
+    org.opencontainers.image.url="https://apeworx.io/framework" \
+    org.opencontainers.image.documentation="https://docs.apeworx.io/ape/stable/userguides/quickstart.html\#installation" \
+    org.opencontainers.image.source="https://github.com/ApeWorX/ape" \
+    org.opencontainers.image.vendor="ApeWorX" \
+    org.opencontainers.image.licenses="Apache-2.0" \
+    org.opencontainers.image.authors="ApeWorX"
+    
+# Install build dependencies needed for installing packages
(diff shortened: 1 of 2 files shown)


Dockerfile.uv
```diff
@@ -0,0 +1,60 @@
...
+RUN useradd --create-home --shell /bin/bash harambe
+
+# Create directory for wheels
+RUN mkdir -p /wheels
+
+# Copy wheels from builder stage
+COPY --from=builder /build/dist/*.whl /wheels/
+
+# Install the wheels with uv (using --system since we're in a Docker container)
+RUN uv pip install --system /wheels/*.whl
```
fubuloubu, line 47: Can it not do `--system`? I am not sure of the security implications of this, if it overrides any system installations that could expose an attack surface inside the container

banteg:
noticed this inside the container, caused by `setuptools-scm`

    - slim: eth-ape 0.8.29.dev0+g4d2d1fd54.d20250226
    - uv: eth-ape 0.8.999

Dockerfile.uv
banteg: i think that's how you are supposed to install things in docker

PR #2540, opened 2025-03-06
fix: issue where `contract_type` would get corrupted in cache during trace enrichment

### LINKED ISSUE #2539: Call trace enrichment method pollutes `contract_type` JSON files for unrelated contracts when `tx.show_trace()` includes an ERC-20 token
### Have you ensured that all of these are up to date?

- [x] Ape
- [ ] Ape plugins

### What version of Ape are you on?

0.8.29

### What command(s) is the bug in?

ape test

### Operating System

Linux

### Ape Configuration

n/a

### Description

I am developing an arbitrage contract that swaps between Uniswap V2 and V4. While debugging various failures, I wrote an Ape test that called the main execution method with the argument `raise_on_revert=False` followed by a check for failure, and a `tx.trace()` if the transaction reverted. 

After some time the tests began to stall, eventually failing to complete within an hour, and the last run killed by the out-of-memory watchdog on my machine.

The `contract_types` directory was quite large, roughly 3 GB. I cleared that directory and ran the test again, finding that the trace was fast. Repeated runs of the test revealed that certain contract_type files were growing in size:

```
btd@dev:~/.ape/ethereum/mainnet/contract_types$ du -sh *
92K    0x000000000004444c5dc75cB358380D2e3dE08A90.json
56K    0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599.json
68K    0xBb2b8038a1640196FbE3e38816F3e67Cba72D940.json

btd@dev:~/.ape/ethereum/mainnet/contract_types$ du -sh *
1.0M    0x000000000004444c5dc75cB358380D2e3dE08A90.json
600K    0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599.json
664K    0xBb2b8038a1640196FbE3e38816F3e67Cba72D940.json

btd@dev:~/.ape/ethereum/mainnet/contract_types$ du -sh *
12M    0x000000000004444c5dc75cB358380D2e3dE08A90.json
6.8M    0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599.json
7.4M    0xBb2b8038a1640196FbE3e38816F3e67Cba72D940.json
```

I suspected some corruption of the contract type cache, and ultimately discovered that the `_enrich_contract_id` method in ecosystem.py causes the JSON to be overwritten with invalid data:

```
if kwargs.get("use_symbol_for_tokens") and "symbol" in contract_type.view_methods:
    # Use token symbol as name
    contract = self.chain_manager.contracts.instance_at(
        address, contract_type=contract_type
    )
```

The `contract_type` object gets corrupted because the `_get_contract_type_for_enrichment` function prioritizes the contract type in `kwargs` instead of the cached contract_type held by the `chain_manager.contracts` container.

For long call traces involving token transfers early in the transaction, the token `contract_type` is placed in `kwargs` and corrupts all lookups after. 

In my case, the WETH9 contract was placed into kwargs in the first loop, and repeatedly found on subsequent enrichment runs by `_get_contract_type_for_enrichment`. When this method finds the 'contract_type' key set in `kwargs`, it skips the fetch from the `chain_manager.contracts` container. There is no address or name verification performed by this method, so the mismatch is not detected.

After this occured, WETH9 persisted in `kwargs`, and was returned as the `contract_type` for the enriched traces for V2 pool, V4 PoolManager, and my arbitrage contract. The `"symbol" in contract_type.view_methods` check triggered because 'symbol' is a standard ERC-20 method. This led to the WETH9 contract type being saved back to the cache in the call to `self.chain_manager.contracts.instance_at()`, since the `contract_type=` argument is provided.

I have naively resolved the issue by popping both 'contract_address' and 'contract_type' from `kwargs` at the top of `_enrich_contract_id` so subsequent lookups aren't corrupted.

I'm ignorant about other effects this might have, so I'm filing the issue now for discussion. 

Draft PR to follow.

### Reproduction code

_No response_

BowTiedDevil:
Fixes #2539


bitwise-constructs (dismissed):
Clean fix, thank you

antazoey:
I'd like a test for this before calling it good, but I am happy to write such test.

src/ape_ethereum/ecosystem.py
```diff
@@ -1200,11 +1204,13 @@ def _enrich_contract_id(self, address: AddressType, **kwargs) -> str:
 
         kwargs["contract_type"] = contract_type
         if kwargs.get("use_symbol_for_tokens") and "symbol" in contract_type.view_methods:
-            # Use token symbol as name
-            contract = self.chain_manager.contracts.instance_at(
-                address, contract_type=contract_type
-            )
+            # Store contract_type in cache only if missing
+            if (contract := self.chain_manager.contracts.instance_at(address)) is None:
```
antazoey, line 1208: I don't think `instance_at()` returns `None`, you probably want `.get_contract_type()` here.
```suggestion
            if (contract_type := self.chain_manager.contracts.get_contract_type(address)) is None:
```

BowTiedDevil:
> I'd like a test for this before calling it good, but I am happy to write such test.

If you have a test in mind, please do! I am traveling today but can get back to the workstation to pick up this (and other) changes on Monday morning.

BowTiedDevil:
Reopening with new commit incorporating fix discussed in https://github.com/ApeWorX/ape/pull/2540#issuecomment-2708499027

antazoey:
> If you have a test in mind, please do! I am traveling today but can get back to the workstation to pick up this (and other) changes on Monday morning.

I opened up a PR against your main branch adding the test: https://github.com/BowTiedDevil/ape/pull/1

antazoey (approved):

PR #2543, opened 2025-03-08
fix: unable to connect to adhoc local network when using IPC

antazoey:
### What I did

In one terminal do:

```
geth --dev
```

**Notice the IPC path it spits out.**

In another do,

```
ape console --network <path/to/gethdev/ipc>
```

It was failing before, thinking the network was hardcoded. But now, it should realize it isn't.

### Checklist

### DIFF
--- a/src/ape_ethereum/provider.py
+++ b/src/ape_ethereum/provider.py
@@ -594,7 +594,9 @@ class Web3Provider(ProviderAPI, ABC):
     @cached_property
     def chain_id(self) -> int:
         default_chain_id = None
-        if (not self.network.is_adhoc and self.network.is_custom) or not self.network.is_dev:
+        if (not self.network.is_adhoc and self.network.is_custom) or (
+            not self.network.is_dev and not self.network.is_adhoc
+        ):
             # If using a live network, the chain ID is hardcoded.
             default_chain_id = self.network.chain_id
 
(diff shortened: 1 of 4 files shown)


src/ape_ethereum/provider.py
```diff
@@ -592,7 +592,9 @@ def estimate_gas_cost(self, txn: TransactionAPI, block_id: Optional["BlockID"] =
     @cached_property
     def chain_id(self) -> int:
         default_chain_id = None
-        if (not self.network.is_adhoc and self.network.is_custom) or not self.network.is_dev:
+        if (not self.network.is_adhoc and self.network.is_custom) or (
+            not self.network.is_dev and not self.network.is_adhoc
+        ):
             # If using a live network, the chain ID is hardcoded.
             default_chain_id = self.network.chain_id
```
fubuloubu, line 599: I wanted to break these two conditions up into 2 so it reads easier. This is okay because `default_chain_id` is set to `None` by default from not matching any condition

```suggestion
        if not self.network.is_adhoc and self.network.is_custom:
            # If using a configured, custom network, the chain ID is hardcoded.
            default_chain_id = self.network.chain_id

        elif not self.network.is_dev and not self.network.is_adhoc:
            # If using a configured public network, the chain ID is hardcoded.
            default_chain_id = self.network.chain_id
```

fubuloubu (approved):

Issue #2528, opened 2025-02-26
Add `--background` flag to `ape networks run`


antazoey:
### Describe

I take inspiration from ICP's `dfx` tool which has a `dfx start --background` and a `dfx stop` command for easily starting and stopping the dev node.

### Specification

when given background, background the process.
still track that is running so we can also add a

```
ape networks stop --network <choice>
```

command

PR #2545, opened 2025-03-10
feat: local network process MGMT CLI

antazoey:
### What I did

Allows flows like this:

```
ape networks run --network ::node --background
ape networks list --running
ape networks ping --network ::node
ape networks kill --all
```

basically Ape now tracks the processes it has started and you can easily kill them all from the command line as well more easily background the processes in the same terminal session and w/o losing the PID

fixes: #2528 

Also allows 

```python
with networks.parse_network_choice("pid://12345") as running_node":
    ...
```

For connecting to local nodes processes via PID.
**Note**: this process must have been started by Ape for this to work currently though.

### How I did it

copied this from the `dfx` tool: https://github.com/dfinity/sdk

### How to verify it

tbd

### Checklist

### DIFF
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -309,10 +309,6 @@ class NetworkManager(BaseManager, ExtraAttributesMixin):
         if not self.running_nodes:
             return {}
 
-        if not process_ids:
-            # Defaults to all managed processes.
-            process_ids = self.running_nodes.process_ids
-
         pids_killed = {}
         for pid in process_ids:
             if not (data := self.running_nodes.nodes.get(pid)):
--- a/src/ape/managers/networks.py
+++ b/src/ape/managers/networks.py
@@ -1,25 +1,156 @@
+import os
+import signal
 from collections.abc import Collection, Iterator
 from functools import cached_property
-from typing import TYPE_CHECKING, Optional, Union
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Optional, Union
 
 from evmchains import PUBLIC_CHAIN_META
+from pydantic import ValidationError
 
 from ape.api.networks import ProviderContextManager
 from ape.exceptions import EcosystemNotFoundError, NetworkError, NetworkNotFoundError
 from ape.managers.base import BaseManager
 from ape.utils.basemodel import (
+    BaseModel,
+    DiskCacheableModel,
     ExtraAttributesMixin,
     ExtraModelAttributes,
     get_attribute_with_extras,
     only_raise_attribute_error,
 )
 from ape.utils.misc import _dict_overlay, log_instead_of_fail
+from ape.utils.os import clean_path
 from ape_ethereum.provider import EthereumNodeProvider
 
 if TYPE_CHECKING:
     from ape.api.networks import EcosystemAPI, NetworkAPI
-    from ape.api.providers import ProviderAPI
+    from ape.api.providers import ProviderAPI, SubprocessProvider
     from ape.utils.rpc import RPCHeaders
+
+
+class NodeProcessData(BaseModel):
+    """
+    Cached data for node subprocesses managed by Ape.
+    """
+
+    network_choice: str
+    """
+    The network triple ``ecosystem:network:node``.
+    """
+
+    ipc_path: Optional[Path] = None
+    """
+    The IPC path this node process communicates on.
+    """
+
+    http_uri: Optional[str] = None
+    """
+    The HTTP URI this node process exposes.
+    """
+
+    ws_uri: Optional[str] = None
+    """
+    The websockets URI this node process exposes.
+    """
+
+    @log_instead_of_fail(default="<NodeProcessData>")
+    def __repr__(self) -> str:
+        if ipc := self.ipc_path:
+            return f"{self.network_choice} - {clean_path(ipc)}"
+
+        elif uri := (self.http_uri or self.ws_uri):
+            return f"{self.network_choice} - {uri}"
+
+        return self.network_choice
+
+    def matches_provider(self, provider: "SubprocessProvider") -> bool:
+        if self.network_choice != f"{provider.network.choice}:{provider.name}":
+            return False
+
+        # Skip if any of the connection paths (IPC, HTTP, WS) differ
+        for attr in ("ipc_path", "http_uri", "ws_uri"):
+            if (
+                getattr(provider, attr)
+                and getattr(self, attr)
+                and getattr(provider, attr) != getattr(self, attr)
+            ):
+                return False
+
+        return True
+
+
+class NodeProcessMap(DiskCacheableModel):
+    """
+    All managed running network subprocesses.
+    """
+
+    nodes: dict[int, NodeProcessData] = {}
+
+    @property
+    def path(self) -> Path:
+        return self._path
+
+    @property
+    def process_ids(self) -> tuple[int, ...]:
+        return tuple(self.nodes.keys())
+
+    def __bool__(self) -> bool:
+        """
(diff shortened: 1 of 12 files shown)


fubuloubu (approved):
Hmm, would this better unlock subprocess testing in CI? At least for the integration tests

antazoey:
> Hmm, would this better unlock subprocess testing in CI? At least for the integration tests

Probably, as it gives you more control to start up any networks you need in the beginning.. Also, you can more easily connect to providers via PID, which I want to think of a better way to incorporate that in `parse_network_choice()` logic, but in any case it would probably be good for silverback back testing so you can always connect to the same running fork node rather than starting up new ones constantly... you can still reset the fork to get a clean state.

fubuloubu:
>  you can still reset the fork to get a clean state.

Why not have this be the default way that `ProviderAPI.disconnect` works with those types of forks? And then just add `atexit` to fully shut it down on session end (if not asked to run in background)

antazoey:
But I agree - when using `.fork()`, we could incorporate this a bit... I think that is the intended use-case of `.fork()` so it makes sense

fubuloubu:
> But I agree - when using `.fork()`, we could incorporate this a bit... I think that is the intended use-case of `.fork()` so it makes sense
> 
> Also, the atexit handler is the what does call the disconnect, not exiting the context.

Would certainly be a lot faster testing this way vs. how I am doing backtests now w/ anvil in https://github.com/ApeWorX/silverback/pull/174

antazoey:
> Currently takes 14 mins to run 2 mins of history, so that needs to come way down

If I was you, I would already be managing the process and just connecting to it and calling `reset_fork()`.
You can connect to it normally and call 1 method each time at the beginning of the simulation...
`.fork()` is the only method that completely tears down the forked network, regularly connecting won't do that.

antazoey:
> Would certainly be a lot faster testing this way vs. how I am doing backtests now w/ anvil in [ApeWorX/silverback#174](https://github.com/ApeWorX/silverback/pull/174)

In any case, this is not super related to this PR so shouldn't be blocking

antazoey:
Testing demo:

I made a config like this:

```yaml
foundry:
  ethereum:
    local:
      uri: http://127.0.0.1:5001

node:
  ethereum:
    local:
      uri: http://127.0.0.1:5000
```


And then these are some commands w/ output:

```
(apedev13) ➜  playground ape networks list --running 
Local node(s) not running.
(apedev13) ➜  playground ape networks run --network ::node --background
Making request HTTP. URI: http://127.0.0.1:5000 Method: web3_clientVersion
Starting geth (HTTP=127.0.0.1:5000, IPC=$HOME/.ape/node/geth.ipc).
(apedev13) ➜  playground ape networks run --network ::foundry --background
Starting 'anvil' process.
(apedev13) ➜  playground ape console --network ::node
(apedev13) ➜  playground ape networks list --running
PID    NETWORK                 IPC                       HTTP                   WS                 
33542  ethereum:local:node     $HOME/.ape/node/geth.ipc  http://127.0.0.1:5000  None               
33564  ethereum:local:foundry  None                      http://127.0.0.1:8545  ws://127.0.0.1:8545
(apedev13) ➜  playground ape console --network ::node
INFO:     Connecting to existing Geth node at $HOME/.ape/node/geth.ipc.

In [1]: exit
(apedev13) ➜  playground ape networks list --running 
PID    NETWORK                 IPC                       HTTP                   WS                 
33542  ethereum:local:node     $HOME/.ape/node/geth.ipc  http://127.0.0.1:5000  None               
33564  ethereum:local:foundry  None                      http://127.0.0.1:8545  ws://127.0.0.1:8545
(apedev13) ➜  playground ape console --network ::foundry

In [1]: exit
(apedev13) ➜  playground ape networks list --running    
PID    NETWORK                 IPC                       HTTP                   WS                 
33542  ethereum:local:node     $HOME/.ape/node/geth.ipc  http://127.0.0.1:5000  None               
33564  ethereum:local:foundry  None                      http://127.0.0.1:8545  ws://127.0.0.1:8545
(apedev13) ➜  playground ape networks kill                                
Stopped the following node(s):
        ethereum:local:node - $HOME/.ape/node/geth.ipc
        ethereum:local:foundry - http://127.0.0.1:8545
(apedev13) ➜  playground ape networks list --running
Local node(s) not running.
(apedev13) ➜  playground 
```

src/ape_networks/_cli.py
```diff
@@ -163,3 +203,44 @@ def _run(cli_ctx, provider: "SubprocessProvider"):
     else:
         provider.disconnect()
         cli_ctx.abort("Process already running.")
+
+
+@cli.command(short_help="Stop node processes")
+@ape_cli_context()
+@click.option("--pid", "process_ids", help="The PID of the process(es) to kill", multiple=True)
+@network_option(default=None)
+def kill(cli_ctx, process_ids, provider):
```
fubuloubu, line 212: I don't see where `provider` gets used (I am supposing that specifying the network option will only kill that one specific provider)

src/ape_networks/_cli.py
```diff
@@ -163,3 +203,44 @@ def _run(cli_ctx, provider: "SubprocessProvider"):
...
+    """
+    Stop node processes
+    """
+    if not cli_ctx.network_manager.running_nodes:
+        echo_rich_text("No running nodes found.")
+
+    elif processes_killed := cli_ctx.network_manager.kill_node_process(*process_ids):
```
fubuloubu, line 219: According to your demo output, if you don't provide any process_ids, it kills all of them? But if you provide 1 then it only kills 1? I think that might not be expected behavior from a user perspective

src/ape/managers/networks.py
```diff
@@ -97,6 +237,106 @@ def ecosystem(self) -> "EcosystemAPI":
...
+        Returns:
+            dict[str, :class:`~ape.managers.networks.NodeProcessData`]: The process data
+            of all terminated processes.
+        """
+        if not self.running_nodes:
+            return {}
+
+        if not process_ids:
+            # Defaults to all managed processes.
+            process_ids = self.running_nodes.process_ids
```
fubuloubu, line 314: This seems a bit magical, wonder if a separate method `kill_all_node_processes` that does this would be more clear

fubuloubu (approved):
Looks pretty good, just have some notes about killing process interface

src/ape_networks/_cli.py
antazoey: OK, I could add a `--all` flag. It kinda seems like 99% of the time, you are trying to kill them all but I could be wrong.

src/ape_networks/_cli.py
antazoey: [e5b468c](https://github.com/ApeWorX/ape/pull/2545/commits/e5b468c59e4b7ab4dd83d63e2d6335bc11bce660)

src/ape_networks/_cli.py
antazoey: [7447e6d](https://github.com/ApeWorX/ape/pull/2545/commits/7447e6dfacbb9e784314022cae1cde7263d478a0)

src/ape/managers/networks.py
antazoey: [7447e6d](https://github.com/ApeWorX/ape/pull/2545/commits/7447e6dfacbb9e784314022cae1cde7263d478a0)

antazoey:
> antazoey

Feedback addressed! thank you

antazoey:
So now you do `--all` to kill them all, else you have to give it one or more `--pid` option

bitwise-constructs (dismissed):
super helpful, nice add

fubuloubu (dismissed):

Issue #2476, opened 2025-01-22
Add `__hash__` to `BaseAddress`


fubuloubu:
### Describe

Adding `__hash__` implementation to `BaseAddress` allows all derivatives (e.g. `ContractInstance`, `AddressAPI`, etc.) to be used as keys in `dict`s and also added to `set`s, simplifying downstream logic

### Specification

```py
...
from ape.utils import to_int
...

class BaseAddress(BaseInterface):
    ...
    def __hash__(self) -> int:
        return to_int(hexstr=self.address)
    ...
```

see: https://github.com/ApeWorX/ape/blob/main/src/ape/api/address.py#L19

### Dependencies

n/a

Oshioke-Salaki:
Gm can i handle this?

fubuloubu:
> Gm can i handle this?

Absolutely! easy code update, just needs a test to check it works with `dict` and `set` for the various types of Ape objects that users will use (`ContractInstance`, `AccountAPI`, and `Address`)

3shv:
Hi, Please take a look at my draft PR and let me know if the direction is okay. 
One thing I found is that these two address are not equal

```
    str_address = TestAddress("0x1234567890123456789012345678901234567890")
    int_address = TestAddress(int("1234567890123456789012345678901234567890", 16))
    assert str_address == int_address
```

PR #2550, opened 2025-03-14
refactor: add hash protocol to BaseAddress

3shv:
### What I did

Add hash function to BaseAddress class

Fixes: #2476

### How to verify it

Add new tests, checking for membership in dict and set

### Checklist

### DIFF
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -90,6 +90,15 @@ class BaseAddress(BaseInterface):
         """
         return self.address
 
+    def __hash__(self) -> int:
+        """Return consistent hash for all address representations with the same value."""
+        value = self.address
+
+        try:
+            return int(self.conversion_manager.convert(value, AddressType), 16)
+        except Exception:
+            raise TypeError(f"Cannot hash address of type: {type(value)}")
+
     def __call__(self, **kwargs) -> "ReceiptAPI":
         """
         Call this address directly. For contracts, this may mean invoking their
--- a/src/ape/api/address.py
+++ b/src/ape/api/address.py
@@ -92,12 +92,7 @@ class BaseAddress(BaseInterface):
 
     def __hash__(self) -> int:
         """Return consistent hash for all address representations with the same value."""
-        value = self.address
-
-        try:
-            return int(self.conversion_manager.convert(value, AddressType), 16)
-        except Exception:
-            raise TypeError(f"Cannot hash address of type: {type(value)}")
+        return hash(int(self.address, base=16))
 
     def __call__(self, **kwargs) -> "ReceiptAPI":
         """
--- a/tests/functional/test_hash_address.py
+++ b/tests/functional/test_hash_address.py
@@ -1,109 +1,20 @@
-import pytest
-
-from ape.api.address import BaseAddress
-from ape.types.address import AddressType
-
-
-class TestAddress(BaseAddress):
-    """Test implementation of BaseAddress with a configurable address value"""
-
-    def __init__(self, address_value):
-        self._address_value = address_value
-        super().__init__()
-
-    @property
-    def address(self) -> AddressType:
-        return self._address_value
+def test_hash_same_as_python_hash(accounts):
+    for account in accounts:
+        assert account.__hash__() == hash(account)
 
 
-@pytest.mark.parametrize(
-    "address_input,expected_hash",
-    [
-        # String with 0x prefix
-        (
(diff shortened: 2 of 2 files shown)


src/ape/api/address.py
```diff
@@ -90,6 +90,15 @@ def __str__(self) -> str:
         """
         return self.address
 
+    def __hash__(self) -> int:
+        """Return consistent hash for all address representations with the same value."""
+        value = self.address
```
fubuloubu, line 95: `self.address` is already defined as `AddressType`, so don't think we need the extra conversion

tests/functional/test_hash_address.py
```diff
@@ -0,0 +1,109 @@
+import pytest
+
+from ape.api.address import BaseAddress
+from ape.types.address import AddressType
+
+
+class TestAddress(BaseAddress):
```
fubuloubu, line 7: We have some built in account objects we can probably test with instead (see `accounts` fixture)

tests/functional/test_hash_address.py
```diff
@@ -0,0 +1,109 @@
...
+
+    expected = int.from_bytes(address_bytes, byteorder="big")
+    assert address.__hash__() == expected
+
+
+def test_hash_with_unsupported_type():
+    """Test that TypeError is raised for unsupported address types"""
+    address = TestAddress(None)
+    with pytest.raises(TypeError, match="Cannot hash address of type"):
+        address.__hash__()
```
fubuloubu, line 81: What is the reasoning behind these test cases?

tests/functional/test_hash_address.py
```diff
@@ -0,0 +1,109 @@
...
+
+def test_hash_is_consistent_with_equality():
+    """Test that hash is consistent with equality for BaseAddress objects"""
+    addr1 = TestAddress("0xabcdef1234567890abcdef1234567890abcdef12")
+    addr2 = TestAddress("0xabcdef1234567890abcdef1234567890abcdef12")
+
+    assert addr1 == addr2
+    assert addr1.__hash__() == addr2.__hash__()
+    assert hash(addr1) == hash(addr2)
+    assert addr1.__hash__() != hash(addr1)
```
fubuloubu, line 105: Wait, these don't return the same?

fubuloubu (commented):
Thank you for taking a crack at this! Had a few questions/comments but it is looking good

tests/functional/test_hash_address.py
antazoey: A class in pytest land is different

antazoey: Oh, it is not actually being used this way... Ya, I agree with @fubuloubu then.

tests/functional/test_hash_address.py
3shv: Yeah. Was surprised too. Later found out [this](https://docs.python.org/3/library/functions.html#hash) 
Now I changed the implementation to make them equal

tests/functional/test_hash_address.py
3shv: I was thinking `AddressType` can be anything such as str, int, bytes, HexStr (Confused with `RawAddress`). 
Later found out that its always HexStr. So now I removed them and simplified the test cases

src/ape/api/address.py
3shv: Thanks. Fixed it!

3shv:
Thanks for your review. I updated as per your suggestions. Please let me know if further changes are required.

tests/functional/test_hash_address.py
fubuloubu: Awesome, thank you!

tests/functional/test_hash_address.py
fubuloubu: man, that is some dev alpha. thanks for sharing!

tests/functional/test_hash_address.py
```diff
@@ -0,0 +1,20 @@
+def test_hash_same_as_python_hash(accounts):
+    for account in accounts:
+        assert account.__hash__() == hash(account)
```
fubuloubu, line 3:
```suggestion
        # NOTE: According to CPython spec, `hash` is truncated according to OS bit width
        #       https://docs.python.org/3/library/functions.html#hash
        assert account.__hash__() == hash(account)
```

fubuloubu (approved):
This is perfect! Thank you!

Issue #2350, opened 2024-10-28
`ape plugins install` doesn't work if ape installed via `uv tool`


fubuloubu:
### Have you ensured that all of these are up to date?

- [X] Ape
- [X] Ape plugins

### What version of Ape are you on?

v0.8.18

### What command(s) is the bug in?

ape plugins install

### Operating System

Linux

### Description

```yaml
plugins:
- name: aws
  veresion: 0.8.0a3
```

```sh
$ uv tool install eth-ape[recommended-plugins]
...
$ ape plugins install .
INFO:     Installing 'aws' plugin ...
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
ERROR:    Failed to install plugin 'aws.
```

bitwise-constructs:
is this intentionally being done with `uv tool install` over `uv pip install`? tool installs globally on the path, so if you run an `ape plugins install .` on a tool version of ape then it will collide with whatever else is there

bitwise-constructs:
but tested with `uv pip install` and it behaves as expected

fubuloubu:
> is this intentionally being done with `uv tool install` over `uv pip install`? tool installs globally on the path, so if you run an `ape plugins install .` on a tool version of ape then it will collide with whatever else is there

Specifically with tool, because I feel like downstream users might find it interesting to use in that way

I think we could do some detection it was installed with tool and mutate the install so that the plugin actually can be added to the install other tool

3shv:
I installed eth-ape using pipx and for me `ape plugins install arbitrum` doesnt work either. 
It works if the virtualenv is inferred by uv.
While debugging, I found the command that's being run is `uv pip install 'ape-arbitrum>=0.8,<0.9'`. We can reproduce this issue by running this command on non-virtualenv locations.

[Raised a draft PR](https://github.com/ApeWorX/ape/pull/2553) to allow users to specify virtual env location. Let me know your thoughts.

There are other approaches such as automatically inferring where eth-ape venv is located, but it felt complicated.

PR #2553, opened 2025-03-16
feat: Add python loc option for installing plugins

3shv:
### What I did

Related to issue #2350 

### How I did it

I installed eth-ape using pipx and noticed that `ape plugins install arbitrum` fails since it can't find the virtual environment
Added an optional python venv location so that user can specify the location incase the install fails.

### Checklist


antazoey:
Nice, this approach makes sense.

3shv:
On digging further, I can see the virtualenv location on the first line in the installed cli file (cat `which ape`). Maybe we can extract it out and use it directly. WDYT?

fubuloubu (approved):

antazoey (dismissed):

antazoey (approved):

PR #2557, opened 2025-03-20
fix(deps): upgrade packaging

fubuloubu:
### What I did

I guess `setuptools` upgraded there pin and it screwed us:
https://github.com/ApeWorX/ape/actions/runs/13958350355/job/39075553268

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
         "cchecksum>=0.0.3,<1",
         # Pandas peer-dep: Numpy 2.0 causes issues for some users.
         "numpy<2",
-        "packaging>=23.0,<24",
+        "packaging>=24.2,<25",
         "pandas>=2.2.2,<3",
         "pluggy>=1.3,<2",
         "pydantic>=2.10.0,<3",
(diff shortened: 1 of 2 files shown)


antazoey (approved):

antazoey (approved):

setup.py
```diff
@@ -94,7 +94,7 @@
         "cchecksum>=0.0.3,<1",
         # Pandas peer-dep: Numpy 2.0 causes issues for some users.
         "numpy<2",
-        "packaging>=23.0,<24",
+        "packaging>=24.2,<25",
```
antazoey, line 97: We have to fix in some downstream plugins before this will succeed: https://github.com/ApeWorX/ape-hardhat/blob/main/setup.py#L81

antazoey: Actually... This is a bit of a circular dependency. The container depends on both core Ape and plugins. Plugins depend on core ape. This why I think `recommened_plugins` type flows don't make much sense.

The only solution to merge with with a failing solution and then update the plugin and then the next update of core ape will have it all fixed.

antazoey:
@fubuloubu can we merge this? ill fix the plugins

fubuloubu:
> @fubuloubu can we merge this? ill fix the plugins

Yes, sounds more complex than I expected, thank you for taking this up!

PR #2560, opened 2025-03-30
refactor: improve error handling

mdqst:


src/ape_plugins/_cli.py
```diff
@@ -341,10 +341,10 @@ def _change_version(spec: str):
     ape_retcode = _install("eth-ape", spec)
     if ape_retcode == 0 and plugin_retcode == 0:
         prefix = "Ape"
-        if plugin_retcode == 0:
+        if plugin_retcode == 0 and get_plugin_dists():
```
fubuloubu, line 344: The other change is good, but what would the benefit of this change be given `plugin_retcode` getting set in the `get_plugin_dists` for loop

fubuloubu (approved):

Issue #2566, opened 2025-04-08
Contract ABI contains duplicated items


fubuloubu:
### Have you ensured that all of these are up to date?

- [x] Ape
- [x] Ape plugins

### What version of Ape are you on?

0.8.30

### Operating System

Linux

### Description

When running an ape script (a silverback bot really) in a containerized environment, I am getting a weird issue where the same contract event definition appears multiple times in a contract type

### Reproduction code

Ran this on the current container version

```sh
$ docker run -e ETHERSCAN_API_KEY=$ETHERSCAN_API_KEY -it --entrypoint sh ghcr.io/apeworx/silverback:stable
$ ape console --network :sepolia
INFO:     Connecting to existing Geth node at https://sepolia.gateway.tenderly.co.

In [1]: from apepay import StreamManager

In [2]: sm = StreamManager("0x7B25175142886e34543247f4583fe034eafDee32")

In [3]: sm.contract.StreamFunded.events
Out[3]: StreamFunded(uint256 indexed stream_id, address indexed funder, uint256 funded_amount, uint256 
time_left)
StreamFunded(uint256 indexed stream_id, address indexed funder, uint256 funded_amount, uint256 
time_left)
StreamFunded(uint256 indexed stream_id, address indexed funder, uint256 funded_amount, uint256 
time_left)
StreamFunded(uint256 indexed stream_id, address indexed funder, uint256 funded_amount, uint256 
time_left)
[, , , ]
```

fubuloubu:
Noticing that every time `ContractInstance._events_` is called, the number of duplicate events grows by 1 (the first time it is just `ContractEvent`, then becomes `ContractEventWrapper`)

fubuloubu:
Okay, seems to stem from the way that contract types are merged with the default, assuming there is a slight naming difference between their ABIs:
https://github.com/ApeWorX/ape/blob/main/src/ape/managers/_contractscache.py#L558
```py
(Pdb) !list(a == b for a, b in zip(default.abi, contract_type.abi))
[True, True, True, True, True, False, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, False, True, True]
(Pdb) default.abi[5]
[...]

PR #2568, opened 2025-04-08
fix(Contracts): only update contract cache with proper merge of ABI

fubuloubu:
### What I did

Noticed that when a contract type was "merged" (by combining ABIs of 2 different types) that it was duplicating the shared ABI between the two. This change instead makes sure that only those ABIs that were missing from the underlying contract type will get merged with the base, ensuring that problems with duplicate methods do not appear.

fixes: #2566

### How I did it

Check relevant ABIs against the set of existing ABIs of the base contract type before merging. Do no merge those that are shared between the existing base contract type and the type to merge it with

### How to verify it

see: #2566

### Checklist

~~- [ ] Documentation is complete~~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
         "web3[tester]>=6.20.1,<8",
         # ** Dependencies maintained by ApeWorX **
         "eip712>=0.2.10,<0.3",
-        "ethpm-types>=0.6.23,<0.7",
+        "ethpm-types>=0.6.25,<0.7",
         "eth_pydantic_types>=0.1.3,<0.2",
         "evmchains>=0.1.0,<0.2",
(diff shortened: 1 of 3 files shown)


fubuloubu:
Need to add a test case that shows a contract type w/ N methods has N+1 methods when 1 of them changes it's method selector

antazoey (approved):

antazoey:
Wow I remember having this problem once, I thought I fixed it.

fubuloubu:
> Wow I remember having this problem once, I thought I fixed it.

Yeah, it was pretty weird, and the reason why I think we ended up with `ContractEventWrapper` since I'm not sure it's possible to define multiple event definitions that have the same name

Issue #2564, opened 2025-04-07
FoundryProject does not detect via-ir config


antazoey:
### Have you ensured that all of these are up to date?

- [ ] Ape
- [ ] Ape plugins

### Operating System

macOS (Apple Silicon)

### Ape Configuration

n/a, only using foundry's config

### Description

Foundry has via_ir set to True but it is not picked up in Ape's foundry project detection logic it seems

shwetd19:
Hey @antazoey , I think the issue can be fixed by updating Ape’s Foundry project detection to check if the `via_ir` setting is set in the `foundry.toml` file, and then pass it to the compiler. I guess this might need a small change in the `ape_foundry` plugin so it can read and use `via_ir=True` while compiling.

Also, I came across this project on Gitcoin (here’s the link: https://explorer.gitcoin.co/#/round/42161/863/12) and I’ve been really excited to find ways to contribute! Could you please let me know if I can work on this issue, or suggest any similar beginner-friendly issues where I can start? Would love to get involved.

Thanks!

antazoey:
@shwetd19 Hello there! Yes, it should not be too complicated actually. Here is the method you have to edit:
https://github.com/ApeWorX/ape/blob/main/src/ape_pm/project.py#L253-L275

After that, it will just work as if it were configured in Ape directly. You shouldn't have to mess with the ape-foundry plugin here.

shwetd19:
Pls review my PR @antazoey , pls let me know if any change is required, also can you pls suggest any similar beginner-friendly issues where I can start? Would love to get involved.

PR #2572, opened 2025-04-15
Added `via_ir` Detection for Foundry Projects in Ape

shwetd19:
## Description

This PR fixes issue #2564  , where Ape’s `FoundryProject` class does not detect the `via_ir` setting from Foundry’s `foundry.toml` configuration file. 

The change updates the `_parse_solidity_config` method in `src/ape_pm/project.py` to read the `via_ir` key and include it in the Solidity compiler configuration as `viaIR`. This ensures Ape respects Foundry’s `via_ir` setting during compilation, aligning with Foundry’s behavior.

### Changes Made
- Modified `FoundryProject._parse_solidity_config` in `src/ape_pm/project.py` to check for `via_ir` in the parsed `foundry.toml` data.
- Added `via_ir` to the `sol_cfg` dictionary as `"viaIR": via_ir` when present, preserving existing logic for other settings (`version`, `evm_version`, `remappings`).
- While ensuring safe handling of missing `via_ir` keys using `data.get("via_ir")`.

---

### Testing

- Created a test Foundry project with:

  ```toml
  [profile.default]
  via_ir = true
  ```
  in `foundry.toml`.

- Ran `ape compile` to verify that the compiler applies the `viaIR` pipeline (confirmed via compiler output).
- Tested with `via_ir` absent to ensure no unintended side effects.
- [Optional] Added a unit test in `tests/test_foundry_project.py` to validate `via_ir` detection.

---

Thank you for considering this contribution! 🚀

### DIFF
--- a/src/ape_pm/project.py
+++ b/src/ape_pm/project.py
@@ -266,6 +266,9 @@ class FoundryProject(ProjectAPI):
         if evm_version := data.get("evm_version"):
             sol_cfg["evm_version"] = evm_version
 
+        if via_ir := data.get("via_ir"):
+            sol_cfg["viaIR"] = via_ir
+
         foundry_remappings = [*data.get("remappings", []), *self._parse_remappings_file()]
         if remappings := self._parse_remappings(
             foundry_remappings, dependencies, lib_paths, contracts_folder=contracts_folder
--- a/src/ape_pm/project.py
+++ b/src/ape_pm/project.py
@@ -267,7 +267,7 @@ class FoundryProject(ProjectAPI):
             sol_cfg["evm_version"] = evm_version
 
         if via_ir := data.get("via_ir"):
-            sol_cfg["viaIR"] = via_ir
+            sol_cfg["via_ir"] = via_ir
 
         foundry_remappings = [*data.get("remappings", []), *self._parse_remappings_file()]
         if remappings := self._parse_remappings(
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -59,6 +59,7 @@ out = 'out'
 libs = ['lib']
 solc = "0.8.18"
 evm_version = 'cancun'
+via_ir = true
 
 remappings = [
     'forge-std/=lib/forge-std/src/',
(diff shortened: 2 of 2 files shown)


src/ape_pm/project.py
```diff
@@ -266,6 +266,9 @@ def _parse_solidity_config(
         if evm_version := data.get("evm_version"):
             sol_cfg["evm_version"] = evm_version
 
+        if via_ir := data.get("via_ir"):
+            sol_cfg["viaIR"] = via_ir
```
antazoey, line 270: It is snake case: https://github.com/ApeWorX/ape-solidity/blob/main/ape_solidity/compiler.py#L142

```suggestion
            sol_cfg["via_ir"] = via_ir
```

antazoey (commented):
please also edit this test and the foundry_toml fixture to include the new value

shwetd19:
Hey @antazoey I did the change for 'via_ir' and updated it but I'm not able to do the change for 'foundry_toml' , pls help me out there

antazoey:
> Hey @antazoey I did the change for `via_ir` and updated it but I'm not able to do the change for `foundry_toml` , pls help me out there

Right now, the foundry toml fixture looks like this:

(from tests/functional/test_project.py)
```
@pytest.fixture(scope="module")
def foundry_toml():
    return """
[profile.default]
src = 'src'
out = 'out'
libs = ['lib']
solc = "0.8.18"
evm_version = 'cancun'

remappings = [
    'forge-std/=lib/forge-std/src/',
    '@openzeppelin/=lib/openzeppelin-contracts/',
]
""".lstrip()
```

it is basically just the string version of a foundry TOML file, for configuring foundry. Foundry-based projects will have these (similar to how Ape has an ape-config.yaml).

The project API layer in Ape converts 1 config to another for the sake of successfully using dependencies that happen to be foundry projects, as well as being able to use `ape` command directly in a foundry project. Much of this integration needs work, so I appreciate you fixing the `via_ir`.

Anyway, you need to update this fixture to have `via_ir = true`, so we can test that our integration sucessfully reads and parses the value thanks to your  change.

The `via_ir` field goes in the `[profile.default]` stanza, so you can plop it right under the `evm_version` part.

After this is done, head over to the `TestFoundryProject::test_extract_config` in the same module - add an assertion under the `evm_version` part like

```python
assert actual_sol["via_ir"] is True
```

fubuloubu (dismissed):

antazoey (approved):

PR #2606, opened 2025-05-02
test: fix out-of-gas problems with test contracts

antazoey:
### What I did

The logic, specifically in the vyper contract, was a bit problematic here. It seems to be less problematic doing two things:

1. not doing an expensive op during the deploy
2. using small dynamic sized arrays.

### Checklist


antazoey:
I am going to make sure this either fixes the flakey tests or will fix them as part of this
this is also pulls out something from my tests update branch, having the damnest time with that so going to have to hoist a bunch up from it that i need so it is not lost, starting with this

fubuloubu (approved):

PR #2608, opened 2025-05-05
feat: test accounts public key

antazoey:
i need to be able to easily access public keys in tests

* makes public_key more of an official api and gives test accounts access to their publi
* adds a couple missing arg docs
* fixes a casing from camelCase to snake_case
* replaces a production assert statement with an actual error
* other small stylistic improvements

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -73,7 +73,7 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         return None
 
     @property
-    def public_key(self) -> Optional[HexBytes]:
+    def public_key(self) -> Optional["HexBytes"]:
         """
         The public key for the account.
         """
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -72,6 +72,14 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         """
         return None
 
+    # TODO: In 0.9, make this an @abstactmetod
+    @property
+    @raises_not_implemented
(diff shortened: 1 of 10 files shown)


src/ape/api/accounts.py
```diff
@@ -72,6 +72,14 @@ def alias(self) -> Optional[str]:
         """
         return None
 
+    # TODO: In 0.9, make this an @abstactmetod
+    @property
+    @raises_not_implemented
```
fubuloubu, line 77: Might I suggest instead that this returns optional, and then None by default? so that account plugins can "opt-in" to sharing their public key (they might not have this property, e.g. Safe account or expose access to it e.g. Ledger)

src/ape/api/accounts.py
```diff
@@ -72,6 +72,14 @@ def alias(self) -> Optional[str]:
...
+    def public_key(self):
```
fubuloubu, line 78:
```suggestion
    def public_key(self) -> Optional[HexBytes]:
```

fubuloubu (approved):

src/ape/api/accounts.py
```diff
@@ -72,6 +72,13 @@ def alias(self) -> Optional[str]:
         """
         return None
 
+    @property
+    def public_key(self) -> Optional[HexBytes]:
+        """
+        The public key for the account.
+        """
```
fubuloubu, line 79:
```suggestion

        ```{notice}
        Account might not have this property if feature is unsupported or inaccessible.
        ```
        """
```

PR #2612, opened 2025-05-07
feat(ethereum): add support for EIP-7702

fubuloubu:
### What I did

Added support to decode EIP-7702 `SET_CODE` transactions (tx type 0x4)

### How I did it

shout out to Web3.py

### How to verify it

```python
$ ape console --network :mainnet:node
INFO:     Connecting to a 'mevblocker' node.

In [1]: tx = chain.get_receipt("0x00645e1eafe945e7a356ccd2be36dcef81eb3bb752035887487cf82b6cc13c28")

In [2]: tx.transaction.authorizations[0].authority
Out[2]: '0x864a5a451126D99647D9fE58b21e124AE6F67E35'
```

### Checklist

- [x] All changes are completed
- [x] Change is covered in tests
~~- [ ] Documentation is complete~~

PR #2613, opened 2025-05-07
feat(api): support signing authorizations

fubuloubu:
### What I did

This adds support for EIP7702-style "Authorizations" which are delegated account access for any account that supports protocol-level delegation

~~requires: #2612~~
~~requires: #2623~~
~~requires: https://github.com/ApeWorX/evm-trace/pull/78~~

### How I did it

Again, thanks to web3.py

### How to verify it

```python
 In[0]: acct = accounts.load(...)

 In[0]: sig = acct.sign_authorization(...)
```

### Checklist


src/ape/api/accounts.py
```diff
@@ -100,6 +103,36 @@ def sign_raw_msghash(self, msghash: "HexBytes") -> Optional[MessageSignature]:
...
+            The signature corresponding to the message.
+
+        ```{caution}
+        This action has the capability to be extremely destructive to the signer, and might lead to
+        full account compromise. All implementations are recommended to ensure that the signer be
+        made aware of the severity and impact of this action through some callout.
+        ```
+        """
+
+        raise NotImplementedError(
```
antazoey, line 132: use `APINotImplementedError` if it is common user facing expected error; it displays nicer.

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+            **txn_kwargs: Additional transaction kwargs passed to
+              :meth:`~ape.api.networks.EcosystemAPI.create_transaction`, such as ``gas``
+              ``max_fee``, or ``max_priority_fee``. For a list of available transaction
+              kwargs, see :class:`~ape.api.transactions.TransactionAPI`.
+        """
+        raise NotImplementedError
+
+    def remove_delegate(self, **txn_kwargs):
+        """
+        Has the account class remove the override for the value of it's `delegate`. For plugins
```
antazoey, line 511:
```suggestion
        Has the account class remove the override for the value of its `delegate`. For plugins
```

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+
+        Returns:
+            Optional[`:class:~ape.contracts.ContractInstance`]:
+                The contract instance of the delegate contract (if available).
+        """
+        return None
+
+    def set_delegate(self, contract: ContractInstance, **txn_kwargs):
+        """
+        Have the account class override the value of it's `delegate`. For plugins that support this
```
antazoey, line 488:
```suggestion
        Have the account class override the value of its `delegate`. For plugins that support this
```

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+        feature, the way they choose to handle it can vary. For example, it could be a call to up-
+        grade itself using some built-in method for a smart wallet (with default txn args) e.g. the
+        Safe smart wallet (https://github.com/ApeWorX/ape-safe), or it could be to use an EIP7702-
+        like feature available on the network to set a delegate for that account. However it a
```
antazoey, line 492:
```suggestion
        like feature available on the network to set a delegate for that account. However if a
```

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+        plugin chooses to handle it, the resulting action (if successful) should make sure that the
+        value that `self.delegate` returns is the same as `contract` after it is completed.
```
antazoey, line 494: you should use double tics in the docstring to do what you want normal tics to do.

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+              kwargs, see :class:`~ape.api.transactions.TransactionAPI`.
+        """
+        raise NotImplementedError
+
+    def remove_delegate(self, **txn_kwargs):
+        """
+        Has the account class remove the override for the value of it's `delegate`. For plugins
+        that support this feature, the way they choose to handle it can vary. For example, on a
+        network using an EIP7702-like feature available it will reset the delegate to empty.
+        However it a plugin chooses to handle it, the resulting action (if successful) should
```
antazoey, line 514:
```suggestion
        However, if a plugin chooses to handle it, the resulting action (if successful) should
```

src/ape/api/accounts.py
```diff
@@ -433,6 +466,109 @@ def get_deployment_address(self, nonce: Optional[int] = None) -> AddressType:
...
+    @contextmanager
+    def delegate_to(
+        self,
+        contract: ContractInstance,
+        set_txn_kwargs: Optional[dict] = None,
+        remove_txn_kwargs: Optional[dict] = None,
+        **txn_kwargs,
+    ) -> Iterator[ContractInstance]:
+        """
+        Temporarily overrides the value of `delegate` for the account inside of a context manager,
```
antazoey, line 537:
```suggestion
        Temporarily override the value of `delegate` for the account inside of a context manager,
```

we've been using the style of "What can I do with this method?", so verbs should be like `override` instead of `overrides`

setup.py
```diff
@@ -118,7 +118,7 @@
         "hexbytes>=0.3.1,<2",
         "py-geth>=5.4.0,<6",
         "trie>=3.0.1,<4",  # Peer: stricter pin needed for uv support.
-        "web3[tester]>=6.20.1,<8",
+        "web3[tester]>=7.11,<8",
```
antazoey, line 121: By doing this, we can get rid of the all the web3 compat imports we have to do. That may be worth doing on a separate PR

src/ape_test/accounts.py
```diff
@@ -126,6 +129,29 @@ def public_key(self) -> "HexBytes":
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}_{self.index} {self.address_str}>"
 
+    def sign_authorization(
+        self,
+        address: "AddressType",
+        chain_id: Optional[int] = None,
```
antazoey, line 135: so you can pass `0` here for all chains? **idea**: also support `"*"` at the top-level because it is clearer to mean "all chains" imo

src/ape/api/address.py
```diff
@@ -171,15 +170,33 @@ def codesize(self) -> int:
...
+    @property
+    def delegate(self) -> Optional["ContractInstance"]:
+        """
+        Check and see if Account has a "delegate" contract, which is a contract that this account
+        delegates functionality to. This could be from many contexts, such as a Smart Wallet like
+        Safe (https://github.com/ApeWorX/ape-safe) which has a Singleton class it forwards to, or
+        an EOA using a EIP7702-style delegate. Returning `None` means that the account does not
+        have a delegate avaiable, which should be overrided in plugins that support ecosystems
+        where such methods are possible natively (e.g. EIP7702 context), or plugins such as smart
+        wallets which have this functionality. The default implementation is to return `None`.
```
antazoey, line 193:
```suggestion
        wallets which have this functionality. The default implementation is to return ``None``.
```

fubuloubu: Actually have to rewrite this entire doc now that I am delegating to `ChainManager.get_delegate`

setup.py
fubuloubu: Actually I think I can get around this via `try..except` on `eth_account.Account.sign_authorization` below (which will raise `AttributeError` on un-supported web3py versions)

src/ape_test/accounts.py
fubuloubu: ac873ef7a

src/ape_accounts/_cli.py
```diff
@@ -222,3 +224,38 @@ def delete(cli_ctx, alias):
     account = cli_ctx.account_manager.load(alias)
     account.delete()
     cli_ctx.logger.success(f"Account '{alias}' has been deleted")
+
+
+@cli.group()
+def auth():
```
fubuloubu, line 230: @antazoey what do you think of this?

antazoey: I guess one thing worth considering is "auth" often refers to "authentication" and not always "authorization", so if ever something closer to wallet-connect became a better thing, it might clash here

antazoey (dismissed):

src/ape_accounts/_cli.py
```diff
@@ -222,3 +224,38 @@ def delete(cli_ctx, alias):
...
+
+    else:
+        click.secho(f"{account.address} has no delegate", fg="red")
+
+
+@auth.command(name="set", cls=ConnectedProviderCommand)
+@ape_cli_context()
+@network_option()
+@account_option()
+@click.argument("contract", type=AddressType)
```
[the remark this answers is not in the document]
fubuloubu: 74c7bc292 larger refactor to accept anything that can be converted into `AddressType` (also changed return type of `.delegate` to just `BaseAddress`, in case address is not verified)

src/ape_accounts/_cli.py
```diff
@@ -222,3 +224,38 @@ def delete(cli_ctx, alias):
...
+@auth.command(name="show", cls=ConnectedProviderCommand)
+@network_option()
+@account_option()
+def show_delegate(account):
+    """Show if an existing delegate is authorized for account"""
+    if contract := account.delegate:
+        click.echo(f"{account.address} is delegated to {contract.address}")
+
+    else:
+        click.secho(f"{account.address} has no delegate", fg="red")
```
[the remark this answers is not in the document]
fubuloubu: Not sure if this is the more dangerous outcome for the api to raise nonzero error

fubuloubu:
stuck on https://github.com/ethereum/go-ethereum/issues/31967

Issue #2604, opened 2025-05-01
Add `--no-recurse` flag to `ape pm install` to avoid installing deps of deps


antazoey:
### Describe

deps of deps are not always necessary, especially if using `--force` just on a single dep to reinstall.

### Specification

@click.option("--no-recursion", is_flag=True)

PR #2614, opened 2025-05-08
feat: `--no-recurse` flag on  `ape pm install` command

antazoey:
### What I did

fixes: #2604 

This can be quite the time saver when you only want to force reinstall the main dependency and not all of its dependencies... like i dont need to install `forge-std` over and over again, but my own projects that are also dependencies, yes i typically need to force reinstall those to get new changes and stuff

### Checklist


antazoey:
tests TBD

fubuloubu (approved):

Issue #2580, opened 2025-04-23
Padding on wrong side when giving a shorter hex str for bytes32 in a struct


antazoey:
### Have you ensured that all of these are up to date?

- [x] Ape
- [ ] Ape plugins

### What version of Ape are you on?

latest

### What command(s) is the bug in?

python

### Operating System

macOS (Apple Silicon)

### Description

So i have a struct with a bytes32 field. I tried to pass it an address sized value, in hopes it would left-pad zeroes... well, it right padded zeroes which is way wrong.

struct MyStruct {
    my_field: bytes32
}

data = {"my_field": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"}

resulted in a `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266000000000000000000000000` value

this might not be in Ape but it probably is...

PR #2620, opened 2025-05-15
test: document and show-case more the right-padded nature of bytes encoding

antazoey:
### What I did

It was counter-intuitive to me that bytes values are r-padded.
I thought it was a bug, it is not.. I tried to document it and show-case it more for next time.

fixes: #2580 

### How I did it

recursively left-pad bytes in struct parser

### Checklist


fubuloubu (approved):

tests/functional/conversion/test_encode_structs.py
```diff
@@ -44,7 +44,7 @@ class SimilarStruct(Struct):
 
 EXPECTED = HexBytes(
     "0000000000000000000000000000000000000000000000000000000000000001"
-    "0200000000000000000000000000000000000000000000000000000000000000"
```
antazoey, line 47: @fubuloubu I am surprised we had so many tests expecting right padded. Surely that can't be correct. I maybe just glossed over this. But I want to make sure...

antazoey: OK so I realize dynamic sized types are right padded, but bytes32 is fixed size so it should remain left padded I think... however, I want to make sure dynamic sizes still work...

antazoey:
I think I was just simply wrong, bytes are r-padded I guess.

fubuloubu (approved):

Issue #2631, opened 2025-05-29
Easier way to get prepped/signed/serialized tx from contract method handle


antazoey:
### Describe

Right now, i have to do this to get tx bytes

```
tx = contract.setNumber.as_transaction(321, sender=user)
prepped_tx = user.prepare_transaction(tx)
signed_tx = user.sign_transaction(prepped_tx)
tx_bytes = signed_tx.serialize_transaction()
```

it is verbose, itd be nice to get them faster and with less lines of code

### Specification

in contract tx/call handler classes, do
```
def serialize_transaction(self, *args, **kwargs):
        contract_transaction = self._as_transaction(*args)
        if "sender" not in kwargs and self.account_manager.default_sender is not None:
            kwargs["sender"] = self.account_manager.default_sender

       # Differs here *****
        return contract_transaction.serialize_transaction(*args, **kwargs)
```

it is very close to `__call__` except `ContactTransactionHandler` calls `ContractTransaction.serialize_transaction` instead of the `ContractTransaction.__call__`

PR #2633, opened 2025-05-30
feat(contracts): easily get signed serialized transactions bytes from contract handlers

antazoey:
### What I did

This helps me turn 4 lines of code into 1

```python
tx = contract.MyMethod.as_transaction(sender=me)
prepped_tx = me.prepare_transaction(tx)
signed_tx = me.sign_transaction(prepped_tx)
tx_bytes  = signed_tx.serialize_transaction()
```

to

```python
tx_bytes = contract.MyMethod.as_transaction_bytes(sender=me)
```

fixes: #2631 

### Checklist

### DIFF
--- a/src/ape/api/accounts.py
+++ b/src/ape/api/accounts.py
@@ -83,6 +83,11 @@ class AccountAPI(BaseInterfaceModel, BaseAddress):
         """
         return None
 
+    def prepare_transaction(self, txn: "TransactionAPI", **kwargs) -> "TransactionAPI":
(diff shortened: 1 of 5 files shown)


fubuloubu (approved):

fubuloubu:
> @fubuloubu I am open to any design feedback here, I just don't want to have to do call 4 Ape methods in a row like this just to get the serialized transaction bytes, which i think is a common enough use-case, as many non-standard RPCs require this.

No this seems great!

PR #2675, opened 2025-08-27
fix: PluginManager to handle single items and iterables

zkpepe:
### What I did

updated `__getattr__` in `PluginManager` so it handles both single objects and collections (lists, tuples, generators) correctly.

### How I did it

replaced the `Generator` check with an `Iterable` check (excluding strings and bytes). Iterates over collections when needed, validates single objects directly.

### How to verify it

call hooks that return either a single object or a collection and ensure they are processed correctly without errors.

### Checklist

* [x] All changes are completed
* [ ] Change is covered in tests
* [ ] Documentation is complete

### DIFF
--- a/src/ape/managers/plugins.py
+++ b/src/ape/managers/plugins.py
@@ -104,16 +104,15 @@ class PluginManager:
 
         for plugin_name, results in map(get_plugin_name_and_hookfn, hookimpls):
             # NOTE: Some plugins return a tuple and some return iterators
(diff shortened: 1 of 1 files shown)


fubuloubu:
Not against an update here, but as you can see from CI this code is kind of fragile and has a large impact when something changes slightly

Feel free to continue working on it, but it might not be worth the time for the update

PR #2685, opened 2025-09-15
fix: issue creating transaction not connected to a provider

antazoey:
### What I did

Just a simple fix to allow creating transactions without requiring a connection to a provider

### Checklist

### DIFF
--- a/src/ape/api/transactions.py
+++ b/src/ape/api/transactions.py
@@ -75,10 +75,11 @@ class TransactionAPI(BaseInterfaceModel):
     @classmethod
     def validate_gas_limit(cls, value):
         if value is None:
-            if not cls.network_manager.connected:
-                raise NetworkError("Must be connected to use default gas config.")
-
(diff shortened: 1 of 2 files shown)


src/ape/api/transactions.py
```diff
@@ -75,10 +75,11 @@ def __init__(self, *args, **kwargs):
...
     def validate_gas_limit(cls, value):
         if value is None:
-            if not cls.network_manager.connected:
-                raise NetworkError("Must be connected to use default gas config.")
-
-            value = cls.network_manager.active_provider.network.gas_limit
+            value = (
+                cls.network_manager.active_provider.network.gas_limit
+                if cls.network_manager.connected
+                else 0
```
antazoey, line 81: Should I make this default to `21_000` instead?

antazoey: my goal is just to be able to create default txns offline

Issue #2624, opened 2025-05-23
Ability to register custom classes to convert to certain types when used as inputs


antazoey:
### Describe

It'd be cool if I could registry my custom class to automatically convert to a certain type when used as in put.
For example, I have a custom class with a `.contract` property. Sometimes I forget to do `obj.contract` when passing stuff around.

### Specification

One idea is to make a registry decorator

```python
import ape.convert

class MyClass:
    @property
    def contract(self) -> ContractInstance:
        ...

    @ape.convert.register
    def as_input(self):
        return self.contract
```

PR #2694, opened 2025-10-01
feat: custom converter types

antazoey:
### What I did

Allow you to implement mixin hooks for a custom class to be convertible to different types.
I have a weird wrapper type that can't easily inherit from `ContractInstance` and I am constantly accidentally passing it to transactions but i of course get conversion errors. With this, I can implement these hooks and my class will work as arguments much like a regular `ContractInstance` does.

fixes: #2624

### Checklist


fubuloubu (dismissed):
I like this! Perhaps we can do something in v0.9/v1.x series to make this more generalizable/performant, but for now this works well!

PR #2717, opened 2025-10-23
refactor: pull out usable `ApeSigner` class as a cheaper `AccountAPI`

antazoey:
### What I did

Extracts out the inner `ApeAccount` which essentially is a wrapper around our eth-account signing utilitiy.
Both keyfile and test use this now.
And also, more importantly, users can use this independently which is really convenient.

### Checklist

### DIFF
--- a/src/ape_accounts/accounts.py
+++ b/src/ape_accounts/accounts.py
@@ -14,6 +14,7 @@ from eth_account.hdaccount import ETHEREUM_DEFAULT_PATH
 from eth_account.messages import encode_defunct
 from eth_keys.datatypes import PrivateKey
 from eth_pydantic_types import HexBytes
+from eth_pydantic_types import HexStr as PydanticHexStr
 from eth_typing import HexStr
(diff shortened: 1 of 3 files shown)


src/ape_accounts/accounts.py
```diff
@@ -66,12 +69,149 @@ def __len__(self) -> int:
         return len([*self._keyfiles])
 
 
+class ApeSigner(AccountAPI):
```
antazoey: We could also plop this right in `ape.api.accounts`

src/ape_test/accounts.py
```diff
@@ -107,7 +102,7 @@ def reset(self):
 class TestAccount(TestAccountAPI):
     index: int
     address_str: str
-    private_key: str
+    signer: ApeSigner
```
fubuloubu: Hmm, what would be wrong with subclassing `ApeSigner` here?

antazoey: nothing, just had to also use the `TestAccountAPI` mixin, which should be fine

fubuloubu (approved):

PR #2736, opened 2025-12-17
refactor(Ethereum): improve efficiency of SafeProxy detection heuristic

fubuloubu:
### What I did

Narrow the heuristic which triggers the `masterCopy()` call to a contract to determine if it is a SafeProxy type. This should reduce/eliminate most of the instances where Ape tries the call and it fails and creates a log like so:

```
reverted without message
0xC8C8...6793.0xa619486e()
```

### How I did it

We can add a couple extra ~free checks on code and slot_0 value before triggering `masterCopy()`:
1. check that the method ID for `masterCopy()` (0xa619486e) is actually in the proxy code (should always be true, but has potential for false positives)
2. check that slot_0 is "address-like" (has 12 leading empty bytes, and at least 12/32 non-empty bytes)

### How to verify it

Anything weird happens

### Checklist

~~- [ ] Documentation is complete~~

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -36,7 +36,6 @@ from ape.exceptions import (
     CustomError,
     DecodingError,
     SignatureError,
-    TransactionError,
 )
 from ape.logging import logger
 from ape.managers.config import merge_configs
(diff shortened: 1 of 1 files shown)


antazoey (approved):

antazoey (dismissed):

banteg:
properly addressed in #2784

PR #2756, opened 2026-02-11
fix: honor skip trace in proxy calls

antazoey:
### What I did

I noticed all the proxy traces still appeared which was annoying and causing confusion in the logs

### Checklist


fubuloubu (approved):
I also had another PR that would prevent it from making repeated calls in the first place, but was having difficulties there

PR #2760, opened 2026-03-02
fix(config): respect ethereum.local default_provider overrides

### LINKED ISSUE #2490: `project.config.ethereum.local.default_provider` has a different value than `networks.ethereum.local.default_provider`
Thankfully `networks.ethereum.local.default_provider` is the correct one but it is still concerning the project config one is wrong

sanjaysudagani:
### What I did

Updated the config model so project/global overrides for `ethereum.local.default_provider` are reflected in `EthereumConfig.local`.

fixes: #2490

### How I did it

- read the raw `local` override from the plugin extras instead of always constructing a hard-coded local config
- merged the configured `local` values with the existing local defaults before validating as `NetworkConfig`
- added regression coverage for `ethereum.local.default_provider` override behavior

### How to verify it

```bash
pytest tests/functional/test_config.py -k ethereum_local_network_config_override -q
pytest tests/functional/test_network_api.py -k "gas_limit_defaults or gas_limit_configured_max or base_fee_multiplier or config_custom_networks_default or config_custom_networks" -q
```

### Checklist

- [x] Documentation is complete; no user-facing docs update was needed for this internal config-model fix

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -283,11 +283,22 @@ class BaseEthereumConfig(PluginConfig):
     @computed_field  # type: ignore[misc]
     @cached_property
     def local(self) -> NetworkConfig:
-        return create_local_network_config(
+        default_local = create_local_network_config(
             default_provider="test",
             default_transaction_type=self.DEFAULT_TRANSACTION_TYPE,
             gas_limit=self.DEFAULT_LOCAL_GAS_LIMIT,
         )
+        if not (override := (self.__pydantic_extra__ or {}).get("local")):
+            return default_local
+
+        if isinstance(override, NetworkConfig):
+            return override
+
+        if isinstance(override, dict):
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):
Thank you so much!

src/ape_ethereum/ecosystem.py
```diff
@@ -283,11 +283,22 @@ def load_network_configs(cls, values):
...
-        return create_local_network_config(
+        default_local = create_local_network_config(
             default_provider="test",
             default_transaction_type=self.DEFAULT_TRANSACTION_TYPE,
             gas_limit=self.DEFAULT_LOCAL_GAS_LIMIT,
         )
+        if not (override := (self.__pydantic_extra__ or {}).get("local")):
+            return default_local
+
+        if isinstance(override, NetworkConfig):
```
antazoey, line 294: do we also want to merge here (like we do for dict)? just have to dump it first if that's the case

PR #2762, opened 2026-03-15
feat(ext): fastapi network context

### LINKED ISSUE #2761: Integrate with FastAPI
### Describe

It'd be really nice to have an easy way to integrate Ape with a framework like FastAPI, so that certain network requests could be handled within a specific network context (including request body validation)

### Specification

Not sure how this would work in practice, but something like:

```py
@app.post("/{ecosystem}/{network}/...")
@ape.networks.wrap_context()
def send_transaction(body: ParsedWithApeConverts):
    tx = body.target.method(*body.args, sender=settings.sender)
    return tx.hash
```

### Dependencies

_No response_

ChickenBenny:
### What I did

Added `ape.ext.fastapi` — a purpose-built FastAPI integration module that provides
a `network_context()` decorator for managing Ape network contexts in route handlers.

This is the first integration under the new `ape.ext` package, which is intended
to house future framework integrations (e.g. Flask, Django).

fixes: #2761

### How I did it

- Created `ape/ext/` as a new package for external framework integrations
- Implemented `@network_context()` in `ape/ext/fastapi.py`
- The decorator automatically injects `ecosystem` and `network` as FastAPI
  `Query` parameters by patching `wrapper.__signature__` at decoration time —
  so route handlers do not need to declare them explicitly
- If the handler does declare `ecosystem`/`network`, they are forwarded as normal
- `fastapi` is kept as an optional dependency, imported lazily inside the decorator
  to avoid making it a hard requirement for all Ape users
- Only synchronous routes are supported for now, as async network context
  is not yet available in Ape

### How to verify it

Install `fastapi` and run:

```bash
pytest tests/functional/test_ext_fastapi.py -v
```

Expected: 9 passed.

Example usage:

```python
from ape.ext.fastapi import network_context

@app.post("/faucet/{token}")
@network_context()
def send_token(token: str, receiver: str, amount: int) -> str:
    # POST /faucet/USDC?ecosystem=ethereum&network=mainnet
    # Ape is automatically connected to the requested network
    ...
```

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,6 +125,8 @@ test = [  # `test` GitHub Action jobs use this
     "ape-vyper>=0.8.10,<0.9",  # Needed for compiling test contracts
     "vyper>=0.4.3,<0.5",  # Avoid having to download Vyper binaries
     "ape-solidity>=0.8.5,<0.9",  # Needed for compiling test contracts
+    "fastapi",  # Needed for ext/fastapi tests
+    "httpx",  # Needed for FastAPI TestClient
 ]
 lint = [  # `lint` GitHub Action jobs use this
     "ruff>=0.12.0",  # Unified linter and formatter
--- a/src/ape/ext/__init__.py
+++ b/src/ape/ext/__init__.py
@@ -1,0 +1,1 @@
+# ape.ext — External framework integrations for Ape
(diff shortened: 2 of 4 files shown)


fubuloubu (dismissed):

fubuloubu:
I think that the tests might not work without adding fastapi to test dependencies

ChickenBenny:
Add the fastapi dependency in [dc658b3](https://github.com/ApeWorX/ape/pull/2762/commits/dc658b3da8250f89bb0619e31f7070531ca59131) 🚀

fubuloubu (approved):

PR #2768, opened 2026-04-21
feat: provider connect ctx

antazoey:


fubuloubu (dismissed):
I was just thinking about something like that. I think eventually we will need a more friendly mulyi-chain story, to switch back and forth without disconnecting, but supporting async (which I think this could also do)

fubuloubu (dismissed):

PR #2785, opened 2026-05-08
perf: skip unnecessary proxy storage checks

banteg:
## Summary
- move bytecode-marker-gated Safe and EIP-897 checks ahead of generic storage-slot probing
- use an executable `DELEGATECALL` opcode gate before Sequence, Safe, and generic storage-backed proxy probes
- keep EIP-897 detection before the opcode gate because forwarding proxies expose `proxyType()`/`implementation()` but may not contain `DELEGATECALL`
- add public bytecode utilities for stripping compiler metadata and PUSH arguments before opcode scans
- keep bytecode normalization at the `get_proxy_info()` boundary with `HexBytes`, so the utilities operate on bytes directly

## Performance implications
Proxy detection is a hot path where contract code is usually already cached, but storage reads and `eth_call`s still hit the node. This change uses cached runtime bytecode to reject contracts that cannot delegate before doing heavier proxy probes, while preserving the EIP-897 forwarding-proxy exception.

That mainly improves non-proxy contracts and Safe/storage-slot paths:

| Case | Before warm RPCs | After warm RPCs | Effect |
|---|---:|---:|---|
| WETH non-proxy | 5 `get_storage` | 0 | avoids all storage probes |
| USDT non-proxy | 5 `get_storage` | 0 | avoids all storage probes |
| USDC proxy | 3 `get_storage` | 3 `get_storage` | unchanged for real storage-backed proxy |
| Yearn Safe | 6 `get_storage` + 1 `eth_call` | 1 `get_storage` + 1 `eth_call` | reaches Safe detection before generic slots |

Cold paths still pay the expected `get_code` lookup first, then follow the same reduced storage/call behavior.

## Validation
Benchmarked against a local mainnet Reth archive node at `http://127.0.0.1:8545`.

- `uv run ruff check src/ape_ethereum/ecosystem.py src/ape_ethereum/utils.py tests/functional/geth/test_proxy.py tests/functional/test_proxy.py`
- `uv run ruff format --check src/ape_ethereum/ecosystem.py src/ape_ethereum/utils.py tests/functional/geth/test_proxy.py tests/functional/test_proxy.py`
- `uv run pytest tests/functional/geth/test_proxy.py tests/functional/test_proxy.py`
- live before/after benchmark against local Reth archive node

### DIFF
--- a/src/ape_ethereum/ecosystem.py
+++ b/src/ape_ethereum/ecosystem.py
@@ -78,6 +78,7 @@ from ape_ethereum.transactions import (
     TransactionStatusEnum,
     TransactionType,
 )
+from ape_ethereum.utils import strip_compiler_metadata, strip_push_data
 
 if TYPE_CHECKING:
     from ethpm_types import ContractType
@@ -472,12 +473,12 @@ class Ethereum(EcosystemAPI):
 
     def get_proxy_info(self, address: AddressType) -> ProxyInfo | None:
         contract_code = self.chain_manager.get_code(address)
-        if isinstance(contract_code, bytes):
-            contract_code = to_hex(contract_code)
+        code_bytes = HexBytes(contract_code)
 
-        if not (code := contract_code[2:]):
+        if not code_bytes:
             return None
 
+        code = code_bytes.hex()
         patterns = {
             ProxyType.Minimal: r"^363d3d373d3d3d363d73(.{40})5af43d82803e903d91602b57fd5bf3",
             ProxyType.ZeroAge: r"^3d3d3d3d363d3d37363d73(.{40})5af43d3d93803e602a57fd5bf3",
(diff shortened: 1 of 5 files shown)


src/ape_ethereum/ecosystem.py
```diff
@@ -93,6 +93,32 @@
...
 }
 BLUEPRINT_HEADER = HexBytes("0xfe71")
+SOLIDITY_METADATA_MARKERS = (
+    "a165627a7a7230",
+    "a264697066735822",
+    "a265627a7a7231",
+)
+
+
+def _has_opcode(code: str, opcode: int) -> bool:
```
fubuloubu: Add a note that this can have false positives in different scenarios where data bytes are included that have the same matching opcode.

That is okay for this scenario, because it's just an additional check to reduce hot path traversal

fubuloubu (dismissed):

src/ape_ethereum/ecosystem.py
banteg: Added the note in `_has_opcode()` documenting that opaque data bytes can still produce false positives, which is acceptable here because this is only a cheap hot-path filter before storage probes.

fubuloubu (approved):

PR #2786, opened 2026-05-08
perf: cache proxy detection misses

banteg:
## Summary

- Cache proxy-detection misses in the existing `proxy_info` disk cache by writing JSON `null` for addresses confirmed not to be proxies.
- Add a proxy-info-specific cache lookup that distinguishes unchecked addresses from checked `ProxyInfoAPI | None` results.
- Skip `ecosystem.get_proxy_info()` when `proxy_info/<address>.json` already exists, whether it contains proxy info or `null`.
- Keep normal proxy-info reads returning `ProxyInfoAPI | None`, so a negative marker never becomes fake proxy info.
- Add backwards-compatibility coverage showing existing on-disk proxy-info JSON objects still load as proxy info.

## Performance implications

This removes repeat proxy-detection work for contracts already known not to be proxies. That matters on the contract lookup hot path because a miss currently still re-enters `get_proxy_info()` every time when no contract type is cached.

For bytecode with no storage-backed proxy shape, the improvement is mostly CPU/cache lookup overhead. For non-proxy bytecode that contains `DELEGATECALL`, the `null` marker also avoids repeated storage probes.

Live local Reth mainnet archive-node check from the previous implementation of the same no-hit behavior:

```text
WEB3_PROVIDER_URI=http://127.0.0.1:8545 uv run python /private/tmp/ape_proxy_miss_bench.py
provider=http://127.0.0.1:8545 chain_id=1
WETH9: before_rpc={"get_code": 0, "get_storage": 0, "send_call": 0} after_rpc={"get_code": 0, "get_storage": 0, "send_call": 0} before_mean_ms=0.381 after_mean_ms=0.032 before_median_ms=0.287 after_median_ms=0.018
Safe singleton v1.3.0: before_rpc={"get_code": 0, "get_storage": 100, "send_call": 0} after_rpc={"get_code": 0, "get_storage": 5, "send_call": 0} before_mean_ms=53.274 after_mean_ms=2.604 before_median_ms=50.451 after_median_ms=0.021
```

Benchmark setup used 20 repeated `chain.contracts.get(..., fetch_from_explorer=False, fetch_from_disk=False)` calls with `ChainManager` code cache pre-warmed. The "before" case clears the negative marker between calls to simulate current behavior; the "after" case keeps it.

## Tests

- `uv run ruff check src/ape/managers/_contractscache.py src/ape/utils/os.py tests/functional/test_contracts_cache.py`
- `uv run ruff format --check src/ape/managers/_contractscache.py src/ape/utils/os.py tests/functional/test_contracts_cache.py`
- `uv run mypy src/ape/managers/_contractscache.py src/ape/utils/os.py`
- `uv run pytest tests/functional/test_contracts_cache.py`

Commit hook note: full-repo `mypy` still fails on unrelated existing errors in `src/ape/cli/choices.py`, `src/ape/api/config.py`, and `src/ape_ethereum/ecosystem.py`.

### DIFF
--- a/docs/userguides/proxy.md
+++ b/docs/userguides/proxy.md
@@ -40,6 +40,16 @@ This allows you to still call methods as you normally do on proxy contracts.
 my_contract.my_method(sender=account)
 ```
 
+Ape caches proxy detection results on live networks, including when no proxy is found.
+If an address becomes a proxy after it was previously checked, delete its cached contract data
+before loading it again:
+
+```python
+from ape import chain
+
+del chain.contracts["0x..."]
+```
+
 ## Manual Proxy Configuration
 
 If you need more control over proxy behavior, you can manually specify proxy information when creating contract instances with `ContractContainer.at()`:
--- a/src/ape/managers/_contractscache.py
+++ b/src/ape/managers/_contractscache.py
@@ -2,6 +2,7 @@ import json
 from collections.abc import Collection
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
+from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
 from typing import TYPE_CHECKING, Generic, TypeVar
@@ -30,6 +31,12 @@ if TYPE_CHECKING:
 _BASE_MODEL = TypeVar("_BASE_MODEL", bound=BaseModel)
 
 
+@dataclass(frozen=True)
+class ProxyInfoCacheEntry:
+    exists: bool
+    value: ProxyInfoAPI | None = None
+
+
 class ApeDataCache(CacheDirectory, Generic[_BASE_MODEL]):
     """
     A wrapper around some cached models in the data directory,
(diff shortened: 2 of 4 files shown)


banteg:
Followed up in f74e9f360 to make the persistent negative-cache trade-off recoverable through the existing mapping API:

- normalize addresses in `ContractCache.__delitem__()`, so `del chain.contracts["0x..."]` works with unchecksummed input
- add a live-network regression proving lowercase deletion removes the persisted proxy no-hit
- document the cache-busting path in the proxy guide

The branch is rebased onto current `main`.

Validation:

- `uv run ruff check src/ape/managers/_contractscache.py src/ape/utils/os.py tests/functional/test_contracts_cache.py`
- `uv run ruff format --check src/ape/managers/_contractscache.py src/ape/utils/os.py tests/functional/test_contracts_cache.py`
- `uv run mypy src/ape/managers/_contractscache.py src/ape/utils/os.py`
- `uv run pytest -q tests/functional/test_contracts_cache.py` (51 passed)
- `uv run mdformat --check docs/userguides/proxy.md`

fubuloubu (approved):

PR #2799, opened 2026-07-11
feat!: support ordered network explorers

banteg:
## Summary

- replace `NetworkAPI.explorer` with an ordered, cached `NetworkAPI.explorers` tuple
- prefer Sourcify when it is installed and supports the active network, while preserving existing order for other explorers
- deduplicate explorer plugins that match through both registration and `supports_chain()`
- fall back across every matching explorer when contract metadata is missing or an explorer raises
- keep publishing on the first preferred explorer, while transaction URL generation falls through explorers that do not implement it

## Breaking change

`NetworkAPI.explorer` is removed. Consumers should use `NetworkAPI.explorers`; index `0` is the preferred explorer when the tuple is non-empty.

## Why

Explorer selection previously depended on installed-distribution and Pluggy registration order, and contract lookup stopped after the first explorer returned no metadata or raised. That made source discovery unreliable when multiple explorer plugins were installed.

Sourcify is preferred because it does not require authentication and has generous rate limits. If Sourcify is absent or unsupported, the existing explorer order is retained. Operations that Sourcify cannot provide, such as transaction URLs on chains without known explorer metadata, can fall through to another installed explorer.

## Validation

- `uv run pytest -q tests/functional --ignore=tests/functional/geth` — 1,650 passed, 1 skipped
- `uv run pytest -q tests/functional/geth/test_contracts_cache.py` — 1 passed
- explorer, receipt, and contract-cache focused suite — 122 passed
- transaction and receipt suite — 98 passed
- Ruff checks and formatting
- targeted mypy checks for all changed source modules
- mdformat check for the updated publishing guide

### DIFF
--- a/docs/userguides/publishing.md
+++ b/docs/userguides/publishing.md
@@ -42,12 +42,15 @@ For more information on accessing contract instances, follow [this guide](./cont
 
 ## Publishing to Explorer
 
-If you want to publish your contracts to an explorer, you can use the [publish_contract](../methoddocs/api.html#ape.explorers.ExplorerAPI.publish_contract) on the `ExplorerAPI`.
+If you want to publish your contracts to an explorer, use the
+[publish_contract](../methoddocs/api.html#ape.explorers.ExplorerAPI.publish_contract) method on
+an `ExplorerAPI`. A network's explorers are ordered by preference, so index `0` selects the
+preferred explorer.
 
 ```python
 from ape import networks
 
-networks.provider.network.explorer.publish_contract("0x123...")
+networks.provider.network.explorers[0].publish_contract("0x123...")
 ```
 
 If you want to automatically publish the source code upon deployment, you can use the `publish=` kwarg on the `deploy` methods:
--- a/src/ape/api/networks.py
+++ b/src/ape/api/networks.py
@@ -611,8 +611,8 @@ class EcosystemAPI(ExtraAttributesMixin, BaseInterfaceModel):
         data["providers"] = []
         network = self[network_name]
 
-        if network.explorer:
-            data["explorer"] = str(network.explorer.name)
+        if network.explorers:
+            data["explorer"] = str(network.explorers[0].name)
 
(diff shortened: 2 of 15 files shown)


banteg:
Downstream blast-radius recon for removing `NetworkAPI.explorer` (read-only; no downstream PRs opened):

- **ape-etherscan — required runtime migration.** [`EtherscanDependency.explorer`](https://github.com/ApeWorX/ape-etherscan/blob/main/ape_etherscan/dependency.py), [`SourceVerifier._wait_for_verification()`](https://github.com/ApeWorX/ape-etherscan/blob/main/ape_etherscan/verify.py), its explorer fixture, and README use the singular property. These call sites must select the `Etherscan` instance from `.explorers` by type/name; using `.explorers[0]` would be incorrect because this PR normally places Sourcify first. Its `eth-ape<0.9` bound also needs a coordinated release update.
- **ape-sourcify — docs/release migration.** Runtime code has no singular-property dependency, but the README currently describes nondeterministic first-explorer selection and references `.explorer`. Its `eth-ape<0.9` bound must also be updated for the 0.9 release.
- **ape-blockscout — tests only.** The fixture assigns/reads `network.explorer`; runtime code has no hit. It is currently pinned to `eth-ape<0.8` and has not been updated since 2024, so 0.9 support would be a broader compatibility decision rather than a small follow-up.
- **ape-safe — no current impact.** A stale local checkout contained a reference, but current upstream `main` no longer does.

Two non-plugin ApeWorX applications also have runtime references: [`faucet`](https://github.com/ApeWorX/faucet/blob/main/faucet/app.py) and [`stablecoin`](https://github.com/ApeWorX/stablecoin/blob/main/app.py). Their generic transaction-URL lookups can use the first available explorer.

The main coordination requirement before merging for 0.9 is therefore maintainer buy-in for the ape-etherscan migration/release, plus the ape-sourcify version-bound and documentation update.

PR #2800, opened 2026-07-19
fix: resolve overloaded methods by argument type, not just count

### LINKED ISSUE #2670: Conversion error when two different data types are accepted for args
### Have you ensured that all of these are up to date?

- [x] Ape
- [x] Ape plugins

### What version of Ape are you on?

_No response_

### What command(s) is the bug in?

_No response_

### Operating System

Linux

### Ape Configuration

_No response_

### Description

A particular contract on Base is causing this issue, it has an overloaded method that accepts either an `int` or `bytes[]` (which would not be convertible between each other), but it will not attempt to call the correct selector.

I believe what should happen is if it can't find a matching set of args for the first selector, it should "fall through" to the second (in this case, a list is not convertible to `int`, but is convertible to `bytes[]`)

### Reproduction code

`$ ape console --network base:mainnet`:
```py
>>> pyth = Contract("0x8250f4aF4B972684F7b336503E2D6dFeDeB1487a")
>>> pyth.getUpdateFee
getUpdateFee(bytes[] updateData) -> uint256
getUpdateFee(uint256 updateDataSize) -> uint256
>>> pyth.getUpdateFee([])
ERROR:    (ConversionError) No conversion registered to handle '[]'.
```

beardthelion:
Calling an overloaded method whose overloads share the same argument count (e.g. `getUpdateFee(bytes[])` and `getUpdateFee(uint256)`) failed with `ConversionError: No conversion registered to handle '[]'` instead of using the overload the arguments fit.

`_select_method_abi` selected by argument count alone and, among equal-count candidates, kept the last one declared, with no type check. So a list argument (valid for `bytes[]`, not `uint256`) was routed to the `uint256` overload and failed at encode time.

This makes selection type-aware: when more than one overload matches the argument count, it picks the first whose input types the arguments encode to, and otherwise keeps the previous behavior (single candidate, no probe, or none fits). Encoding is pure, so the probe adds no network calls. The `multicall` caller, which does not pass the probe, is unaffected.

Validation:
- New test in `tests/functional/test_contract_instance.py`: an overloaded method with same-count `bytes[]` and `uint256` overloads selects `bytes[]` for a list argument, in both declaration orders.

Closes #2670


fubuloubu (dismissed):

fubuloubu:
Overall this is a good fix for ambiguously overloaded ABIs, but it introduces a few risks and a handful of small mitigations I'd recommend before merging:

1. Changed tie-breaker semantics
 - Risk: previously the last-declared overload was chosen; now the first ABI that successfully encodes is chosen. This can change runtime behavior for ambiguous overload sets.
 - Suggestion: document the new rule in a comment

2. Broad exception masking in _can_encode
 - Risk: catching Exception hides programming/runtime errors (not only encoding/conversion failures) and makes debugging harder.
 - Suggestion: narrow the except to the specific conversion/encoding exceptions (e.g., ConversionError / EncodingError) and log probe failures at debug level with the ABI signature and exception type.

3. Testing / observability
- Add tests for:
  * multiple overloads that both can encode the same args (deterministic selection)
  * probe failure paths (conversion raising unexpected exceptions)
- Add debug logging for probe failures including ABI signature and exception class.

fubuloubu (approved):

PR #2804, opened 2026-08-18
fix: reject path separators in account aliases

SashaMIT:
`_write_and_return_account` joins the alias into the accounts data folder as `{alias}.json`. `_validate_account_alias` already rejects long hex strings so a private key is not stored as a filename, but it still accepted `../name` and `nested/alias`.

`generate_account("../pwned_ape_alias", ...)` wrote the encrypted keystore to `DATA_FOLDER/pwned_ape_alias.json` instead of `DATA_FOLDER/accounts/`. Same for import-from-mnemonic and import-from-private-key.

## Change

- Require `Path(alias).name == alias` (and a non-empty alias) in `_validate_account_alias`.
- Keep the existing hex and in-use checks.

## Test

`pytest tests/functional/test_accounts.py::test_generate_account_rejects_path_alias tests/functional/test_accounts.py::test_generate_account_invalid_alias` (2 passed). Revert-tested: the new test fails with `DID NOT RAISE` when the path-segment guard is removed.


fubuloubu (approved):
Good find, thank you!

SashaMIT:
Thanks for merging this.

PR #2807, opened 2026-08-24
fix: jail extract_archive members to the destination

SashaMIT:
`extract_archive` strips a shared top-level directory from zip/tar members, then writes `destination / relative_path`. A member such as `pkg/../escaped.txt` becomes `../escaped.txt` after the strip, so the write lands outside `destination`.

`_fetch_from_pypi` downloads a PyPI sdist and calls `extract_archive`. A malicious or custom-index archive can write outside the extract cache.

## Change

- Resolve each extracted path and reject members that are not under `destination`.
- Keep the existing top-level-dir strip for happy-path sdists.
- Drop the unjailed `extractall` fallback.

## Test

`pytest tests/functional/utils/test_os.py::test_extract_archive_zip_strips_top_level_dir tests/functional/utils/test_os.py::test_extract_archive_zip_rejects_parent_escape tests/functional/utils/test_os.py::test_extract_archive_tar_rejects_parent_escape`

Revert-tested: without the jail, the zip escape test writes `dest.parent/escaped.txt`.


fubuloubu (approved):

SashaMIT:
Thanks for merging this, @fubuloubu. The extract_archive jail is a nice companion to the alias-path fix.

PR #2808, opened 2026-09-12
ci: add release artifact provenance notarization via ProofCore Action

ProofCore-Protocol:
Hi Ape team! 👋

Following up on recent CI hardening around PyPI trusted publishing and OIDC boundaries, this PR adds automated cryptographic provenance attestation for compiled release artifacts (dist/*) using [ProofCore Action](https://github.com/ProofCore-Protocol/proofcore-action).

How it works:
    Runs directly after uv build and before PyPI publishing.
    Computes SHA-256 digests of the generated wheel and sdist files locally on the runner.
    Leverages the existing id-token: write permission to anchor the build manifest and GitHub OIDC token directly to the blockchain.
    Allows developers and downstream smart contract tooling relying on ape to independently verify artifact integrity and origin against public ledger records.

Zero disruption:
    Pure zero-storage model: raw code or binaries are never stored externally.
    Runs with continue-on-error: true to guarantee that publishing is never blocked in case of external network timeouts.

Thanks for building the best Python framework for smart contracts! 🚀


fubuloubu:
Thank you for your contribution!

Unfortunately, we do not allow access to our publishing pipeline to 3rd party tools

